PR #617, opened 2017-09-26
Announcements - make announcements directly from events, meetings and interest groups

liseph:
Text type:
image

Button type:
image 1

When clicked:
image 2
image 3

Meetings:
screenshot from 2017-11-02 23-08-00

Interest groups:
screenshot from 2017-11-02 23-09-23

Events:
screenshot from 2017-11-02 23-10-05

### DIFF
--- a/app/components/AnnouncementInLine/index.js
+++ b/app/components/AnnouncementInLine/index.js
@@ -9,13 +9,14 @@ import { Form, TextArea } from 'app/components/Form';
 import { reduxForm, Field, reset } from 'redux-form';
 import { createAnnouncement } from 'app/actions/AnnouncementsActions';
 import styles from './AnnouncementInLine.css';
+import type { CreateAnnouncement, ID } from 'app/models';
 
 type Props = {
   placeholder?: string,
-  event?: number,
-  meeting?: number,
-  group?: number,
-  createAnnouncement: (...any) => void,
+  event?: ID,
+  meeting?: ID,
+  group?: ID,
+  createAnnouncement: CreateAnnouncement => Promise<*>,
   handleSubmit: Function => void,
   actionGrant: boolean,
   hidden?: boolean,
@@ -80,8 +81,7 @@ class AnnouncementInLine extends Component<Props, State> {
                   onClick={this.handleHide}
                   className={styles.announcementButton}
                 >
-                  {' '}
-                  Ny kunngjøring{' '}
+                  Ny kunngjøring
                 </Button>
               )}
               {showLabel && (
@@ -112,8 +112,7 @@ class AnnouncementInLine extends Component<Props, State> {
                     className={styles.fieldText}
                   />
                   <Button submit className={styles.button}>
-                    {' '}
-                    SEND{' '}
+                    SEND
                   </Button>
                 </Form>
               )}
--- a/app/components/AnnouncementInLine/index.js
+++ b/app/components/AnnouncementInLine/index.js
@@ -10,17 +10,21 @@ import { createAnnouncement } from 'app/actions/AnnouncementsActions';
 import styles from './AnnouncementInLine.css';
 
 type Props = {
-  label?: string,
+  placeholder?: string,
   event?: number,
   meeting?: number,
   group?: number,
   createAnnouncement: (...any) => void,
-  handleSubmit: () => void,
+  handleSubmit: Function => void,
   actionGrant: boolean,
   hidden?: boolean
 };
 
-class AnnouncementInLine extends Component {
+type State = {
+  hidden: boolean
+};
+
+class AnnouncementInLine extends Component<Props, State> {
   props: Props;
   state = {
     hidden: true
@@ -36,33 +40,28 @@ class AnnouncementInLine extends Component {
     });
   };
 
-  handleHide = current => {
-    this.setState({
-      hidden: !current
-    });
+  handleHide = () => {
+    this.setState(prevState => ({
+      hidden: !prevState.hidden
+    }));
   };
 
   render() {
     const {
       actionGrant,
       handleSubmit,
-      label,
+      placeholder,
       event,
       meeting,
       group
     } = this.props;
-
     return (
       <div>
         {actionGrant &&
           (event || meeting || group) && (
             <div>
-              <a
-                onClick={() => this.handleHide(this.state.hidden)}
-                className={styles.label}
-              >
-                {' '}
-                {label || 'Ny kunngjøring'}{' '}
+              <a onClick={this.handleHide} className={styles.label}>
+                Ny kunngjøring
               </a>
               {!this.state.hidden && (
                 <Form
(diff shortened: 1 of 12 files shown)


lib/state_example.js
```diff
@@ -0,0 +1,1397 @@
+module.exports = {
+  location: {
+    pathname: '/events',
+    search: '',
```
finnss: I'm guessing this file should be in `.gitignore`?

yarn.lock
```diff
@@ -1,33 +1,24 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 # yarn lockfile v1
-
-
 "@types/node@^6.0.46":
```
finnss: I'm not an expert on `yarn.lock`, but 1000 lines changed seems excessive?

app/components/ProfilePicture/Readme.md
```diff
@@ -0,0 +1,7 @@
+Example profilePicture:
+```
+<div>
+  <ProfilePicture user={state.currentUser} />
+  <p> {state.currentUser.fullName} </p>
```
finnss: Guessing this isn't intended to be included in this PR?

app/components/Search/Readme.md
```diff
@@ -0,0 +1,11 @@
+```
+const { Provider } = require('react-redux');
+const configureStore = require('app/utils/configureStore').default;
+const store = configureStore({ state });
```
finnss: Guessing this isn't intended to be included in this PR?

app/routes/announcements/components/AnnouncementItem.js
```diff
@@ -84,19 +84,22 @@ const AnnouncementItem = ({
         </Flex>
       </Flex>
       {!announcement.sent &&
-        actionGrant.includes('send') && (
+        actionGrant.includes('send') &&
+        actionGrant.includes('delete') && (
```
finnss: Is there a difference between who can send and who can delete an announcement?

app/routes/announcements/components/AnnouncementItem.js
```diff
@@ -84,19 +84,22 @@ const AnnouncementItem = ({
...
           <Flex className={styles.wrapperSendButton}>
             <Button
               className={styles.sendButton}
               onClick={() => deleteAnnouncement(announcement.id)}
             >
-              Slett
+              {' '}
+              Slett{' '}
```
finnss: Are these spaces necessary? I think they're kinda fine to use in the middle of a sentence where setting a margin just to imitate a space would be akward, like 
```
Han er{' '}
{longFunctionDetermeningHowOldTheUserIs()}
{' '}år
```
but they aren't a good way to imitate margin or padding.

app/routes/announcements/components/AnnouncementsCreate.js
```diff
@@ -0,0 +1,136 @@
...
+import React from 'react';
+import styles from './AnnouncementsList.css';
+import Flex from 'app/components/Layout/Flex';
+import { Form, SelectInput, TextArea } from 'app/components/Form';
+import { reduxForm, Field, reset } from 'redux-form';
+import Button from 'app/components/Button';
+import { ContentMain } from 'app/components/Content';
+
+type Props = {
+  createAnnouncement: (...any) => void,
```
finnss: Would be best if you made an `AnnouncementEntity` that determines what fields an announcement should have. Probably need one for displaying them and one to determine what fields are sent when creating one, like `CreateAnnouncementEntity` or some such. See reducers for examples. Also, this function probably returns `Promise<*>`, not void.

app/components/AnnouncementInLine/index.js
```diff
@@ -0,0 +1,148 @@
...
+import { reduxForm, Field, reset } from 'redux-form';
+import { createAnnouncement } from 'app/actions/AnnouncementsActions';
+import styles from './AnnouncementInLine.css';
+
+type Props = {
+  placeholder?: string,
+  event?: number,
+  meeting?: number,
+  group?: number,
+  createAnnouncement: (...any) => void,
```
finnss: See other comment about this

app/routes/announcements/components/AnnouncementsCreate.js
```diff
@@ -0,0 +1,136 @@
...
+        ? announcement.meetings.map(meeting => meeting.value)
+        : [],
+      events: announcement.events
+        ? announcement.events.map(event => event.value)
+        : [],
+      send
+    });
+  };
+  return (
+    <ContentMain>
```
finnss: @ekmartin would be proud

app/routes/announcements/components/AnnouncementsList.js
```diff
@@ -1,18 +1,15 @@
...
-import { reduxForm, Field } from 'redux-form';
-import Button from 'app/components/Button';
 import AnnouncementItem from './AnnouncementItem';
+import AnnouncementsCreate from './AnnouncementsCreate';
+import { Content, ContentMain } from 'app/components/Content';
 
 type Props = {
   announcement: Object,
-  announcements: Array</*TODO: Announcement */ any>,
+  announcements: Array<Object>,
```
finnss: These should both be `AnnouncementEntity`, see other comment.

app/routes/announcements/components/AnnouncementsCreate.js
```diff
@@ -0,0 +1,136 @@
...
+}: Props) => {
+  const disabledButton = invalid || pristine || submitting;
+
+  const onSubmit = (announcement, send = false) => {
+    return createAnnouncement({
+      ...announcement,
+      users: announcement.users
+        ? announcement.users.map(user => user.value)
+        : [],
+      groups: announcement.groups
```
finnss: You could probably avoid all these checks if you passed AnnouncementCreate some `initialValues`:

```
const AnnouncementsCreate = ...;

const initialValues = {
    events: [],
    meetings: [],
    ...
}

export default reduxForm({
   initialValues,
   ...
})(AnnouncementCreate);
```

app/routes/announcements/components/AnnouncementsList.js
```diff
@@ -1,18 +1,15 @@
...
   actionGrant: /* TODO: ActionGrant */ any,
   sendAnnouncement: () => void,
   createAnnouncement: (announcement: any) => void,
```
finnss: These both probably return promises. announcement should be an entity, not any.

app/routes/announcements/components/AnnouncementsCreate.js
odinuge: Create a model for it here: https://github.com/webkom/lego-webapp/blob/master/app/models.js

app/components/Search/Readme.md
odinuge: This was removed upstream, and should be removed. The `state-example` should also be removed from the commit it was introduced.

app/routes/announcements/components/AnnouncementsCreate.js
finnss: I guess I need to move all of my Entities there.. 😅

PR #722, opened 2017-10-19
[Discussion] Footer text update

finnss:
Not personally a fan of current text in the footer. Talking about LEGO with the codename and with the context of it being new feels pretty temporary and out of place imo.


ekmartin:
What if we keep the existing section, but replace the `Vi i Webkom har arbeidet lenge med denne siden og håper den faller i smak.` sentence with something about the website being open source and a link to the GitHub organization instead?

PR #725, opened 2017-10-19
Joblistings responsive

tmsolber:


finnss:
I would suggest having a delete button in the Navigation Tab, similar to what's done in [meetings](https://abakus.no/meetings/437). If you think having the delete button only in the edit view is intuitive enough then that's fine too, just my two cents.

PR #793, opened 2017-10-21
Add 'loggedIn' as a default watchProp in prepare

odinuge:
It is annoying when you as a user stays on a page, and then logges in - and you still can't see any content. This can be achived to add `loggedIn` as a watchProp in all routes, but imo this is a better approach. When a user logges out, we clear the whole state (and then setting it to `initialState`). We could (and probably should) do something similar with login, but this is how far I got today.

Thoughts anyone?

### DIFF
--- a/app/utils/prepare.js
+++ b/app/utils/prepare.js
@@ -13,12 +13,13 @@ type PrepareFn = (props: Object, dispatch: Dispatch<*>) => Promise<*>;
  */
 export default function prepare(
   prepareFn: PrepareFn,
-  watchProps: ?Array<string>
+  watchProps?: Array<string> = []
 ) {
   // Returns true if any of the given watchProps have changed:
   const componentWillReceiveProps = (oldProps, newProps) =>
-    watchProps &&
-    watchProps.some(key => get(oldProps, key) !== get(newProps, key));
+    watchProps
+      .concat('loggedIn')
+      .some(key => get(oldProps, key) !== get(newProps, key));
 
   return dispatched(prepareFn, { componentWillReceiveProps });
 }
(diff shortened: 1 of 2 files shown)


odinuge:
@hanswilw, any thoughts about how this will work with SSR?

odinuge:
From the docs of `react-prepare`:
- `opts.componentWillReceiveProps` (default: `true`): on the client, `sideEffect` is called again whenever the component receive props.
- `dispatched`: when server-side rendered using prepare, `sideEffect` will be run and awaited before the component is rendered; if `sideEffect` throws, prepare will also throw.

Looks like this will have no effect on SSR, because `componentWillReceiveProps` is only used on the client.

app/utils/prepare.js
```diff
@@ -13,12 +13,13 @@ type PrepareFn = (props: Object, dispatch: Dispatch<*>) => Promise<*>;
...
-  watchProps: ?Array<string>
+  watchProps?: Array<string> = []
 ) {
   // Returns true if any of the given watchProps have changed:
   const componentWillReceiveProps = (oldProps, newProps) =>
-    watchProps &&
-    watchProps.some(key => get(oldProps, key) !== get(newProps, key));
+    watchProps
+      .concat('loggedIn')
+      .some(key => get(oldProps, key) !== get(newProps, key));
```
finnss: Won't this crash if `watchProps` is undefined?

skjermbilde 2017-10-30 kl 15 20 52

finnss: Could this act weirdly if someone passes in `loggedIn` as a prop manually? You'd get a list like `['loggedIn', 'loggedIn']`. If that could cause issues I suppose we should remove any cases of `loggedIn` being passed as `watchProps` in this PR.

odinuge: The only thing that will happen is that the `loggedIn` prop will be checked twice.

PR #806, opened 2017-10-22
POC: Refactor JoinEventForm Proposal

Hanse:
Move the countdown logic from the form component, create a reducer and make it a little bit more testable.

In order to use enzyme's `state()` to test the state of a component, the component must be the root of `mount() etc`. To use redux-form <Field />, the component must be wrapped in `reduxForm` and thus the component holding the state is no longer the root. The easiest is therefore to decouple everything so you can test the countdown isolated without redux and redux-form.

Should probably add a test to assert what is actually rendered also though.


hanswilw:
Closing in favor of #957

PR #820, opened 2017-10-24
Fix semester handling in CompanyInterest

odinuge:
~~Requires https://github.com/webkom/lego/pull/886~~

And show companyInterest in BDB navigation.

### DIFF
--- a/app/actions/CompanyActions.js
+++ b/app/actions/CompanyActions.js
@@ -333,12 +333,16 @@ export function deleteCompanyContact(
   });
 }
 
-export function fetchSemesters({ companyInterest }: Object = {}) {
+export function fetchSemestersForInterestform() {
+  return fetchSemesters({ company_inerest: 'True' });
+}
+
+export function fetchSemesters(
+  queries: { [key: string]: ?string | ?number } = {}
+) {
   return callAPI({
     types: Company.FETCH_SEMESTERS,
-    endpoint: `/company-semesters/${createQueryString({
-      company_interest: companyInterest
-    })}`,
+    endpoint: `/company-semesters/${createQueryString(queries)}`,
     schema: [companySemesterSchema],
     meta: {
       errorMessage: 'Henting av semestre feilet'
--- a/app/routes/bdb/components/bdb.css
+++ b/app/routes/bdb/components/bdb.css
@@ -4,6 +4,34 @@
   margin: 5px 0 25px;
 }
 
+.root select {
+  font-size: 100%;
+  font-family: 'proxima-nova', 'Helvetica Neue', Arial, sans-serif;
+  cursor: pointer;
+  transition: color 0.2s;
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  text-indent: 1px;
+  text-overflow: '';
+  background-color: inherit;
+  border: none;
+  outline: none;
+
+  &:hover {
+    color: #777;
+    cursor: pointer;
+  }
+
+  &:focus {
+    border: none;
+    outline: none;
+  }
+
+  &::-ms-expand {
+    display: none;
+  }
+}
+
 .companyList table {
   border-collapse: collapse;
 }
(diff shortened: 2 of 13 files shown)


finnss:
Maybe you could add this new `activeInterestForm` field to the `CompanySemesterEntity` in `app/reducers/companySemesters` while you're at it?

odinuge:
Smart @finnss! Will do :ok_hand:

app/routes/companyInterest/CompanyInterestEditRoute.js
```diff
@@ -11,13 +11,13 @@ import CompanyInterestPage, {
...
 import { selectCompanyInterestById } from 'app/reducers/companyInterest';
-import { selectCompanySemesters } from 'app/reducers/companySemesters';
+import { selectCompanySemestersForInterestform } from 'app/reducers/companySemesters';
 import { push } from 'react-router-redux';
 import prepare from 'app/utils/prepare';
 
 const loadCompanyInterests = (props, dispatch) => {
   const { companyInterestId } = props.params;
-  return dispatch(fetchSemesters()).then(() =>
+  return dispatch(fetchSemesters({ companyInterest: 'True' })).then(() =>
```
hanswilw: Hmm, I dont like the 'True' value – maybe better to pass a boolean and fix the query in the action? Add better flow typing would be nice as well, because it is easy to make errors with these strings.

odinuge: Agree,

This was just a quickfix by having the same behavior in the edit and create form. Currently the edit form lists all semesters in the db. Aka. 19. I also added a custom selector to only select active semesters, so this isn't that necessary. This action should just be something like `fetchSemestersForCompanyIntrests`, that is a small wrapper around the `fetchSemesters` action.

odinuge: screenshot from 2017-10-26 13-03-53

finnss:
Hmm, couldn't semester just be a dropdown? How often do companies request the same content for  several semesters at once? And even if they do, another dropdown could appear once the first is selected, or the dropdown could work like the one in the bdb. Having all of these checkboxes seems like an odd solution imo.

odinuge:
It is supposed to be four checkboxes, and imo that looks just fine..

app/routes/companyInterest/CompanyInterestEditRoute.js
```diff
@@ -27,7 +27,9 @@ const mapStateToProps = (state, props) => {
   const companyInterest = selectCompanyInterestById(state, {
     companyInterestId
   });
-  const semesters = selectCompanySemesters(state);
+  // TODO show an intersection of the currently active semesters, and
+  // the semester chosen when the interestform was created
```
finnss: You wanna get this done in this PR or just merge it and get back to this later?

odinuge: Will fix in this PR. None of the things currently is important (Fixed the important stuff in the backend).

app/routes/bdb/components/bdb.css
```diff
@@ -1,48 +1,9 @@
...
-  font-weight: bold;
-  display: block;
-  padding-left: 0;
-}
-
 .root h3 {
   margin: 5px 0 25px;
 }
 
-.root select {
```
odinuge: @finnss What of this should not be removed?

app/routes/companyInterest/CompanyInterestEditRoute.js
odinuge: WIll fix this at a later stage...

app/actions/CompanyActions.js
```diff
@@ -333,12 +333,16 @@ export function deleteCompanyContact(
   });
 }
 
-export function fetchSemesters({ companyInterest }: Object = {}) {
+export function fetchSemestersForInterestform() {
+  return fetchSemesters({ company_inerest: 'True' });
```
odinuge: @hanswilw Is this better?

PR #830, opened 2017-10-24
Fix waiting registration and loadingindicator

hanswilw:
- Handle non existing waiting registrations gracefully
- Fix loading indicator hiding on other users registrations

### DIFF
--- a/app/reducers/events.js
+++ b/app/reducers/events.js
@@ -57,7 +57,7 @@ function mutateEvent(state: any, action: any) {
       if (!stateEvent) {
         return state;
       }
-      let waitingRegistrations = stateEvent.waitingRegistrations;
+      let waitingRegistrations = stateEvent.waitingRegistrations || [];
       if (!registration.pool) {
         waitingRegistrations = [...waitingRegistrations, registration.id];
       }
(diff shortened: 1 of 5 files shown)


app/reducers/events.js
```diff
@@ -68,6 +68,7 @@ function mutateEvent(state: any, action: any) {
           [eventId]: {
             ...stateEvent,
             loading: false,
+            registrationCount: stateEvent.registrationCount + 1,
```
ekmartin: why do you need the + 1 here? shouldn't the number sent by the server when you sign up be up to date?

hanswilw: It is up to date when you fetch the event, but when you and others receive socket messages for other registrations it could be updated like this. So when you are on the calendar or event listview the registrationCount for different events will increase when others sign up. 

This gray one:

ekmartin: ah, misread - thought the registration count was sent with the websocket event. i guess that's something to look at in the future, to make sure it stays up to date even if a user doesn't receive all the websocket events?

hanswilw: When I'm thinking about it we should probably cap this at 30/30, so the waiting registrations does not increase this number. Give me a second to fix that

PR #840, opened 2017-10-25
Add support for admin registration to waiting list

edvardvb:
Kinda hacky, but since there is no pool with `id=0` we can use that as an "id" for the waiting list.

Depends on this PR in the backend: https://github.com/webkom/lego/pull/894

### DIFF
--- a/app/actions/EventActions.js
+++ b/app/actions/EventActions.js
@@ -193,7 +193,7 @@ export function unregister(
 export function adminRegister(
   eventId: number,
   userId: number,
-  poolId: number,
+  poolId?: number,
   feedback: string,
   reason: string
 ) {
@@ -203,7 +203,7 @@ export function adminRegister(
     method: 'POST',
     body: {
       user: userId,
-      pool: poolId ? poolId : undefined,
+      pool: poolId,
       feedback,
       admin_reason: reason
     },
--- a/app/actions/EventActions.js
+++ b/app/actions/EventActions.js
@@ -203,7 +203,7 @@ export function adminRegister(
     method: 'POST',
     body: {
       user: userId,
-      pool: poolId,
+      pool: poolId ? poolId : undefined,
       feedback,
       admin_reason: reason
     },
--- a/app/routes/events/components/EventAdministrate/AdminRegisterForm.js
+++ b/app/routes/events/components/EventAdministrate/AdminRegisterForm.js
@@ -40,7 +40,7 @@ const AdminRegister = ({
           label="Pool"
           options={pools
             .map(pool => ({ value: pool.id, label: pool.name }))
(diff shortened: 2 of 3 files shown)


app/actions/EventActions.js
```diff
@@ -203,7 +203,7 @@ export function adminRegister(
     method: 'POST',
     body: {
       user: userId,
-      pool: poolId,
+      pool: poolId ? poolId : undefined,
```
hanswilw: Instead of this ternary, maybe set the flowtype to `poolId?: number` instead and not pass poolId when calling the function? That way it will implicitly be undefined.

app/routes/events/components/EventAdministrate/AdminRegisterForm.js
```diff
@@ -35,7 +35,9 @@ const AdminRegister = ({
           name="pool"
           component={SelectInput.Field}
           placeholder="Pool"
-          options={pools.map(pool => ({ value: pool.id, label: pool.name }))}
+          options={pools
+            .map(pool => ({ value: pool.id, label: pool.name }))
+            .concat([{ value: 0, label: 'Venteliste' }])}
```
hanswilw: Maybe -1 is a more natural value than 0 when referring to a non existant pool? 0 is an indice in arrays etc – even though that have no effect here. I guess the option wont stick in the input if you use null or does not define it - right?

app/actions/EventActions.js
hanswilw: You may handle the value in the component here instead if you want to change it https://github.com/webkom/lego-webapp/blob/master/app/routes/events/components/EventAdministrate/index.js#L80

app/routes/events/components/EventAdministrate/AdminRegisterForm.js
```diff
@@ -63,7 +65,7 @@ function validateForm(data) {
     errors.reason = 'Forklaring er påkrevet';
   }
 
-  if (!data.pool) {
+  if (!data.pool && data.pool !== 0) {
```
hanswilw: -1

app/routes/events/components/EventAdministrate/AdminRegisterForm.js
```diff
@@ -38,7 +38,9 @@ const AdminRegister = ({
           component={SelectInput.Field}
           placeholder="Pool"
           label="Pool"
-          options={pools.map(pool => ({ value: pool.id, label: pool.name }))}
+          options={pools
+            .map(pool => ({ value: pool.id, label: pool.name }))
+            .concat([{ value: 0, label: 'Venteliste' }])}
```
hanswilw: -1

PR #843, opened 2017-10-26
Fix incrementing and decrementing beyond max cap

hanswilw:
Follow up on #830 

Fixes an error that occurs when no state exist in form as well.

### DIFF
--- a/app/reducers/events.js
+++ b/app/reducers/events.js
@@ -68,14 +68,16 @@ function mutateEvent(state: any, action: any) {
           [eventId]: {
             ...stateEvent,
             loading: false,
-            registrationCount: stateEvent.registrationCount + 1,
+            registrationCount: registration.pool
+              ? stateEvent.registrationCount + 1
+              : stateEvent.registrationCount,
             waitingRegistrations
           }
         }
       };
     }
     case Event.SOCKET_UNREGISTRATION.SUCCESS: {
-      const { eventId, activationTime } = action.meta;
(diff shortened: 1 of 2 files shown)


app/reducers/events.js
```diff
@@ -68,14 +68,16 @@ function mutateEvent(state: any, action: any) {
           [eventId]: {
             ...stateEvent,
             loading: false,
-            registrationCount: stateEvent.registrationCount + 1,
+            registrationCount: registration.pool
+              ? stateEvent.registrationCount + 1
+              : stateEvent.registrationCount,
```
ekmartin: why would the pool not exist if registrationCount exists?

hanswilw: When it is a socket registration to waiting list – the pool is not set.

app/reducers/forms.js
```diff
@@ -22,6 +22,7 @@ export default formReducer.plugin({
         };
       case Event.SOCKET_UNREGISTRATION.SUCCESS:
       case Event.SOCKET_REGISTRATION.SUCCESS: {
+        if (!state) return undefined;
```
ekmartin: `if (!state) return` should be the same if flow doesn't complain

hanswilw: Didnt complain, but better off writing it like that. Fixed.

app/reducers/events.js
ekmartin: alright

hanswilw:
Agreed, I'll create an issue for it.

PR #859, opened 2017-10-26
Content rendering

odinuge:


app/components/DisplayContent/index.js
```diff
@@ -14,12 +14,38 @@ type Props = {
   style?: Object
 };
 
-const isValidNode = () => true;
+const legalTags = [
```
larseen: Isnt `href` a legal tag? For Links

larseen:
Looks good, would it also be an idea to have some kind of consolidated typography styling thats used by the content renderer and editor, such that all content is styled similar?

ekmartin:
@larseen: wanna make a separate issue for that?

PR #865, opened 2017-10-26
Mark notifications as read after opening notification box

### LINKED ISSUE #832: Mark notifications as read when users open the notification box
Slightly opinionated, but I think having a mark all as read button is kind of unnecessary.

JonasBak:
Fixes #832
Hides unread count from notification icon when opening the notification box and marks all notifications as read after closing, removed "Marker alle som sett"
notification
notificationbox

### DIFF
--- a/app/components/HeaderNotifications/index.js
+++ b/app/components/HeaderNotifications/index.js
@@ -6,7 +6,6 @@ import Icon from '../Icon';
 import { activityRenderers } from '../Feed';
 import Time from 'app/components/Time';
 import styles from './HeaderNotifications.css';
-import { Link } from 'react-router';
 
 type Props = {
   notificationsData: Object,
@@ -37,7 +36,7 @@ const NotificationElement = ({
           styles.notification,
           !notification.read ? styles.unRead : null
         )}
-        onClick={() => markNotification(notification.id)}
+        onClick={() => this.props.markNotification(notification.id)}
       >
         <div className={styles.innerNotification}>
           <div className={styles.icon}>{renders.icon(notification)}</div>
@@ -58,7 +57,8 @@ const NotificationElement = ({
 
 export default class NotificationsDropdown extends Component<Props, State> {
   state: State = {
-    notificationsOpen: false
+    notificationsOpen: false,
+    hideUnreadCount: false
   };
 
   seeAll = () => {
@@ -100,12 +100,20 @@ export default class NotificationsDropdown extends Component<Props, State> {
         toggle={() =>
           this.setState(
             {
-              notificationsOpen: !this.state.notificationsOpen
+              notificationsOpen: !this.state.notificationsOpen,
+              hideUnreadCount: true
             },
-            () => (this.state.notificationsOpen ? fetchNotifications() : null)
+            () =>
+              this.state.notificationsOpen
+                ? fetchNotifications()
+                : this.props.markAllNotifications().then(this.fetch)
           )}
         triggerComponent={
-          <Icon.Badge name="notifications" size={30} badgeCount={unreadCount} />
+          <Icon.Badge
+            name="notifications"
+            size={30}
+            badgeCount={!this.state.hideUnreadCount && unreadCount}
+          />
         }
         contentClassName={styles.notifications}
       >
@@ -114,11 +122,6 @@ export default class NotificationsDropdown extends Component<Props, State> {
           <div style={{ width: '100%' }}>
             <div style={{ maxHeight: '300px', overflowY: 'overlay' }}>
               {this.renderNotifications(notifications)}
-            </div>
-            <div className={styles.seeAllWrapper}>
-              <Link onClick={this.seeAll} className={styles.seeAll}>
-                Marker alle som sett
-              </Link>
             </div>
           </div>
         ) : (
--- a/app/components/HeaderNotifications/index.js
+++ b/app/components/HeaderNotifications/index.js
@@ -12,7 +12,6 @@ type Props = {
   fetchNotifications: () => void,
   notifications: Array<Object>,
   markAllNotifications: () => Promise<void>,
-  markNotification: number => Promise<void>,
   fetchNotificationData: () => Promise<void>
 };
 
@@ -20,13 +19,7 @@ type State = {
   notificationsOpen: boolean
 };
 
-const NotificationElement = ({
-  notification,
-  markNotification
-}: {
-  notification: Object,
-  markNotification: number => void
-}) => {
+const NotificationElement = ({ notification }: { notification: Object }) => {
   const renders = activityRenderers[notification.verb];
 
   if (renders) {
@@ -36,7 +29,6 @@ const NotificationElement = ({
           styles.notification,
           !notification.read ? styles.unRead : null
         )}
-        onClick={() => this.props.markNotification(notification.id)}
       >
         <div className={styles.innerNotification}>
           <div className={styles.icon}>{renders.icon(notification)}</div>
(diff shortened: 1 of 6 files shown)


app/components/HeaderNotifications/index.js
```diff
@@ -37,7 +36,7 @@ const NotificationElement = ({
           styles.notification,
           !notification.read ? styles.unRead : null
         )}
-        onClick={() => markNotification(notification.id)}
+        onClick={() => this.props.markNotification(notification.id)}
```
ekmartin: do you still need this line?

app/components/HeaderNotifications/index.js
```diff
@@ -100,12 +100,20 @@ export default class NotificationsDropdown extends Component<Props, State> {
         toggle={() =>
           this.setState(
             {
-              notificationsOpen: !this.state.notificationsOpen
+              notificationsOpen: !this.state.notificationsOpen,
+              hideUnreadCount: true
```
ekmartin: I don't think you need to keep this in state? `markAllNotifications` should update the state in redux - if you want it to update right away (and e.g. not wait until the response comes back) you can add that manually to the `notificationsFeed` reducer on `NotificationsFeed.MARK_ALL.SUCCESS`

app/components/HeaderNotifications/index.js
```diff
@@ -100,12 +100,20 @@ export default class NotificationsDropdown extends Component<Props, State> {
...
             },
-            () => (this.state.notificationsOpen ? fetchNotifications() : null)
+            () =>
+              this.state.notificationsOpen
+                ? fetchNotifications()
+                : this.props.markAllNotifications().then(this.fetch)
```
ekmartin: Do you need to fetch the notifications again after marking them as read?

app/components/HeaderNotifications/index.js
```diff
@@ -115,11 +123,6 @@ export default class NotificationsDropdown extends Component<Props, State> {
             <div style={{ maxHeight: '300px', overflowY: 'overlay' }}>
               {this.renderNotifications(notifications)}
             </div>
-            <div className={styles.seeAllWrapper}>
-              <Link onClick={this.seeAll} className={styles.seeAll}>
```
ekmartin: Should also remove the `seeAll` function and styles from the CSS file, right?

ekmartin:
Agree with the change, just needs some small tuning :)

app/components/HeaderNotifications/index.js
JonasBak: No, and should I remove markNotification from props? It's unnecessary now anyways

app/components/HeaderNotifications/index.js
JonasBak: I'll do that :ok_hand:

app/components/HeaderNotifications/index.js
ekmartin: Yeah, you can remove it from everywhere (search for it in the codebase) - actions, action types, reducers etc.

JonasBak:
I'll try redoing it with your suggestions, I wanted the new notifications to still be stylized in the dropdown when opening it the first time, but hide the count from the icon once it was opened, and this was the easiest way I could think of

app/components/HeaderNotifications/index.js
```diff
@@ -80,11 +61,7 @@ export default class NotificationsDropdown extends Component<Props, State> {
     return (
       <div>
         {notifications.map((notification, i) => (
-          <NotificationElement
-            key={i}
-            notification={notification}
-            markNotification={this.markNotification}
-          />
+          <NotificationElement key={i} notification={notification} />
```
finnss: Using the index of an array as key has turned out to be dangerous. If an element is deleted, the elements are moved around so that a new element has the same key as the deleted one, which can lead to weird behaviour. Might not be an issue here as I don't know if notifications can be deleted, but it seems like a good plan to use something like `key={notifiaction.id}` instead just in case.

app/reducers/notificationsFeed.js
```diff
@@ -20,6 +20,8 @@ export default function notificationsFeed(
         ...state,
         ...action.payload
       };
+    case NotificationsFeed.MARK_ALL.SUCCESS:
+      return initialState;
```
finnss: :+1:

app/components/HeaderNotifications/index.js
JonasBak: Ok, I'll fix that

hanswilw: It is not necessarily better @finnss since the notification id might not be unique either. A better example would be something like
``` key={`notification-${i}`} ``` which would create unique keys like `notification-1` etc. (As long as nobody else creates the same string as key somewhere else)

Comment (ekmartin) [reply app/components/HeaderNotifications/index.js]: isn't the notification id from the database? o.o
and why is `notification-${i}` any different from just `i`? (you still have the same bug with sorting/deleting etc.)

pretty sure the id needs to be something that is unique only for the siblings of that dom element, not globally in the app

Comment (odinuge) [reply app/components/HeaderNotifications/index.js]: @hanswilw & @ekmartin 

Using the index from mapping/looping as the react-key is an anti pattern that creates strange bugs, that are hard to identify. Since it we never "remove" one notification it isn't "that" important, but when new notifications are added, using indexes will result in way more unnecessary rendering - aka. it only removes the warning. It will then create new elements in the bottom of the list, and moves the old elements further down.

Keys only have to be unique among their siblings, and they _don't_ need to be unique globally. 

I used quite a bit of time to track down #848, that was a result of using mapping indexes as keys. When deleting an element in the middle, it is the react-component in the bottom that is unmounted and removed, and the state-elements are popped upwards to fill the gap.

This is generally not a huge problem with other than performance when using stateless components, but when state is added to components - this __will__ create bugs.

Comment (hanswilw) [reply app/components/HeaderNotifications/index.js]: I bow, I was told otherwise at some point earlier – thats why I commented. Should've checked it out better myself though 😌

File: app/reducers/notificationsFeed.js (hunk 6 above)
Comment (ekmartin) [reply app/reducers/notificationsFeed.js]: Wouldn't you want the red background color from the unread events to disappear as well? While testing this locally it seemed like the counter got cleared, but the background color didn't disappear until the next server fetch.

Maybe you want to map over the notifications in `reducers/notifications.js` and set them all to read? Something like:

```js
    case Notifications.MARK_ALL.SUCCESS:
      return {
        ...state,
        items: state.items.map(item => ({ ...item, read: true }))
      };
```

finnss: I don't think that's what you would want, no? If so there would be no way to tell which notifications are new. I think it might be unintuitive if your notification bells tells you you have three new notifications, but when you open the dropdown you see a list of all notifications you've ever gotten that all look alike. Removing the color on the next server fetch seems as good a way as any to determine when a notification shouldn't be "new" anymore.

ekmartin: I thought we were marking them as read when the notification box is closed - not when we open it? I agree that they should be read when you open, but if they're marked as read when you close the box that would work right?

finnss: Hm, removing the color when the box is closed could work. Facebook might not be the ideal to strive for, but just for reference they keep notifications marked with a different background color for a while even after you open and close the notification list. In any case, I believe `MARK_ALL_SUCCESS` happens when the list is opened, not closed, as this is what happens when the list is opened or closed:

```
<Dropdown
        show={this.state.notificationsOpen}
        toggle={() =>
          this.setState(
            {
              notificationsOpen: !this.state.notificationsOpen
            },
            () =>
              this.state.notificationsOpen
                ? fetchNotifications()
                : this.props.markAllNotifications()
          )}
```

So if we want to reset the background color after closing the dropdown we could look at the reducer that handles `fetchNotifications()` or some such. I don't really have any strong feelings for whether that's the desired functionality or not.

JonasBak: The notifications are marked as read when closing the box. It was intentional to keep the background color even though the counter is cleared, but I can see that it might make sense to make the color disappear. I also see that if we decide to only make the counter disappear, I can make the code a bit cleaner by calling `markAllNotificaitons` when the box opens, then I won't need to check `badgeCount={this.state.notificationsOpen ? 0 : unreadCount}`

PR #867, opened 2017-10-26
Add functionality to delete files in semester statuses

### LINKED ISSUE #771: Remove files from semester statuses in BDB
Once you've uploaded a file there is currently no way to remove it.

finnss:
skjermbilde 2017-10-26 kl 23 37 49

Fixes #771

### DIFF
--- a/app/routes/bdb/components/BdbDetail.js
+++ b/app/routes/bdb/components/BdbDetail.js
@@ -22,6 +22,7 @@ import type {
   SemesterStatusEntity
 } from 'app/reducers/companies';
 import type { CompanySemesterEntity } from 'app/reducers/companySemesters';
+import type { UserEntity } from 'app/reducers/users';
 import Button from 'app/components/Button';
 import type { CompanySemesterContactedStatus } from 'app/models';
 import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
@@ -133,6 +134,12 @@ export default class BdbDetail extends Component<Props, State> {
     };
 
     return editSemesterStatus(sendableSemester, { detail: true });
+  };
+
+  studentContactLink = (studentContact?: UserEntity): string => {
+    return studentContact
+      ? 'abakus.no/users/' + String(studentContact.username)
+      : '';
   };
 
   render() {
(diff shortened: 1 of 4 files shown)


app/routes/bdb/components/BdbDetail.js
```diff
@@ -275,6 +292,10 @@ export default class BdbDetail extends Component<Props, State> {
                 company.studentContact.fullName) ||
                 '-'}`}
               meta="Studentkontakt"
+              link={`${String(
```
hanswilw: Maybe create a util function for this? Gets a bit messy with template literals, String()'s and the && shortcutting to join the strings.

app/routes/bdb/components/SemesterStatusDetail.js
```diff
@@ -73,6 +78,34 @@ export default class SemesterStatusDetail extends Component<Props, State> {
     if (!semesterStatus) return <LoadingIndicator />;
 
     const humanReadableSemester = this.semesterToHumanReadable();
+
+    const renderFile = (type: string) => {
```
hanswilw: This will recreate the renderFile function on every render, it is better to put it out as a class function and passing the semesterStatus as an argument instead! Maybe split it up a bit and make it a bit more readable. 

```
const fileName = this.fileNameToShow(semesterStatus[type], ...);
const showDelete = this.state.editing && semesterStatus[type];
const showUpload = !showDelete && this.state.editing;

return (
  ...
  <span>
    {showDelete && <span className={...}><span>{fileName}</span>... }
    {showUpload && <UploadButton />}
  </span>
```
Maybe even make the whole renderFile thing as a component, to make the `{['contract', 'statistics', 'evaluation'].map(type => (` more readable?

finnss: They're mutually exclusive though, `showDelete` and `showUpload` are never true at the same time. Doesn't that make your solution confusing? If I saw that I wouldn't assume there was an if/else relationship going on.

finnss: I moved it to its own component though, I think that makes sense :+1:

hanswilw: Didn't dive completely into when you wanted to show it and not - was just a suggestion for readability 👍🏻

PR #869, opened 2017-10-26
Add support for reversed sorting of comments

odinuge:
The prop 'newOnTop' for CommentView allows comments to be sorted (only
the first level in the tree) with the newest comments on top. The
comment form will then also be on the top.

screenshot from 2017-10-27 01-01-10

### DIFF
--- a/app/components/Comments/CommentView.js
+++ b/app/components/Comments/CommentView.js
@@ -4,6 +4,7 @@ import React from 'react';
 import { generateTreeStructure } from 'app/utils';
 import LoadingIndicator from 'app/components/LoadingIndicator';
 import CommentForm from 'app/components/CommentForm';
+import { Flex } from 'app/components/Layout';
 import type { CommentEntity } from 'app/reducers/comments';
 import type { UserEntity } from 'app/reducers/users';
 import CommentTree from './CommentTree';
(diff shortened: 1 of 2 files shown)


edvardvb:
This does not sort replies, only the top-level comments, right?

app/components/Comments/CommentView.js
```diff
@@ -26,28 +31,38 @@ const CommentView = (props: Props) => {
...
+      <Title displayTitle={displayTitle} />
+      <Flex
+        style={{
+          flexDirection: newOnTop ? 'column-reverse' : 'column'
+        }}
+      >
+        <LoadingIndicator loading={!comments}>
+          {comments && (
+            <CommentTree
+              comments={newOnTop ? tree.reverse() : tree}
```
odinuge: Yeah @edvardvb. It reverses the first layer of comments in the generated tree, not the rest of them.

ekmartin:
Couldn't we just always return comments sorted like this from the backend? Newest first seems reasonable to me - i.e. do we need to support both options?

finnss:
We can always remove support for oldest first later. I'll get this merged so we have the functionality for sorting by newest first at all.

PR #872, opened 2017-10-26
Fix issue where callAPI promise isn't fulfilled

### LINKED ISSUE #775: Bdb doesn't always trigger a fetchAll() when navigating to the list page with more than zero companies in the redux state

### LINKED ISSUE #847: BDB does not show full company information when navigating from the BDB list
Example:
- Visit https://abakus.no/bdb

- Click on a Company
- Information is missing

- Refresh the company page to get the detailed view

odinuge:
When callAPI used a cached value, the returned promise never fulfilled,
and therefore all the '.then' functions chained together with the
promise was never executed. This fixes the issue, by returning a promise
that when executed will resolve with the cached value.

Since the action doesn't contain a promise, the `promiseMiddleware` doesn't do it's magic - and therefore the promise isn't fulfilled via `resolve`.

And isn't it closer to `cachedResponse`, instead of `cachedRequest`??

This fixes #847 and fixes #775 :smile:

### DIFF
--- a/app/actions/callAPI.js
+++ b/app/actions/callAPI.js
@@ -105,7 +105,7 @@ export default function callAPI({
     if (shouldUseCache) {
       const cachedRequest = getCachedRequest(state, endpoint, cacheSeconds);
       if (cachedRequest) {
-        return new Promise(() => dispatch(cachedRequest));
+        return Promise.resolve(dispatch(cachedRequest));
       }
     }
 


odinuge:
By design this only runs the action `SUCCESS` (eg. `Company.FETCH_SEMESTERS.SUCCESS`), not the `BEGIN`. This is feels odd, and makes debugging kinda hard.

odinuge:
We can eg. do:
```jsx
    const state = getState();
    if (shouldUseCache) {
      const cachedRequest = getCachedRequest(state, endpoint, cacheSeconds);
      if (cachedRequest) {
        return new Promise(() =>
          dispatch({ ...cachedRequest, types, promise: Promise.resolve() })
        );
      }
    }

```

app/actions/callAPI.js
```diff
@@ -105,7 +105,7 @@ export default function callAPI({
     if (shouldUseCache) {
       const cachedRequest = getCachedRequest(state, endpoint, cacheSeconds);
       if (cachedRequest) {
-        return new Promise(() => dispatch(cachedRequest));
+        return new Promise(resolve => resolve(dispatch(cachedRequest)));
```
ekmartin: isn't this the same as `return Promise.resolve(dispatch(cachedRequest))`?

odinuge: Yeah.. :smile:

odinuge:
Merging this now. We can look at the behavior later.

PR #875, opened 2017-10-27
Proposal function for adding meta tags to routes

larseen:


odinuge:
Awesome idea @larseen. Putting the metadata in the routes directly makes it a lot cleaner than adding a `<Helmet>`-tag in the render function! 😄

larseen:
Maybe find a better way of doing this?
I would like to have it in the route instead of having it in the components, however it can be quite overwhelming when there are many tags, as the example shows.

ekmartin:
@larseen lets get this merged and improve it later? Better than not having meta tags at all - looks pretty weird when events are shared on FB now.

app/utils/helmet.js
```diff
@@ -0,0 +1,33 @@
...
+  ) => Array<{
+    property?: string,
+    content?: string,
+    element?: string,
+    children?: string,
+    rel?: string,
+    href?: string
+  }>
+) {
+  // $FlowFixMe React.Node in >= 0.53
```
ekmartin: can remove this flowfixme and use React.Node now probably

odinuge:
Events are still only visible for authenticated users (students?) so it doesn't get that much better. Adding a nice picture and better tags will still improve it a lot!

ekmartin:
Ugh that's a good point. Should fix that soon :p Could fix the description in the meantime so it says something generic about Abakus instead of the lego info:
image

odinuge:
We should also upgrade `react-helment`. We currently use `helment.rewind()` in ssr, https://github.com/webkom/lego-webapp/blob/3f30e2dbdd9ffdf4f607e3581cb0c374a55d1860/server/render.js#L69
and it looks like it isn't a good approach.. There are some issues in the main repo: https://github.com/nfl/react-helmet/issues/216. I cannot understand how that approach works with async requests. Looks like this works the same way as `react-cookie`, by saying something like `as long as you do these two things together; it will be no problem`.

It should be no problem to stress test it, to see how it behaves in an async context. But someone still has to do it... :smile:

odinuge:
Since events are just a few days/hours away from being public, ill merge and use this to add better seo/metatags. Or do anyone have a better idea now?

Current look:
screenshot from 2018-02-06 12-37-49

odinuge:
I did a rebase, and cleaned up the flow annotations. I think this solution is ok. We only need this much meta-information in public content. For private content like meetings etc., just a few is enough.

We need it for:
- Events
- Flatpages
- Articles (Done in this PR)
- Commitees (Not yet public)
- Interestgroups (Not yet public)
- Undergrupper (Not yet public)

PR #877, opened 2017-10-27
Add timeout to server side rendering

odinuge:
This add a 15sec timeout to prepare. Sometimes react-prepare halts
and never retuns a response, resulting in a browser timeout. This will
stop server side rendering when the timeout occures, and return an
normal non SSR response.


server/render.js
```diff
@@ -21,6 +21,21 @@ require('../app/assets/icon-256x256.png');
 require('../app/assets/icon-384x384.png');
 require('../app/assets/icon-512x512.png');
 
+const TimeoutError = new Error(
+  'React prepare timeout when server side rendering.'
+);
+
+function prepareWithTimeout(app, ms = 15 * 1000) {
```
ekmartin: Might as well move ms to a constant here. Also, 15 seconds is probably way too high? Maybe something like 3-5 seconds is more reasonable.

I guess the optimal thing would be if we could figure out why prepare times out, and somehow have it bubble up the error instead?

odinuge: +1 on the constant thing.

Totally agree about the `3-5sec` too, but we have had some discussions regarding timeouts - and `fetchJSON` currently have a `15sec` timeout...

The latency between the server-side renderer and the backend is more or less equal to zero, and that (up to 15sec) slow queries shouln't be run on the server.

odinuge: Will fix

odinuge: And yes, we should do some debugging. But when chaining 2-3-4 actions, with 15sec timeouts - the browser will timeout, and the user will be unable to open the page at all...

server/render.js
```diff
@@ -55,25 +72,25 @@ function render(req, res, next) {
...
-    const respond = () => {
+    const respond = (timeout = false) => {
       const body = renderToString(app);
 
       return res.send(
         renderPage({
-          body,
-          state: store.getState(),
+          body: timeout ? body : '',
+          state: timeout ? store.getState() : {},
```
finnss: What kind of result does this eventually return to the user?

odinuge: When the request is timed out, it returns the same as if SSR was disabled, and that is handled here:
https://github.com/webkom/lego-webapp/blob/2b24f8b9d19cfe77426d3dbd99c13bbcbb84b518/app/index.js#L29

Looks like the logic was broken in my last change :laughing:

server/render.js
```diff
@@ -21,6 +21,23 @@ require('../app/assets/icon-256x256.png');
 require('../app/assets/icon-384x384.png');
 require('../app/assets/icon-512x512.png');
 
+const serverSideTimeoutInMs = 5000;
```
eirsyl: The ingress times out after 5s, maye set this to 4000?

odinuge:
Lets test this in staging

PR #880, opened 2017-10-27
Bdb improvements

### LINKED ISSUE #876: [BDB] Show most recent events first
image

Events from 7 years ago might not be very relevant here :)

finnss:
- Order events in bdb by newest first (Fixes #876)
- Scroll to top when navigating to detail view
- Use `prepare` in all the routes
- Autonavigate when search yields only one result (type Bekk -> only one result -> press enter -> get navigated to Bekk's detail view)
- use `fetchAdmin` and `fetchAllAdmin` in all bdb routes instead of `fetch` and `fetchAll`

### DIFF
--- a/app/routes/bdb/BdbRoute.js
+++ b/app/routes/bdb/BdbRoute.js
@@ -13,6 +13,7 @@ import { selectCompanies } from 'app/reducers/companies';
 import { LoginPage } from 'app/components/LoginForm';
 import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
 import { selectCompanySemesters } from 'app/reducers/companySemesters';
+import { push } from 'react-router-redux';
 
 const loadData = (props, dispatch) =>
   dispatch(fetchSemesters()).then(() => dispatch(fetchAllAdmin()));
@@ -24,11 +25,16 @@ const mapStateToProps = (state, props) => ({
   fetching: state.companies.fetching
 });
 
-const mapDispatchToProps = {
-  editSemesterStatus,
-  addSemesterStatus,
-  addSemester
-};
+function mapDispatchToProps(dispatch) {
+  return {
+    editSemesterStatus: (semesterStatus, options) =>
+      dispatch(editSemesterStatus(semesterStatus, options)),
+    addSemesterStatus: (semesterStatus, options) =>
+      dispatch(addSemesterStatus(semesterStatus, options)),
+    addSemester: semester => dispatch(addSemester(semester)),
+    push: uri => dispatch(push(uri))
+  };
+}
 
 export default compose(
   replaceUnlessLoggedIn(LoginPage),
--- a/app/routes/bdb/BdbRoute.js
+++ b/app/routes/bdb/BdbRoute.js
@@ -25,16 +25,12 @@ const mapStateToProps = (state, props) => ({
   fetching: state.companies.fetching
 });
 
-function mapDispatchToProps(dispatch) {
-  return {
-    editSemesterStatus: (semesterStatus, options) =>
-      dispatch(editSemesterStatus(semesterStatus, options)),
-    addSemesterStatus: (semesterStatus, options) =>
-      dispatch(addSemesterStatus(semesterStatus, options)),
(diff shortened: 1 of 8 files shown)


app/routes/bdb/components/BdbDetail.js
```diff
@@ -53,6 +53,10 @@ export default class BdbDetail extends Component<Props, State> {
     eventsToDisplay: 3
   };
 
+  componentDidMount() {
+    window.scrollTo(0, 0);
+  }
```
finnss: Is doing this fine? While searching for a solution it seemed like a lot of projects added this functionality to Router so that it happened for all navigations, but I don't know if we want that.

finnss:
Not sure how to make these tests pass

skjermbilde 2017-10-27 kl 18 22 33

app/routes/bdb/BdbRoute.js
```diff
@@ -24,11 +25,16 @@ const mapStateToProps = (state, props) => ({
   fetching: state.companies.fetching
 });
 
-const mapDispatchToProps = {
-  editSemesterStatus,
-  addSemesterStatus,
-  addSemester
-};
+function mapDispatchToProps(dispatch) {
+  return {
```
odinuge: Why are you doing this?

app/routes/bdb/components/BdbDetail.js
```diff
@@ -181,7 +185,7 @@ export default class BdbDetail extends Component<Props, State> {
     const events =
       companyEvents &&
       companyEvents
-        .sort((a, b) => new Date(b.startTime) - new Date(a.startTime))
+        .sort((a, b) => Date.parse(b.startTime) - Date.parse(a.startTime))
```
odinuge: How does this help when you get the first events for a certain company, not the newest ones?

odinuge:
Looks like this works:
```jsx
export const selectCompanyContactById = createSelector(
  (state, props) => selectCompanyById(state, props),
  (state, props) => props.companyContactId,
```
No idea why tho

app/routes/bdb/BdbRoute.js
finnss: To get `push`, which needs to be dispatched, right?

finnss: Maybe I could have simply added `push` to the simpler version? Didn't even strike me as an opportunity

app/routes/bdb/components/BdbDetail.js
finnss: With pagination you mean? Isn't events sorted backend so that you get the 30 most recent events? The alternative wouldn't make much sense

odinuge: When looking at BEKK in prod, I get the "first" 30 events from BEKK, not the 30 most recent ones. Looks like the backend sorts by pk, not by date.

app/routes/bdb/components/BdbPage.js
```diff
@@ -178,13 +179,24 @@ export default class BdbPage extends Component<Props, State> {
       this.state.startSem
     );
 
+    const filteredCompanies = this.filterCompanies(sortedCompanies);
+
+    const searchKeyPress = (event: Object) => {
+      if (event.key === 'Enter' && filteredCompanies.length === 1) {
+        push(`/bdb/${filteredCompanies[0].id}`);
```
odinuge: Smart!

app/routes/bdb/components/BdbDetail.js
finnss: Surely the events from the generic `/events` endpoint are sorted by date with the most recent first, right? Otherwise the front page and the calendar wouldn't do their job at all. So I guess that means the queryset isn't sorted the same way when you add a filter..?

finnss:
Haha that flow workaround worked nicely @odinuge. Bad solutions aren't bad if they work 🤷‍♀️

app/routes/bdb/components/BdbDetail.js
odinuge: Hmm. Looks like it works well here, but imo it should not be used everywhere.. It is nice to click back, and retain your scrolling position.

PR #886, opened 2017-10-28
Fix issues in pages for unauthorized users

odinuge:
This also moves a lot of the logic to selectors

### DIFF
--- a/app/reducers/pages.js
+++ b/app/reducers/pages.js
@@ -4,6 +4,8 @@ import { createSelector } from 'reselect';
 import { Page } from '../actions/ActionTypes';
 import createEntityReducer from 'app/utils/createEntityReducer';
 import { selectGroupsWithType } from './groups';
+import { selectGroup } from 'app/reducers/groups';
+import { selectMembershipsForGroup } from 'app/reducers/memberships';
 
 export type PageEntity = {
   id: number,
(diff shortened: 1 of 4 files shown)


app/reducers/pages.js
```diff
@@ -68,3 +70,55 @@ export const selectGroupsForHierarchy = createSelector(
...
+    )
+);
+
+export const selectFlatpageForPages = createSelector(
+  selectPageBySlug,
+  (state, props) => props.pageSlug,
+  (selectedPage, pageSlug) => ({
+    selectedPage,
+    selectedPageInfo: {
+      actionGrant: selectedPage.actionGrant || [],
```
ekmartin: shouldn't we always at least return an empty array from the backend?

app/reducers/pages.js
```diff
@@ -68,3 +70,55 @@ export const selectGroupsForHierarchy = createSelector(
...
+      title: selectedPage.title,
+      editUrl: `/pages/info/${pageSlug}/edit`
+    }
+  })
+);
+
+export const selectComiteeForPages = createSelector(
```
ekmartin: committee ;)

app/reducers/pages.js
```diff
@@ -68,3 +70,55 @@ export const selectGroupsForHierarchy = createSelector(
...
+  (state, props) =>
+    selectMembershipsForGroup(state, { groupId: Number(props.pageSlug) }),
+  (group, memberships) => {
+    const selectedPageInfo = group && {
+      actionGrant: group.actionGrant || [],
+      title: group.name,
+      editUrl: `/admin/groups/${group.id}/settings`
+    };
+    return {
+      selectedPage: group && { ...group, memberships },
```
ekmartin: this will make selectedPage and selectedPageInfo be `false` if group doesn't exist - which seems a bit weird?

app/reducers/pages.js
odinuge: lol. copypase much

app/reducers/pages.js
odinuge: Good point

app/reducers/pages.js
ekmartin: ah, didn't realize you just moved this logic from another file - feel free to look at in another PR

odinuge: But i wrote the logic (the night before launch) the first time, so i should fix it... :laughing:

odinuge: But after some testing, it turns out that 
```js
const group = undefined;
group && {...group, memberships} // Evaluates to undefined, not false
```

app/routes/pages/components/PageDetail.js
```diff
@@ -95,8 +97,11 @@ function PageDetail<T: Object>({
         )}
       </NavigationTab>
       <Flex className={styles.page} wrap>
-        <PageRenderer page={selectedPage} />
-
+        {isComplete ? (
```
odinuge: Did this because `LoadingIndicator wrapps it's content children in a div, and I wan't no div.

PR #888, opened 2017-10-28
Hide EventListGroup if there are no events in group

### LINKED ISSUE #750: Hide "Denne uken" when there are no upcoming events
https://abakus.no/events

JonasBak:
Fix #750
Only display the EventListGroup if there are events to display
hide

### DIFF
--- a/app/routes/events/components/EventList.js
+++ b/app/routes/events/components/EventList.js
@@ -15,6 +15,7 @@ import styles from './EventList.css';
 import EventFooter from './EventFooter';
 import { Flex } from 'app/components/Layout';
 import EmptyState from 'app/components/EmptyState';
+import { isEmpty } from 'lodash';
 
 // Kinda works
 function groupEvents(events) {
@@ -101,12 +102,12 @@ type EventListGroupProps = {
 };
 
 function EventListGroup({ name, events = [] }: EventListGroupProps) {
-  return events.length ? (
+  return isEmpty(events) ? null : (
     <div className={styles.eventGroup}>
       <h2 className={styles.heading}>{name}</h2>
       {events.map((event, i) => <EventItem key={i} event={event} />)}
     </div>
-  ) : null;
+  );
 }
 
 type EventListProps = {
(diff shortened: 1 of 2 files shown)


app/routes/events/components/EventList.js
```diff
@@ -123,11 +123,19 @@ const EventList = (props: EventListProps) => {
     <div className={styles.root}>
       <Helmet title="Arrangementer" />
       <Toolbar actionGrant={props.actionGrant} />
-      <EventListGroup name="Denne uken" events={events.currentWeek} />
+      {events.currentWeek && (
+        <EventListGroup name="Denne uken" events={events.currentWeek} />
```
odinuge: If you check inside the `<EventListGroup` component, you can do the same there instead (return `null`). That makes the code cleaner, and requires less code.

app/routes/events/components/EventList.js
```diff
@@ -123,11 +123,19 @@ const EventList = (props: EventListProps) => {
...
-      <EventListGroup name="Neste uke" events={events.nextWeek} />
+      {events.nextWeek && (
+        <EventListGroup name="Neste uke" events={events.nextWeek} />
+      )}
 
-      <EventListGroup name="Senere" events={events.later} />
+      {events.later && <EventListGroup name="Senere" events={events.later} />}
+
+      {!props.events && (
+        <h2 className={styles.heading}>Ingen kommende arrangementer</h2>
```
odinuge: Here you can use the `EmptyState`  component. Example here: https://github.com/webkom/lego-webapp/blob/86f6e649ed68e8b7d15e18b7a2b2a1ca8d870f4e/app/routes/overview/components/Feed.js#L56

app/routes/events/components/EventList.js
JonasBak: Smart :+1:

app/routes/events/components/EventList.js
```diff
@@ -128,6 +129,12 @@ const EventList = (props: EventListProps) => {
       <EventListGroup name="Neste uke" events={events.nextWeek} />
 
       <EventListGroup name="Senere" events={events.later} />
+
+      {!props.events && (
```
odinuge: Since `props.events` is always an Array of events, it will always evaluate to `true` - `Boolean([]) == true`. There are many ways to check if an array is empty, but using [isEmpty](https://lodash.com/docs/4.17.4#isEmpty) from lodash (javascript lib) is a nice and explicit way to do it. 
 
```jsx
import { isEmpty } from 'lodash';
[...]
{ isEmpty(props.events) && <EmptyState .... /> }
```

JonasBak: Yes, I found out when changing the EventListGroup, but forgot to change it here, thanks :+1:

Issue #848, opened 2017-10-26
When deleting a single semester status, multiple statuses get deleted in descending order


kotlarz:
When testing, 3 instead of 1 statuses were deleted when deleting a specific semester status.

finnss:
How to reproduce?

kotlarz:
Delete a semester status  from a company in bdb

finnss:
well duh << doesn't happen when I try that, got anything more specific? I tried Create Company -> add tons of semester statuses -> start deleting, which seems to work as intended.

odinuge:
Looks like it is a strange bug. I am able to reproduce it @finnss. When clicking "delete", the semsterStatus is deleted, but the modal is still visable, and clicking again deletes the next one. We should autoclose the modal when clicking delete too (Using a promise.then(), or just when clicked...).

finnss:
That's more of a problem with the modal, right? There used to be a bug with deleting semester statuses before the confirmation modal was even made, but I think I fixed it. Wondering if that's what @kotlarz is experiencing.

kotlarz:
I can't recall 100% what happen, but something among the lines of what @odinuge described sounds about right. Let me know if you fix and push anything to staging, I will try to reproduce it then @finnss.

odinuge:
This is a hell of a bug :laughing: - And it is not the modal's fault!

This is basically the key to the whole bug.
https://github.com/webkom/lego-webapp/blob/7887524df397cb06d5740c044c428cafec3523b1/app/routes/bdb/components/BdbDetail.js#L142

When deleting a semesterstatus, the status is removed from the semester and the global state. On the next render, the "next" `Semesterstatus` is rendered in a `SemesterStatusComponent` with the __same__ `key` as the deleted one. Therefore it is using the same component as the deleted one, and it is the bottom component in the list that is removed. And since we had a delete modal open for the semester we just deleted, it will still be open, with the new data it got in the next render...

When thinking about it, using indexes as keys is more or less just a way to remove the warning...
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

React __0__ - BugBoys __1__

finnss:
Hahah really? Would never have guessed, I'll have to go through other places I might have used index as key. I'll get this fixed with a PR soon-ish. Good detective work @odinuge 👍

odinuge:
I have fixed quite a few bits in the Modal PR, so you can start by looking
at that :+1:

[...]

PR #889, opened 2017-10-28
Make confirm modal take promises instead of funcs

odinuge:
This adds autoClose functionality, useful when the route isn't changed
in the action executed. The closing is executed when the given promises
resolve.

### DIFF
--- a/app/actions/CompanyActions.js
+++ b/app/actions/CompanyActions.js
@@ -228,27 +228,18 @@ export function editSemesterStatus(
   };
 }
 
-export function deleteSemesterStatus(
-  companyId: number,
-  semesterId: number
-): Thunk<*> {
-  return dispatch => {
-    return dispatch(
-      callAPI({
-        types: Company.DELETE_SEMESTER_STATUS,
-        endpoint: `/companies/${companyId}/semester-statuses/${semesterId}/`,
-        method: 'DELETE',
-        meta: {
-          companyId,
-          semesterId,
-          errorMessage: 'Sletting av semesterstatus feilet'
-        }
-      })
-    ).then(() => {
-      dispatch(addNotification({ message: 'Semester Status slettet' }));
-      dispatch(push(`/bdb/${companyId}/`));
-    });
-  };
+export function deleteSemesterStatus(companyId: number, semesterId: number) {
+  return callAPI({
+    types: Company.DELETE_SEMESTER_STATUS,
+    endpoint: `/companies/${companyId}/semester-statuses/${semesterId}/`,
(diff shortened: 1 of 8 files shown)


odinuge:
This should fix #848.

@finnss i also added the "name" of the semester the user is about to delete :sunglasses:

app/components/Modal/ConfirmModal.js
```diff
@@ -6,8 +6,9 @@ import Modal from 'app/components/Modal';
 import Button from '../Button';
 
 type ConfirmModalProps = {
-  onConfirm: () => any,
-  onCancel?: () => any,
+  onConfirm?: () => Promise<*>,
+  onCancel?: () => Promise<*>,
+  autoClose?: boolean,
```
odinuge: Why...?

app/actions/CompanyActions.js
```diff
@@ -328,24 +319,18 @@ export function editCompanyContact({
...
+  return callAPI({
+    types: Company.DELETE_COMPANY_CONTACT,
+    endpoint: `/companies/${companyId}/company-contacts/${companyContactId}/`,
+    method: 'DELETE',
+    meta: {
+      companyId,
+      companyContactId,
+      errorMessage: 'Sletting av bedriftskontakt feilet',
+      successMessage: 'Bedriftskontakt slettet'
+    }
```
finnss: Same here

app/actions/CompanyActions.js
```diff
@@ -228,27 +228,18 @@ export function editSemesterStatus(
...
+    types: Company.DELETE_SEMESTER_STATUS,
+    endpoint: `/companies/${companyId}/semester-statuses/${semesterId}/`,
+    method: 'DELETE',
+    meta: {
+      companyId,
+      semesterId,
+      errorMessage: 'Sletting av semesterstatus feilet',
+      successMessage: 'Semesterstatus slettet'
+    }
+  });
```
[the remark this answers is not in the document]
odinuge: Didn't see the reason for the push, becuse the only place you used the action was the same route as the push. It is also cleaner to have the push in the component instead of the actions, becuase it forces you to make one action for each place you are supposed to do a "thing"


And since `successMessage` is set (I set it), there will be a notification when the action succeeds.

odinuge:
I have noticed that the modal moves the content to the bottom... I suppose that is a general modal issue, not only the ConfirmModal. `bottom: 0` or something?

finnss:
Yeah true, I guess that oughta be looked at more generally. I'll set up this branch locally and click around a bit, then merge if it seems good :+1:

PR #895, opened 2017-10-30
Add initial support for error handling in confirm

odinuge:


app/components/Modal/ConfirmModal.js
```diff
@@ -91,46 +120,61 @@ export default function withModal<Props>(
...
-const ChildrenWithProps = ({ children, ...restProps }: { children: Node }) => (
-  <div>
-    {React.Children.map(children, child =>
-      React.cloneElement(child, { ...restProps })
-    )}
-  </div>
-);
+const ChildrenWithProps = ({ children, ...restProps }: { children: Node }) =>
+  React.Children.map(children, child =>
+    React.cloneElement(child, { ...restProps })
```
ekmartin: doesn't each child need a key when you remove the div?

app/components/Modal/ConfirmModal.js
```diff
@@ -91,46 +120,61 @@ export default function withModal<Props>(
...
+                this.stopWorking();
+                this.hideModal();
+                this.resetError();
+                return result;
+              }
+            : result => result;
+
+          const onError = error => {
+            this.stopWorking();
+            const errorMessage =
```
ekmartin: we should probably make a `getErrorMessage` helper later

app/components/Modal/ConfirmModal.js
```diff
@@ -31,20 +34,26 @@ export const ConfirmModal = ({
...
+        <p style={{ color: 'red' }}>{errorMessage} </p>
       </div>
     </div>
   </div>
 );
 
 type State = {
   modalVisible: boolean,
-  working: boolean
+  working: boolean,
```
ekmartin: Probably something for a separate PR, but could we make stuff like `working` use cancellable promises instead, and cancel promises when the component unmounts instead of having to keep track of a separate bool? 

Something along the lines of https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

app/components/Modal/ConfirmModal.js
odinuge: Yeah, that might be true..

odinuge: > key and ref from the original element will be preserved
- Docs (https://reactjs.org/docs/react-api.html#createelement)

app/components/Modal/ConfirmModal.js
odinuge: Hmm...

I guess that would be a bit overkill for this... We know when the modal will unmount (due to redirect etc), and when the modal have to do it by it self, and we can set `closeOnConfirm` and `closeOnCancel` according to that.

The best would be to use modals like we currently use Toasts, but that is way out of scope now..

odinuge:
yolo

PR #899, opened 2017-10-30
Add bekk logo to the footer

finnss:
skjermbilde 2017-10-30 kl 21 22 39


edvardvb:
LGTM. It's separated from the quotes, and it's better to just put it in there for now than to get more complaints from Bedkom. If they want it somewhere else, we can deal with that later.

PR #903, opened 2017-11-01
Rewrite pinned item to allow for articles as well

edvardvb:
Follows the same structure as `OverviewItem`, which already works for both types of objects

### DIFF
--- a/app/routes/overview/components/Overview.js
+++ b/app/routes/overview/components/Overview.js
@@ -13,12 +13,15 @@ import Feed from './Feed';
 import CompactEvents from './CompactEvents';
 import { EVENT_TYPE_TO_STRING, colorForEvent } from 'app/routes/events/utils';
 import Button from 'app/components/Button';
+import type { Event, Article } from 'app/models';
 
 const TITLE_MAX_LENGTH = 50;
 const DESCRIPTION_MAX_LENGTH = 140;
 const IMAGE_HEIGHT = 192;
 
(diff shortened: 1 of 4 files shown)


app/routes/overview/components/Overview.js
```diff
@@ -13,59 +13,59 @@ import Feed from './Feed';
...
 import Button from 'app/components/Button';
-import type { Event } from 'app/models';
 
 const TITLE_MAX_LENGTH = 50;
 const DESCRIPTION_MAX_LENGTH = 140;
 const IMAGE_HEIGHT = 192;
 
-function PrimaryItem({ event }: { event: ?Event }) {
-  if (!event) {
+function PrimaryItem({ item }: { item: ?Object }) {
```
hanswilw: Better to use a flow type that accepts Event or Article. `Event | Article` instead of Object.

app/routes/overview/components/Overview.js
```diff
@@ -13,59 +13,59 @@ import Feed from './Feed';
...
           />
           <div className={styles.pinnedHeading}>
-            <h2 className={styles.itemTitle}>Ingen arrangementer</h2>
+            <h2 className={styles.itemTitle}>Ingen oppslag</h2>
           </div>
         </Flex>
       </Flex>
     );
   }
+  const base = item.eventType ? 'events' : 'articles';
```
hanswilw: Maybe better to have a small helperfunction, that accepts different items, outside the render and returns the complete link?

PR #922, opened 2017-11-02
Add (forgotten password) and (create new user) to every page

### LINKED ISSUE #851: Put forgot password and register on every login screen
Could only find the forgot password on the homepage login right now. Took up some of my time finding it. 
screenshot_20171026-141242

SmithPeder:
Fixes #851 
Now you can press the profile icon on every page and be able to create a new user or reset your password. 
screenshot from 2017-11-02 21-05-23
screenshot from 2017-11-02 21-05-47
screenshot from 2017-11-02 21-05-58
Small device
screenshot from 2017-11-02 21-06-37

### DIFF
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -109,11 +109,9 @@ class Header extends Component<Props, State> {
 
   render() {
     const { loggedIn } = this.props;
-
-    const { registerUser, forgotPassword, mode } = this.state;
-
+    const isLogin = this.state.mode === 'login';
     let title, form;
-    switch (mode) {
+    switch (this.state.mode) {
       case 'login':
         title = 'Logg inn';
         form = <LoginForm />;
@@ -223,7 +221,7 @@ class Header extends Component<Props, State> {
                       style={{ whiteSpace: 'nowrap' }}
                     >
                       {title}
-                      {!(registerUser || forgotPassword) && (
+                      {isLogin && (
                         <div>
                           <button
                             onClick={this.toggleForgotPassword}
@@ -240,7 +238,7 @@ class Header extends Component<Props, State> {
                           </button>
                         </div>
                       )}
-                      {(registerUser || forgotPassword) && (
+                      {!isLogin && (
                         <button
                           onClick={this.toggleBack}
                           className={styles.toggleButton}
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -39,8 +39,7 @@ type Props = {
 type State = {
   accountOpen: boolean,
   shake: boolean,
-  forgotPassword: boolean,
-  registerUser: boolean
+  mode: 'login' | 'register' | 'forgotPassword'
 };
 
 type AccountDropdownItemsProps = {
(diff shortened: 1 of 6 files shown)


finnss:
Get the links added to the LoginPage from `app/components/LoginForm` that we use in `replaceUnlessLoggedIn` as well. Once that's done this looks good! 👍

app/components/Header/index.js
```diff
@@ -81,12 +89,43 @@ function AccountDropdownItems({
...
     accountOpen: false,
-    shake: false
+    shake: false,
+    forgotPassword: false,
+    registerUser: false
+  };
+
+  toggleRegisterUser = (e: Event) => {
+    this.setState({ registerUser: true });
+    e.stopPropagation();
```
ekmartin: why do you need `stopPropagation` here?

app/components/Header/index.js
```diff
@@ -81,12 +89,43 @@ function AccountDropdownItems({
 class Header extends Component<Props, State> {
   state: State = {
     accountOpen: false,
-    shake: false
+    shake: false,
+    forgotPassword: false,
```
ekmartin: This should probably be something like an enum instead of multiple booleans, since only one of them can be active at a time. You could do something like: `mode: 'login' | 'register' | 'forgotPassword'`, and flow will make sure that those are the only valid values.

app/components/Header/index.js
```diff
@@ -81,12 +89,43 @@ function AccountDropdownItems({
...
+    e.stopPropagation();
   };
 
   render() {
     const { loggedIn } = this.props;
 
+    const { registerUser, forgotPassword } = this.state;
+
+    let title, form;
+    if (registerUser) {
```
ekmartin: then you could use a switch on `mode` here

app/components/Header/index.js
SmithPeder: toggleRegisterUser, toggleForgotPassword and toggleBack all need the stopPropagation to prevent the whole modal from closing every time the Event is triggered

ekmartin: hm, that seems like a funky part of the modal, but if that's the case it's usually common to use `event.preventDefault()` instead (at the start of the function)

app/components/Header/index.js
SmithPeder: Something like this?
<code>
type State = {
  accountOpen: boolean,
  shake: boolean,
  mode: 'login' | 'register' | 'forgotPassword'
};
</code>
and then
<code>
  state: State = {
    accountOpen: false,
    shake: false,
    mode: 'login'
  };
</code>
then use a switch case when deciding what form will be displayed?

ekmartin: yep!

app/components/Header/index.js
ekmartin: nevermind, I was wrong :)

SmithPeder:
@finnss Wanne check out the Loginpage.js, and see what needs to be changed?
screenshot from 2017-11-08 12-54-13

app/components/LoginForm/LoginPage.js
```diff
@@ -1,15 +1,71 @@
...
+        break;
+      case 'forgotPassword':
+        title = 'Glemt passord';
+        form = <ForgotPasswordForm />;
+        break;
+    }
+    return (
+      <Container className={styles.window}>
+        {isLogin ? (
+          <NavigationTab title={title} className={styles.detailTitle}>
```
finnss: The `{isLogin}` check can be done further in to reduce redundancy. `<NavigationTab title={title} className={styles.detailTitle}>` at least is common for both cases.

app/components/LoginForm/LoginPage.js
```diff
@@ -1,15 +1,71 @@
...
+        title = 'Registrer bruker';
+        form = <RegisterForm />;
+        break;
+      case 'forgotPassword':
+        title = 'Glemt passord';
+        form = <ForgotPasswordForm />;
+        break;
+    }
+    return (
+      <Container className={styles.window}>
```
finnss: The loginpage should certainly have a white background, and it doesn't seem to act too nicely on smaller screens. I'm not sure how `Content` would work here, it might make the component 100% width which you might not want, but you should test it out. Gotta rebase with master to get access to it. If `Content` doesn't make sense then you could compose some style or manually style it, just make sure it gets a nice white background and works well on all screen sizes.

app/components/LoginForm/LoginPage.js
```diff
@@ -1,15 +1,71 @@
...
+class LoginPage extends Component<null, State> {
+  state: State = {
+    mode: 'login'
+  };
+
+  toggleRegisterUser = (e: Event) => {
+    this.setState({ mode: 'register' });
+  };
+
+  toggleForgotPassword = (e: Event) => {
```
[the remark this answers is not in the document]
SmithPeder: Ahh, ofc i dont need that. I copy-pasted from the login modal, where the e.stopPropagation was needed, and forgot to remove the parameter :100:

PR #925, opened 2017-11-02
Change error messages

### LINKED ISSUE #545: Humanize shown errors
E.g. 403 Forbidden -> "Denne siden har du ikke tilgang på".
Can still show the fancy red number though, should just have a readable error message next to it.

SmithPeder:
Fixes #545

### DIFF
--- a/app/routes/errors/HTTPError.js
+++ b/app/routes/errors/HTTPError.js
@@ -6,10 +6,10 @@ import { Link } from 'react-router';
 import renderAbakus from './renderAbakus';
 
 const HTTPMapping = {
-  '400': '400 Bad Method',
-  '401': '401 Unauthorized',
-  '403': '403 Forbidden',
-  '404': 'Webkom finner ikke det du leter etter!',
+  '400': 'Noe gikk galt med forespørselen',
+  '401': 'Du er ikke logget inn',
+  '403': 'Denne siden har du ikke tilgang på',
+  '404': 'Denne siden finnes ikke. Er du logget inn?',
   '500': 'Noe gikk veldig galt, Webkom er på saken!'
 };
 


app/routes/errors/HTTPError.js
```diff
@@ -6,9 +6,9 @@ import { Link } from 'react-router';
 import renderAbakus from './renderAbakus';
 
 const HTTPMapping = {
-  '400': '400 Bad Method',
-  '401': '401 Unauthorized',
-  '403': '403 Forbidden',
+  '400': 'Noe gikk veldig galt, Webkom er på saken!',
```
ekmartin: 400 isn't that bad :) it _should_ mean that the user did something wrong, and not us. i'm not sure where this would pop up, but maybe something like `Noe gikk galt med forespørselen`?

app/routes/errors/HTTPError.js
```diff
@@ -6,9 +6,9 @@ import { Link } from 'react-router';
...
+  '401': 'Du er ikke logget inn',
+  '403': 'Denne siden har du ikke tilgang på',
   '404': 'Webkom finner ikke det du leter etter!',
```
ekmartin: maybe we should add something about logging in to 404 as well? e.g. `Denne siden finnes ikke. Er du logget inn?`

SmithPeder: Can you get the 404 error if you're not logged in? Shouldn't 401 trump 404?
-Try to access a page when you are not logged in -> 401
-Try to access a page that does not exit -> 404
-Try to access a page that does not exit when you're not a user -> 401
Or is this radical thinking?

ekmartin: It's usually common to return a 404 for specific items if the user isn't logged in, since you're otherwise revealing information (i.e. if the user gets a 404 for some events, but a 401 for others they'll know that there is an event with the ID they tried).

If on the other hand you have functionality that you always need to log in for, where the existence of the functionality itself isn't secret, returning a 401 is fine (you just have to do it for all IDs, no matter if it exists or not).

PR #927, opened 2017-11-02
Add an InfoList component for sidebars

ekmartin:
image

Currently used by event and joblistings.

### DIFF
--- a/app/routes/events/components/EventDetail/index.js
+++ b/app/routes/events/components/EventDetail/index.js
@@ -14,6 +14,7 @@ import {
 } from 'app/components/UserAttendance';
 import Tag from 'app/components/Tags/Tag';
 import { FormatTime, FromToTime } from 'app/components/Time';
+import InfoList from 'app/components/InfoList';
 import LoadingIndicator from 'app/components/LoadingIndicator';
 import { Flex } from 'app/components/Layout';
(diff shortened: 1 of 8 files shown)


app/routes/events/components/EventDetail/index.js
```diff
@@ -138,6 +139,38 @@ export default class EventDetail extends Component<Props> {
...
+      }
+    ].filter(Boolean);
+
+    const paidItems = [
+      { key: 'Pris', value: `${event.priceMember / 100},-` },
+      event.paymentDueDate && {
+        key: 'Betalingsfrist',
+        value: <FormatTime time={event.paymentDueDate} />
+      }
+    ].filter(Boolean);
```
finnss: Is there any merit to the idea of moving this filter functionality to the component itself? I imagine doing `someCondition && item` is common enough that most people who use the component will have to add the filter. While it's not a lot of code to write, it might be easy to forget and it might lead to weird error message for those less comfortable with javascript. 

Something like
```
type Props = {
   items: Array<Item | boolean>
 };
...
function InfoList({ items }: Props) {
   return (
     <ul>
       {items.filter(Boolean).map(({ key, value }) => (
 ...
```

ekmartin: Thought about it, but it seemed like a bit of a weird API. Can do it though, it's a reasonable argument.

PR #931, opened 2017-11-02
Add delete modal to events

odinuge:


app/routes/events/components/EventDetail/index.js
```diff
@@ -233,11 +230,7 @@ export default class EventDetail extends Component<Props> {
                     registration={currentRegistration}
                     isPriced={event.isPriced}
                   />
-                  <Admin
-                    actionGrant={actionGrant}
-                    event={event}
-                    deleteEvent={deleteEvent}
-                  />
+                  <Admin actionGrant={actionGrant} event={event} />
```
finnss: We don't want to be able to delete an event from its detail page?

hanswilw: Should change this anyways, but better off having this in the meantime. We should place edit and delete in the header like the other pages.

ekmartin: noo

ekmartin: having it in the sidebar (or something else) is better, navigation should be reserved for actual navigation - like changing pages, not actions

hanswilw: Sure – but then joblistings should be changed. Should be similar in the different pages :)

app/routes/events/components/Admin.js
```diff
@@ -1,64 +1,64 @@
...
+  deleteEvent: (eventId: ID) => Promise<*>,
+  eventId: ID
+}) => (
+  <li>
+    <ConfirmModalWithParent
+      title="Slett arrangement"
+      message="Er du sikker på at du vil slette dette arrangementet?"
+      onConfirm={() => deleteEvent(eventId)}
+    >
+      <a>Slett</a>
```
ekmartin: `a`s should only be used with `href`s - could we style the confirm component like a link instead?

app/routes/events/components/EventDetail/index.js
ekmartin: yeah, https://github.com/webkom/lego-webapp/pull/932

finnss: Hm, what about pages without sidebars? You say navigation should be `reserved` for actual navigation, and while I don't disagree with the idea, should pages implement sidebars if their only purpose would be such actions?

app/routes/events/components/Admin.js
odinuge: 👍

app/routes/events/components/EventDetail/index.js
odinuge: I would prefer to have as few "admin" stuff in the detail view. No need for a delete button in the detail view, because it isn't used at all. It is also annoying that it takes up space and can be clicked on by accident on mobile.

finnss: Don't you think it would be intuitive for an event admin to look for a delete button in the detail view, not list view, if they wanted to delete an event they made? If it's annoying and unnecessary on mobile we can just remove it there, that shouldn't determine desktop functionality imo.

odinuge: It would be intuitive for admins to go to the events admin-page/admin drop down etc. to do admin stuff. We should redesign our permission system to not make superusers admins everywhere - just like you don't use the `root` user for everyday tasks you do on your computer. The Facebook engineers  don't see "delete" buttons on every page they visit either, and imo we shouldn't do that either. It is ok to have a link to the admin page, but having all the admin stuff in the everyday components is bad design. 

We have already had to fix a few "bad" bugs that occurred because we are superusers when both developing and when running in PROD.  Eg: Abakom can see all events, but have no indication about who else can see them. I guess we can rename `Edit` to `Admin` if that makes more sense.

__TL;DR__: It is better to look for the delete button in 10sec, instead of having to send us an email to recover a deleted item. And when i visit an event, I just don't want to see the delete button... 👍

finnss: Oh so you want it in the `administrate` view? That's fine ofc. Most routes don't have that though, only a list view and a detail view. What then?

PR #936, opened 2017-11-03
Skip server-side rendering on errors

ekmartin:
try-catches `renderToString` so we avoid 504 on errors like https://github.com/webkom/lego-webapp/commit/aff37ec81b53965f35f4668f1e8de35068aa05fa

### DIFF
--- a/server/render.js
+++ b/server/render.js
@@ -73,19 +73,29 @@ function render(req, res, next) {
     );
 
     const respond = (error = null) => {
-      const renderResult = error !== TimeoutError;
-
-      const body = renderResult ? renderToString(app) : '';
-      const state = renderResult ? store.getState() : {};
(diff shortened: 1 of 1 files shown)


server/render.js
```diff
@@ -73,19 +73,29 @@ function render(req, res, next) {
...
+      const render = (body = '', state = {}) =>
+        res.send(
+          renderPage({
+            body: '',
+            state: {},
+            helmet
+          })
+        );
+
+      // Skip server side rendering on errors:
```
odinuge: This only applies if the error is a timeout error

odinuge: When we get other errors (eg. 404) we still want to use the SSR render.

ekmartin: fixed

PR #938, opened 2017-11-03
Verify student email

hanswilw:
Quickfix when entering full email, like `abcd@stud.ntnu.no` when we require username only. 

Add stripe remember cookie as well

### DIFF
--- a/app/routes/events/components/JoinEventForm.js
+++ b/app/routes/events/components/JoinEventForm.js
@@ -314,7 +314,7 @@ class JoinEventForm extends Component<Props, State> {
               description={event.title}
               image={logoImage}
               currency="NOK"
-              allowRememberMe={false}
+              allowRememberMe={true}
               locale="no"
               token={onToken}
               stripeKey={config.stripeKey}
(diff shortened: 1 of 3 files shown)


app/routes/events/components/JoinEventForm.js
```diff
@@ -314,7 +314,7 @@ class JoinEventForm extends Component<Props, State> {
               description={event.title}
               image={logoImage}
               currency="NOK"
-              allowRememberMe={false}
+              allowRememberMe={true}
```
ekmartin: this can be just `allowRememberMe` right?

app/routes/users/components/StudentConfirmation.js
```diff
@@ -119,7 +119,7 @@ const StudentConfirmation = ({
 };
 
 const validate = createValidator({
-  studentUsername: [required()],
+  studentUsername: [isNotEmail('Kun NTNU-brukernavn')],
```
ekmartin: couldn't we just do `input.replace('@stud.ntnu.no', '') or something before passing it to the API instead?

ekmartin:
👍

PR #944, opened 2017-11-05
Update the uncoordinated design of the titles/headers

SmithPeder:
So the title/header of each page, with its links and buttons are a bit of a mess, to put it mildly.
We got a mix of buttons, uppercase, lowercase, Square brackets, bold font. MeetingDetail has a top-margin, MeetingList does not use NavigationTab + NavigationLink and so forth.

I suggest we make it a bit more coordinated like this, left being the old and right being the new. 
Was supposed to make the same changes to "New Meeting", but for some reason i cant access the page on Staging :(
edit

### DIFF
--- a/app/routes/interestgroups/components/InterestGroupDetail.js
+++ b/app/routes/interestgroups/components/InterestGroupDetail.js
@@ -36,12 +36,12 @@ type TitleProps = {
 
 const Title = ({ group: { name, id }, showEdit }: TitleProps) => (
   <NavigationTab title={name}>
+    <NavigationLink to={`/interestgroups/`}>
+      <i className="fa fa-angle-left" /> Tilbake
+    </NavigationLink>
     {showEdit && (
-      <NavigationLink to={`/interestgroups/${id}/edit`}>
-        [Rediger]
-      </NavigationLink>
+      <NavigationLink to={`/interestgroups/${id}/edit`}>Rediger</NavigationLink>
     )}
-    <NavigationLink to={`/interestgroups/`}>[Tilbake]</NavigationLink>
   </NavigationTab>
 );
 
--- a/app/routes/interestgroups/components/InterestGroupDetail.js
+++ b/app/routes/interestgroups/components/InterestGroupDetail.js
@@ -36,7 +36,7 @@ type TitleProps = {
 
 const Title = ({ group: { name, id }, showEdit }: TitleProps) => (
   <NavigationTab title={name}>
-    <NavigationLink to={`/interestgroups/`}>
+    <NavigationLink to="/interestgroups/">
       <i className="fa fa-angle-left" /> Tilbake
     </NavigationLink>
     {showEdit && (
--- a/app/routes/interestgroups/components/InterestGroupList.js
+++ b/app/routes/interestgroups/components/InterestGroupList.js
@@ -24,7 +24,7 @@ const InterestGroupList = (props: Props) => {
       <div className={styles.section}>
         <div>
           <NavigationTab title="Interessegrupper">
-            <NavigationLink to={`/`}>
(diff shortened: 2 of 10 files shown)


app/routes/interestgroups/components/InterestGroupDetail.js
```diff
@@ -36,12 +36,12 @@ type TitleProps = {
 
 const Title = ({ group: { name, id }, showEdit }: TitleProps) => (
   <NavigationTab title={name}>
+    <NavigationLink to={`/interestgroups/`}>
```
finnss: You only need these ```{`...`} ``` if you use variables in there. With simple strings like this, just write `to="/interestgroups/"`.

app/routes/interestgroups/components/InterestGroupList.js
```diff
@@ -24,7 +24,9 @@ const InterestGroupList = (props: Props) => {
       <div className={styles.section}>
         <div>
           <NavigationTab title="Interessegrupper">
-            <NavigationLink to={`/`}>Hjem</NavigationLink>
+            <NavigationLink to={`/`}>
```
finnss: same here

app/routes/meetings/components/MeetingEditor.js
```diff
@@ -98,13 +98,11 @@ function MeetingEditor({
...
     <div className={styles.root}>
-      <h2>
-        <Link to={isEditPage ? `/meetings/${meeting.id}` : `/meetings/`}>
-          <i className="fa fa-angle-left" />
-          {isEditPage ? ` ${meeting.title}` : ' Mine møter'}
-        </Link>
-      </h2>
-      <h1>{isEditPage ? 'Endre møte' : 'Nytt møte'} </h1>
+      <NavigationTab title={'Nytt møte'} className={styles.detailTitle}>
+        <NavigationLink to={`/meetings/`}>
```
finnss: and here and here

app/routes/meetings/components/Toolbar.js
```diff
@@ -1,25 +1,19 @@
...
-
-      <div className={styles.section}>
-        <Link to="/meetings/create/">
-          <Button>Nytt møte</Button>
-        </Link>
-      </div>
-    </div>
+    <Flex wrap alignItems="center" justifyContent="space-between">
+      <NavigationTab title={'Dine Møter'} className={styles.detailTitle}>
+        <NavigationLink to={'/meetings/create/'}>Nytt møte</NavigationLink>
```
finnss: you know it

app/routes/pages/components/PageDetail.js
```diff
@@ -91,9 +91,9 @@ function PageDetail<T: Object>({
     <Content>
       <NavigationTab title={title}>
         {actionGrant.includes('edit') &&
-          editUrl && <NavigationLink to={`${editUrl}`}>ENDRE</NavigationLink>}
+          editUrl && <NavigationLink to={`${editUrl}`}>Endre</NavigationLink>}
```
finnss: This is kinda the opposite problem – the purpose of the ```{`someString${someVar}`}``` syntax is to combine strings and variables. My earlier comments tell you to use only "" when all you have is strings, but here you only have a variable. Thus, you can just write `to={editUrl}`.

app/routes/pages/components/PageDetail.js
```diff
@@ -91,9 +91,9 @@ function PageDetail<T: Object>({
...
         {actionGrant.includes('create') && (
-          <NavigationLink to={`/pages/new`}> NY </NavigationLink>
+          <NavigationLink to={`/pages/new`}>Ny</NavigationLink>
```
finnss: Last one! `to="..."`

PR #950, opened 2017-11-07
Make sure events in bdb detail doesn't get mutated on each render

finnss:


app/routes/bdb/components/BdbDetail.js
```diff
@@ -185,6 +185,7 @@ export default class BdbDetail extends Component<Props, State> {
     const events =
       companyEvents &&
       companyEvents
+        .slice()
```
ekmartin: for future reference, lodash methods usually don't mutate, so they're often good alternatives (`sort` and `slice`)

PR #963, opened 2017-11-08
 Fix infinite fetch after 401 response

odinuge:
Only run logout after getting a 401 response when the user is loggedIn
(has authentication info in the state). The previous behavior resulted
in an infinite loop.

### DIFF
--- a/app/actions/UserActions.js
+++ b/app/actions/UserActions.js
@@ -5,7 +5,7 @@ import { normalize } from 'normalizr';
 import config from 'app/config';
 import cookie from 'js-cookie';
 import moment from 'moment-timezone';
-import { push, replace } from 'react-router-redux';
+import { push } from 'react-router-redux';
 import { userSchema } from 'app/reducers';
 import callAPI from 'app/actions/callAPI';
 import { User } from './ActionTypes';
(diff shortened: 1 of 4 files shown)


odinuge:
This should fix all the sentry issues `Attempt to use history.replaceState() more than 100 times per 30.000000 seconds`

app/actions/UserActions.js
```diff
@@ -63,11 +63,13 @@ export function login(
     });
 }
 
-export function logout() {
+export function logout(redirect: boolean = true) {
   return (dispatch: Dispatch<*>) => {
     removeToken();
     dispatch({ type: User.LOGOUT });
-    dispatch(replace('/'));
+    if (redirect) {
```
ekmartin: wouldn't it be just as good to just dispatch this after dispatching `logout`? not a fan of boolean arguments :P

ekmartin: `push` that is

odinuge: :+1:

app/reducers/index.js
```diff
@@ -89,7 +89,12 @@ const appReducer = combineReducers(reducers);
 
 export default function rootReducer(state: State, action: Action) {
   if (action.type === User.LOGOUT) {
-    return appReducer(undefined, action);
+    return appReducer(
+      {
+        routing: state.routing
```
odinuge: This will save the router info and history, but that shouldn't matter since we now use `push` instead of just `replace`.

app/actions/UserActions.js
```diff
@@ -63,11 +63,17 @@ export function login(
     });
 }
 
+export function logoutWithRedirect() {
```
odinuge: This is certainly better than the previous one.. What do you think @ekmartin?

ekmartin: 👍

PR #964, opened 2017-11-08
Fix height of captcha

### LINKED ISSUE #762: `Meld deg på` button is only clickable on text
Should be able to click on the whole button.

odinuge:
This fixes the issue when only button text was clickable due to the
captcha.

New:
screenshot from 2017-11-08 11-31-44

Old:
screenshot from 2017-11-08 11-32-18

Fixes #762

Caused by https://github.com/jgthms/minireset.css/blob/master/minireset.css#L64


ekmartin:
Brilliant!

PR #967, opened 2017-11-08
Use actionGrant for interest group edit

### LINKED ISSUE #965: Use actionGrant to toggle interestgroup edit
Currently hardcoded to `true`
https://github.com/webkom/lego-webapp/blob/6121a489a129ea161833cb5e3c2cbeced59e7b4b/app/routes/interestgroups/components/InterestGroupDetail.js#L129

ekmartin:
Fixes #965

### DIFF
--- a/app/routes/interestgroups/components/InterestGroupDetail.js
+++ b/app/routes/interestgroups/components/InterestGroupDetail.js
@@ -123,10 +123,11 @@ type Props = {
 
 function InterestGroupDetail(props: Props) {
   const { group } = props;
+  const canEdit = group.actionGrant && group.actionGrant.includes('edit');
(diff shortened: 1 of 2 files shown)


app/routes/interestgroups/components/InterestGroupDetail.js
```diff
@@ -123,10 +123,11 @@ type Props = {
 
 function InterestGroupDetail(props: Props) {
   const { group } = props;
+  const canEdit = group.actionGrant && group.actionGrant.includes('edit');
```
hanswilw: If group exists, I believe actionGrant is always an array - so we probably dont need this short-circuit check?

PR #971, opened 2017-11-09
Return [] groups when the user object don't contains the abakusGroups property

eirsyl:


eirsyl:
Profile page crashes when visiting users with public profile stored in the state

PR #973, opened 2017-11-10
Bump prettier and update linting

odinuge:
Now Prettier also formats markdown :rocket:


odinuge:
When using the eslint prettier addon, it only checks the js files. Should we just remove it, or run `lint:pretteier` on only css and md? It may be useful for people using eslint with vim etc., but with format on autosave, it shouldn't be necessary...

package.json
```diff
@@ -15,10 +15,11 @@
     "test": "NODE_ENV=test jest",
     "test:coverage": "yarn run test -- --coverage",
     "test:watch": "yarn run test -- --watch",
-    "lint": "yarn run lint:js && yarn run lint:css",
+    "lint": "yarn run lint:js && yarn run lint:css && yarn run lint:prettier",
```
ekmartin: isn't this already done in `lint:js` with eslint-plugin-prettier? could probably remove that then?

ekmartin: lol 2slow, yeah, just remove eslint-plugin-prettier :)

odinuge:
Guess I should update `README` too

PR #975, opened 2017-11-10
Sort frontpage events by `id`, after `pinned` and start time.

martinhath:
The functions passed to `lodash.sortBy` are ran on each element,
producing a new list `[f(x), g(x), h(x), ...]`, and it is _this_ list
that is sorted. With the previous attempted solution to this problem we
had the `id` as the first element. This causes the list to only look at
`pinned` and start time, if there are collisions in the IDs (whic is
unlikely). With this change, we first sort by pinned. If there are
multiple items with the same value (which there will be, since `pinned`
is either `true` or `false`), we look at the start time abs thing.
Then, if there are elements that still compare as equals, we look at the
ID. This is the indended semantics of the previous PR (#961).


ekmartin:
👍

PR #976, opened 2017-11-10
Fix login bugs

odinuge:
This fixes login bug when HTTPError is rendered, and the bug when
navigation between 404 pages that doesn't have a router entry. It also
clears the fetchHistory on login.

### DIFF
--- a/app/actions/ActionTypes.js
+++ b/app/actions/ActionTypes.js
@@ -275,7 +275,8 @@ export const Feed = {
 };
 
 export const FetchHistory = {
-  SET_HISTORY: 'FetchHistory.SET_HISTORY'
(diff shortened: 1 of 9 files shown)


odinuge:
Aaand, what does people think about having `replaceUnlessLoggein` on `/events/pk`. Have heard a lot of complains about it.. This atleast allows the user to login when having a 404 before login. :smile:

odinuge:
This will also fix the strange `undefined` is not a function from the `PhantomJS` browser:

https://sentry.abakus.no/webkom/lego-webapp/issues/3908/

ekmartin:
yeah, that sounds like a good temporary measure (until we've sorted out the permission stuff for events)

odinuge:
Smooth. Will add that too

PR #978, opened 2017-11-11
Rename bubble notification to toast

### LINKED ISSUE #952: Rename bubble notifications to something else
`reducers/notifications.js` and `NotificationContainer` are pretty confusing names considering we have actual notifications on the page. We should call these toasts or alerts or something to differ between them (for the small popups that show up in the bottom left corner).

JonasBak:
Fix #952
Renames `notification` and `NotificationContainer` to `toast` and `ToastContainer`


ekmartin:
Nice!

PR #980, opened 2017-11-11
Add a safer way to handle jwt decode

### LINKED ISSUE #624: Don't crash on invalid tokens

odinuge:
It now handles an invalid jwt token (token that cannot be decoded) as if
no token exists.

This fixes #624

### DIFF
--- a/app/actions/UserActions.js
+++ b/app/actions/UserActions.js
@@ -16,9 +16,9 @@ import { setStatusCode } from './RoutingActions';
 
 const USER_STORAGE_KEY = 'lego.auth';
 
-type encodedToken = string;
+type EncodedToken = string;
 
-type decodedToken = {
+type DecodedToken = {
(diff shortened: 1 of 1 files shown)


app/actions/UserActions.js
```diff
@@ -31,6 +46,25 @@ function removeToken() {
   return cookie.remove(USER_STORAGE_KEY, { path: '/' });
 }
 
+function getToken(getCookie?: string => encodedToken): ?token {
+  const encodedToken = getCookie
+    ? getCookie(USER_STORAGE_KEY)
```
ekmartin: might be slightly cleaner (and more explicit) to pass in `cookie.get` as well, so you don't need the conditional here?

odinuge: Agree! Should be good to go now.

PR #981, opened 2017-11-11
Update toast styling on mobile

odinuge:
This makes the toast look at least a bit civilized on mobile. It also
removes the default "lukk" action that is totally useless. I'll do like
the author of the 'react-notification' library (that is shit, and has a
horrible api) and link to the material design guidelines about the
toast - even tho the library doesn't follow a single on of them.

http://www.google.com/design/spec/components/snackbars-toasts.html#snackbars-toasts-usage

screenshot_20171111-212708
screenshot_20171111-214108


SmithPeder:
That looks alot better :) Nice work :)

PR #983, opened 2017-11-13
Fix feedback initialValue when feedbackrequired

hanswilw:
Feedback was not prefilled when using required feedback.


odinuge:
Smooth! :+1:

Issue #904, opened 2017-11-01
Highlight fields that needs to be filled in before registration to event


cherrybeans:
"NB: Dette arrangementet krever tilbakemelding: Melding til arrangører:" is not clear enough. In a rush you'd want everything you need to submit be highlighted so you understand why the Register button is still deactivated.

Suggested fix: Put highlight around fields that are required.

PR #985, opened 2017-11-13
Add support for validating fields without touch

odinuge:
This is _one_ way of solving #904. Not sure if it is the right way of doing it tho. Feel free to suggest a better name than `validateUntouched` - but it was the best i was able to find... :laughing: 

This is how it looks when you open an event with registration open, and with required feedback..
screenshot from 2017-11-13 16-36-20

### DIFF
--- a/app/components/Form/Field.js
+++ b/app/components/Form/Field.js
@@ -25,7 +25,8 @@ type FieldProps = {
   fieldStyle: any,
   fieldClassName: string,
   labelClassName: string,
-  showErrors: boolean
+  showErrors: boolean,
+  validateUntouched?: boolean
 };
 
 /**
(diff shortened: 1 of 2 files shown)


hanswilw:
I thought of the same, but it was discussed if we should have a red asterisk, red text or something instead. This "screams" form error, before it is touched – and that might not be wanted behavior?

odinuge:
Yeah, it kinda screams towards the user. This is a "general probelm" in the whole app. It is annoying that field error moves the content, and it takes up quite a lot of space.

If we redesign the whole field error to eg. took like this, it would look a lot better!

(Still a bit too much for the register form, but much better)
screenshot from 2017-11-13 16-53-26
 

screenshot from 2017-11-13 16-55-45

Some inspiration:
https://material.io/guidelines/patterns/errors.html#errors-user-input-errors

odinuge:
Or we can just trigger validation on the `feedback` field when the `Captcha` is validated.

finnss:
Might not be appropriate for this PR, but while we're on the topic: A lot of users have complained about a lack of responsiveness when trying to submit and an error occurs. We should either scroll the page to the first field with an error (best imo) or at least display some sort of "You need to fix errors in the form" message near the submit button. Atm it feels like the site is broken or something when you try to click submit with an error.

odinuge:
Smart @finnss! Started "working" on that in #990.

PR #987, opened 2017-11-13
Show submission errors in user confirmation form

kristeine:
image

I tried to create a user in staging but got a toast saying "Opprettelse av bruker feilet" which doesn't really explain to the user what's wrong. 

Next step: i18n 😉 

(Also, I added `name` props to some Field elements which were logging console warnings)

### DIFF
--- a/app/routes/users/components/UserConfirmation.js
+++ b/app/routes/users/components/UserConfirmation.js
@@ -3,7 +3,7 @@
 import styles from './UserConfirmation.css';
 import React from 'react';
 import { Container, Flex } from 'app/components/Layout';
-import { reduxForm } from 'redux-form';
+import { reduxForm, SubmissionError } from 'redux-form';
 import {
   Form,
   TextInput,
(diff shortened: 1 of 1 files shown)


finnss:
The idea of this is nice, but the error messages from the back-end are in english and we want messages the user sees to be in norwegian, no?

hanswilw:
We just discussed that. :) We do already have the english backend errors in other submission errors in master now, we should however add some localization in the future.

app/routes/users/components/UserConfirmation.js
```diff
@@ -98,16 +103,19 @@ const UserConfirmation = ({
           />
           <RadioButtonGroup name="gender">
             <Field
+              name="genderMale"
```
hanswilw: These names are determined by the RadioButtonGroup's `name="gender"` and should be removed or set explicitly to `name="gender"`. They refer to the field `gender` that gets posted to the backend, and redux-form uses the `name` variable to be the same to ensure that the radiobuttons switches correctly.

ekmartin:
I think @odinuge fixed a generic version of this, with `legoForm` - right?

odinuge:
Should work out of the box if 
https://github.com/webkom/lego-webapp/blob/a7b051448592540a08cc199e44a6a824469d4cc6/app/routes/users/components/UserConfirmation.js#L137
is switched to `legoForm` instead

ekmartin:
I'll close this then - wanna send a PR fixing that @odinuge?

odinuge:
Will do

Issue #946, opened 2017-11-06
ReduxForm initialValues is set even though the field is not initialized


hanswilw:
E.g. `mergeTime` in event creation is actually set to the initialValue even the field is not initialized (when no pools or a single pool is selected). I think we should only pass the field values to the submit callback if the field is initialized, so e.g. `mergeTime` would be undefined when not initialized.

PR #990, opened 2017-11-13
Add legoForm

### LINKED ISSUE #746: Lack of feedback when submitting in forms and actions around the webapp

odinuge:
This is much better than having this in each form.

It would also be much easier to refactor later.

This fixes #746

### DIFF
--- a/app/components/Form/legoForm.js
+++ b/app/components/Form/legoForm.js
@@ -2,11 +2,13 @@
 
 import { reduxForm, SubmissionError } from 'redux-form';
 import { handleSubmissionError } from './utils';
-import { pick, difference } from 'lodash';
+import { pick } from 'lodash';
 type Props = {
   onSubmitFail: any => any,
   onSubmit: any => Promise<*>,
-  /* Enable auto submissionError */
+  /* Enable auto submissionError
+   * This only works if 'onSubmit' is passed into legoForm,
+   * and not if it is passed into 'handleSubmit()'*/
   enableSubmissionError?: boolean,
   /* Move the screen to the first error in the list on SubmissionError */
   enableFocusOnError?: boolean,
@@ -14,8 +16,9 @@ type Props = {
    * Should be used when there is no visible filed for the value,
    * but that it should still pass the value further.
    * 
-   * The id is always picked
+   * The id is always picked, together with all registered fields
    */
+  enableValuePicking?: boolean,
   pickAdditionalValues?: Array<string>,
   ...any
 };
@@ -25,16 +28,15 @@ const legoForm = ({
   onSubmit,
   enableSubmissionError = true,
   enableFocusOnError = true,
+  enableValuePicking = true,
   pickAdditionalValues = [],
   ...rest
 }: Props) =>
   reduxForm({
     ...rest,
     onSubmitFail: errors => {
-      onSubmitFail(errors);
-
       if (!enableFocusOnError) {
-        return;
+        return onSubmitFail(errors);
       }
 
       try {
@@ -48,19 +50,17 @@ const legoForm = ({
       } catch (e) {
         //
       }
+      return onSubmitFail(errors);
     },
     onSubmit: (values, dispatch, props) => {
-      const pickedValues = pick(
-        values,
-        Object.keys(props.registeredFields)
-          .concat('id')
-          .concat(pickAdditionalValues)
-      );
-
-      console.log(
-        'Diff (lost in translation): ',
-        difference(Object.keys(values), Object.keys(pickedValues))
-      );
+      const pickedValues = enableValuePicking
+        ? pick(
+            values,
+            Object.keys(props.registeredFields)
+              .concat('id')
+              .concat(pickAdditionalValues)
+          )
+        : values;
 
       return onSubmit(pickedValues, dispatch, props).catch(error => {
         if (error instanceof SubmissionError || !enableSubmissionError) {
--- a/app/components/Form/legoForm.js
+++ b/app/components/Form/legoForm.js
@@ -8,17 +8,30 @@ type Props = {
   onSubmit: any => Promise<*>,
   /* Enable auto submissionError
    * This only works if 'onSubmit' is passed into legoForm,
-   * and not if it is passed into 'handleSubmit()'*/
+   * and not if it is passed into 'handleSubmit()
+   *
+   * This will add a "generic" error message to the "_error" value that can
+   * be ectracted from the form props as "props.error"
+   */
   enableSubmissionError?: boolean,
   /* Move the screen to the first error in the list on SubmissionError */
   enableFocusOnError?: boolean,
+  /* This will pick the values assosiated with only the visible
+   * fields at the time of submitting. This makes it possible to add
+   * default values to the `initialValues` to redux form, and only send
+   * them if a field is registered/visible. It therefore also allows you to
+   * do object spreading into initialValues, without sending all the values
+   * into the onSubmit action.
+   *
+   * about picking: https://lodash.com/docs/4.17.4#pick
+   */
+  enableValuePicking?: boolean,
   /* Also pick these values based on the keys given on the array
-   * Should be used when there is no visible filed for the value,
-   * but that it should still pass the value further.
+   * Should be used when there isn't a visible field for the value,
+   * at the time of submitting
    * 
    * The id is always picked, together with all registered fields
    */
-  enableValuePicking?: boolean,
   pickAdditionalValues?: Array<string>,
   ...any
 };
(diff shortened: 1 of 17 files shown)


app/components/Form/utils.js
```diff
@@ -0,0 +1,18 @@
...
+  return (...data: Array<any>) =>
+    func(...data).catch(error => {
+      const errPayload = get(error, ['payload', 'response', 'jsonData']);
+      if (!errPayload) {
+        throw error;
+      }
+
+      throw new SubmissionError({
+        ...errPayload,
+        _error: errPayload.detail
```
odinuge: A lot of django errors returns a general error as `detail`.The `_error` is sent to the form as the `error` prop, so it can be shown together with the submit button...

app/routes/articles/components/ArticleEditor.js
```diff
@@ -121,8 +106,28 @@ const ArticleEditor = ({
...
 };
 
-export default reduxForm({
+const onSubmit = (
+  data,
+  dispatch,
+  { currentUser, isNew, articleId, submitArticle }: Props
+) => {
+  const body = {
+    ...(isNew ? {} : { id: articleId }),
```
hanswilw: When you are already omitting using spread, you might as well do short-circuiting
`...(!isNew && { id: articleId })`

odinuge: :+1:
I just moved the function, so didn't read it.. But yours look way better !

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,54 @@
...
+
+        const field = document.querySelector(`[name="${firstErrorField}"]`);
+        if (field && field.focus) {
+          field.focus();
+        }
+      } catch (e) {
+        //
+      }
+    },
+    onSubmit: (values, dispatch, props) =>
```
hanswilw: Maybe this might be the place to handle #946 ?

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,75 @@
...
+  enableSubmissionError?: boolean,
+  /* Move the screen to the first error in the list on SubmissionError */
+  enableFocusOnError?: boolean,
+  /* Also pick these values based on the keys given on the array
+   * Should be used when there is no visible filed for the value,
+   * but that it should still pass the value further.
+   * 
+   * The id is always picked
+   */
+  pickAdditionalValues?: Array<string>,
```
odinuge: add `enablePicking`?? With a default of `true` or `false`?

app/components/Form/utils.js
```diff
@@ -0,0 +1,34 @@
...
+/*
+ * Simple utility that handles submission errors
+ *
+ * Usage:
+ * onSubmit(data).catch(handleSubmissionError)
+ * or
+ * onSubmit(data).then(result=>{...}, handleSubmissionError)
+ */
+export const handleSubmissionError = (error: any) => {
+  const errPayload = get(error, ['payload', 'response', 'jsonData']);
```
odinuge: Move this to a separate reducer/util that handles errors.

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,75 @@
...
+    onSubmitFail: errors => {
+      onSubmitFail(errors);
+
+      if (!enableFocusOnError) {
+        return;
+      }
+
+      try {
+        // We should instead check if error is SubmissionError. Does not work now
+        const [firstErrorField] = Object.keys(errors);
```
odinuge: Cleanup this, and add flow :no_smoking:

app/routes/users/components/UserSettings.js
```diff
@@ -51,11 +52,7 @@ const UserSettings = (props: Props) => {
   const showAbakusMembership = user.isStudent;
 
   const onSubmit = data =>
-    updateUser(omit(data, 'profilePicture')).catch(err => {
-      if (err.payload && err.payload.response) {
-        throw new SubmissionError(err.payload.response.jsonData);
-      }
-    });
+    updateUser(omit(data, 'profilePicture')).catch(handleSubmissionError);
```
odinuge: This `omit` can be removed when we introduce "picking" in `legoForm`?? Since there is no `profilePicture` field..

app/routes/meetings/components/MeetingEditor.js
```diff
@@ -186,33 +167,49 @@ function MeetingEditor({
...
+  onSubmit: (
+    data,
+    dispatch,
+    { push, handleSubmitCallback, inviteUsersAndGroups }: Props
+  ) =>
+    handleSubmitCallback(data).then(result => {
+      const id = data.id || result.payload.result;
+      const { groups, users } = data;
+      if (groups || users) {
+        return inviteUsersAndGroups({ id, users, groups }).then(() =>
```
odinuge: Chain this better

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,75 @@
...
+        if (field && field.focus) {
+          field.focus();
+        }
+      } catch (e) {
+        //
+      }
+      return onSubmitFail(errors);
+    },
+    onSubmit: (values, dispatch, props) => {
+      const pickedValues = enableValuePicking
```
odinuge: Can this be a ok solution on #946 @hanswilw ? Not sure if we should enable it for everything tho...

app/components/Form/legoForm.js
hanswilw: `enableValuePicking` is better imo. Default to false maybe? Better to explicitly handle it don't u think? Might be other arguments sent in addition to the registered ones, which could lead to issues if one is not aware of this picking.

app/components/Form/legoForm.js
hanswilw: Yep, I dont see any issues with this approach. As mentioned we could just keep it false by default - atleast initially.

app/components/Form/utils.js
```diff
@@ -0,0 +1,34 @@
...
+  if (!errPayload) {
+    throw error;
+  }
+
+  throw new SubmissionError({
+    ...errPayload,
+    _error: errPayload.detail
```
finnss: I'm not sure I get this. Isn't this payload included by default when `error` is thrown by itself?

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,75 @@
...
+import { pick } from 'lodash';
+type Props = {
+  onSubmitFail: any => any,
+  onSubmit: any => Promise<*>,
+  /* Enable auto submissionError
+   * This only works if 'onSubmit' is passed into legoForm,
+   * and not if it is passed into 'handleSubmit()'*/
+  enableSubmissionError?: boolean,
+  /* Move the screen to the first error in the list on SubmissionError */
+  enableFocusOnError?: boolean,
```
finnss: Do we ever *not* want this?

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,75 @@
...
+  /* Also pick these values based on the keys given on the array
+   * Should be used when there is no visible filed for the value,
```
finnss: "no visible filed for the value"?

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,75 @@
...
+      try {
+        // We should instead check if error is SubmissionError. Does not work now
+        const [firstErrorField] = Object.keys(errors);
+
+        const field = document.querySelector(`[name="${firstErrorField}"]`);
+        if (field && field.focus) {
+          field.focus();
+        }
+      } catch (e) {
+        //
```
finnss: Do you think there's any reason to actually handle this? If I understand correctly the error that can be caught is from `querySelector` not finding anything? If this happens enough to warrant a backup plan, I guess we could add "There are errors in the field that need to be corrected" or some such near the Submit button?

app/components/Form/legoForm.js
odinuge: Short: _NO_ :laughing:

app/components/Form/legoForm.js
finnss: I guess filed = field. This comment should probably be rewritten to better document exactly what `enableValuePicking` and `pickAdditionalValues` do.

app/components/Form/utils.js
odinuge: Added a comment about this in the `Props` section.

app/components/Form/legoForm.js
```diff
@@ -0,0 +1,88 @@
...
+  enableFocusOnError?: boolean,
+  /* This will pick the values assosiated with only the visible
+   * fields at the time of submitting. This makes it possible to add
+   * default values to the `initialValues` to redux form, and only send
+   * them if a field is registered/visible. It therefore also allows you to
+   * do object spreading into initialValues, without sending all the values
+   * into the onSubmit action.
+   *
+   * about picking: https://lodash.com/docs/4.17.4#pick
+   */
```
finnss: Excellenté!

PR #1007, opened 2017-11-17
Abacard

ekmartin:
Adds a view where users can search for themselves and select that they're attending. Uses the existing `SearchPage`, but also shows a green checkmark and plays a sound when someone is marked as attending. We can consider removing search in favor of just usernames later, but works for now.

image


app/routes/events/components/EventAdministrate/Abacard.js
```diff
@@ -1,62 +1,69 @@
...
+    this.props.handleSelect(result).then(
+      () => {
+        const sound = new window.Audio(goodSound);
+        sound.play();
+        this.showCompleted();
+        return this.props.clearSearch();
+      },
+      err => {
+        const payload = err.payload.response.jsonData;
+        if (payload.error_code === 'not_registered') {
```
odinuge: These are "camelCased" by the backend before they are sent. So you should use `payload.errorCode`.

app/routes/events/components/EventAdministrate/Abacard.js
```diff
@@ -1,62 +1,69 @@
...
+  handleSelect = (result: SearchResult) => {
+    this.props.handleSelect(result).then(
+      () => {
+        const sound = new window.Audio(goodSound);
+        sound.play();
+        this.showCompleted();
+        return this.props.clearSearch();
+      },
+      err => {
+        const payload = err.payload.response.jsonData;
```
odinuge: Guess you should check that `jsonData` ins't undefined, so you don't crash the thing.. :/

ekmartin:
Fixed the two error comments - thanks @odinuge. Tested it against the staging API now as well. There's definitely other improvement areas here, but I think it should be decent enough for tomorrow.

PR #1012, opened 2017-11-18
Add penalty notice

hanswilw:
Adds notice when trying to unregister when event heeds penalties and has passed unregistration deadline

image


app/components/Modal/ConfirmModal.js
```diff
@@ -8,7 +8,7 @@ import Button from '../Button';
 type ConfirmModalProps = {
   onConfirm?: () => Promise<*>,
   onCancel?: () => Promise<*>,
-  message: string,
+  message: string | Node,
```
ekmartin: this can be just `Node`

PR #1016, opened 2017-11-19
Abacard fixes

ekmartin:
Fixes a few random issues with Abacard:
* Makes sure that the input is always in focus
* Adds a custom placeholder prop to SearchPage
* Changes the title of the admin page to the event title
* Caps `selectedIndex` to `0 <= index <= results.length`

### DIFF
--- a/app/components/Search/SearchPage.js
+++ b/app/components/Search/SearchPage.js
@@ -9,6 +9,7 @@ import type { SearchResult } from 'app/reducers/search';
 type Props = {
   searching: boolean,
   location: Object,
+  inputRef?: (?HTMLInputElement) => void,
   onQueryChanged: string => void,
   results: Array<SearchResult>,
   handleSelect: SearchResult => void
@@ -46,10 +47,14 @@ class SearchPage extends Component<Props, State> {
 
       case Keyboard.ENTER:
         e.preventDefault();
-        this.setState({ query: '' });
-        this.props.handleSelect(this.props.results[this.state.selectedIndex]);
+        this.handleSelect(this.props.results[this.state.selectedIndex]);
         break;
     }
+  };
+
+  handleSelect = (result: SearchResult) => {
+    this.setState({ query: '', selectedIndex: 0 });
+    this.props.handleSelect(result);
   };
 
   handleQueryChange = ({ target }: SyntheticInputEvent<HTMLInputElement>) => {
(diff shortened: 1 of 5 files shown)


app/components/Search/SearchPage.js
```diff
@@ -25,6 +27,16 @@ class SearchPage extends Component<Props, State> {
     query: this.props.location.query.q || ''
   };
 
+  componentWillReceiveProps(nextProps: Props) {
+    // Make sure the selectedIndex is within 0 <= index < results.length:
+    const selectedIndex = Math.min(
```
odinuge: How about just:
```js
const selectedIndex = (selectedIndex < resullts.length) ? selectedIndex : 0;
// Or magic with lodash
const selectedIndex = has(result, selectedIndex) ? selectedIndex : 0;
```
The `min` and `max` makes it kinda hard to quickly understand the flow.

Or just reset the index to 0  when new search-results appear.

app/routes/events/components/EventAdministrate/Abacard.js
```diff
@@ -25,26 +26,33 @@ class Abacard extends React.Component<*, State> {
...
-        if (payload && payload.errorCode === 'not_registered') {
-          alert('Bruker er ikke påmeldt på eventet!');
-        } else if (payload && payload.errorCode === 'already_present') {
-          alert('Bruker er allerede satt som tilstede.');
-        } else {
-          alert(
-            `Det oppsto en uventet feil: ${JSON.stringify(payload || err)}`
-          );
+    this.props.clearSearch();
+    this.props
```
odinuge: Is there a reason to not return this promise?

app/routes/events/components/EventAdministrate/Abacard.js
```diff
@@ -25,26 +26,33 @@ class Abacard extends React.Component<*, State> {
   };
 
   handleSelect = (result: SearchResult) => {
```
odinuge: This should be `(result: ?SearchResult=>{}` since it is allowed to click enter when there is no searchResult, resulting in an empty object here..

app/components/Search/SearchPage.js
```diff
@@ -38,6 +38,7 @@ class SearchPage extends Component<Props, State> {
   }
 
   handleKeyDown = (e: KeyboardEvent) => {
+    if (this.props.results.length === 0) return;
```
odinuge: :+1:

app/routes/events/components/EventAdministrate/Abacard.js
ekmartin: I changed the `handleKeyDown` logic to only do stuff if there's at least one result, `handleSelect` shouldn't be called without a result.

app/routes/events/components/EventAdministrate/Abacard.js
ekmartin: Nah, not really - fixed

PR #1018, opened 2017-11-20
Add the new readme release

odinuge:
screenshot from 2017-11-20 10-43-51

And YES, we should talk with `readme` to automate/force them to do this this..


larseen:
There have been several talks in the past of digitalizing readme articles. 
There is a way of doing it where we not only can get all new readme articles in a database but also all old articles.
This need to be a joint effort with readme and webkom, where we create a mapping from the readme layout used in indesign to html. This has been investigated before, and is possible to do, but been down prioritised due to lego, etc.
https://helpx.adobe.com/indesign/using/export-content-html-cc.html

PR #1020, opened 2017-11-20
Use lodash get when resolving data for table

odinuge:
This is "already" used in event administrate

Now:
screenshot from 2017-11-20 15-04-14

vs. the old
screenshot from 2017-11-20 15-04-47


odinuge:
We should make shortnames for classes, so we can print the value in one (short)line

PR #1022, opened 2017-11-23
Disable submit button in joblistings when working

odinuge:
This will help bedkom to avoid creating 10+ duplicates of a joblisting

### DIFF
--- a/app/routes/joblistings/components/JoblistingEditor.js
+++ b/app/routes/joblistings/components/JoblistingEditor.js
@@ -5,7 +5,13 @@ import styles from './JoblistingEditor.css';
 import LoadingIndicator from 'app/components/LoadingIndicator';
 import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
 
-import { reduxForm, Field, SubmissionError, change } from 'redux-form';
+import {
+  type FormProps,
+  reduxForm,
+  Field,
+  SubmissionError,
+  change
+} from 'redux-form';
 import { httpCheck } from 'app/routes/bdb/utils';
 import {
   TextInput,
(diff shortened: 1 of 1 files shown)


app/routes/joblistings/components/JoblistingEditor.js
```diff
@@ -228,7 +241,11 @@ class JoblistingEditor extends Component<Props, State> {
                 <Button dark>Slett</Button>
               </ConfirmModalWithParent>
             )}
-            <Button className={styles.submit} submit>
+            <Button
+              disabled={pristine || submitting}
```
ekmartin: shouldn't pristine be replaced with `invalid`? since we have validations for the form

odinuge: Yeah, that's possible. It doesn't matter that much to me, as long as it is consistent. #990 adds support for focusing on the invalid fields when clicking submit, and that wouldn't work if the button is disabled.

ekmartin: Yeah, that's a good point. Maybe a good user experience would be to disable the button for frontend validation errors, and jump to fields for backend errors (i.e. if a field is unique etc.). Since you see the frontend errors inline right after you type something in a field, I'm hoping most people would fix it right away instead of scrolling past and trying to submit.

odinuge: Sounds good :+1:

It does however affect untouched fields.

It isn't easy to understand why the submit button is disabled here..
screenshot from 2017-11-23 18-26-22

We don't show field errors (even tho the error exists) if the field is untouched https://github.com/webkom/lego-webapp/blob/86f6e649ed68e8b7d15e18b7a2b2a1ca8d870f4e/app/components/Form/Field.js#L52 Redux-form "force" touches all fields inside the `handleSubmit` function https://github.com/erikras/redux-form/blob/8c6ec32936893cccbaa75bca3d1c9ca8b8603413/src/handleSubmit.js#L33 , but that doesn't matter if the button is disabled.

PR #1024, opened 2017-12-08
Add error boundary to feed

odinuge:
This fixes the error caused by feed activities
without the context (eg. the joined group is deleted).
This should therefore be properly fixed in the backend.

We can switch back to a single error boundary per feed when the backend issue is resolved. :smile:

### DIFF
--- a/app/components/Feed/index.js
+++ b/app/components/Feed/index.js
@@ -3,6 +3,7 @@ import React from 'react';
 import Activity from './activity';
 import type { AggregatedActivity } from './types';
 import EmptyState from 'app/components/EmptyState';
+import ErrorBoundary from 'app/components/ErrorBoundary';
 
 export const activityRenderers = {
   comment: require('./renders/comment'),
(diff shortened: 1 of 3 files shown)


app/components/Feed/index.js
```diff
@@ -16,13 +17,17 @@ export const activityRenderers = {
...
+
+const Feed = ({ items }: Props) => (
   <div style={{ width: '100%' }}>
     {items.length ? (
       items.map((item, i) => {
         const renders = activityRenderers[item.verb];
         return renders ? (
-          <Activity aggregatedActivity={item} key={i} renders={renders} />
+          <ErrorBoundary>
+            <Activity aggregatedActivity={item} key={i} renders={renders} />
```
ekmartin: shouldn't the `key` be passed to `ErrorBoundary` now?

odinuge: Yeah, testing it now

odinuge:
screenshot from 2017-12-08 18-51-27

Guess we don't need to show the error to the user..

ekmartin:
Yep, seems reasonable! 👍

PR #1030, opened 2017-12-16
Refactor release script

odinuge:
~~During build for previous release, not all sourcemaps was uploaded to sentry. This adds more verbose logging, and limits to 10 file uploads at once. Maybe 15-20 is a better number, but we currently have around 60 chunks, and the number is growing.~~

## Includes:
- Multistage docker build (only leaves the `RELEASE` arg to the final image)
- Upload working sourcemap for SSR
- Use sentry-cli in multistage docker build, instead of custom script
- Only one chunk on server
- Minimal docker file
- Replace release script with sentry-cli
- Alpine based image in the (330MB vs old ~1.2GB)
- Does not run `yarn build` ci task when creating docker image.
- Small speedup in docker build (~7min vs ~11min)
- Separate server and client assets output.
- Fix missing assets in client build

## TODO
 - Include all libs in webpack output, so we don't need `node_modules` in the final image. (out of scope of this PR)


odinuge:
To get SSR source references to work, we should also upload the source file (not only the sourcemap). They also need the prefix they have in the filesystem, `/app/dist/server.js{,.map}` in prod.

PR #1034, opened 2017-12-20
[WIP] Fix raven context in node for SSR

odinuge:
Add better sentry logging in SSR.

screenshot from 2017-12-22 00-46-17

## Includes:
- "Universal" Raven interface that can be used everywhere; and that works on SSR and client
- Breadcrumbs
- User context
- State dump
- lastAction
- logger: 'node' for easier filtering. `raven-js` defaults to `javascript`, while `raven` has no default.
- Return the correct request status code from SSR, depending on `state.routing.statusCode`
- Use the new `getUserContext` in raven middleware: https://github.com/captbaritone/raven-for-redux/pull/49


odinuge:
Looks like we have a "huge" memory-leak running in PROD. The `raven-for-redux` middleware adds a global reference to the store while doing SSR, becuase it is given the `raven-js` object. Did a small benchmark with `memory-usage`

Command: `ab -t 120 -c 50 -n 1000000 -k http://127.0.0.1:3000/`

The huge drop in the end is a forced gc using the chrome dev tool.

Left: this branch
Right: master
screenshot from 2018-01-08 18-41-54

odinuge:
Guess this should be more or less good to go. 

Have had some discussion with the maintainer of `raven-for-redux` in [https://github.com/captbaritone/raven-for-redux/issues/50](https://github.com/captbaritone/raven-for-redux/issues/50) about how to handle this in a general way.

#1056 should be fixed in another PR, together with a small cleanup to make it simpler. :smile:

app/utils/configureStore.js
```diff
@@ -50,26 +51,43 @@ const trackerMiddleware = createTracker({
...
 const loggerMiddleware = createLogger({
   level: 'info',
   collapsed: true
 });
 
-const messageMiddleware = createMessageMiddleware(message =>
-  addToast({ message })
-);
+export default function configureStore(
+  initialState: State | {||},
```
ekmartin: o.o what is `{||}`

app/utils/configureStore.js
```diff
@@ -50,26 +51,43 @@ const trackerMiddleware = createTracker({
...
+  Raven: any
```
ekmartin: https://github.com/flowtype/flow-typed/blob/master/definitions/npm/raven-js_v3.17.x/flow_v0.38.x-/raven-js_v3.17.x.js

app/utils/configureStore.js
ekmartin: ah, a sealed empty object

server/render.js
```diff
@@ -53,62 +76,68 @@ function render(req, res, next) {
...
+        <Provider store={store}>
+          <RouterContext {...renderProps} createElement={createElement} />
+        </Provider>
+      );
+
+      const reportError = (error: Error) => {
+        try {
+          // $FlowFixMe
+          const err = error.error ? error.payload : error;
+          log.error(err, 'render_error');
```
ekmartin: maybe the log should at least be outside the try catch, so we don't completely ignore the error

app/utils/universalRaven.js
```diff
@@ -0,0 +1,28 @@
+//@flow
+
+export interface UniversalRaven {
```
ekmartin: couldn't this be replaced with [raven-node](https://docs.sentry.io/clients/node/) if we only use it on the server anyway?

ekmartin:
Good job finding the memory leak :) This looks good to me, with a few small nitpicks.

app/utils/configureStore.js
odinuge: Yeah, tried to use that one.. But it has terrible typeings, together with function signatures that returns `this`. It is therefore only compatible with "itself", and cannot be used as an interface for a generic raven client (one for ssr/node and one for client).

app/utils/universalRaven.js
odinuge: **TL;DR**: Nope...

Because we use `react-prepare` and a lot of promises to do things, it is impossible to handle it directly using raven node. The promises "forks" away from the main thread, making it impossible for raven-node to understand what is connected to what. The reason we can use `raven` like we does in the client, is because is uses the global scope to store breadcrumbs/contexts etc. This approach works by binding the `UniversalRavenNode` functions to the context defined here (using the implicit binding of arrow functions): https://github.com/webkom/lego-webapp/pull/1034/files#diff-bec864430eb5752a683c8798d0c6bd6cR79. Then it passes the "custom" raven instance to all the components that use it during ssr.

I have used way too much time on this PR now :laughing: Still having some issues with sourcemaps when the `ENVIRONMENT` isn't development, so I guess I have to investigate it some more..


__PS__: The `RavenOptions` is from the flow typeing you mentioned over. :smile:

server/render.js
odinuge: Yeah. But I think it is nice to guarantee that `reportError` is a nothrow. It is kinda boring if our error reporting function throws an exception, resulting in a timeout for the user.

server/render.js
```diff
@@ -11,33 +12,55 @@ import serialize from 'serialize-javascript';
...
+      );
     })
   ]);
-}
 
-function render(req, res, next) {
+function render(req: $Request, res: $Response, next: Middleware) {
+  const render = (
+    body: string = '',
+    state: State | {||} = Object.freeze({})
```
odinuge: The sealed empty object looked like a good idea, but this `Object.freeze` looks aweful :laughing:

server/render.js
```diff
@@ -11,33 +12,55 @@ import serialize from 'serialize-javascript';
...
-      throw TimeoutError;
+      throw new TimeoutError(
+        'React prepare timeout when server side rendering.'
+      );
     })
   ]);
-}
 
-function render(req, res, next) {
+function render(req: $Request, res: $Response, next: Middleware) {
```
odinuge: Guess we should add stuff like request path to the sentry context.. Hmm...

app/utils/universalRaven.js
```diff
@@ -0,0 +1,31 @@
...
+  captureException(ex: Error, options?: RavenOptions): *;
+  setDataCallback(data: any, orig?: any): *;
+  captureException(ex: Error, options?: RavenOptions): *;
+}
+
+export const createNewRavenInstance = (Raven: any): UniversalRaven =>
+  Raven.context(() => new UniversalRavenNode(Raven));
```
odinuge: @ekmartin 

This makes it a lot cleaner. Now you can call this to get a "custom" raven instance that can be passed around to all components dealing with the current request. Looks like it works well, but it should be tested some more I guess.

ekmartin:
Feel free to merge whenever you feel it's done, looks good :) if some of the sourcemap stuff is still weird you can always fix that later.

PR #1039, opened 2018-01-11
Penalties improved

### LINKED ISSUE #968: Show penalty expiration date in user profile

cherrybeans:
screenshot from 2018-01-11 23-15-42
screenshot from 2018-01-11 23-16-44

Fixes #968

### DIFF
--- a/app/routes/users/components/Penalties.js
+++ b/app/routes/users/components/Penalties.js
@@ -2,14 +2,13 @@
 
 import React from 'react';
 import moment from 'moment-timezone';
-import Time, { FormatTime } from 'app/components/Time';
-
-
+import { FormatTime } from 'app/components/Time';
 
 type Penalty = {
   id: number,
   reason: string,
-  weight: number
+  weight: number,
+  exactExpiration: string
 };
 
 type Props = {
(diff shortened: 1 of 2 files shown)


app/routes/users/components/Penalties.js
```diff
@@ -22,11 +25,17 @@ function Penalties({ penalties }: Props) {
...
         const word = penalty.weight > 1 ? 'prikker' : 'prikk';
         return (
-          <li key={penalty.id}>
+          <li key={penalty.id} style={{ paddingBottom: '10px' }}>
             <strong>
-              {penalty.weight} {word}:{' '}
+              Du har fått {penalty.weight} {word}
             </strong>
-            {penalty.reason}
+            <br />
```
odinuge: You can use `<p>`-tags instead of these breaks :smile:

app/routes/users/components/Penalties.js
```diff
@@ -22,11 +25,17 @@ function Penalties({ penalties }: Props) {
...
+            Begrunnelse: <i>{penalty.reason}</i>
+            <br />
+            Utgår:{' '}
+            <i>
+              <FormatTime time={moment(penalty.exactExpiration)} />
```
odinuge: Looks like the `<FormatTime />` component converts the `time` prop to a moment instance automatically. https://github.com/webkom/lego-webapp/blob/387b70e6bb357326d05dfa6798d5a11fe6791dbe/app/components/Time/index.js#L52

So you can write `<FormatTime time={penalty.exactExpiration} />` instead.

cherrybeans: Nice! I'll get onto that.

app/routes/users/components/Penalties.js
cherrybeans: Might be a better solution, yes.

odinuge: Ahh, i see.. Does wrapping each section in a `div` work? Otherwise, breaktags are ok

cherrybeans: Wrapping in  `div` 's did indeed work.

PR #1040, opened 2018-01-11
Clickable notifications field

liseph:
Notifications
screenshot from 2018-02-05 23-22-08

Feed
screenshot from 2018-02-05 23-22-52

Should the red text in the feed be links as well (clicking webkom webkom links to its profile etc.)? They are in the feed, but not in notifications now, but not sure if it's better to be consistent.

### DIFF
--- a/app/components/Feed/context.js
+++ b/app/components/Feed/context.js
@@ -11,31 +11,39 @@ export function lookupContext(
   return aggregatedActivity.context[key];
 }
 
-const renderUser = (context: Object) => (
-  <Link to={`/users/${context.username}/`}>
-    {`${context.firstName} ${context.lastName}`}
-  </Link>
-);
+export const linkAndText = (
+  link: string,
+  text: string,
+  notLink: boolean = false
+) => ({
+  link,
+  text,
+  notLink
+});
 
-const renderEvent = (context: Object) => (
-  <Link to={`/events/${context.id}/`}>{`${context.title}`}</Link>
-);
+const renderUser = (context: Object) =>
+  linkAndText(
+    `/users/${context.username}/`,
+    `${context.firstName} ${context.lastName}`
+  );
 
-const renderMeetingInvitation = (context: Object) => (
-  <Link to={`/meetings/${context.meeting.id}/`}>{context.meeting.title}</Link>
-);
+const renderEvent = (context: Object) =>
+  linkAndText(`/events/${context.id}/`, context.title);
 
-const renderArticle = (context: Object) => (
-  <Link to={`/articles/${context.id}/`}>{context.title}</Link>
-);
+const renderMeetingInvitation = (context: Object) =>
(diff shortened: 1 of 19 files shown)


finnss:
Looks good in general!

About whether the red links should be clickable in addition to the whole notification: it's subjective, but I think we should at least be consistent in whatever we choose. I understand the argument that the dropdown and the frontpage feed are different features, but I still feel like they're similar enough that the red links should be clickable in either both or neither. 

Also, if we decide the red links *should* be clickable, there needs to be some hover effect for proper ux. For example, add a text underline on hover.

finnss:
I kinda feel like the `/timeline/` page is both ugly and unnecessary. Doesn't feel like it adds much?

skjermbilde 2018-02-08 kl 18 54 46


That leaves the question of what announcements should link to. It might be out of scope of this PR (we can easily merge and then change this later), but I still personally feel like it would be fun to have a announcement detail page – with comments 😄 "hold kjeft syll det er like lang kø for alle"

finnss:
How did the links to comments turn out? Does it jump to just the event/whatever, the comment section, or the comment itself?

odinuge:
Moving the screen to the comment mentioned in the notification/feed has to be done inside/together with the router. We now have several reasons to switch to react-router v4 (eg. the hot-reloading, and modals when leaving dirty forms), and imo., it is better to wait till after we do that.

app/components/Feed/context.js
```diff
@@ -44,3 +53,16 @@ export const contextRender = {
...
+
+export function toLink(linkAndText: linkAndText) {
+  return linkAndText.notLink ? (
+    toSpan(linkAndText)
+  ) : (
+    <Link to={linkAndText.link}>{linkAndText.text}</Link>
+  );
+}
+
+export function toSpan(linkAndText: linkAndText) {
```
odinuge: What do you mean with the fact that the parameter `linkToText` has the type `linkToText` (the same name)? Since the global `linkAndText` is an object, i guess that is the reason why flow doesn't care.

app/components/Feed/renders/adminRegistration.js
```diff
@@ -8,7 +8,10 @@ import type { AggregatedActivity } from '../types';
 /**
  * Normal grouping by target and date
  */
-export function activityHeader(aggregatedActivity: AggregatedActivity) {
+export function activityHeader(
+  aggregatedActivity: AggregatedActivity,
+  htmlTag: Function => string
```
odinuge: Isn't the flow typing for `htmlTag` here supposed to be something like:
```jsx
htmlTag: ({
  link: string,
  text: string,
  notLink: boolean
}) => React.Node<*>
```

Since you use this a lot, you can create a custom type ala:

```j
type MyType = {
  link: string,
  text: string,
  notLink: boolean
};

Issue #986, opened 2017-11-13
Surveys


finnss:
Add surveys to Abakus.no, giving us a way to gather feedback after events.

Requires https://github.com/webkom/lego/pull/525. 

I'm using this PR to track progress across pull requests.

- [x] Actions to fetch, add, edit, and delete surveys. (#901)
- [x] Reducer for surveys including selectors and EntityTypes. (#901)
- [x] Routes and components for list view, detail view, editing, and adding surveys. (#901)
- [x] All necessary intermediate files such as the `index.js` making the feature available at  `/surveys/`. (#901)
- [x] Components to render actual forms: Question, Answer, and Option.
- [x] Using those components in detail view, and making them editable and sendable in `SurveyEditor`.
- [x] Actions, Routes, and components for adding Submissions, so that forms can actually be answered.
- [x] Actions, Routes, and components for viewing Submissions, so that those answers can be read.
- [x] Making surveys from templates 
- [ ] Generating graphs or charts or some such to make viewing Submissions manageable. 
- [x] Functionality for sending mails to attendees of an event with link to the Survey
- [x] Tie surveys closer with events somehow. At least a link for the event admins to make a survey from an event page, which means I also need functionality to pass an event to the "add survey" page.

PR #1058, opened 2018-01-22
Add and Edit Surveys

finnss:
This PR continues the Survey project #986, checking the next two todo items:
- [x] Components to render actual forms: Question, Answer, and Option.
- [x] Using those components in detail view, and making them editable and sendable in SurveyEditor.

With this PR, the Editor ( `surveys/add` and `surveys/someId/edit`) should be mostly finished.

skjermbilde 2018-01-22 kl 20 45 54

skjermbilde 2018-01-22 kl 20 46 47

In addition, a basic display of the questions in each survey has been added to the detail page.

skjermbilde 2018-01-22 kl 20 52 40

### DIFF
--- a/app/routes/surveys/components/Question.js
+++ b/app/routes/surveys/components/Question.js
@@ -21,28 +21,28 @@ type State = {
 };
 
 type QuestionTypeSelectorProps = {
-  questionType: number
+  questionType: string
 };
 
 function QuestionTypeSelector({ questionType }: QuestionTypeSelectorProps) {
-  const questionTypeIntToString = {
-    '1': 'Multiple Choice',
-    '2': 'Sjekkboks',
-    '3': 'Fritekst'
+  const questionTypeToString = {
+    single_choice: 'Multiple Choice',
+    multiple_choice: 'Sjekkboks',
+    text_field: 'Fritekst'
   };
-  const questionTypeIntToIcon = {
-    '1': 'radio-button-on',
-    '2': 'checkbox',
-    '3': 'more'
+  const questionTypeToIcon = {
+    single_choice: 'radio-button-on',
+    multiple_choice: 'checkbox',
+    text_field: 'more'
   };
   return (
     <Link className={styles.questionTypeSelector}>
       <span style={{ width: '170px', display: 'inline-block' }}>
         <Icon
-          name={questionTypeIntToIcon[String(questionType)]}
+          name={questionTypeToIcon[questionType]}
           style={{ marginRight: '15px' }}
         />
-        {questionTypeIntToString[String(questionType)]}
+        {questionTypeToString[questionType]}
       </span>
       <Icon name="arrow-down" size={24} className={styles.typeIcon} />
     </Link>
@@ -50,26 +50,26 @@ function QuestionTypeSelector({ questionType }: QuestionTypeSelectorProps) {
 }
 
 type DropdownItemsProps = {
-  setQuestionType: number => void
+  setQuestionType: string => void
 };
 
 function DropdownItems({ setQuestionType }: DropdownItemsProps) {
   return (
     <Dropdown.List>
       <Dropdown.ListItem>
-        <Link onClick={() => setQuestionType(1)}>
+        <Link onClick={() => setQuestionType('single_choice')}>
           <strong>Multiple Choice</strong>
           <Icon name="radio-button-on" size={24} />
         </Link>
       </Dropdown.ListItem>
       <Dropdown.ListItem>
-        <Link onClick={() => setQuestionType(2)}>
+        <Link onClick={() => setQuestionType('multiple_choice')}>
           Sjekkboks
           <Icon name="checkbox" size={24} />
         </Link>
       </Dropdown.ListItem>
       <Dropdown.ListItem>
-        <Link onClick={() => setQuestionType(3)}>
+        <Link onClick={() => setQuestionType('text_field')}>
           Fritekst
           <Icon name="more" size={24} />
         </Link>
@@ -101,7 +101,7 @@ class Question extends Component<Props, State> {
     }));
   };
 
-  setQuestionType = (questionType: number) => {
+  setQuestionType = (questionType: string) => {
     this.props.updateQuestion(
       { ...this.props.question, questionType },
       this.props.index
@@ -129,7 +129,7 @@ class Question extends Component<Props, State> {
               }
             />
           </div>
-          {this.props.question.questionType === 3 ? (
+          {this.props.question.questionType === 'text_field' ? (
             <TextArea
               className={styles.freeText}
               placeholder="Fritekst"
--- a/app/routes/surveys/components/Question.js
+++ b/app/routes/surveys/components/Question.js
@@ -8,6 +8,7 @@ import styles from './surveys.css';
 import Icon from 'app/components/Icon';
 import { Link } from 'react-router';
 import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
+import { QuestionTypes, PresentableQuestionType } from '../utils';
 
 type Props = {
   updateQuestion: (Object, number) => void,
@@ -25,11 +26,6 @@ type QuestionTypeSelectorProps = {
 };
 
 function QuestionTypeSelector({ questionType }: QuestionTypeSelectorProps) {
-  const questionTypeToString = {
-    single_choice: 'Multiple Choice',
-    multiple_choice: 'Sjekkboks',
-    text_field: 'Fritekst'
-  };
   const questionTypeToIcon = {
     single_choice: 'radio-button-on',
     multiple_choice: 'checkbox',
@@ -42,7 +38,7 @@ function QuestionTypeSelector({ questionType }: QuestionTypeSelectorProps) {
           name={questionTypeToIcon[questionType]}
           style={{ marginRight: '15px' }}
         />
-        {questionTypeToString[questionType]}
+        {PresentableQuestionType(questionType)}
       </span>
       <Icon name="arrow-down" size={24} className={styles.typeIcon} />
     </Link>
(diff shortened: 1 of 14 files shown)


app/routes/surveys/components/Question.js
```diff
@@ -0,0 +1,237 @@
...
+import styles from './surveys.css';
+import Icon from 'app/components/Icon';
+import { Link } from 'react-router';
+import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
+
+type Props = {
+  updateQuestion: Object => void,
+  deleteQuestion: number => Promise<*>,
+  question: Object,
+  nr: number
```
ekmartin: what's an `nr` (give it a better name :p)

app/routes/surveys/components/Question.js
```diff
@@ -0,0 +1,237 @@
...
+import Option from './Option';
+import { TextInput, TextArea, CheckBox } from 'app/components/Form';
+import Dropdown from 'app/components/Dropdown';
+import styles from './surveys.css';
+import Icon from 'app/components/Icon';
+import { Link } from 'react-router';
+import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
+
+type Props = {
+  updateQuestion: Object => void,
```
ekmartin: `QuestionEntity => void`?

app/routes/surveys/components/Question.js
```diff
@@ -0,0 +1,237 @@
...
+      .map((option, i) => ({ ...option, nr: i }))
+      .concat({
+        nr: this.props.question.options.length,
+        optionText: ''
+      });
+    this.setState({ options });
+  }
+
+  componentDidUpdate() {
+    if (this.state.shouldUpdate) {
```
ekmartin: this `shouldUpdate` thing seems fishy - could you try explaining what it does?

app/routes/surveys/components/Question.js
```diff
@@ -0,0 +1,237 @@
...
+        options: this.state.options
+      });
+      this.setState({ shouldUpdate: false });
+    }
+  }
+
+  updateOptions = (option: Object) => {
+    const options = this.state.options.slice();
+
+    const oldOption = options.find(o => o.nr === option.nr);
```
ekmartin: isn't `nr` the index? do you need a `find`?

app/routes/surveys/components/Question.js
```diff
@@ -0,0 +1,237 @@
...
+    const changedOptionIndex = options.indexOf(oldOption || {});
+
+    if (changedOptionIndex !== -1) {
```
ekmartin: this logic is pretty confusing - maybe it could be written in a better way? wanna try explaining what the function's purpose is in english? (maybe put that in a comment above `updateOptions` as well?)

app/routes/surveys/components/Question.js
```diff
@@ -0,0 +1,237 @@
...
+              onInput={e =>
+                updateQuestion({
+                  ...this.props.question,
+                  questionText: e.target.value,
+                  nr
+                })
+              }
+            />
+          </div>
+          {this.props.question.questionType === 3 ? (
```
ekmartin: It'd be nice to use something enum like for these question type constants, i.e. something like

```js
const QUESTION_TYPE = {
  TEXT: 1,
  EXISTENTIAL: 2,
  BLABLA: 3
};
```

app/routes/surveys/components/Question.js
finnss: Well I've been unsure about how to handle problems like this for a while. Popping in your suggestion won't work right of the bat, since `QuestionEntity` has a couple of mandatory fields like ID that aren't contructed before the object reaches the back-end. I suppose I could make a second entity like `MakeQuestionEntity`, but that would have to be done for almost all entities and would mostly be the same as the original entities but with all fields optional. Doesn't really seem too useful to me?

app/routes/surveys/components/Question.js
finnss: Yeah I wasn't 100% sure what it would end up as when I implemented it, I just needed some identifier for objects that didn't have an assigned id from the backend yet. I agree index is a better name, just wasn't sure it would end up being akin to an array index which might have made the name confusing. Probably still better than `nr` tho

app/routes/surveys/components/Question.js
finnss: As explained above, I thought of `nr` as an ID, meaning an option would keep its `nr` even if previous elements in the `options` array were deleted.

app/routes/surveys/components/Question.js
finnss: Yeah I figured it might be a bit difficult to understand the functionality from reading the code, so I added a quick comment, but I agree it's not the most clarifying. 

The basic functionality is simply to take an object as parameter and figure out if its a brand new option or an updated version of an old one, then add or change that option in state, respectively. Then, in addition, I implemeted a (hopefully) smoother version of having a "Add another option" button. If the incoming parameter a new option and this change is the very first change to that option, that means the user just started typing something in the last empty option seen in part 1 and 2 [here](https://user-images.githubusercontent.com/14221386/35241320-60d25bd6-ffb6-11e7-8d05-7e0f3dae2a6f.png). If so, that means we need a new blank option, because that's the only way to add more options.

Was that a better explanation?

app/routes/surveys/components/Question.js
finnss: Yeah I figured we needed some more obvious representation, I just wasn't sure if the current standard we used was an object like you suggest or just changing the whole variable to a string or something else. So I held off on commiting to either one, but if you think objects like this should be the way to go then I'll gladly use that.

app/routes/surveys/components/Question.js
ekmartin: Not completely sure I understand, but maybe you were calling `setState` after mutating `options`? `setState` is sort of like Redux in that regard: you need a new reference, i.e.:

```js
// Bad:
this.state.options.push(newOption);
this.setState({ options: this.state });

// Good:
const options = [...this.state.options, newOption];
this.setState({ options });
```

I saw that you used `.slice()` to create a copy further down, but this only creates a shallow copy of the array - the option objects inside the array are still the same, so if you mutate any of those it won't matter (not sure if you do though).

https://medium.com/pro-react/a-brief-talk-about-immutability-and-react-s-helpers-70919ab8ae7c

If that's not the case we can discuss it later today, at the meeting.

finnss:
New and better – now without any `nr` or `shouldUpdate`s 8)

app/routes/surveys/components/SurveyDetail.js
```diff
@@ -35,6 +36,40 @@ const SurveyDetail = (props: Props) => {
...
+
+            {question.questionType === 3 ? (
+              <TextArea
+                value=""
+                placeholder="Fritekst..."
+                className={styles.freeText}
+              />
+            ) : (
+              <ul className={styles.detailOptions}>
+                {(question.options || []).map(option => (
```
ekmartin: I guess this one needs to be there because of the concat later on?

app/routes/surveys/components/SurveyEditor.js
```diff
@@ -72,43 +134,59 @@ class SurveyEditor extends Component<Props> {
...
-            name="activeFrom"
-            className={styles.editEvent}
-            component={DatePicker.Field}
-          />
+          <ul className={styles.questions}>
+            {this.state.questions.map((question, i) => (
+              <Question
+                key={i}
+                index={i}
+                question={question || {}}
```
ekmartin: can question be false here? do you need the || {}?

app/routes/surveys/components/SurveyEditor.js
```diff
@@ -72,43 +134,59 @@ class SurveyEditor extends Component<Props> {
...
+                updateQuestion={this.updateQuestion}
+                deleteQuestion={this.deleteQuestion}
+              />
+            ))}
+          </ul>
+
+          <Link
+            onClick={() => {
+              const questions = this.state.questions.slice();
```
ekmartin: this should be a concat/spread fo sure:
```js
const newQuestion = { stuff };
const questions = [...this.state.questions, newQuestion];
this.setState({ questions });
```

ekmartin:
looking good :)

app/routes/surveys/components/SurveyEditor.js
```diff
@@ -3,42 +3,104 @@
...
 import {
   TextInput,
-  RadioButton,
+  CheckBox,
   SelectInput,
-  RadioButtonGroup,
   DatePicker
 } from 'app/components/Form';
 import { createValidator, required } from 'app/utils/validation';
 import { reduxForm } from 'redux-form';
```
odinuge: legoForm :smile:

package.json
```diff
@@ -23,8 +27,10 @@
     "lint:prettier": "prettier '**/*.{js,css,md}' --list-different",
```
odinuge: Looks like the changes in here are because of prettier. You can add `json` into both the `lint:prettier` command and the `prettier` command here, so we ensure proper formatting. :smile:

app/routes/surveys/components/SurveyEditor.js
```diff
@@ -3,42 +3,104 @@
...
+    questions: []
+  };
+
+  componentDidMount() {
+    const questions =
+      this.props.survey.questions.length === 0
+        ? [
+            {
+              questionText: '',
+              questionType: 'single_choice',
```
odinuge: It would be nice to have all of these "constants" in one place, so you don't have to copy strings all the time while coding. Will help type-safety aswell

app/routes/surveys/components/SurveyEditor.js
```diff
@@ -3,42 +3,104 @@
...
+      questions.push(question);
+    }
+    this.setState({ questions });
+  };
+
+  deleteQuestion = (index: number) => {
+    this.setState(state => ({
+      questions: state.questions.filter((question, i) => i !== index)
+    }));
+    return Promise.resolve();
```
odinuge: :+1: Keeping the modal happy :smile:

PR #1060, opened 2018-01-22
Add a field that displays the event creator and make company name a link

SmithPeder:
Depends on https://github.com/webkom/lego/pull/1019

### DIFF
--- a/app/routes/events/components/EventDetail/index.js
+++ b/app/routes/events/components/EventDetail/index.js
@@ -28,6 +28,7 @@ import {
   ContentSidebar
 } from 'app/components/Content';
 import type { ID } from 'app/models';
+import { Link } from 'react-router';
 
 type InterestedButtonProps = {
   isInterested: boolean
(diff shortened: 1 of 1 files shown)


app/routes/events/components/EventDetail/index.js
```diff
@@ -141,7 +142,19 @@ export default class EventDetail extends Component<Props> {
...
         key: 'Arrangerende bedrift',
-        value: event.company.name
+        value: (
+          <Link to={`/companies/${event.company.id}`}>
+            {event.company.name}
+          </Link>
+        )
+      },
+      event.createdBy && {
+        key: 'Ansvarlig',
```
finnss: Is the creator of an event always responsible for it? Does this also apply to events for interest groups and other such non-standard cases? Not a huge deal of course, but I can imagine Bedkom requesting the ability to change who's responsible if we include this functionality

SmithPeder: @edvardvb Finn got a point here

edvardvb: Could just use "forfatter", but I believe we used a similar term on nerd, and Bedkom never had an issue with that. Maybe check the old repo and use the same term?

SmithPeder: Digging through NERD was to hard. So i agree that we can just go with "forfatter", and if Bedkom wants something else we'll just change it

Issue #1045, opened 2018-01-15
Navigation autocomplete search should display and sort by date


finnss:
Right now searching for "Bekk" yields a bunch of results with no indication of whether it's an event from 4 years ago, the company's page, or some other content. A first step to solve this should be to display the most recent first, as that's likely the most relevant. That information should also be visible to the user. 

This sorting should be done back-end because of pagination, and if the date isn't already being sent then that needs to be added to the serializer as well.

odinuge:
This is partly solved by #1063, and will be further improved with better search results :smile:

PR #1063, opened 2018-01-23
Make the autocomplete search more user friendly

JonasBak:
This is a bit WIP, but i wanted to get some feedback, working on #1045
Events in the list of results will now show information about when it starts
acin
scago

### DIFF
--- a/app/components/Search/Search.css
+++ b/app/components/Search/Search.css
@@ -43,7 +43,7 @@
 }
 
 .searchResultItemIcon {
-  margin-right: 10px;
+  margin-right: 15px;
 }
 
 .closeButton {
--- a/app/components/Search/Search.css
+++ b/app/components/Search/Search.css
@@ -44,6 +44,7 @@
 
 .searchResultItemIcon {
   margin-right: 15px;
+  margin-left: 3px;
 }
 
 .closeButton {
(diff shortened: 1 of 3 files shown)


finnss:
So you've only changed results that are events? Do you have any idea about what we should do about other content such as companies, if anything? Is that information available for display? Just saying that the result is an event or a user or whatever could be a big step.

JonasBak:
So far I've only added the date for events, I'm thinking of adding some text about pressing enter for a complete search. You can currently distinguish different types of results (events, users, companies...) by their logo, but the briefcase used for companies and the calendar used for events are too similar imo. We could easily add some more text somewhere to say what type of result you get.

app/components/Search/SearchResults.js
```diff
@@ -33,6 +34,13 @@ const SearchResultItem = ({
           <Icon className={styles.searchResultItemIcon} name={result.icon} />
         )}
       {result.label}
+      {result.date && (
+        <Time
+          time={Date.parse(result.date)}
```
odinuge: It is enough to just write `time={result.date}`, since it is converted to the proper moment-object (with "our" timezone) inside the `<Time />` component.

app/components/Search/SearchResults.js
```diff
@@ -33,6 +34,13 @@ const SearchResultItem = ({
...
+          wordsAgo
+          className={styles.resultDate}
```
odinuge: If we add other information in this spot, it would be better with a more generic name. If we only use it for dates/time, this is nice! :smile:

JonasBak:
Now you also see what type of result you get in the search, and events in SelectInput display what date the event is/was.
screenshot from 2018-01-29 21-02-05
screenshot from 2018-01-29 21-02-34

app/components/Search/Search.css
```diff
@@ -140,12 +141,33 @@
 .resultDate {
   color: var(--lego-font-color);
   margin-left: auto;
+  white-space: nowrap;
```
finnss: Make sure to check stuff like this on Safari and its iOS developer mode. Should be good but those have acted up in the past, and making sure the site looks good on iPhone is pretty important.

Issue #1015, opened 2017-11-19
Remove all unnecessary enableReinitialize in redux-form


odinuge:
Remove all unnecessary enableReinitialize in redux-form.

It's much better to "wait" sending initialValues to the form until all data exists. It may be useful in some cases, but not always...

https://redux-form.com/7.1.2/docs/api/reduxform.md/#-code-enablereinitialize-boolean-code-optional-

PR #1066, opened 2018-01-24
Survey submissions

finnss:
Adds submission handling to surveys: Users can answer surveys and admins can view them.

Continues the Survey project #986: 
- [x] Actions, Routes, and components for adding Submissions, so that forms can actually be answered.
- [x] Actions, Routes, and components for viewing Submissions, so that those answers can be read.

Also does some general fixes to surveys, such as moving admin functionality from the navigation menu to a sidebar.

skjermbilde 2018-01-24 kl 20 56 25

skjermbilde 2018-01-24 kl 20 56 34

Pictures not representative of the finished product.

### DIFF
--- a/app/reducers/surveySubmissions.js
+++ b/app/reducers/surveySubmissions.js
@@ -1,0 +1,64 @@
+// @flow
+
+import { SurveySubmission } from '../actions/ActionTypes';
+import createEntityReducer from 'app/utils/createEntityReducer';
+import { createSelector } from 'reselect';
+import type { OptionEntity, SurveyEntity, QuestionEntity } from './surveys';
+import type { UserEntity } from 'app/reducers/users';
+
+export type SubmissionEntity = {
+  id: number,
+  user: UserEntity,
+  survey: SurveyEntity,
+  submitted: boolean,
+  submittedTime?: string,
+  answers: Array<AnswerEntity>
+};
+
+export type AnswerEntity = {
+  id: number,
+  submission: SubmissionEntity,
+  question: QuestionEntity,
+  answerText: string,
+  selectedOptions: Array<OptionEntity>
+};
+
+function mutateSurveySubmissions(state, action) {
+  switch (action.type) {
+    case SurveySubmission.FETCH.SUCCESS: {
+      return {
+        ...state,
+        items: state.items.filter(id => id !== action.meta.id)
+      };
+    }
+
+    default:
+      return state;
+  }
+}
+
+export default createEntityReducer({
+  key: 'surveySubmissions',
+  types: {
+    fetch: SurveySubmission.FETCH,
+    mutate: SurveySubmission.ADD
+  },
+  mutate: mutateSurveySubmissions
+});
+
+export const selectSurveySubmissions = createSelector(
+  (state, props) => props.surveyId,
+  state => state.surveySubmissions.items,
+  state => state.surveySubmissions.byId,
+  state => state.users.byId,
+  (surveyId, surveySubmissionIds, surveySubmissionsById, usersById) =>
+    surveySubmissionIds
+      .map(surveySubmissionId => {
+        const submission = surveySubmissionsById[surveySubmissionId];
+        return {
+          ...submission,
+          user: usersById[submission.user]
+        };
+      })
+      .filter(surveySubmission => surveySubmission.survey === surveyId)
+);
--- a/app/routes/surveys/components/AdminSideBar.js
+++ b/app/routes/surveys/components/AdminSideBar.js
@@ -20,8 +20,7 @@ const AdminSideBar = ({
   actionGrant,
   push
 }: Props) => {
-  // const canEdit = actionGrant.includes('edit');
-  const canEdit = true;
+  const canEdit = actionGrant.includes('edit');
 
   const onConfirm = () => {
     if (deleteFunction && push) {
@@ -42,7 +41,10 @@ const AdminSideBar = ({
         <strong>Admin</strong>
         <ul>
           <li>
-            <Link to={`/surveys/${surveyId}/edit`}>Endre</Link>
+            <Link to="/surveys/add">Ny undersøkelse</Link>
+          </li>
+          <li>
+            <Link to={`/surveys/${surveyId}/edit`}>Endre undersøkelsen</Link>
           </li>
           {deleteFunction && (
             <ConfirmModalWithParent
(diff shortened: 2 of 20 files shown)


finnss:
I think this should be ready for some review!

app/routes/surveys/components/AdminSideBar.js
```diff
@@ -0,0 +1,64 @@
...
+  actionGrant,
+  push
+}: Props) => {
+  // const canEdit = actionGrant.includes('edit');
+  const canEdit = true;
+
+  const onConfirm = () => {
+    if (deleteFunction && push) {
+      return (
+        deleteFunction &&
```
odinuge: Isn't `push` and `deleteFunction` always supposed to be truthy (to be functions), since they are always mapped into props in the `mapDispatchToProps`?
```jsx
type Props = {
  surveyId: number,
  deleteFunction: number => Promise<*>,
  push: string => void,
  actionGrant: ActionGrant
};
```

app/routes/surveys/components/SubmissionEditor.js
```diff
@@ -0,0 +1,140 @@
...
+          };
+        })
+        .filter(answer => answer)
+    };
+
+    return submitFunction(toSubmit);
+  };
+
+  if (fetching || !survey || !survey.event || !survey.event.cover) {
+    return <LoadingIndicator loading />;
```
odinuge: Not a big deal, but this could be replaced with the `loggedIn` HOC inside the route.

app/routes/surveys/components/SubmissionEditor.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+const validate = createValidator({
+  user: [required()]
+});
+
+export default reduxForm({
+  form: 'submissionEditor',
+  validate,
+  enableReinitialize: true
```
odinuge: Is there a reason you use this? ref. #1015

app/routes/surveys/components/SubmissionIndividual.js
```diff
@@ -0,0 +1,89 @@
...
+type Props = {
+  submissions: Array<SubmissionEntity>,
+  addSubmission: SubmissionEntity => Promise<*>,
+  deleteSurvey: number => Promise<*>,
+  survey: SurveyEntity
+};
+
+const SubmissionPage = ({ submissions, deleteSurvey, survey }: Props) => {
+  if (!submissions || !survey || !survey.event)
+    return <LoadingIndicator loading />;
```
odinuge: This can also be solved via the `loadingIndicator`-util inside the route.

app/routes/surveys/components/SubmissionsPage.js
```diff
@@ -0,0 +1,66 @@
...
+  submissions: Array<SubmissionEntity>,
+  addSubmission: SubmissionEntity => Promise<*>,
+  deleteSurvey: number => Promise<*>,
+  survey: SurveyEntity,
+  children: React.Element<*>,
+  actionGrant: ActionGrant,
+  isSummary: boolean
+};
+
+const SubmissionPage = (props: Props) => {
```
odinuge: The file is called `SubmissionsPage.js`, and this is called `SubmissionPage` without an `s`.. Typo?

app/reducers/surveySubmissions.js
```diff
@@ -0,0 +1,64 @@
...
+});
+
+export const selectSurveySubmissions = createSelector(
+  (state, props) => props.surveyId,
+  state => state.surveySubmissions.items,
+  state => state.surveySubmissions.byId,
+  state => state.users.byId,
+  (surveyId, surveySubmissionIds, surveySubmissionsById, usersById) =>
+    surveySubmissionIds
+      .map(surveySubmissionId => {
```
odinuge: Hmmm. This will map over all submissions in the state, and then filter... With the current implementation, this looks like the only solution, but i guess we can do better?

Maybe we can:
- Add a list of all `submissionIds` into the submission.
- Or maybe do something like this :thinking:  https://github.com/webkom/lego-webapp/blob/106e835d7f08d4f845e0e74bf985ba93832bae42/app/reducers/index.js#L176

app/routes/surveys/components/SubmissionEditor.js
```diff
@@ -0,0 +1,140 @@
...
+      </form>
+    </Content>
+  );
+};
+
+const validate = createValidator({
+  user: [required()]
+});
+
+export default reduxForm({
```
odinuge: `#legoForm`

app/routes/surveys/components/AdminSideBar.js
finnss: Looks like I messed up those Props. I intended to have `deleteFunction` as an optional prop so that not every component that wanted to use this sidebar wouldn't need to include the deleteFunction in its route and pass it down and all that.

odinuge: Yeah, that looks like the reason. Because then flow will force you to write it like this

app/routes/surveys/components/SubmissionEditor.js
finnss: It's kind of a workaround, since it seemed like `survey` gets populated as an object without these values in early renders. I don't know if the Route-based `loadingIndicator` handles that. This solution doesn't look pretty though, I'll give you that, it was simply my response to getting `survey.event is not defined` despite having checked that `survey` is defined.

app/routes/surveys/components/SubmissionEditor.js
finnss: Nope, copypasta. Will remove.

app/routes/surveys/components/SubmissionsPage.js
finnss: Heh yeah, good point. I'll choose one and stick to it.

app/routes/surveys/components/SubmissionEditor.js
odinuge: It should be possible to do it like this:
https://github.com/webkom/lego-webapp/blob/46191ef7cae05cb65e60362b5dba8689a1ecb423/app/routes/admin/email/EmailUsersRoute.js#L22
```jsx
loadingIndicator(['survey.event.cover', 'notFetching']) 
```

app/routes/surveys/components/SubmissionEditor.js
odinuge: You can try to refresh while in this route, to see if that still works.

app/routes/surveys/components/SubmissionsPage.js
odinuge: Aaand, there is another `SubmissionPage`

PR #1067, opened 2018-01-24
Use timezone when formatting date/time in datepicker

### LINKED ISSUE #1050: Event edit seems to ignore timezones when server rendering
Try going to https://abakus.no/events/2145/, click edit and then refresh. 

With client-side render:
image

With server-side render:
image

Couldn't reproduce locally though (with server-rendering), so might be a configuration related issue.

odinuge:
The containers we run the webapp in use "no" timezone, aka. they just
use +00:00. This results in strange behavior when the server side
renderer formats the date/time on the server. When the datepicker is
re-rendered in the browser, it use out timezone (Europe/Oslo). This
results in a wrong date label, and an annoying flash when clicking
around. :clock1: :clock2: :clock3: 

image
Source: [youtube](https://www.youtube.com/watch?v=-5wpm-gesOY)

This fixes #1050. Nice catch @ekmartin. This will save all of us (including bedkom & fagkom) from scratching our head, wondering why the hell we shifted all the timestamps with 1 (or 2) hour(s).

**PS**: Do we format time other places, without using the `<Time />` component?? :thinking:

### DIFF
--- a/app/components/Form/DatePicker.js
+++ b/app/components/Form/DatePicker.js
@@ -10,6 +10,7 @@ import { createField } from './Field';
 import TextInput from './TextInput';
 import TimePicker from './TimePicker';
 import styles from './DatePicker.css';
+import config from 'app/config';
 
 type Props = {
   onChange: string => void,
@@ -103,7 +104,9 @@ class DatePicker extends Component<Props, State> {
         triggerComponent={
           <TextInput
             className={cx(styles.inputField, className)}
-            value={this.state.value.format(this.props.dateFormat)}
+            value={moment
+              .tz(this.state.value, config.timezone)
+              .format(this.props.dateFormat)}
           />
         }
         componentClass="div"


odinuge:
screenshot from 2018-01-24 23-11-02
Left: Container running current `master`
Right: Container running `this` branch

ekmartin:
Nice! Grepping after `moment` reveals a few other places that probably should include timezones. Maybe we could replace all this with a call to [`moment.tz.setDefault()`](https://momentjs.com/timezone/docs/#/using-timezones/default-timezone/) in `server.js`? In the frontend we probably don't need to set the timezone at all, like you're saying.

odinuge:
Yeah! Sounds like a good idea to use the `moment.tz.setDefault()` to force (Europe/Oslo) for everyone - even on frontend. Without much investigation, I imagine it will be hard to get the clients timezone in SSR, without some kind of strange guessing. Imo forcing our own timezone (more or less what we do) looks like a good solution for our needs.

PR #1072, opened 2018-01-28
Add pretty console output for tty

odinuge:
Use a pretty output instead of json when the server is running in a tty
(valid terminal). Will still output json when piped, written to a file
or just running in headless.

screenshot from 2018-01-28 13-40-31
Left: json
Right: pretty-print

This is a way better approach than just using __DEV__ or NODE_ENV,
becuase it also prints pretty when running a containerized version of
the server with an pseudo-TTY.
```bash
$ docker run -t lego-webapp   # pretty
$ docker run lego-webapp      # json
$ node dist/server.js         # pretty
$ node dist/server.js | cat   # json
$ yarn start                  # pretty
$ yarn start | cat            # json
```

This can also be achived using: `yarn start | ./node_modules/.bin/bunyan`, but that is rather annoying. When you have a tty, you will always prefer pretty-print over json. :tada:

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@webkom/react-prepare": "^0.5.3",
     "animate.css": "3.5.2",
     "bunyan": "^1.8.12",
+    "bunyan-pretty": "^0.0.1",
     "child-process-promise": "^2.2.1",
     "classnames": "2.2.5",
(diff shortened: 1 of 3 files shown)


ekmartin:
Nice 👍

PR #1088, opened 2018-02-05
Refactor Image component

ekmartin:
I'm not completely sure, but I think we were doing the `componentDidMount` stuff to add an opacity transition to the images, however since we weren't setting `opacity: 0` this wasn't actually happening. This removes it altogether and simplifies the component. That also fixes a bug where we'd try to load an undefined image while the `src` prop was getting ready.

### DIFF
--- a/app/components/Image/Image.js
+++ b/app/components/Image/Image.js
@@ -7,47 +7,25 @@ import styles from './Image.css';
 type Props = {
   src: string,
   className?: string,
-  alt: string
+  alt: string,
+  style: Object
 };
 
-type State = {
-  loaded: boolean
-};
-
-class Image extends Component<Props, State> {
+class Image extends Component<Props> {
   static defaultProps = {
     alt: 'image'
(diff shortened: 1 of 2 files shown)


app/components/Image/Image.js
```diff
@@ -7,47 +7,25 @@ import styles from './Image.css';
...
-
-  componentWillUnmount() {
-    this._isMounted = false;
-  }
-
   render() {
-    const { src, className, alt, ...props } = this.props;
+    const { src, className, alt, style, ...props } = this.props;
+    const finalClass = cx(styles.image, className);
+    if (!src) return <div className={finalClass} style={style} />;
```
odinuge: Smooth!

We should try to switch almost all images to this

ekmartin: we could also show a specific thing while the image itself loading by using `onLoad` on the `img`, but not sure if it's worth it

PR #1090, opened 2018-02-05
Show a loading indicator on the frontpage

### LINKED ISSUE #1053: Frontpage loading
Don't display `INGEN OPPSLAG` while loading content for frontpage. A loading indicator would be better.

output

ekmartin:
(doesn't happen with the server side render, so usually you won't see it) - fixes #1053


app/routes/overview/components/Overview.js
```diff
@@ -21,24 +22,7 @@ const IMAGE_HEIGHT = 192;
 
 const itemUrl = item => `/${item.eventType ? 'events' : 'articles'}/${item.id}`;
 
-function PrimaryItem({ item }: { item?: Event | Article }) {
-  if (!item) {
```
ekmartin: removed this, it's not really that useful

PR #1098, opened 2018-02-06
Add initial support for public events

### LINKED ISSUE #1096: Update event detail for new permission system
Depends on this: https://github.com/webkom/lego/pull/1027

- Remove login-requirement on both detail and list
- Make the detail view pretty when not showing registrations
  - Show number of registrations in each pool, but not users (also waiting regs)
  - Show "login to register" when needed
  - Hide all components about registered users (icons, lists of registered users etc)
- Add is_abakom_only checkbox in editor

odinuge:
This fixes #1096.

### DIFF
--- a/app/components/UserAttendance/AttendanceStatus.js
+++ b/app/components/UserAttendance/AttendanceStatus.js
@@ -8,35 +8,53 @@ import type { EventPool } from 'app/models';
 type AttendanceElementProps = {
   pool: EventPool,
   index: number,
-  toggleModal: number => void
+  toggleModal: number => void,
+  useModal: boolean,
+  isCount: boolean
 };
 
 const AttendanceElement = ({
-  pool,
+  // $FlowFixMe
+  pool: { name, registrations = 0, capacity = '∞' },
   index,
-  toggleModal
+  toggleModal,
+  useModal,
+  isCount
 }: AttendanceElementProps) => {
-  const capacity = pool.capacity ? pool.capacity : '∞';
+  const registrationCount =
+    isCount || !registrations ? registrations : registrations.length;
+  const Status = () => <strong>{`${registrationCount}/${capacity}`}</strong>;
+
   return (
     <div className={styles.poolBox}>
-      <strong>{pool.name}</strong>
-      <a onClick={() => toggleModal(index)}>
-        <strong>
-          {pool.registrations
-            ? `${pool.registrations}/${capacity}`
-            : `0/${capacity}`}
-        </strong>
-      </a>
+      <strong>{name}</strong>
+      {useModal ? (
+        <a onClick={() => toggleModal(index)}>
+          <Status />
+        </a>
+      ) : (
+        <Status />
+      )}
     </div>
   );
 };
 
 export type AttendanceStatusProps = {
   pools: Array<EventPool>,
-  toggleModal: number => void
+  /*
+   * The elemnts will only be clickabel if 'toggleModal' is truthy
+   */
+  useModal?: boolean,
+  toggleModal: number => void,
+  isCount?: boolean
 };
 
-const AttendanceStatus = ({ pools, toggleModal }: AttendanceStatusProps) => {
+const AttendanceStatus = ({
+  pools,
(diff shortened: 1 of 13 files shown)


odinuge:
Did some more work today.

Should be more or less good to go, but still need some more testing :smile: 
screenshot from 2018-02-06 15-18-16
screenshot from 2018-02-06 15-19-16

app/reducers/fetchHistory.js
```diff
@@ -33,6 +33,7 @@ export default function fetchHistory(state: State = initialState, action: any) {
       };
     }
     case FetchHistory.CLEAR_HISTORY:
+    case User.LOGIN.SUCCESS:
```
odinuge: This will clear the `FetchHistory` on login, and it will force actions to run again after login.

app/routes/events/components/EventDetail/index.js
```diff
@@ -210,35 +212,56 @@ export default class EventDetail extends Component<Props> {
...
+                <Admin
+                  actionGrant={actionGrant}
+                  event={event}
+                  deleteEvent={deleteEvent}
+                />
+              </Flex>
             </ContentSidebar>
           </ContentSection>
+          <div className={styles.joinHeader}>
+            Bli med på dette arrangementet
```
edvardvb: A link for users to log in to register would be nice! :)

edvardvb:
@odinuge Could you add a checkbox in the editor for abakom-only events as well? :smile:

app/routes/events/components/EventDetail/index.js
odinuge: :+1:

Just need a nice way to do it

odinuge:
Yeah

app/routes/events/components/EventDetail/index.js
odinuge: Any idea about how we can do it? Should we just create a custom route with redirect back to the event? An inline login form (visible after eg. a click) will look kinda shady imo.

edvardvb: Not sure what the best way is, but I think the behaviour I'd expect would be a simple link "Click here to login", which would take me to a login page that directs me back to the event after logging in.

app/routes/events/components/EventEditor/index.js
```diff
@@ -239,6 +239,16 @@ function EventEditor({
                 className={styles.formField}
               />
             </Tooltip>
+            <Tooltip content="Kun la medlemmer i abakom se arrangement">
+              <Field
+                label="Kun for abakom"
```
edvardvb: Capital A in Abakom :sunglasses:

app/components/UserAttendance/AttendanceStatus.js
```diff
@@ -8,27 +8,23 @@ import type { EventPool } from 'app/models';
...
   index,
-  toggleModal,
-  useModal,
-  isCount
+  toggleModal
 }: AttendanceElementProps) => {
-  const registrationCount =
-    isCount || !registrations ? registrations : registrations.length;
-  const Status = () => <strong>{`${registrationCount}/${capacity}`}</strong>;
+  const totalCount = registrations ? registrations.length : registrationCount;
```
edvardvb: Is this necessary? Shouldn't these be the same number?

app/routes/events/components/EventEditor/index.js
odinuge: :sunglasses:

app/components/UserAttendance/AttendanceStatus.js
odinuge: Ish.

We can use `pool.registrationCount` always, but then we need to update it when websocket-actions with new registrations occurs (and i guess we should do that).

odinuge: https://github.com/webkom/lego-webapp/blob/master/app/reducers/registrations.js

odinuge:
Running in prod as we speek, so I guess it is a good `Add initial support [...]` :smile:

PR #1109, opened 2018-02-13
Only cache requests on the server

ekmartin:
We don't really need to cache stuff on the frontend. This should hopefully fix some of the cache issues, while still keeping the benefits of not having to fetch stuff again after the server side render. 

In the future we could probably replace this altogether with something where we make sure that we never fetch initial data if we have it in the store from before, as that's mostly what we use the cache for anyway.

Also removed the `SET_HISTORY` action, as it's not being used anywhere.

### DIFF
--- a/app/actions/ActionTypes.js
+++ b/app/actions/ActionTypes.js
@@ -293,7 +293,6 @@ export const Feed = {
 };
 
 export const FetchHistory = {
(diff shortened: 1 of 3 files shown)


hanswilw:
@ekmartin Tests are breaking 😄

ekmartin:
Fixed :)

Issue #831, opened 2017-10-24
UI for changing semesters in companyInterestForm


odinuge:
Bedkom should be able to change the visible semesters in the companyInterestForm.

mrfarstad:
We figured out that it would be sufficient to have the semesters change depending on the current semester. For instance, spring 2018 will disappear 1. Jan 2018 💯

PR #1115, opened 2018-02-13
UI for changing semesters in companyInterestForm

mrfarstad:
3e7893e0e0b4177f72e9225cba97c4e3

This PR solves #831. It is now possible to add/activate and deactive semesters used in companyInterestForm. It is not possible to delete semesters, as an overflow of semesters will not be likely. 

The PR also fixes some styling for the companyInterestForm.

### DIFF
--- a/app/routes/companyInterest/components/CompanySemesterGUI.js
+++ b/app/routes/companyInterest/components/CompanySemesterGUI.js
@@ -22,10 +22,10 @@ type Props = FieldProps & {
   onSubmit: CompanySemesterEntity => Promise<*>,
   push: string => void,
   events: Array<Object>,
-  semesters: Array<Object>,
+  semesters: Array<CompanySemesterEntity>,
   autoFocus: any,
   edit: boolean,
-  editSemester: (Array<Object>) => void
+  editSemester: CompanySemesterEntity => void
 };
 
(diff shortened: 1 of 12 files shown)


finnss:
Looks like you have some flow errors you need to fix, and you should probably squash a couple of these commits. `Fix style`, for instance. Doesn't matter that much, but have you checked this on mobile?

mrfarstad:
I have tried it on mobile and it looks well imo 💯 The tests are good now as well :)

app/routes/companyInterest/components/CompanySemesterGUI.js
```diff
@@ -0,0 +1,124 @@
...
+};
+
+const validate = createValidator({
+  year: [required()],
+  semester: [required()]
+});
+
+export default legoForm({
+  form: 'addCompanySemester',
+  onSubmitSuccess: (result, dispatch) => dispatch(reset('addCompanySemester')),
```
odinuge: To avoid having to hardcode the form name twice, you can use this (example c): https://redux-form.com/6.0.0-alpha.4/docs/faq/howtoclear.md/

```jsx
submitMyForm(data) {
  const {createRecord, resetForm} = this.props;
  return createRecord(data).then(() => {
    resetForm();
    // do other success stuff
  });
}
```

PR #1116, opened 2018-02-14
Validate survey forms and check if user already answered

finnss:
- Reformats the survey app folder structure
- Adds form validation
- Adds a redirection from `/surveys/1` to `surveys/1/answer` for users without admin permissions
- Adds a separate page for when you try answering a survey you've already answered (requires https://github.com/webkom/lego/pull/1075)
- Moves some repeated code into a generic component (`StaticSubmission`)


app/routes/surveys/AddSubmissionRoute.js
```diff
@@ -27,9 +43,7 @@ const mapDispatchToProps = { submitFunction: addSubmission };
 
 export default compose(
   replaceUnlessLoggedIn(LoginPage),
-  prepare(({ params: { surveyId } }, dispatch) => dispatch(fetch(surveyId)), [
-    'params.surveyId'
-  ]),
+  prepare(loadData, ['params.surveyId', 'currentUser.id']),
   connect(mapStateToProps, mapDispatchToProps),
   loadingIndicator(['survey.questions', 'survey.event.cover'])
```
odinuge: Maybe add the `fetching` for `Suervey.FETCH_FOR_USER` here? To avoid jumping when the the data is reduced

app/routes/surveys/components/StaticSubmission.js
```diff
@@ -0,0 +1,93 @@
...
+import type { SurveyEntity, QuestionEntity } from 'app/reducers/surveys';
+import { RadioButton, CheckBox, TextArea } from 'app/components/Form';
+import { QuestionTypes } from '../utils';
+
+type Props = {
+  survey: SurveyEntity,
+  submission?: SubmissionEntity
+};
+
+const StaticSubmission = ({ survey, submission }: Props) => {
```
odinuge: Isn't it possible to reuse the answer form, by having all the fields readonly?

odinuge:
screenshot from 2018-02-16 13-41-40

Left: State after answering.
Right: State after refresh

odinuge:
Aand, what does the `isClone` checkbox do?

app/routes/surveys/components/StaticSubmission.js
finnss: Would probably generate more cluttercode than it would save I think. There is quite a bit of logic, styling, etc in the form that would need something like `readOnly ? <h1>...</h1> : somethingCompletelyDifferent`. One file would be serving two pretty different functions. That's why it better to separate them imo. But if you can find a clean way of combining them and proving me wrong I'm all ears, giving them the exact same styling would be nice.

finnss:
`isClone` will be repurposed once I get a Templates branch going.

finnss:
Closing this in favor of #1119

PR #1117, opened 2018-02-15
Fix radio button and checkbox styling

finnss:
This finally fixes radio button and checkbox styling!

**How**
~~I added a second parameter to `createField` that adds some extra styling if passed as true.~~ 
**edit:** On suggestion from @ekmartin, I send a custom style to `Field` as a prop it already receives, `fieldClassName`. If the user wants to use this prop for custom styling, that is still supported. 

This styling makes both radio buttons and checkboxes look good when used in Redux-form (or legoForm heh). I also tweaked the styling in `RadioButtonGroup` to make sure those interact nicely. In addition, the baseline styling of radio buttons and checkboxes have been updated, meaning that both should look nice even when used "by themselves".

So both this:
```
<CheckBox checked={this.state.checked} /> Går det bra?
```
and this:
```
<Field
    component={CheckBox.Field}
    label="Går det bra?"
    name="fine"
/>
```
should now look good without any custom styling required.

I did a search for every component that used radio buttons and/or checkboxes, and I made an effort to make sure they were all updated to look good with the new styling. I found and fixed the following:

**Checkbox**
- Joblistings
- Company interest 
- Email
- Bdb
- Articles
- Contact
- Events
- Gallery
- User settings

**Radio buttons**
- Bdb
- User settings

If you know you've used either radio buttons or checkboxes and your app isn't on this list, you might want to check to make sure it still looks good with this PR.

Finally, some photos of the result. Several of these uses custom margin between list elements, which is why the spacing is slightly inconsistent, but they all look intentional at least.

skjermbilde 2018-02-15 kl 01 09 04

skjermbilde 2018-02-15 kl 01 15 14

skjermbilde 2018-02-15 kl 01 09 16

skjermbilde 2018-02-15 kl 01 53 04

### DIFF
--- a/app/components/Form/Field.js
+++ b/app/components/Form/Field.js
@@ -19,7 +19,7 @@ function renderErrorMessage(error: Array<string> | string) {
   return <FieldError error={error} key={error} />;
 }
 
-type FieldProps = {
+export type FieldProps = {
   className: string,
   input: Object,
   meta: Object,
@@ -38,10 +38,7 @@ type FieldProps = {
  *
  * http://redux-form.com/6.0.5/docs/api/Field.md/
  */
-export function createField(
-  Component: ComponentType<*>,
-  radioOrCheckbox: boolean = false
-) {
+export function createField(Component: ComponentType<*>) {
   const Field = (field: FieldProps) => {
     const {
       input,
@@ -59,14 +56,7 @@ export function createField(
     const { error, touched } = meta;
     const hasError = showErrors && touched && error && error.length > 0;
     return (
-      <div
-        className={cx(
-          styles.field,
-          fieldClassName,
-          radioOrCheckbox && styles.radioOrCheckbox
-        )}
-        style={fieldStyle}
-      >
+      <div className={cx(styles.field, fieldClassName)} style={fieldStyle}>
         <label className={cx(styles.label, labelClassName)}>
           <Flex>
             {label && <div>{label}</div>}
--- a/app/components/Form/Field.js
+++ b/app/components/Form/Field.js
@@ -38,7 +38,10 @@ type FieldProps = {
  *
  * http://redux-form.com/6.0.5/docs/api/Field.md/
  */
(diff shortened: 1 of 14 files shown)


app/components/Form/Field.js
```diff
@@ -56,7 +59,14 @@ export function createField(Component: ComponentType<*>) {
     const { error, touched } = meta;
     const hasError = showErrors && touched && error && error.length > 0;
     return (
-      <div className={cx(styles.field, fieldClassName)} style={fieldStyle}>
+      <div
+        className={cx(
+          styles.field,
+          fieldClassName,
+          radioOrCheckbox && styles.radioOrCheckbox
```
ekmartin: Couldn't you make `CheckBox.Field` always pass this prop as the `fieldClassName` instead?
Something like:

```js
const RawField = createField(CheckBox);
CheckBox.Field = ({ fieldClassName, ...props }: FieldProps) => (
  <RawField fieldClassName={cx(fieldClassName, styles.radioOrCheckbox)} {...props} />
);
```

Would be nice to avoid adding a boolean prop to `createField`.

finnss: I might have misunderstood, but I thought `fieldClassName` was meant to be a prop the user could use for styling. Ie
```
<Field
    component={CheckBox.Field}
    label="Går det bra?"
    name="fine"
    fieldClassName={styles.myField}
/>
```
As such, I didn't want to override it and remove that option from users. Am I wrong?

ekmartin: Yes, but if you look at the code above you'll see that it still takes in `fieldClassName`, it just appends on the `styles.radioOrCheckbox` classname to it as well.

ekmartin: i.e. `fieldClassName={cx(fieldClassName, styles.radioOrCheckbox)}`

finnss:
That seemed to work nicely, @ekmartin!

ekmartin:
Good job @finnss :)

PR #1119, opened 2018-02-16
Use Redux-form in Surveys, validate the submission form, ++

odinuge:
- Reformats the survey app folder structure
- Adds form validation to submissions
- Adds a redirection from `/surveys/1` to `surveys/1/answer` for users without admin permissions
- Adds a separate page for when you try answering a survey you've already answered
- Moves some repeated code into a generic component (`StaticSubmission`)
- Reworks `SurveyEditor` from a custom solution using state to pure redux-form by utilizing `FieldArray`


app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -199,12 +179,50 @@ class SurveyEditor extends Component<Props, State> {
...
+          question={question}
+          question_data={questions && questions[i]}
+          updateQuestion={this.updateQuestion}
+          deleteQuestion={() => Promise.resolve(fields.remove(i))}
+        />
+      ))}
+    </ul>,
+
+    <Link
+      key="remove"
```
finnss: add?

odinuge: :+1:

app/actions/ActionTypes.js
```diff
@@ -282,6 +282,7 @@ export const Survey = {
  */
 export const SurveySubmission = {
   FETCH: generateStatuses('Survey.FETCH_SUBMISSIONS'),
+  FETCH_FOR_USER: generateStatuses('Survey.FETCH_FOR_USER'),
```
finnss: Maybe call this something with Survey or SurveySubmission so that the action log is more obvious?

app/routes/surveys/AddSubmissionRoute.js
```diff
@@ -27,9 +43,7 @@ const mapDispatchToProps = { submitFunction: addSubmission };
 
 export default compose(
   replaceUnlessLoggedIn(LoginPage),
-  prepare(({ params: { surveyId } }, dispatch) => dispatch(fetch(surveyId)), [
-    'params.surveyId'
-  ]),
+  prepare(loadData, ['params.surveyId', 'currentUser.id']),
```
finnss: Add `fetching` like we discussed in the other PR

app/routes/surveys/components/SurveyEditor/Question.js
```diff
@@ -0,0 +1,167 @@
...
+  >
+    <Icon
+      name={questionTypeToIcon[props.option && props.option.value]}
+      style={{ marginRight: '15px' }}
+    />
+    {props.children}
+  </div>
+);
+
+const QuestionTypeValue = (props: Object) => (
```
finnss: What's the difference between `QuestionTypeValue` and `QuestionTypeOption`? What are the `optionComponent` and `valueComponent` passed to `SelectInput.Field`?

app/routes/surveys/components/SurveyEditor/Question.js
```diff
@@ -0,0 +1,167 @@
...
+  <ul className={styles.options}>
+    {fields.map((option, index) => {
+      const isLast = fields.length - 1 === index;
+      return (
+        <Option
+          index={index}
+          onChange={
+            isLast
+              ? value => {
+                  if (value) fields.push({ optionText: '' });
```
finnss: nice

app/routes/surveys/components/SurveyEditor/Question.js
```diff
@@ -0,0 +1,167 @@
...
+                }
+              : undefined
```
finnss: Couldn't you do `isLast && ...` instead of this?

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -0,0 +1,189 @@
...
+
+const renderQuestions = ({ fields, meta: { touched, error }, questions }) => {
+  return [
+    <ul className={styles.questions} key="questions">
+      {fields.map((question, i) => (
+        <Question
+          key={i}
+          index={i}
+          question={question}
+          questionData={questions ? questions[i] : {}}
```
finnss: ugh. stupid `FieldArray`. Could we perhaps make it more clear what `question` is here? `questionId` or whatever is it?

app/routes/surveys/components/SubmissionEditor/SubmissionEditor.js
```diff
@@ -91,46 +102,69 @@ const SubmissionEditor = ({
...
-
-        return {
-          ...answer,
-          question: question.id,
-          selectedOptions,
-          answerText: answer.answerText || ''
-        };
-      })
-      .filter(answer => answer)
+    answers: formatAnswers(formContent.answers, survey).filter(answer => answer)
```
finnss: `filter(Boolean)`

app/routes/surveys/components/SubmissionEditor/SubmissionEditor.js
```diff
@@ -91,46 +102,69 @@ const SubmissionEditor = ({
...
+    const question = survey.questions[i];
+    const selected = answer.selectedOptions || [];
+    const selectedOptions =
+      question.questionType === QuestionTypes('single')
+        ? selected.map(Number).filter(Boolean)
+        : selected
+            .map(
+              (optionSelected, j) => optionSelected && question.options[j].id
+            )
+            .filter(Boolean);
```
finnss: We could write this as 
```
const selectedOptions = (
    question.questionType === QuestionTypes('single')
        ? selected.map(Number)
        : selected
            .map(
                (optionSelected, j) => optionSelected && question.options[j].id
            )
).filter(Boolean);

app/actions/ActionTypes.js
odinuge: Agree!

I guess `FETCH_SUBMISSIONS` is ok, or is that a bit misleading?

app/routes/surveys/AddSubmissionRoute.js
odinuge: :+1:

app/routes/surveys/components/SubmissionEditor/SubmissionEditor.js
odinuge: :+1:

app/routes/surveys/components/SurveyEditor/Question.js
odinuge: The `Value`-one is for the selected value that is always visable. The `Option` is for the dropdown-list. More or less copied from the  "Custom Placeholder, Option, Value, and Arrow Components" section from https://jedwatson.github.io/react-select/

app/routes/surveys/components/SurveyEditor/Question.js
odinuge: Did that intially, but `redux-form` didn't like `false` as a `onChange` prop.. :cry:

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
odinuge: Yeah, this naming schema isn't ideal. The `questionData` comes from here: https://github.com/webkom/lego-webapp/pull/1119/files#diff-abb76d391c7bf50ade8ddf7cd0d08d97R185 (and forwarded to `renderQuestions` here https://github.com/webkom/lego-webapp/pull/1119/files#diff-abb76d391c7bf50ade8ddf7cd0d08d97R94), and is selected from the `redux(-form)` state. We have to do that, since we need the `questionType` inside of the `QuestionComponent`.

The `question` prop is just the form-name that the current `question` is given.

odinuge: I guess it would be a bit easier to read (and understand) with more flow entities :smile: Redux-form v7 will also help, with way better flow support.

odinuge:
Yeah, that sounds like a good idea @finnss. You can just take this branch, and ill stay away (no more `redux-form`-programming for me the next few weeks now :laughing: ).

`do you want surveyQuestions to be its own redux state thingy` - yeah, i guess so. If you look at the picture in the other PR, you se that the state differs. When submitting the form, the form values ain't placed into the form. The states also differs if you look at the questions.


Nice work btw! Looks like this is almost ready to go live! :smile:

app/actions/ActionTypes.js
finnss: I guess these actionTypes should actually be rewritten to `SurveySubmission.FETCH`, etc instead of `Survey.FETCH_SUBMISSION`. I don't actually use this `FETCH_FOR_USER` atm iirc, I just figured the content of the action was separate enough from the `FETCH` that they should be different actions. We could call them `FETCH_ALL` and `FETCH`? I'm just aware that `FETCH` usually does something like `endpoint="someApp/{id}"`, while we're using a filter to do some custom stuff. But I guess that's fine, sine we don't really ever want to do the original kind of single fetch?

odinuge: That sounds reasonable!
`SurveySubmission.FETCH` is indeed better that `Survey.FETCH_XXX`!

app/routes/surveys/components/SurveyEditor/Question.js
finnss: Should be possible to reduce the amount of reused code though, surely. Is it significant that one has a `<span>` while the other doesn't? If it's just for the classname we can send both that and the surrounding div's classname as props and put the rest in a shared generator-component

finnss:
For `surveyQuestions`, isn't the problem you described with the photos in the last PR just an effect of poor reducing? `submission.answer` have `question`s as full nested objects back-end exactly to solve this kind of issue by avoiding dependency on the survey endpoint. So the question id on the left picture should be an able to be made into the full object without normalizing questions, that seems a bit extreme to me. I'll look into it

app/routes/surveys/components/SurveyEditor/Question.js
odinuge: Yeah, that is possible! The hardcoded classNames are required tho. They are special for `react-select`

app/routes/surveys/AddSubmissionRoute.js
finnss: Does adding 'fetching' work out of the box here? Other places I've sent `state.myApp.fetching` down as a prop and returned a `<LoadingIndicator />` if it was true.

finnss: Something like this?

https://github.com/webkom/lego-webapp/blob/master/app/routes/surveys/SurveyRoute.js

PR #1121, opened 2018-02-16
Fix date formatting for dates outside current year

odinuge:
screenshot from 2018-02-16 20-23-05
:arrow_down: 
screenshot from 2018-02-16 20-22-39

if `year === currentYear`, the year will be omitted.

### DIFF
--- a/app/components/Time/Time.spec.js
+++ b/app/components/Time/Time.spec.js
@@ -63,9 +63,7 @@ describe('<FromToTime />', () => {
     const wrapper = shallow(<FromToTime from={from} to={to} />);
     const time = wrapper.find('span');
 
-    const fromTime = (
-      <FormatTime time={moment(from)} format="dddd DD. MMMM, HH:mm" />
-    );
+    const fromTime = <FormatTime time={moment(from)} />;
     const toTime = <FormatTime time={moment(to)} format="HH:mm" />;
 
     expect(time.prop('children')[0]).toEqual(fromTime);
--- a/app/components/Time/Time.spec.js
+++ b/app/components/Time/Time.spec.js
@@ -4,6 +4,12 @@ import Time, { FromToTime, FormatTime } from '../Time';
 import { shallow } from 'enzyme';
 import moment from 'moment-timezone';
 
+jest.mock('app/config', () => {
+  return {
+    timezone: 'GMT'
+  };
+});
+
 describe('<Time />', () => {
   it('should show a date with a default format', () => {
     const dateTime = '2016-02-02T22:17:21.838103Z';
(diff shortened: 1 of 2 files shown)


finnss:
Oh, because the `format` prop in `fromToTime` was overriding the custom stuff that makes other years look good? So without this PR other years look good alone, but ugly in `fromToTime`? Makes sense.

That means we lose the custom formatting of `fromToTime`, though. 
`30. oktober, 18.15` instead of `30. okt. 18.15` was deliberate design. We do need to shorten months when year has to be displayed as well, as implemented previously. But I think we should keep the custom format of `fromToTime` otherwise, which is lost with your solution. 

We could keep it by doing something like this?
```
      const toTime = moment(to);
      const fromFormat = moment().isSame(dateTime, 'year')
          ? 'dddd DD. MMMM, HH:mm'
          : 'dddd DD. MMM YYYY HH:mm';
      return fromTime.isSame(toTime, 'day') ? (
        <span>
          <FormatTime time={fromTime} format={fromFormat} /> -{' '}
          <FormatTime time={toTime} format="HH:mm" />
        </span>
      ) : (
```

odinuge:
Fair point @finnss. Rewrote the code a bit, and added some more tests :smile:

app/components/Time/Time.spec.js
```diff
@@ -15,7 +21,7 @@ describe('<Time />', () => {
   it('should show a date with a custom format', () => {
     const dateTime = '2016-02-02T22:17:21.838103Z';
     const wrapper = shallow(<Time format={'HH:mm'} time={dateTime} />);
-    const expected = <time dateTime={dateTime}>23:17</time>;
+    const expected = <time dateTime={dateTime}>22:17</time>;
```
finnss: haha, I specifically picked this time to make sure we tested the midnight crossover with +1 timezones :P

app/components/Time/index.js
```diff
@@ -70,14 +70,14 @@ export const FromToTime = ({
...
   const toTime = moment(to);
-  return fromTime.isSame(toTime, 'day') ? (
-    <span>
-      <FormatTime time={fromTime} format="dddd DD. MMMM, HH:mm" /> -{' '}
-      <FormatTime time={toTime} format="HH:mm" />
-    </span>
-  ) : (
+  const fromFormat =
+    moment().isSame(fromTime, 'year') && 'dddd DD. MMMM, HH:mm';
+  const toFormat = fromTime.isSame(toTime, 'day') ? 'HH:mm' : fromFormat;
```
finnss: Smart!

app/components/Time/Time.spec.js
finnss: I assume you changed it because it was acting up, is there a real issue here we should be testing?

ekmartin: @odinuge

odinuge: :+1: Got tired of dealing with timezones. Will look at this with @finnss today! :smile:

ekmartin: no worries, was just trigger happy on the merge button :p

PR #1123, opened 2018-02-18
Fix minor admin regsitration bugs

odinuge:
This resets the form if the submit succeeds. It also fixes registration
to the waiting list, by sending no poolId; instead of pk=-1. Also add
legoForm <3

Toghether with a change in the backend, this is achived: 
screenshot from 2018-02-18 12-03-55

~~There are still a few `statusCode==500` on the endpoint...~~ Fixed in: https://github.com/webkom/lego/pull/1087.


odinuge:
This "depends" (will not crash without it) on https://github.com/webkom/lego/pull/1087. 
screenshot from 2018-02-18 13-05-35

PR #1130, opened 2018-02-20
Server-side rendering of the editor

ekmartin:
This makes the editor work when it's server-side rendered. It's a bit of a hack, and passes `jsdom` to draft-convert. It seems to work though, and allows us to just render the editor in read-only mode instead of using `DisplayContent` - which means that stuff like quote blocks works out of the box:

image

The editor needs some slight styling for it to look perfect though, but after that I think it would be nice (with some extensive testing).


config/webpack.client.js
```diff
@@ -50,6 +50,7 @@ module.exports = {
     compact([
       // Explicitly import the moment locales we care about:
       new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
+      new webpack.IgnorePlugin(/^jsdom$/),
```
ekmartin: This ensures that we don't include `jsdom` in the frontend bundle.

app/components/DisplayContent/index.js
```diff
@@ -17,88 +14,19 @@ type Props = {
...
-    isValidNode,
-    processingInstructions
-  );
-
   return (
-    <div className={cx(styles.base, className)} id={className} style={style}>
-      {react}
+    <div htmlId={id} style={style} className={className}>
+      <Editor
+        simple={false}
```
ekmartin: @larseen I think some of these prop types should be on the form of:

```js
type Props = {
  autoFocus?: boolean
};

// instead of:
type Props = {
  autoFocus: ?boolean,
};
```

so that you don't have to pass in `={false}` etc. explicitly?

ekmartin:
This should be good to go now. Fixed up the styling and changed the types of the editor props slightly.
There's going to be some weird looking events/articles lying around from the old editor (i.e. with too much spacing and so on), but I don't think that matters too much.

This is how the previous screenshot looks now (removed some empty lines and so on from the actual content):

image

ekmartin:
eh, forgot the comment boxes - one sec

ekmartin:
Done now :)

ekmartin:
Added a commit to fix the comment collapsing/expanding (https://github.com/webkom/lego-webapp/pull/1130/commits/c54addcba60786db509d0bdbacda847bef28b716). It's slightly hacky, but @larseen can fix it properly eventually ;)

odinuge:
You can remove this too:

https://github.com/webkom/lego-webapp/blob/961b80763817e4d32f383449686a3d7bca45852f/app/utils/urlifyString.js#L20

ekmartin:
Done in https://github.com/webkom/lego-webapp/pull/1130/commits/98c4219fe220c6b8b4d450e211e27ac13cebc45b - @odinuge

PR #1135, opened 2018-02-20
Fix huge iCal error 😱

sarmil:
This error ruined the booking round completely, as no one would take us seriously 🤑


ekmartin:
Thanks thanks :)

PR #1138, opened 2018-02-21
Survey templates

finnss:
Adds the option to load templates in surveys to reduce the amount of repetetive work survey creators have to do.

Does so using url params: `/surveys/add/?templateType=company_presentation` will issue a fetch for the template used for company presentations, and then use that to initialize the form. Regular users do this by clicking a new "Use template" button, which has a dropdown containing each type from the possible event types found in `events/utils`.

skjermbilde 2018-02-21 kl 02 21 46

Clicking an item simply links to an url on the form above and reinitializes the form.

Because clicking such an item overrides all changes users have made, I figured there should be a confirm modal. However, the confirm modal component didn't seem to cooperate at all, so for now I use the native `if (confirm("er du sikker?")) { ... }`.

skjermbilde 2018-02-21 kl 02 32 35

This works for both adding and editing: `/surveys/someId/edit/?templateType=other` is also implemented. Overriding existing surveys with templates is probably less common functionality than while creating a new survey, but it's there. Loading from an event while editing is **not** supported, as I didn't see what kind of use case this would cover.

Also adds a list view of all templates, separating them from the regular list view (templates aren't included in the regular list view). 

skjermbilde 2018-02-21 kl 02 26 47

Because templates are just regular surveys with an additional field set, they can be edited like any other survey. 

**Also adds another cool feature**: initializing surveys from an event id. `/surveys/add/?event=50` will not only set the `survey.event` as expected, but also initialize the form _based on the template for the eventType_ of event 50. `/surveys/add/?event=50` without any inputs:

skjermbilde 2018-02-21 kl 02 30 38

It also sets the "Active from" field to `event.endTime`.

### DIFF
--- a/app/routes/surveys/AddSurveyRoute.js
+++ b/app/routes/surveys/AddSurveyRoute.js
@@ -15,18 +15,17 @@ import { selectSurveyTemplate } from 'app/reducers/surveys';
 import { fetchEvent } from 'app/actions/EventActions';
 import { selectEventById } from 'app/reducers/events';
 import { defaultActiveFrom } from './utils';
+import loadingIndicator from 'app/utils/loadingIndicator';
 
 const loadData = (props, dispatch) => {
-  const { template, event } = props.location.query;
+  const { templateType, event } = props.location.query;
   if (event) {
-    return dispatch(fetchEvent(event)).then(
-      result =>
-        result.success &&
-        dispatch(fetchTemplate(result.payload.entities.events[event].eventType))
+    return dispatch(fetchEvent(event)).then(result =>
+      dispatch(fetchTemplate(result.payload.entities.events[event].eventType))
     );
   }
-  if (template) {
-    return dispatch(fetchTemplate(template));
+  if (templateType) {
+    return dispatch(fetchTemplate(templateType));
   }
 };
 
@@ -35,12 +34,10 @@ const mapStateToProps = (state, props) => {
   const { templateType, event } = props.location.query;
 
   const fullEvent = selectEventById(state, { eventId: event });
-
   const template = selectSurveyTemplate(state, {
     ...props,
     templateType: templateType || fullEvent.eventType
   });
-
   const initialEvent = event
     ? {
         value: fullEvent.id,
@@ -50,12 +47,7 @@ const mapStateToProps = (state, props) => {
   const activeFrom = event ? fullEvent.endTime : defaultActiveFrom(12, 0);
 
   let initialValues = null;
-  if (
-    !(
-      ((templateType || event) && !template) ||
-      (event && Object.keys(fullEvent).length === 0)
-    )
-  ) {
+  if (notFetching) {
     if (template) {
       initialValues = {
         ...template,
@@ -91,10 +83,7 @@ const mapDispatchToProps = {
 
 export default compose(
   replaceUnlessLoggedIn(LoginPage),
-  prepare(loadData, [
-    'notFetching',
-    'location.query.template',
-    'location.query.event'
-  ]),
-  connect(mapStateToProps, mapDispatchToProps)
+  prepare(loadData, ['location.query.templateType', 'location.query.event']),
+  connect(mapStateToProps, mapDispatchToProps),
+  loadingIndicator(['notFetching'])
 )(SurveyEditor);
--- a/app/routes/surveys/AddSurveyRoute.js
+++ b/app/routes/surveys/AddSurveyRoute.js
@@ -1,31 +1,85 @@
 import { connect } from 'react-redux';
 import { compose } from 'redux';
-import { addSurvey, deleteSurvey } from '../../actions/SurveyActions';
+import {
+  addSurvey,
+  deleteSurvey,
+  fetchTemplate
+} from '../../actions/SurveyActions';
 import SurveyEditor from './components/SurveyEditor/SurveyEditor';
 import { LoginPage } from 'app/components/LoginForm';
 import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
 import { push } from 'react-router-redux';
-import moment from 'moment-timezone';
 import { initialQuestion } from './components/SurveyEditor/SurveyEditor';
+import prepare from 'app/utils/prepare';
+import { selectSurveyTemplate } from 'app/reducers/surveys';
+import { fetchEvent } from 'app/actions/EventActions';
+import { selectEventById } from 'app/reducers/events';
+import { defaultActiveFrom } from './utils';
 
-const time = (hours, minutes) =>
-  moment()
-    .startOf('day')
-    .add({ hours, minutes })
-    .toISOString();
+const loadData = (props, dispatch) => {
+  const { template, event } = props.location.query;
+  if (event) {
+    return dispatch(fetchEvent(event)).then(
+      result =>
+        result.success &&
+        dispatch(fetchTemplate(result.payload.entities.events[event].eventType))
+    );
+  }
+  if (template) {
+    return dispatch(fetchTemplate(template));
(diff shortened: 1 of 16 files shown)


app/routes/surveys/AddSurveyRoute.js
```diff
@@ -1,31 +1,85 @@
...
-  moment()
-    .startOf('day')
-    .add({ hours, minutes })
-    .toISOString();
+const loadData = (props, dispatch) => {
+  const { template, event } = props.location.query;
+  if (event) {
+    return dispatch(fetchEvent(event)).then(
+      result =>
+        result.success &&
```
ekmartin: do you need the `result.success` here? wouldn't it throw if the result wasn't successful?

app/routes/surveys/AddSurveyRoute.js
```diff
@@ -1,31 +1,85 @@
...
+        dispatch(fetchTemplate(result.payload.entities.events[event].eventType))
```
ekmartin: if the only reason you're fetching the event is to retrieve the template, maybe we should have an API endpoint on the backend for doing so? `/api/v1/events/10/survey-template` or something like that? up to you, but might save you some code :p

app/routes/surveys/AddSurveyRoute.js
```diff
@@ -37,5 +91,10 @@ const mapDispatchToProps = {
 
 export default compose(
   replaceUnlessLoggedIn(LoginPage),
+  prepare(loadData, [
+    'notFetching',
```
ekmartin: why do you need `notFetching` in here?

app/routes/surveys/AddSurveyRoute.js
```diff
@@ -1,31 +1,85 @@
...
 
 const mapStateToProps = (state, props) => {
+  const notFetching = !state.surveys.fetching && !state.events.fetching;
+  const { templateType, event } = props.location.query;
+
+  const fullEvent = selectEventById(state, { eventId: event });
+
+  const template = selectSurveyTemplate(state, {
+    ...props,
+    templateType: templateType || fullEvent.eventType
```
ekmartin: if you're always fetching the template, when would `fullEvent.eventType` be used?

app/routes/surveys/AddSurveyRoute.js
```diff
@@ -1,31 +1,85 @@
...
+  });
+
+  const initialEvent = event
```
ekmartin: why can an event be both an object and a string?

app/routes/surveys/AddSurveyRoute.js
ekmartin: or is that in the case where there's no template at all?

app/routes/surveys/AddSurveyRoute.js
```diff
@@ -1,31 +1,85 @@
...
+        value: fullEvent.id,
+        label: fullEvent.title
+      }
+    : '';
+  const activeFrom = event ? fullEvent.endTime : defaultActiveFrom(12, 0);
+
+  let initialValues = null;
+  if (
+    !(
+      ((templateType || event) && !template) ||
```
ekmartin: this seems painful, what's going on here?

app/routes/surveys/AddSurveyRoute.js
```diff
@@ -1,31 +1,85 @@
...
-      isClone: true,
-      questions: [initialQuestion]
-    },
+    template,
+    initialValues,
     survey: {
-      questions: []
-    }
+      questions: template ? template.questions : [],
+      event: event && fullEvent
```
ekmartin: couldn't this just be `event: fullEvent`?

app/routes/surveys/TemplatesRoute.js
```diff
@@ -0,0 +1,34 @@
...
+  addSurvey,
+  deleteSurvey,
+  push
+};
+
+export default compose(
+  replaceUnlessLoggedIn(LoginPage),
+  prepare(loadData),
+  connect(mapStateToProps, mapDispatchToProps),
+  loadingIndicator(['notFetching', 'surveys'])
```
ekmartin: do you need `'surveys` here? doens't `notFetching` cover that

ekmartin:
good job :) nice that it fetches the template automatically for events.

app/routes/surveys/AddSurveyRoute.js
finnss: Okay? I figured since `result` includes a boolean `success` there were cases where it was false I'd have to handle, but I wouldn't be surprised if that's taken care of in middleware or something.

app/routes/surveys/AddSurveyRoute.js
finnss: I suppose that should be in a `loadingIndicator`

app/routes/surveys/AddSurveyRoute.js
finnss: `templateType` is the url param, which is undefined in the `surveys/add/?event=50` case. So this is for the case where template is fetched based on event type.

app/routes/surveys/AddSurveyRoute.js
finnss: I use it for setting initialValues `survey.event` (which requires title for the autocomplete selector)  and `activeFrom = event.endTime` too. I suppose I fetch quite a bit more information about the event than I need, but I figured that wasn't a huge deal? You'll only ever do one such fetch at a time.

app/routes/surveys/AddSurveyRoute.js
finnss: I think that's how the `SelectInput` autocomplete thing works? Sending an empty string makes it empty, while sending an object on the form
```
{
        value: fullEvent.id,
        label: fullEvent.title
}
```
makes it select that event by default. I suppose that instead of an empty string I could send
```
{
        value: '',
        label: ''
}
```
I just didn't see the need or advantage

app/routes/surveys/AddSurveyRoute.js
finnss: The event selector returns an empty object if it doesn't find anything, and I think I do some `if (event) { ... }` later on which would evaluate to true. I didn't want to deal with any `object 'event' has no attribute 'title'` error messages. Think that's bad practise?

app/routes/surveys/AddSurveyRoute.js
finnss: Yeah I know, I've been thinking about how to fix this. The purpose is to make sure to return `initialValues: null` if either the event or the template is being fetched. Without this check, the logic would use `event: undefined` and/or `template: undefined` to assume we were in a regular case without any url params. That was a problem, because once `initialValues` have been passed to redux-form as anything else than null, it didn't seem to automatically reinitialize when `initialValues` changed. I suppose `if (!notFetching) { ... }` would be equivalent, I've just had some bad experience with the `fetching` flag being inconsistent. I'll check if it works

app/routes/surveys/EditSurveyRoute.js
```diff
@@ -1,43 +1,82 @@
...
+  const templateType = props.location.query.templateType;
+  const template = selectSurveyTemplate(state, { ...props, templateType });
+
+  const initialEvent = survey.event && {
+    value: survey.event.id,
+    label: survey.event.title
+  };
+
+  let initialValues = null;
+  if (notFetching && !(templateType && !template)) {
```
finnss: I tried having just `notFetching` here, but it really seemed unpredictable and sporadic. It _usually_ does the trick, but I can't seem to see the pattern to when it decides to say it's done fetching despite `template` not being available in state yet. Could be a problem somewhere else in the stack but I couldn't find it.

PR #1140, opened 2018-02-22
Add a position indicator in the event waiting list

finnss:
Waiting list positions:
skjermbilde 2018-02-23 kl 00 12 54

Regular user attendance lists remain unchanged:
skjermbilde 2018-02-23 kl 00 09 52


hanswilw:
This does not show the correct position completely, since 3.-5th grade members could be bumped before 1-2th grade. We have to calculate the position based on the pool, which we should do backend.

finnss:
Does that mean the waiting list order in prod is currently incorrect? I assumed that was obviously correct so I didn't get why this would be a hard fix, but I guess I was wrong.

edvardvb:
The order of registration (based on reg time) is correct, so it's almost good, but there is no way to have a single list be correct when people are waiting for two different pools. You could be number 10 in that list, but still be bumped first, because everyone above you is waiting for the other pool(s)

 The hard part is when some people are able to join several pools, because you they could be number 3 in line for pool A, and number 7 for pool B. Luckily, such events rarely happen.

finnss:
So to be accurate, we'd have to include one waiting list for each pool that's at max capacity, correct? If so, are there any arguments against doing so? Seems like something that could be done front-end too

edvardvb:
For events with pools with overlapping members, that would place some people in several waiting lists, although that's not really a problem. Your solution shows all information to the user, and is pretty straight forward. 

However, I think it's cleaner is to only show the user their position (and if they have several pools they can join, only their highest position), and to not actually show the numbered list (i.e. keep the list we have there as is). I think we'll have a hard time finding space for several waiting lists in that sidebar, and knowing who is ahead of you in line shouldn't really matter to the user.

PR #1142, opened 2018-02-24
Sort event registrations by number of shared memberships

finnss:
Front-end for https://github.com/webkom/lego/pull/1105

### DIFF
--- a/app/routes/events/components/EventDetail/index.js
+++ b/app/routes/events/components/EventDetail/index.js
@@ -15,7 +15,11 @@ import Tag from 'app/components/Tags/Tag';
 import { FormatTime, FromToTime } from 'app/components/Time';
 import InfoList from 'app/components/InfoList';
 import { Flex } from 'app/components/Layout';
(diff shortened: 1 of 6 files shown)


app/routes/events/components/EventDetail/index.js
```diff
@@ -236,7 +240,7 @@ export default class EventDetail extends Component<Props> {
               <Flex column>
                 <h3>Påmeldte</h3>
                 {registrations ? (
-                  [
+                  registrations.sort(sortRegistrationsBySharedMemberships) && [
```
finnss: This depends on `.sort()` being in-place, which it is documented as being [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

odinuge: Imo this should be done inside the selector, not here.

It is also easier to use lodash `sortBy`, something like:
```jsx
sortBy(registrations, 'sharedMemberships')
```

PR #1147, opened 2018-02-26
I wanted tab search, I created tab search

larseen:


ekmartin:
@larseen I think you can add a line like [this](https://github.com/webkom/lego-webapp/blob/master/.flowconfig#L19) to the `.flowconfig` to have it ignore `.xml` files.

odinuge:
I guess a `// $FlowFixMe` would work.

https://github.com/webkom/lego-webapp/blob/ca28b7e00602e46d525dae86d88aee57584a1dde/app/routes/events/components/EventAdministrate/Abacard.js#L11

ekmartin:
Yeah, @odinuge is right - I remember that this didn't actually work now (for the mp3 stuff). Really confused why it works for `png`s and stuff though, but dunno - maybe just add a flowfixme.

PR #1151, opened 2018-02-27
Your upcoming events in profile

### LINKED ISSUE #1074: Show list of upcoming events you're attending on User profile
There should be an easy way to quickly get an overview of all events you have upcoming.

cherrybeans:
Fixes #1074.

Now shows the events you're registered for in your own user profile for easy access. Modified the existing EventItem to be able to not show the tags of an event. Also an style update for the message shown in the feed section when there is no feed to show. Extended the existing events endpoint. 

Note: If you have registered for 1000 events, it will show all 1000 events and the page will get quite long. In reality the length won't be able to get unreasonably long though.

With registered events:
screenshot from 2018-02-26 22-48-39

Without registered events:
screenshot from 2018-02-26 22-19-36

Won't show your registered events when viewing others profile.
screenshot from 2018-02-26 22-37-29

### DIFF
--- a/app/components/Feed/index.js
+++ b/app/components/Feed/index.js
@@ -4,6 +4,7 @@ import Activity from './activity';
 import type { AggregatedActivity } from './types';
 import EmptyState from 'app/components/EmptyState';
 import ErrorBoundary from 'app/components/ErrorBoundary';
+import styles from './Feed.css';
 
 export const activityRenderers = {
   comment: require('./renders/comment'),
@@ -32,7 +33,7 @@ const Feed = ({ items }: Props) => (
       })
     ) : (
       <EmptyState>
-        <h1>Ingen aktiviteter i feeden</h1>
+        <h2 className={styles.emptyState}>Ingen aktiviteter i feeden</h2>
       </EmptyState>
     )}
   </div>
--- a/app/routes/users/UserProfileRoute.js
+++ b/app/routes/users/UserProfileRoute.js
@@ -15,7 +15,7 @@ import { LoginPage } from 'app/components/LoginForm';
 
 const loadData = ({ params: { username } }, dispatch) => {
   return dispatch(fetchUser(username)).then(action =>
-    dispatch(fetchUpcoming(action.payload.result))
+    dispatch(fetchUpcoming())
   );
   // TODO: re-enable when the user feed is fixed:
   // .then(action =>
@@ -31,7 +31,7 @@ const mapStateToProps = (state, props) => {
   const user = selectUserWithGroups(state, { username });
   let feed;
   let feedItems;
-  let upcomingEvents = selectUpcomingEvents(state);
+  const upcomingEvents = selectUpcomingEvents(state);
   if (user) {
     feed = { type: 'user', activities: [] };
     feedItems = [];
--- a/app/routes/users/UserProfileRoute.js
+++ b/app/routes/users/UserProfileRoute.js
@@ -4,15 +4,19 @@ import { compose } from 'redux';
 import { connect } from 'react-redux';
 import UserProfile from './components/UserProfile';
 import { fetchUser } from 'app/actions/UserActions';
+import { fetchUpcoming } from 'app/actions/EventActions';
 import { fetchUserFeed } from 'app/actions/FeedActions';
 import { selectUserWithGroups } from 'app/reducers/users';
+import { selectUpcomingEvents } from 'app/reducers/events';
 import loadingIndicator from 'app/utils/loadingIndicator';
 import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
 import prepare from 'app/utils/prepare';
 import { LoginPage } from 'app/components/LoginForm';
 
 const loadData = ({ params: { username } }, dispatch) => {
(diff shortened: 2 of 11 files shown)


cherrybeans:
As of now, the flow fails, and I need help on fixing it.
Also, the two commits named "some change" were accidentally pushed with no message (also don't know if fixable) but they take care of the EventActions and reducer.

app/routes/users/UserProfileRoute.js
```diff
@@ -27,6 +31,7 @@ const mapStateToProps = (state, props) => {
   const user = selectUserWithGroups(state, { username });
   let feed;
   let feedItems;
+  let upcomingEvents = selectUpcomingEvents(state);
```
ekmartin: this can be a `const` since you never change the variable

ekmartin:
looks nice! good job :)

app/routes/users/components/UserProfile.js
```diff
@@ -160,8 +188,17 @@ export default class UserProfile extends Component<Props> {
                 </div>
               )}
           </div>
-
           <div className={styles.rightContent}>
+            {isMe && (
+              <div>
+                <h3>Dine kommende arrangementer</h3>
```
finnss: How about simply doing `{isMe && upcomingEvents.length && ( ...` here, instead of doing the same check later and returning EmptyState in UpcomingEvents? That would remove the "Dine kommende arrangementer" altogether when you have no upcoming events, which seems better to me. Would let you remove a bit of code as well.

app/components/Feed/index.js
```diff
@@ -32,7 +33,7 @@ const Feed = ({ items }: Props) => (
       })
     ) : (
       <EmptyState>
-        <h1>Ingen aktiviteter i feeden</h1>
+        <h2 className={styles.emptyState}>Ingen aktiviteter i feeden</h2>
```
finnss: If you're overriding pretty much every style from `<h2>` in `styles.emptyState` anyway, I guess you might as well make this a `<div>` or a `<p>` or something? I agree with the change btw, looks less intrusive, well done!

cherrybeans: Actually not sure what would be the best choice, as <h2> has a semantic meaning, but as it's not actually a title or important text, ,so maybe a <p> would be better.

app/routes/users/components/UserProfile.js
cherrybeans: I did it that way because it was consistent with how it was done in the Feed. Actually I think that if the feed is showing when there is nothing there, I don't see a problem with the upcoming events staying there if empty. Just for consistency.

cherrybeans:
There is a known bug to this. After unregistering from an event after having visited your profile first,  it will only update to remove that event if you manually refresh the page because there is no trigger to fetch the update from the endpoint. Which I think should not be necessary. Any thoughts or improvements? @ekmartin @finnss Thanks a lot for the help!

ekmartin:
It would be possible to fix by removing the `isUsersUpcoming` flag from events in the event reducer when you get an unregistration where the user matches yourself [here](https://github.com/webkom/lego-webapp/blob/12db7d2f071883fec298b268c2a42596e012855d/app/reducers/events.js#L80-L103), but I don't think it's worth the hassle. 

Let the people refresh :shipit:

cherrybeans:
Noted. Thanks!

app/routes/users/components/UserProfile.js
finnss: I see what you mean, but I think the Feed EmptyState is there to make the page feel less empty. Having _two_ such emptystates doesn't really help solve that same issue. 

However, at least initially it might be a good idea to keep what you've made just so that people see that the feature has been added. We can remove it later if we decide that's best, but let's keep it for now. The visibility might be enough reason to keep it long term too.

Would you mind adding a bit of margin between the feed and the upcoming events though? :P The event list items have some margin built in so it mainly looks too tight when both lists are empty, but adding some margin for both cases wouldn't hurt.

app/routes/users/components/UserProfile.js
```diff
@@ -28,7 +30,8 @@ type Props = {
   showSettings: boolean,
   feedItems: Array<any>,
   feed: Object,
-  isMe: Boolean
+  isMe: Boolean,
```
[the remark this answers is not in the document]
cherrybeans: EventBase is not being exported, but there is an Event that is EventBase + more. I could just export EventBase.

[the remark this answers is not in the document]
cherrybeans: I think using Event might be better than EventBase as the objects I get have variables seen in the Event type.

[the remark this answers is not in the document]
ekmartin: 👍

app/routes/users/components/UserProfile.js
cherrybeans: Agree on the thought that people can see more easily that the feature has been added if it stays. Increased the margin :)

cherrybeans:
Ready for merge. Thanks for all the help!

Now also has a loading indicator while the endpoint is being fetched.
screenshot from 2018-03-04 00-06-11

ekmartin:
Nice 👍

PR #1157, opened 2018-03-05
Various style fixes and quote deleting

### LINKED ISSUE #1131: You have to open the popover twice to delete quotes
It closes after the first click, and then you have to re-open it to press "Er du sikker?".

image

finnss:
- Fixes the z-index of the editor so that it's not visible over modals
- Adds border to the disabled Checkbox
- Removes `cursor: pointer` from disabled radio buttons
- Fixes #1131 
- Removes the dot on the frontpage (meant to separate event info) from articles
- Fixes the z-index of the Toast so that it appears above anything with z-index 1 (radio buttons and checkboxes, for instance)

### DIFF
--- a/app/components/Toast/ToastContainer.js
+++ b/app/components/Toast/ToastContainer.js
@@ -47,7 +47,8 @@ function toastStyleFactory(index, style) {
   }
   return {
     ...style,
-    bottom: `${2 + index * 4}rem`
+    bottom: `${2 + index * 4}rem`,
+    zIndex: 2
   };
 }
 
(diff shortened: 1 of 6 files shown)


app/components/Toast/ToastContainer.js
```diff
@@ -47,7 +47,8 @@ function toastStyleFactory(index, style) {
   }
   return {
     ...style,
-    bottom: `${2 + index * 4}rem`
+    bottom: `${2 + index * 4}rem`,
+    zIndex: 2
```
odinuge: Add zIndex to the CSS for mobile (line 37 to 45) too. :smile:

finnss: done

PR #1161, opened 2018-03-12
Initial small rewrite of event reducers

odinuge:


app/reducers/events.js
```diff
@@ -94,9 +100,14 @@ function mutateEvent(state: any, action: any) {
...
               ? stateEvent.registrationCount - 1
               : stateEvent.registrationCount,
-            waitingRegistrations: (
-              stateEvent.waitingRegistrations || []
-            ).filter(id => id !== action.payload.id)
+            waitingRegistrationCount: fromPool
+              ? stateEvent.waitingRegistrationCount
+              : stateEvent.waitingRegistrationCount - 1,
+            waitingRegistrations:
+              stateEvent.waitingRegistrations &&
```
odinuge: This is the best way I can think of (atm.) to keep the behavior somewhat logical. If the event doesn't have `waitingRegistrations` (not defined), we should not "make it" into an array when receiving a socket update.

odinuge: This will still introduce the key `waitingRegistrations` with a value of `undefined` into the event object.

hanswilw: `...(stateEvent.waitingRegistrations && ` can avoid that, not the most readable though

odinuge: Aha @hanswilw.

Didn't think about that. That solution will be even better :smile:

odinuge:
Any thoughts @hanswilw. These reducers are causing a quite a few bugs, so it would be nice to get this into production.

PR #1163, opened 2018-03-12
Survey mail

finnss:
Adds a note about how mails are sent when a survey activates (for Bedkom and Fagkom). Also adds a separate page for when users try to fetch a survey they have the permissions to answer but that isn't active yet.


mrfarstad:
Nice work! 💯🎉

Issue #1165, opened 2018-03-12
Mega issue for survey creation


edvardvb:
- [x] Title error message is all caps
- [x] Prefilling title with template name could be misleading
- [x] Placeholder for event field could be misleading, replace with `Enter event here...`
- [x] No way of deleting alternatives, no intuitive way of seeing that empty alternatives are allowed
- [x] Looks like you can click the radio buttons and check boxes when creating
- [ ] Text box for free text questions looks like you can write in it 
- [x] #1164 
- [x]  Creating duplicate surveys gives `This filed must be unique`
  - Agreed that a redirect to the existing survey would be a good solution
- [ ]  Poor error message after creating a new survey for an event, after deleting the previous survey

SmithPeder:
Outdated survey issue. Migrate to https://github.com/webkom/lego-webapp/issues/1321

Issue #1164, opened 2018-03-12
Going to the same (answered) survey twice [As normal user]


SmithPeder:
It creates a toast, but it looks like you can answer the survey again.
- You cant answer twice, it gives a new toast when you try to send in a new answer

![screenshot from 2018-03-12 18-21-03](https://user-images.githubusercontent.com/23152018/37299085-920da2d6-2622-11e8-9685-5ab2e6deed7e.png)

finnss:
This is a permissions issue backend, will fix

PR #1168, opened 2018-03-12
Launch(?) Survey adjustments

finnss:
- Adds surveys to navigation menu
- Adds "Lag spørreundersøkelse" to event detail
- Adds functionality for removing options when creating or editing an event
- Fixes various other small issues outlined in #1165

Deleting options:
skjermbilde 2018-03-12 kl 23 39 03

With this PR and #1163, I believe Surveys could potentially be launched. Doing so would only involve giving sudo survey permissions to fagkom and bedkom, plus loading the template fixtures in the prod database (@odinuge knows which ones). 

The last four issues in #1165 remain unsolved and will likely stay that way until I return in april (except disabled forms heh @SmithPeder), so you'll have to consider whether they are necessary for launching the feature. In particular, I think #1164 might actually warrant delaying the launch. #1170 and #1171 are also worth considering.

Just to be clear, surveys doesn't need to be launched for this to be merged.

### DIFF
--- a/app/components/Search/utils.js
+++ b/app/components/Search/utils.js
@@ -104,6 +104,12 @@ const LINKS: Array<Link> = [
     key: 'email',
     title: 'E-post',
     url: '/admin/email'
+  },
+  {
+    admin: true,
+    key: 'surveys',
+    title: 'Spørreundersøkelser',
+    url: '/surveys'
   }
 ];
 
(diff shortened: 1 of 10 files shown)


app/components/Search/utils.js
```diff
@@ -104,6 +104,12 @@ const LINKS: Array<Link> = [
     key: 'email',
     title: 'E-post',
     url: '/admin/email'
+  },
```
odinuge: https://github.com/webkom/lego-webapp/pull/1168/files#diff-92b3a2ff5a180c3afe6e17a36eabb378R76

#yolo

PR #1180, opened 2018-03-19
Add header for articles ect. on frontpage

### LINKED ISSUE #906: Add header for articles/news on frontpage
It looks like "readme-utgaver" is a header for the articles below.

screenshot from 2017-11-01 13-57-26

amalieeh:
fix #906 by adding header "Andre oppslag".

image


app/routes/overview/components/Overview.css
```diff
@@ -31,6 +31,23 @@
...
+  padding-right: 10px;
+  flex: 1;
+  width: 100%;
+  color: #666;
+  text-transform: uppercase;
+  font-weight: 400;
+  letter-spacing: 1px;
+
+  @media (--mobile-device) {
+    padding: 0;
```
SmithPeder: If you give the media version of the header 0 padding if will float all the way to left.
screenshot from 2018-03-22 16-37-36
The other headings on the front page get left paddings from its parent elements
screenshot from 2018-03-22 16-37-55

I think setting `padding: 0 0 0 10px` will have the intended effect.

It would also be nice if `Andre oppslag` had the same spacing down to its content as `Festet oppslag`
screenshot from 2018-03-22 16-46-04
 The front page events under `Andre oppslag` have a preset padding-top, and the `h2` header gets a padding, and with the addition of the `margin-bottom: 10px` it total's up to a lot spacing.

PR #1187, opened 2018-03-20
Add latest readme and fix styling

odinuge:
This also fixes the aspect ratio...
screenshot from 2018-03-20 02-54-11


eirsyl:
Maybe integrate readme.abakus.no with lego and and make it easier to publish readme articles on abakus.no?

PR #1192, opened 2018-03-23
[WIP] Fix bug with duplicated registrations

odinuge:


SmithPeder:
For educational purposes only, how does treating the `AsyncActionType` as an array fix the bug with duplicated registrations? And why is `Fetchtypes` no longer a optional value for `entities()` and `paginationReducer()`, but still a optional value for `fetching()`?

app/utils/createEntityReducer.js
```diff
@@ -10,36 +10,36 @@ import type { Reducer, AsyncActionType } from 'app/types';
...
+      }
     }
+    return state;
   };
 }
 
 const isNumber = id => !isNaN(Number(id)) && !isNaN(parseInt(id, 10));
 
-export function entities(key: string, fetchType?: ?AsyncActionType) {
+export function entities(key: string, fetchTypes: ?Array<AsyncActionType>) {
```
odinuge: Should not be a [maybe type](https://flow.org/en/docs/types/maybe/), in order to simplify it.

export function entities(key: string, fetchTypes: Array<AsyncActionType>) {

app/utils/createEntityReducer.js
```diff
@@ -66,11 +66,13 @@ export function entities(key: string, fetchType?: ?AsyncActionType) {
     }
     const actionGrant = get(action, ['payload', 'actionGrant'], []);
 
+    const isFetchSuccess =
+      fetchTypes &&
```
odinuge: Remove this..

app/utils/createEntityReducer.js
odinuge: We can also accept both a value, and an array. Like we to for `createEntityReducer`.

odinuge:
Good question @SmithPeder! 

Duplicated registrations occur because we now run the 'mutate' function once per fetchType.
The mutate-reducer does non care about the fetchType, and when the frontend received a websocket
action with a new registration, the registration was therefore appended twice to the list of waiting registrations.
The issue is therefore just a bug in 'createEntityReducer', and not the backend.

Just like you can see here:
https://github.com/webkom/lego-webapp/blob/ffcf9b945ea7a849825e48a9f822c17791b3b6ca/app/utils/createEntityReducer.js#L174

We can just rewrite the the `joinReducers`-section to run all the base reducers once per fetchType, and the mutate-reducer only once.
That would also solve the problem:

Example here: 
```jsx
const reduce = joinReducers(
 ...fetchTypes.map(fetchType =>
   joinReducers(
     fetching(fetchType),
     entities(key, fetchType),
     optimistic(mutateType),
     paginationReducer(key, fetchType)
   )
 ),
 mutate
);
```

The optimistic reducer dosn't care about the `fetchType` either, and it is therefore a "waste" to run it multiple times per action. It would also create similar bugs as mutate when executed multiple times. Aaand, guess we should support multiple `mutateTypes` too.. :smile: 

The reason we should not have `fetchType(s)` as an optional parameter in `entities`, `paginationReducer` and
`fetching`, is simply because they don't make sense without it. They only do something with the state if the
type of the action matches the `SUCCESS` version of the fetchType. Imo it would be simpler
to understand (and to program) if `fetchType` isn't optional.

Ill add WIP to the PR, since there are a few other minor tweaks that should be added. :smile:

app/utils/createEntityReducer.js
```diff
@@ -10,36 +10,36 @@ import type { Reducer, AsyncActionType } from 'app/types';
...
-    fetch?: ?AsyncActionType | Array<?AsyncActionType>,
+    fetch?: AsyncActionType | Array<AsyncActionType>,
     mutate?: ?AsyncActionType
   },
   mutate?: Reducer,
   initialState?: Object
 };
 
-export function fetching(fetchType?: ?AsyncActionType) {
+export function fetching(fetchTypes?: Array<AsyncActionType>) {
```
SmithPeder: Regarding you comment about `fetchType(s)` being optional should the `?` be removed here?

odinuge: Imo. yes.

Without a fetchType, this does nothing

SmithPeder: My thoughts exactly

odinuge:
Any thoughts on this @hanswilw?

Guess we can just use:
```jsx
const reduce = joinReducers(
 ...fetchTypes.map(fetchType =>
   joinReducers(
     fetching(fetchType),
     entities(key, fetchType),
     optimistic(mutateType),
     paginationReducer(key, fetchType)
   )
 ),
 mutate
);
```
instead..

hanswilw:
That is probably better, the mutate shouldn't run one time per fetchType anyways. 👍🏻

odinuge:
Can look at this later. Closing in favor of https://github.com/webkom/lego-webapp/pull/1199

Issue #1155, opened 2018-03-05
Gallery issues


SmithPeder:
- [x] Action Grants for both list, pictures and so forth

- [x] Navigation when you view a image. Both with arrow-keys and with buttons

- [x] Sort albums by date 

- [x] Pagination error album list

- [x] Add "confirm delete" modal

- [ ] Make a better highlight of what pictures are public and not (edit). Maybe add a filter too?

- [ ] Add a not to edit view about hiding pictures. Hidden pictures are only visible for people with edit access to the gallery.

- [x] Edit gallery title is currently "Nytt album"

mrfarstad:
![bitmoji](https://render.bitstrips.com/v2/cpanel/2e29187e-812d-4803-bdec-9e49ef28a2d5-ddbe6452-5767-4bdf-8a07-6750a1523753-v1.png?transparent=1&palette=1&width=246)

mrfarstad:
Sorting albums by date: Do we want the albums to be sorted by the date it was taken or the date when the album was created?

- It is now currently sorted by when it was taken and not when it was created.

- It is a bit risky to sort on when it was taken, as this field can be null in the back end. Therefore, there might be several previously created albums with null values for this field.

odinuge:
Gallery pagination was fixed in https://github.com/webkom/lego-webapp/commit/59974743b81746946ebfef01079cc69b54fc7b15

mrfarstad:
- Action grants for both list, pictures and so forth
- Add "confirm delete" modal

Solved in #1182. :100:

finnss:
Anything still unsolved here? @odinuge @mrfarstad @SmithPeder

SmithPeder:
Closing this

PR #1195, opened 2018-04-02
Navigation for galleries

mrfarstad:
- [x] Add link for next and previous galleryPicture
- [x] Add keylisteners for each link
- [x] Add swiping functionality for mobile use

Is a part of #1155 🎉 

<a href="https://gyazo.com/73a685670ddf3cfa8925957d6f92c7c1">https://gyazo.com/73a685670ddf3cfa8925957d6f92c7c1</a>


larseen:
Great! What about adding a link for previous as well, would also would be cool is to add keylisteneres so you can scroll with left and right. 
If you also want it to take it a step further is to also add listneres for swipes, so one could browse on mobile https://github.com/dogfessional/react-swipeable.

PR #1196, opened 2018-04-03
Compress the mobile feed

SmithPeder:
When the feed is empty / only contains a few elements it takes a lot of space when on mobile.
I get that it should have a set height on desktop for the design, but on mobile screen real estate is precious.

screenshot from 2018-04-03 22-38-41

This also removes a margin which pushed the feed out of alignment on reactive screen sizes.  
So this no longer happens.
screenshot from 2018-04-03 22-47-41

### DIFF
--- a/app/routes/overview/components/Feed.css
+++ b/app/routes/overview/components/Feed.css
@@ -24,6 +24,11 @@
   height: 459px;
   flex: 1;
   background-color: rgba(255, 255, 255, 0.4);
+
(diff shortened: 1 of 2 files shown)


SmithPeder:
As an afterthought i don't really know if this will effect anybody? Maybe everybody got at least a couple posts in the feed? Or maybe not, i got 4 items in my feed, and its just rubbish from you guys, so maybe every normal user has an empty feed.

PR #1204, opened 2018-04-10
Add some throttling and fix swipe direction

odinuge:


mrfarstad:
LGTM 😸

app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -6,6 +6,7 @@ import GalleryDetailsRow from './GalleryDetailsRow';
 import { Flex } from 'app/components/Layout';
 import { Content } from 'app/components/Content';
 import Icon from 'app/components/Icon';
+import throttle from 'lodash/throttle';
```
hanswilw: We do have a babel plugin (https://github.com/webkom/lego-webapp/blob/master/package.json#L100) to automatically cherry pick lodash functions, so you can write `import { throttle } from 'lodash';`. Not important here ofc, but if you were to use more lodash functions you can avoid multiple lines.

odinuge: :+1:

Just a (bad?) habit from another project.

hanswilw: Not bad without the plugin, tho 😉

PR #1212, opened 2018-04-11
Gallery upload

odinuge:
Add batch uploading of images, instead of running all uploads at the same time.

Also introduce a status monitor (left bottom corner). :smile: 

screenshot from 2018-04-11 20-52-04


odinuge:
Prokrastinering på høyt nivå.... :coffee:

ekmartin:
bedre enn prokrastinering på lavt nivå

app/components/Dropdown/index.js
```diff
@@ -85,7 +85,7 @@ function List({ children }: { children: any }) {
   return <ul className={styles.dropdownList}>{children}</ul>;
 }
 
-function ListItem(props) {
+function ListItem(props: any) {
```
finnss: helpful ;o

app/components/PhotoUploadStatus/index.js
```diff
@@ -0,0 +1,126 @@
...
+const ConnectedUploadStatusCard = connect(mapStateToProps, mapDispatchToProps)(
+  UploadStatusCard
+);
+
+export default ConnectedUploadStatusCard;
+
+export const addUploadStatusCard = <T>(Component: ComponentType<T>) => (
+  props: T
+) =>
+  true && (
```
finnss: ?

app/routes/admin/email/EmailUserRoute.js
```diff
@@ -19,8 +20,8 @@ const mapStateToProps = (state, { routeParams }) => {
     initialValues: {
       ...emailUser,
       user: {
-        label: emailUser.user.fullName || '',
-        value: emailUser.user.id || ''
+        label: get(emailUser, 'user.fullName', ''),
+        value: get(emailUser, 'user.id', '')
```
finnss: I agree that lodash `get` is nicer than `||`, nice. It is yet another inconsistency in the project, though:p

app/routes/bdb/components/BdbDetail.js
```diff
@@ -364,7 +365,7 @@ export default class BdbDetail extends Component<Props, State> {
           <div className={styles.files}>
             <h3>Filer</h3>
             <ul>
-              {!company.files || company.length === 0 ? (
+              {!company.files || company.files.length === 0 ? (
```
finnss: <3

app/routes/photos/components/GalleryDetail.js
```diff
@@ -22,7 +23,8 @@ type Props = {
   fetch: (galleryId: Number, args: { next: boolean }) => Promise<*>,
   push: string => Promise<*>,
   uploadAndCreateGalleryPicture: (ID, File | Array<DropFile>) => Promise<*>,
-  actionGrant: Array<string>
+  actionGrant: Array<string>,
```
finnss: There's a separate `ActionGrant` type in `apps/models`, in case you didn't know. 100% doesn't matter, though.

app/routes/surveys/components/SurveyEditor/Question.js
```diff
@@ -31,11 +31,11 @@ const questionTypeToIcon = {
 const QuestionTypeOption = (props: Object) => (
   <div
     className={cx(props.className, styles.dropdown)}
-    onMouseDown={() => {
+    onMouseDown={event => {
       props.onSelect && props.onSelect(props.option, event);
     }}
-    onMouseEnter={() => props.onFocus && props.onFocus(props.option, event)}
-    onMouseMove={() => {
+    onMouseEnter={event => props.onFocus && props.onFocus(props.option, event)}
```
finnss: I remember having a discussion with someone about why I wasn't taking in `event` here, and how the code shouldn't work without it as I use the variable later. Did you change it just because it "should" be there, or did you run into any symptoms? Seems to work fine when I check out to this branch locally, I just want to be sure.

app/reducers/galleryPictures.js
```diff
@@ -13,17 +31,72 @@ export type GalleryPictureEntity = {
...
+          showStatus: false
+        }
+      };
+    }
+    case GalleryPicture.CREATE.SUCCESS: {
+      const successCount = uploadStatus.successCount + 1;
+
+      // Only update thumbnail in upload status for each 10th picture
+      const lastUploadedImage =
+        !uploadStatus.lastUploadedImage || successCount % 10 == 0
```
finnss: Hmm, what's the motivation behind this? Performance, I guess? It feels a bit.. unresponsive? Maybe that's just me. Having it update on every upload probably doesn't work too well if the upload is as fast as we want it to be. Maybe a progress bar is better? 

Very nitpicky of course, could easily merged as is. Looks good overall!

app/components/PhotoUploadStatus/index.js
```diff
@@ -0,0 +1,126 @@
...
+    uploadStatus.successCount + uploadStatus.failCount ===
+    uploadStatus.imageCount;
+
+  const hasFailedUploads = uploadStatus.failCount !== 0;
+
+  return (
+    <Card style={style}>
+      <Button onClick={hideUploadStatus}>
+        <Icon name="close" />
+      </Button>
```
finnss: This button is both kinda ugly and could rather easily be interpreted as "Cancel upload". That's what I thought it was at first. Just having the Icon without the button and moving it to the right corner would probably help with both problems.

app/components/PhotoUploadStatus/index.js
```diff
@@ -0,0 +1,126 @@
...
+      {uploadDone ? (
+        <Fragment>
+          <h3>{uploadStatus.successCount} bilder ble lastet opp </h3>
+        </Fragment>
```
finnss: What do you think about removing the popup a few seconds after the upload is done? Very "nice-to-have", but I think it could make sense.

app/reducers/galleryPictures.js
```diff
@@ -13,17 +31,72 @@ export type GalleryPictureEntity = {
...
+  const { uploadStatus = initialUploadStatus } = state;
   switch (action.type) {
     case GalleryPicture.DELETE.SUCCESS: {
       return {
         ...state,
         items: state.items.filter(id => id !== action.meta.id)
       };
     }
+    case Gallery.UPLOAD.BEGIN: {
+      const imageCount = uploadStatus.imageCount + action.meta.imageCount;
```
finnss: This imagecount includes pictures from previous uploads, even if they were in different albums. The only way to reset it is to refresh the page. You could argue that's a feature I suppose, but I would argue each upload is separate and should have its own count. From a user perspective, a redux state "session" is a pretty weird way to decide when to reset the counter.

app/components/Dropdown/index.js
odinuge: Nah. The new version of flow failed without this... :/

app/components/PhotoUploadStatus/index.js
odinuge: FUck eslint. Say no more :laughing: 

Guess I can replace this with an eslint ignore rule..

app/routes/admin/email/EmailUserRoute.js
odinuge: Yeah. We should do this in a more generic way, so all mappings are equal.

The reason I did this was because the new version of flow didn't like the old version.. :cry:

app/routes/bdb/components/BdbDetail.js
odinuge: Haha :grinning: 

The updated flow version told me to do this, and it sounds reasonable too. :smirk:

app/routes/photos/components/GalleryDetail.js
odinuge: :+1: Yeah, that sounds like a good idea.

app/components/PhotoUploadStatus/index.js
odinuge: Yeah. Lazyness it is :laughing:

app/routes/surveys/components/SurveyEditor/Question.js
odinuge: The updated flow version told me to do it. And it makes no sense to don't have it. If it doesn't send anything, it will just behave the same way as before, making event undefined.

app/components/PhotoUploadStatus/index.js
odinuge: Guess we can do that! :smile: Out of scope of this PR tho.

app/reducers/galleryPictures.js
odinuge: Yeah.

Kinda copied the functionality from google photos. If someone uploads to two galleries at the same time, what should we do then? :thinking: Or if a user does multiple uploads to the same gallery, at the same time?

finnss: Can't we just disallow that? "Vennligst fullfør opplastingen din først"?

odinuge:
This looks a bit better @finnss 
screenshot from 2018-04-26 22-40-36

PR #1213, opened 2018-04-14
Add event info and pie charts to the survey result page

finnss:
Requires https://github.com/webkom/lego/pull/1192 for event info

skjermbilde 2018-04-14 kl 18 39 34

### DIFF
--- a/app/routes/surveys/components/Submissions/SubmissionPage.js
+++ b/app/routes/surveys/components/Submissions/SubmissionPage.js
@@ -21,7 +21,7 @@ type Props = {
 };
 
 const SubmissionPage = (props: Props) => {
-  const { submissions, deleteSurvey, survey, actionGrant, isSummary } = props;
+  const { deleteSurvey, survey, actionGrant, isSummary } = props;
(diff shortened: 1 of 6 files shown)


SmithPeder:
Looks really nice :+1: What is the current situation when there is more then one written feedback at he bottom of the summary? Maybe the feedback should have link to the users individual feedback?

finnss:
skjermbilde 2018-04-14 kl 19 18 53

@SmithPeder Right now it just lists them without any styling. Could absolutely add some styling to separate them better, like Online does (https://online.ntnu.no/feedback/events/event/441/25/results/a9eec3caf2b046b384dd707f9267fcfb/), I just didn't prioritize it as I don't really know if it's necessary. 

As for the linking, maybe? The individual answers are all on one page so it would involve navigating there and jumping to the correct post, which I guess you could do with ctrl+f. Open for input, and it's also something we could add in a later PR

mrfarstad:
Gr8 work m8 I r8 8/8 🎉

PR #1215, opened 2018-04-15
Fix mission critical failure

### LINKED ISSUE #1214: readme-title is not displayed in accordance with readme style guidelines in the burger menu
The name "readme", as it describes Abakus' student magazine, should always be written in lower case and the font "OCR A Std" where ever possible, according to readme style guidelines. It is currently displayed in both the wrong font _and_ in upper case.

It is preferable to have both the right font and the right case, but should this not be possible the correct case is permissible.
image

ekmartin:
image

closes #1214


ekmartin:
please review with utmost scrutiny

PR #1217, opened 2018-04-16
Abacard stat counter

### LINKED ISSUE #1079: Abacard stats counter/view
It would be nice with a small counter/view with the stats about the attendees, so they don't have to switch to the administrate view to check the list.

amalieeh:
fix #1079 Lets users see the stats about the attendees.

before:
image

after:
image

### DIFF
--- a/app/routes/events/components/EventAdministrate/Abacard.js
+++ b/app/routes/events/components/EventAdministrate/Abacard.js
@@ -10,6 +10,7 @@ import styles from './Abacard.css';
 // $FlowFixMe
 import goodSound from '../../../../assets/good-sound.mp3';
 
+import type { EventRegistration } from 'app/models';
 type State = {
   showCompleted: boolean
 };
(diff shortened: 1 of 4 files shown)


app/routes/events/components/EventAdministrate/Abacard.js
```diff
@@ -79,6 +85,10 @@ class Abacard extends React.Component<*, State> {
             this.input = input;
           }}
         />
+        <div className={styles.counter}>
+          {' '}
```
ekmartin: do you need this empty space or did prettier just put it in?

ekmartin:
nice! this would be good to have on the `Påmeldinger` page as well (the other admin tab). we should merge this first though :)

app/routes/events/components/EventAdministrate/Abacard.js
amalieeh: Prettier put it in. Should I remove it?

ekmartin: yeah, I don't think you need it (it's on its own line anyway right, so a space shouldn't matter)

PR #1219, opened 2018-04-22
Survey token

finnss:
Requires https://github.com/webkom/lego/pull/1193/files. 

Adds a button that lets admin users get a shareable link that makes survey results publicly available, mainly for the purpose of allowing companies to see the result of their own events.

skjermbilde 2018-04-24 kl 18 36 45

skjermbilde 2018-04-24 kl 18 36 51

Also adds an entirely separate page to view those public results. This looks basically the same as the admin version, except all navigation and other admin-only stuff is gone.

skjermbilde 2018-04-24 kl 18 37 35

### DIFF
--- a/app/actions/SurveySubmissionActions.js
+++ b/app/actions/SurveySubmissionActions.js
@@ -5,14 +5,11 @@ import callAPI from 'app/actions/callAPI';
 import { surveySubmissionSchema } from 'app/reducers';
 import type { Thunk } from 'app/types';
 
-export function fetchSubmissions(surveyId: number, token?: string): Thunk<*> {
+export function fetchSubmissions(surveyId: number): Thunk<*> {
   return callAPI({
     types: SurveySubmission.FETCH_ALL,
     endpoint: `/surveys/${surveyId}/submissions`,
     schema: [surveySubmissionSchema],
-    query: {
-      token
-    },
     meta: {
       errorMessage: 'Henting av svar på spørreundersøkelse feilet'
     },
--- a/app/actions/SurveySubmissionActions.js
+++ b/app/actions/SurveySubmissionActions.js
@@ -5,11 +5,14 @@ import callAPI from 'app/actions/callAPI';
 import { surveySubmissionSchema } from 'app/reducers';
 import type { Thunk } from 'app/types';
 
-export function fetchSubmissions(surveyId: number): Thunk<*> {
+export function fetchSubmissions(surveyId: number, token?: string): Thunk<*> {
   return callAPI({
     types: SurveySubmission.FETCH_ALL,
     endpoint: `/surveys/${surveyId}/submissions`,
     schema: [surveySubmissionSchema],
+    query: {
+      token
+    },
     meta: {
       errorMessage: 'Henting av svar på spørreundersøkelse feilet'
     },
--- a/app/routes/surveys/components/AdminSideBar.js
+++ b/app/routes/surveys/components/AdminSideBar.js
@@ -40,7 +40,7 @@ const AdminSideBar = ({
   const shareLink = !token
     ? ''
     : `https://www.abakus.no/surveys/${surveyId}/submissions/summary/?token=${token}`;
-
+  console.log('admin side bar', canEdit, 'token:', token);
   return (
     canEdit && (
       <ContentSidebar className={styles.adminSideBar}>
@@ -63,11 +63,7 @@ const AdminSideBar = ({
               </li>
             </ConfirmModalWithParent>
           )}
-          {token && (
-            <li>
-              Delbar link: <a href={shareLink}>{shareLink}</a>
-            </li>
-          )}
+          {token && <Link to={shareLink}>Delbar link</Link>}
         </ul>
       </ContentSidebar>
     )
--- a/app/routes/surveys/components/AdminSideBar.js
+++ b/app/routes/surveys/components/AdminSideBar.js
@@ -75,7 +75,10 @@ export class AdminSideBar extends React.Component<Props, State> {
               <li>
                 <CopyToClipboard
                   text={shareLink}
-                  onCopy={() => this.setState({ copied: true })}
+                  onCopy={() => {
+                    this.setState({ copied: true });
+                    setTimeout(() => this.setState({ copied: false }), 2000);
+                  }}
                   style={{ marginTop: '5px' }}
                 >
                   <Button>
--- a/app/routes/surveys/components/AdminSideBar.js
+++ b/app/routes/surveys/components/AdminSideBar.js
@@ -11,14 +11,16 @@ type Props = {
   surveyId: number,
   deleteFunction?: number => Promise<*>,
   push?: string => void,
-  actionGrant: ActionGrant
+  actionGrant: ActionGrant,
+  token?: string
 };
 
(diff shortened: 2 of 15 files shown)


app/routes/surveys/components/Submissions/Results.js
```diff
@@ -0,0 +1,121 @@
...
+import type { SurveyEntity, QuestionEntity } from 'app/reducers/surveys';
+import { QuestionTypes, CHART_COLORS } from '../../utils';
+import InfoBubble from 'app/components/InfoBubble';
+import { VictoryPie, VictoryTheme } from 'victory';
+
+type Props = {
+  survey: SurveyEntity,
+  graphData: Object,
+  numberOfSubmissions: number,
+  generateTextAnswers: QuestionEntity => any
```
odinuge: Guess this can be written using react-types

app/routes/surveys/components/Submissions/Results.js
```diff
@@ -0,0 +1,121 @@
...
+  generateTextAnswers,
+  survey,
+  numberOfSubmissions
+}: Props) => {
+  return (
+    <div>
+      <div className={styles.eventSummary}>
+        <h3>Arrangementet</h3>
+        <div className={styles.infoBubbles}>
+          <InfoBubble
```
odinuge: Maybe write this as an array of data, and map over it?

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
```diff
@@ -0,0 +1,86 @@
...
+import { QuestionTypes, TokenNavigation } from '../../utils';
+
+type Props = {
+  survey: SurveyEntity,
+  actionGrant: ActionGrant,
+  results: Object
+};
+
+const SubmissionPublicResultsPage = ({ survey, actionGrant }: Props) => {
+  const results = survey.results || {};
```
odinuge: const { results = {} } = survey;

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const generateTextAnswers = question => {
+    let texts = [];
+    Object.keys(results[String(question.id)])
+      .map(name => {
+        if (name !== 'questionType') {
+          texts = results[String(question.id)][name].map((answer, i) => (
```
odinuge: You can just use `results[question.id]`. Object keys are always strings

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
```diff
@@ -0,0 +1,86 @@
...
+            <li key={i}>{answer}</li>
+          ));
+        }
+      })
+      .filter(Boolean);
```
odinuge: Smooth!

app/routes/surveys/utils.js
```diff
@@ -55,6 +58,25 @@ export const DetailNavigation = ({
   </NavigationTab>
 );
 
+export const TokenNavigation = ({
+  title,
+  surveyId,
+  actionGrant
```
odinuge: You can use: `actionGrant = []` to give it a default value. Then you don't have to do `actionGrant && moreCode`

app/routes/surveys/components/AdminSideBar.js
```diff
@@ -6,61 +6,89 @@ import { Link } from 'react-router';
...
+              >
+                <li>
+                  <Link to="">Slett undersøkelsen</Link>
+                </li>
+              </ConfirmModalWithParent>
+            )}
+            {token && (
               <li>
-                <Link to="">Slett undersøkelsen</Link>
+                <CopyToClipboard
```
odinuge: Smooth. Maybe add a delay that removes the `Kopiert!` after a few sec.?

app/routes/surveys/components/AdminSideBar.js
```diff
@@ -6,61 +6,89 @@ import { Link } from 'react-router';
...
-            >
+  render() {
+    const { surveyId, deleteFunction, actionGrant, push, token } = this.props;
+
+    const canEdit = actionGrant.includes('edit');
+
+    const onConfirm = () => {
+      if (deleteFunction && push) {
+        return (
+          deleteFunction &&
```
odinuge: Isn't deleteFunction always truthy? If you have it inside mapDispatchTo props, you don't have to double check

app/actions/SurveySubmissionActions.js
```diff
@@ -5,11 +5,14 @@ import callAPI from 'app/actions/callAPI';
 import { surveySubmissionSchema } from 'app/reducers';
 import type { Thunk } from 'app/types';
 
-export function fetchSubmissions(surveyId: number): Thunk<*> {
+export function fetchSubmissions(surveyId: number, token?: string): Thunk<*> {
```
odinuge: Do you ever need the token here?

app/routes/surveys/components/AdminSideBar.js
```diff
@@ -6,61 +6,89 @@ import { Link } from 'react-router';
...
+            {token && (
               <li>
-                <Link to="">Slett undersøkelsen</Link>
+                <CopyToClipboard
+                  text={shareLink}
+                  onCopy={() => this.setState({ copied: true })}
+                  style={{ marginTop: '5px' }}
+                >
+                  <Button>
+                    {this.state.copied ? 'Kopiert!' : 'Kopier delbar link'}
```
odinuge: We should support enable/disable this public sharing. We should also support invalidating the old token.

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
```diff
@@ -0,0 +1,86 @@
...
+
+    return texts.length === 0 ? <i>Ingen svar.</i> : texts;
+  };
+
+  const generateQuestionData = questionId => {
```
odinuge: Would be nice if all this logic was in a separate file. The components are quite large, and hard to read from A to B in order to understand what component does what.

app/routes/surveys/components/AdminSideBar.js
finnss: I believe the idea was that the AdminBar should have different functionality in different places. For instance, having a "Slett" button in the results page could be misinterpreted as deleting the results, even if we called it "Slett spørreundersøkelse".

app/routes/surveys/components/AdminSideBar.js
finnss: 😞

app/routes/surveys/components/Submissions/Results.js
finnss: What would that look like?

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
finnss: You sure? In terms of what the purpose of each file is, I think the structure is pretty nice. I understand readability concerns when reviewing, but is it really any easier to get a good sense of what's going on when functionality very clearly tied to the purpose of a file is moved out to a `utils` file or something where they have to take in a boatload of parameters that would otherwise be available as props?

Issue #1221, opened 2018-04-22
PR-committee request for album improvements


SmithPeder:
- [ ] Button for downloading a entire album

- [ ] Search in album titles

- [ ] Able to right click an album and open in new tab

- [ ] Warning when deleting a album

- [ ] Don't return to the top of a album when you exit, page holds at album position

PR #1222, opened 2018-04-22
Pr album improvments

SmithPeder:
Should fix the 'Warning when deleting a album' part of #1221 , and some other small stuff.

### DIFF
--- a/app/routes/photos/components/GalleryEditor.js
+++ b/app/routes/photos/components/GalleryEditor.js
@@ -180,7 +180,7 @@ class GalleryEditor extends Component<Props, State> {
 
     return (
       <Content>
-        <NavigationTab title="Nytt album">
+        <NavigationTab title={`Redigerer: ${gallery.title}`}>
           <NavigationLink to={'/photos'}>
             <i className="fa fa-angle-left" /> Tilbake
           </NavigationLink>
--- a/app/routes/photos/components/GalleryEditor.js
+++ b/app/routes/photos/components/GalleryEditor.js
@@ -27,6 +27,8 @@ import GalleryComponent from 'app/components/Gallery';
 import styles from './Overview.css';
 import { pull, find } from 'lodash';
 
+import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
+
 import type { ID } from 'app/models';
 import type { GalleryEntity } from 'app/reducers/galleries';
 import type { GalleryPictureEntity } from 'app/reducers/galleryPictures';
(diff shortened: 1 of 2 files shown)


app/routes/photos/components/GalleryEditor.js
```diff
@@ -180,7 +181,7 @@ class GalleryEditor extends Component<Props, State> {
 
     return (
       <Content>
-        <NavigationTab title="Nytt album">
+        <NavigationTab title={`Redigerer: ${gallery.title}`}>
```
finnss: `Editor` components are often used for both Adding and Editing, and this is no exception. This line causes a crash in `/photos/new`, as `gallery` is undefined. You could do something like `gallery ? gallery.title : 'Nytt album'`.

app/routes/photos/components/GalleryEditor.js
```diff
@@ -256,19 +257,20 @@ class GalleryEditor extends Component<Props, State> {
...
+                <Button className={styles.deleteButton}>Slett album</Button>
+              </ConfirmModalWithParent>
             )}
             <Button
               disabled={submitting}
               className={styles.submitButton}
               type="submit"
             >
-              {isNew ? 'Create' : 'Save'}
+              {isNew ? 'Lag nytt' : 'Lagre'}
```
finnss: 👍

finnss: Maybe "Opprett" instead of "Lag nytt"? Subjective, of course

app/routes/photos/components/GalleryEditor.js
```diff
@@ -256,19 +257,20 @@ class GalleryEditor extends Component<Props, State> {
             justifyContent="flex-end"
           >
             {!isNew && (
-              <Button
-                onClick={this.onDeleteGallery}
-                className={styles.deleteButton}
+              <ConfirmModalWithParent
+                title="Slett album"
+                message="Dette vil slette hele albumet og alle bildene inni"
```
finnss: "inni" sounds very casual, how about "alle bildene det inneholder"? Only if you agree, of course. Also, you should probably add a followup à la "...inneholder. Er du sikker?", since the buttons are "Avbryt" and "Ja" which doesn't make that much sense without having a question first.

app/routes/photos/components/GalleryEditor.js
SmithPeder: Agreed

app/routes/photos/components/GalleryEditor.js
SmithPeder: Agreed

app/routes/photos/components/GalleryEditor.js
SmithPeder: I see, didn't know that component was used for adding :grin:

PR #1246, opened 2018-06-27
Fix filename too long 'issue'

SmithPeder:
The issue of a filenames begin to long in the bdb detail is already fixed by  `{truncateString(name, FILE_NAME_LENGTH)}`, so on refresh Jonathan Eriksen issue would have disappeared. This just limits the width of the span, so on upload the filename will line-break.


odinuge:
Smooth! Da svarer du på mailen også?

PR #1255, opened 2018-08-15
Profile change grade

SmithPeder:
Adds the ability for an admin to change a users grade from that users profile.

screenshot from 2018-08-15 23-48-52


odinuge:
I don't think we need the `18k` lines with npm package-lock.json. Otherwise, awesome! :smile: :+1: :1st_place_medal:

PR #1267, opened 2018-08-26
Fix bug going from administrate to user

odinuge:
Because of stupid serializer/selector :upside_down_face:


odinuge:
Will fix https://sentry.abakus.no/webkom/lego-webapp/issues/5660/?query=is:unresolved

PR #1284, opened 2018-09-23
Add sentry reporting to surveys

odinuge:
This also scrolls the user to the question that isn't answered.

We have a few strange issues, so this should report all the things to sentry :tm:

### DIFF
--- a/app/components/Form/legoForm.js
+++ b/app/components/Form/legoForm.js
@@ -54,9 +54,18 @@ const legoForm = ({
 
       try {
         // We should instead check if error is SubmissionError. Does not work now
-        const [firstErrorField] = Object.keys(errors);
-
+        let [firstErrorField] = Object.keys(errors);
+        if (firstErrorField == '_error') {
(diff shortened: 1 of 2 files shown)


odinuge:
Looks like this is the result of:
https://github.com/erikras/redux-form/blob/e7ce5aec2bf0e24574e6f5f90bc04b9815c8e52e/src/structure/plain/deleteIn.js#L56

Because:
```js
const arr = [,,,]; // (3) [empty × 3]
arr.slice() // (3) [empty × 3]
[].concat(arr); // // (3) [empty × 3]
const arrCopy = [...arr] // (3) [undefined, undefined, undefined]

arr.forEach(console.log) // Prints nothing
arrCopy.forEach(console.log) // prints:
// undefined 0 (3) [undefined, undefined, undefined]
// undefined 1 (3) [undefined, undefined, undefined]
// undefined 2 (3) [undefined, undefined, undefined]
```

js-life it is


just copy it into your terminal console (or node repl) to check :smile:

PR #1301, opened 2018-10-11
Add support for bulk admin register

odinuge:
Add support for multiselect when doing admin-reg. :small_airplane: 
screenshot from 2018-10-11 18-50-29


odinuge:
@andrinelo :arrow_up:

Issue #1282, opened 2018-09-19
Fix date time picker component


mrfarstad:
Find a new one, or change the old one into accepting keystrokes for choosing date and time. Right now bedkom is frustrated by the fact that they have to click manually for every minute they want to increment (inside info) :):)

PR #1312, opened 2018-10-17
Fix time picker

mrfarstad:
Solves #1282. Fix keyboard input handling in the time picker 🎉

### DIFF
--- a/app/components/Form/TimePicker.js
+++ b/app/components/Form/TimePicker.js
@@ -27,7 +27,11 @@ type Props = {
 };
 
 type State = {
-  value: moment
+  value: moment,
+  fieldValue: {
+    hour: string,
+    minute: string
+  }
 };
 
 function parseValue(value) {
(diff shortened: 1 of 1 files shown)


odinuge:
:heart_eyes:

app/components/Form/TimePicker.js
```diff
@@ -27,57 +27,98 @@ type Props = {
...
+    minute: string
+  }
 };
 
 function parseValue(value) {
   return value ? moment(value) : moment();
 }
 
 class TimePicker extends Component<Props, State> {
+  constructor(props: Props) {
```
odinuge: Is there a specific reason that made you re-add the `constructor`?

SmithPeder: This was my mistake. I read somewhere that you should only use the `state = {...}` way when the state would not depend on initial props. Like the stackoverflow example below.
**Use constructor when you want to save props data into state**
```js
export default class LoginScreen extends React.Component {
  constructor(props){
    super(props);
    this.state = {
      loading: props.loading,
      loggedIn: props.loggedIn,
    }
  }
}
```
**Otherwise you can directly set the state for hard coded data**
```js
export default class LoginScreen extends React.Component {
  state = {
    loading: false,
    loggedIn: false,
  }
}
```

But this guy says its not the case, so idk https://hackernoon.com/the-constructor-is-dead-long-live-the-constructor-c10871bea599

odinuge: Aha, that makes sense. Guess we can just skip using the constructor then.
Nice article by `Donavon West` btw.!

app/components/Form/TimePicker.js
```diff
@@ -27,57 +27,98 @@ type Props = {
...
-      {
-        value: this.state.value.clone()[unit].call(null, 1)
-      },
-      this.commit
-    );
+  onChange = (unit: 'hour' | 'minute') => (
+    e: SyntheticInputEvent<HTMLInputElement>
+  ) => {
+    const { value } = e.target;
+    if (Number(value) || !Number(value)) {
```
orhanhenrik: What does this accomplish?

app/components/Form/TimePicker.js
```diff
@@ -27,57 +27,98 @@ type Props = {
...
+      if (
+        (unit === 'hour' && Number(value) < 24) ||
+        (unit === 'minute' && Number(value) < 60)
+      ) {
```
orhanhenrik: What happens if you write something like `-1`?

app/components/Form/TimePicker.js
mrfarstad: The reason for using constructor was for creating the variable `const value = parseValue(this.props.value);` so that I didn't have to create the same variable three times and make sure that the variable was created before the state was initially set. I can try to do a workaround =)

mrfarstad:
By using the variable `fieldValue` (renamed from `tempValue`), I am ensuring that it is possible to use backspace for clearing out numbers in the field.

Without it, values like 10 would instantly become 01 after pressing backspace once. 🎉

app/components/Form/TimePicker.js
mrfarstad: You cannot write -1 specifically, since only numbers are allowed in the field. However, by copy pasting `-1` into the field will cause some weird behavior. I think this goes under the category "user error" and should not be handled. :):)

PR #1316, opened 2018-10-18
Update readmeFetching function

odinuge:


SmithPeder:
Why is this needed?

odinuge:
Readme is slow

odinuge:
Hmm.
Looks like it is way faster now, than last time I was checking. Guess we don't need this then :upside_down_face:

odinuge:
We can still move the `first`-count out from the query, but we can keep the SSR fetching. :smile:

PR #1347, opened 2018-10-31
Update ssr-config && docs

odinuge:
Add

```
$ yarn start:staging
```
to make it easier to work without running the backend..


mrfarstad:
Nice work! Seems to work just fine on my end 🎉

PR #1348, opened 2018-10-31
Fix formatting & ordering in articles

### LINKED ISSUE #1325: Poor article ordering
https://abakus.no/articles

This pages tries to order articles by latest first, but only does so if you refresh once first. The first time you open the page the 15(?) latest articles are fetched from the API, but they are sorted by oldest first, which leads to weird results as you scroll further down and fetch more articles.

LudvigHz:
Fixes #1325


app/reducers/articles.js
```diff
@@ -57,7 +58,11 @@ export const selectArticles = createSelector(
   state => state.articles.byId,
   state => state.articles.items,
   (articlesById, articleIds) =>
-    articleIds.map(id => transformArticle(articlesById[id]))
+    orderBy(
+      articleIds.map(id => transformArticle(articlesById[id])),
+      ['cretedAt', 'id'],
```
SmithPeder: creeeeettteeeeeeeeeedAAAAt

Issue #1226, opened 2018-04-24
Error opening quote url


odinuge:
When visiting a quote url eg. https://abakus.no/quotes/255 without being logged in, you can see a login screen. When you login successfully, the page crashes.

https://sentry.abakus.no/webkom/lego-webapp/issues/5157/

SmithPeder:
I don't know why this wasn't autoclosed 🐳

PR #1349, opened 2018-10-31
Fix undefined quotes causing error

mrfarstad:
Solves #1226 🎉


SmithPeder:
MFW you write one line and fail the lint check 💯

kotlarz:
:fire:

PR #1352, opened 2018-10-31
Add extra steps to deleting an event

fireflower1:
To make sure no one deletes an event by accident, add a few extra steps to the process

### DIFF
--- a/app/routes/events/components/Admin.js
+++ b/app/routes/events/components/Admin.js
@@ -61,7 +61,7 @@ class DeleteButton extends React.Component<ButtonProps, State> {
             <input
               type="text"
               id="slettArrangement"
-              placeholder="Arrangement du vil slette"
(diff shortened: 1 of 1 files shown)


app/routes/events/components/Admin.js
```diff
@@ -13,21 +13,65 @@ type Props = {
...
+            Slett
+          </span>
+        )}
+        {this.state.show && (
+          <form>
+            <span> Skriv inn navnet på arrangementet du vil slette: </span>
+            <input
+              type="text"
+              id="slettArrangement"
+              placeholder="Arrangement du vil slette"
```
SmithPeder: This placeholder is a little too long, maybe do something like `Arrangementnavn`

fireflower1: Ok, replaced it with your suggestion

PR #1354, opened 2018-11-01
Temporary frontpage fix - remove feeds, insert weekly

cherrybeans:
In light of the wishes of Abakom (and personal agreement to remove it before its useful) to remove "kunngjøringer" from the frontpage (feeds basically, but only announcements as is), I've removed it and replaced it with the prettier Weekly articles section. 

This is not meant as a permanent fix, rather a quick fix to improve upon the existing before we can work more on the structure (which we know it will change in the future).

Announcements can still be viewed from your notifications.

Please approve of this change <3 

screenshot from 2018-11-01 12-35-26

screenshot from 2018-11-01 12-35-54

screenshot from 2018-11-01 12-36-11

### DIFF
--- a/app/routes/overview/components/Overview.css
+++ b/app/routes/overview/components/Overview.css
@@ -68,6 +68,35 @@
   flex-wrap: wrap;
 }
 
+.weeklyArticles {
+  border: 1px solid rgba(0, 0, 0, 0.09);
+  box-shadow: 0 1px 4px rgba(0, 0, 0, 0.04);
+  border-radius: 3px;
+  overflow-y: auto;
+  height: 480px;
+  flex: 1;
+  background: var(--lego-card-color);
+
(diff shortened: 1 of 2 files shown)


cherrybeans:
So you mean changing back from 0.8 to 2 again? The reason was simply that it looked like it was too big, but it's a fair counter argument, so sure.

SmithPeder:
screenshot 2018-11-01 at 13 12 07

And this Weekly dont line up very good with the other headers, seem a bit low

cherrybeans:
screenshot from 2018-11-01 13-37-20

This is as good as I'm gonna get it without doing some major restructuring of the frontpage css, which would be better to take at a later time. In the future, refactoring to  use CSS Grid would, I think, make it a lot easier.

PR #1357, opened 2018-11-01
Display position of registration in waiting registration list

JamiesWhiteShirt:
I'd like to get a spot on the Julebord, but I don't know where I am in the waiting registration list! Here's something to make waiting to get moved out of the waiting registration list a little more exciting.

bilde

I have implemented a prop for EventDetail that specifies the position of the current registration in its registration pool. This value is displayed accordingly if the pool is "Venteliste".

### DIFF
--- a/app/routes/events/EventDetailRoute.js
+++ b/app/routes/events/EventDetailRoute.js
@@ -96,6 +96,7 @@ const mapStateToProps = (state, props) => {
     );
     currentRegistration = currentPool.registrations[currentRegistrationIndex];
   }
+  const hasSimpleWaitingList = poolsWithRegistrations.length <= 1;
 
   return {
     comments,
@@ -106,7 +107,8 @@ const mapStateToProps = (state, props) => {
     pools,
     registrations,
     currentRegistration,
-    currentRegistrationIndex
+    currentRegistrationIndex,
+    hasSimpleWaitingList
   };
 };
 
--- a/app/routes/events/EventDetailRoute.js
+++ b/app/routes/events/EventDetailRoute.js
@@ -27,9 +27,6 @@ import {
 import loadingIndicator from 'app/utils/loadingIndicator';
 import helmet from 'app/utils/helmet';
 
-const findCurrentRegistration = (registrations, currentUser) =>
-  registrations.find(registration => registration.user.id === currentUser.id);
-
 const mapStateToProps = (state, props) => {
   const {
     params: { eventId },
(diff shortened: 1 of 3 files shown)


SmithPeder:
I did some more testing, and I see you have implemented separate lists when you enter the `Venteliste` modal. This is very nice, but it seem to have a `bug` 💯 , where you land in a random class each time you open it.

screenshot 2018-11-03 at 16 10 46
screenshot 2018-11-03 at 16 11 23
screenshot 2018-11-03 at 16 11 28
screenshot 2018-11-03 at 16 11 37
screenshot 2018-11-03 at 16 11 44

SmithPeder:
Disregard my last comment _(now deleted),_ i checked out to the wrong branch. There was another branch with a similar name in origin, so there was nothing wrong with your implementation.

JamiesWhiteShirt:
Yes, the first matching pool will be found.

Seeing as there is a single "current registration" I assumed a user could be registered in only one pool per event. With this assumption, overlapping pools should not cause any issues. Is this assumption correct?

SmithPeder:
I just bumped myself to 3.grade for this example. It says im number 7 in the list, and im now registred in both pools, so i will be bumped once one 3. or 4. grader leaves, so actully im nr 1 in the waiting list.

**before**
screenshot 2018-11-03 at 16 34 05

**after kicking two people in 4.grade**
screenshot 2018-11-03 at 16 37 35

app/routes/events/EventDetailRoute.js
```diff
@@ -96,7 +102,8 @@ const mapStateToProps = (state, props) => {
     eventId,
     pools,
     registrations,
-    currentRegistration
+    currentRegistration,
+    currentRegistrationIndex
```
SmithPeder: This is missing in the Props Flow validation. 
`currentRegitrationIndex: number` needs to be added

app/routes/events/components/RegistrationMeta.js
```diff
@@ -5,10 +5,15 @@ import { hasPaid } from '../utils';
 
 type Props = {
   registration: Object,
-  isPriced: boolean
+  isPriced: boolean,
+  registrationIndex?: Number
```
SmithPeder: `number` needs to be in lowercase or flow complains

JamiesWhiteShirt:
I see. The position in the waiting list is therefore technically correct, but the position in the list may not be representative of when you will be moved out of it. That makes this much more complex than what I set out to do...

JamiesWhiteShirt:
I believe we don't even need overlapping pools for the number to be misleading. Let's say you have a pool for 1-3 and one for 4-5. The waiting list can contain users who will be assigned to different pools.

SmithPeder:
So in a case where you're allocated two different waiting list spots I see two outcomes.

1. You get the position in a pool and is bumped when the number gets to 0
2. You get the position in a pool and is either bumped when it reaches 0 or when the other pool reaches 0.

In either scenario, you are bumped when the number goes to 0, or earlier(which is fine)

SmithPeder:
So we discussed this some at our weekly meeting, and we figured we could just merge this with some minor changes. If you fix the flow stuff, and then add a chick check `number of pools > 2` which means there is more then one pool + waiting list pool. We did some CI changes, so if you push the new changes it should run and we can see if this works. 🥇

JamiesWhiteShirt:
@SmithPeder I was about to point out how impressively fast the CI was, but it looks like the CI only checked out the code, nothing else.

SmithPeder:
Yehh we changed the CI so it would run on all pull-requsts. Ref (https://github.com/webkom/lego-webapp/commit/93495d40e02e5e84567cbcb46688563385850ecf) So if you rebase with the current master your should get the current CI settings and it should run. We also have a new prettier verison since the creation of your fork, but i dont think it will affect any of your changes.

JamiesWhiteShirt:
There we go! Too bad I already got my spot on the Julebord now...

SmithPeder:
Nice work 👍

PR #1367, opened 2018-11-08
Don't show 0/0 attendance pills for events

JamiesWhiteShirt:
Pills showing "0/0" look slightly ugly in my opinion, so here I have simply removed the pills when the total capacity of the event is zero (which I assume is what qualifies as some event not requiring registration). Showing something else is an alternative - let me know if you would prefer that.

bilde

Note that this also applies to the tooltips in the event calendar.

Speaking of events that do not require registration - When researching how to approach this, I stumbled upon similar logic elsewhere in the application from #1293 which does not appear to be correct:

https://github.com/webkom/lego-webapp/blob/be4a72a10df25a9f489e2a86c842909eef271a6d/app/routes/overview/components/EventItem.js#L32-L38

I believe this would result in "Ingen påmelding" when the event has nonzero capacity but no participants. That does not look like the intended behavior. If it is, it should be applied in this PR too. If not, I'll probably make a separate PR to fix that as well.

### DIFF
--- a/app/routes/events/components/CalendarCell.js
+++ b/app/routes/events/components/CalendarCell.js
@@ -46,9 +46,11 @@ const renderEvent = ({
     <div>
       <h3 className={styles.eventItemTitle}>
         {title}
-        <Pill style={{ marginLeft: 10 }}>
-          {`${registrationCount} / ${totalCapacity}`}
-        </Pill>
+        {totalCapacity > 0 && (
+          <Pill style={{ marginLeft: 10 }}>
+            {`${registrationCount} / ${totalCapacity}`}
+          </Pill>
+        )}
       </h3>
 
       {description}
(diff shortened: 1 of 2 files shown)


SmithPeder:
Nice. When it comes to the "Ingen påmelding" showing it was intended behavior. I
I created it as a quick fix before we implement a larger fix for this. The thing is event capacity is a little wired. There is alot of different checks for totalcapacity. Some events have no registration due to the merits of the event. Others have registations, but an infinite number of spots. For some events BedKom has jet to create pools, so they will have 0/0, and some events will not have opened yet so that will have 0/x.

The thought is to change a little in the backend and implement the new and better way for users to see events. There are methods for getting the specific registration_open_for_user, factoring in things like "prikker", grade and so forth. This is going to allow us to create a "The next event you can sign up for" section, and maybe some filtering for what events the user actually can sign up for.

odinuge:
Nice work! This runs in production (https://abakus.no) now! :smile:

PR #1374, opened 2018-11-18
Remove global css

SmithPeder:
All ul list elements in the webapp got this new ul styling. CSS-modules only hash/scope the custom classNames


SmithPeder:
@estensen 💯

PR #1387, opened 2019-01-11
Add support for specific recipients

### LINKED ISSUE #1385: Allow for specific recipients in contact form
The contact form should be updated to match https://github.com/webkom/lego/pull/1345 

- Title should be changed to "Kontaktskjema for Abakus"
- A paragraph should be added between the title and the first field, containing the following text
> Dette skjemaet kan benyttes for å kontakte Abakus sine komiteer eller Hovedstyret. Sender du til en spesifikk komité er det kun lederen av komiteen som vil motta meldingen. Om du sender til Hovedstyret vil hele styret få meldingen. 

> Send oss gjerne spørsmål, feedback eller ting du ønsker å fortelle oss. Du kan også benytte dette skjemaet om du ønsker å avtale tid for innsyn i HSP-avtalen.
- A new field should be added to select the recipient. This dropdown should contain a list of all committees. The `None`-choice should be labeled `Hovedstyret`. See the grade-change dropdown on the user profiles.
- The lower text should be changed to:
> Du kan velge å sende meldingen med anonym avsender. De som mottar meldingen vil ikke få vite hvem som har opprettet meldingen. De vil da heller ikke ha mulighet til å svare på meldingen. Ønsker om innsyn kan ikke sendes inn anonymt.

edvardvb:
Fix #1385

### DIFF
--- a/app/routes/contact/components/ContactForm.js
+++ b/app/routes/contact/components/ContactForm.js
@@ -9,7 +9,8 @@ import {
   TextInput,
   TextArea,
   CheckBox,
-  Captcha
+  Captcha,
+  SelectInput
 } from 'app/components/Form';
 import type { ContactForm as ContactFormType } from 'app/reducers/contact';
 import type { FieldProps } from 'redux-form';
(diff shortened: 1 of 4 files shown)


app/routes/contact/components/ContactForm.js
```diff
@@ -19,27 +21,62 @@ type Props = FieldProps & {
...
 const ContactForm = (props: Props) => {
-  const { invalid, pristine, submitting } = props;
+  const { invalid, pristine, submitting, groups } = props;
   const disabledButton = invalid || pristine || submitting;
 
   const submit = data => {
     return props
-      .sendContactMessage(data)
+      .sendContactMessage({
+        ...data,
```
SmithPeder: Does this make sense? Wrapping the spread version of the `data` object and setting the `recipient_group` as `data.recipient_group.value`.

edvardvb: Not sure, but it works 😅 Switching their order might make more sense?

SmithPeder: I haven't looked at the `sendContactMessage` function, but it used to take an object, and now it takes the same object with one more value set to `data.recipient_group.value`. So in my mind, you could just send the data object and extract the `recipient_group.value` later, as it will be available. But if it works it's good :)

PR #1390, opened 2019-01-15
Fix Ingen påmelding tag

SmithPeder:
Long ago I wrote some bad code that made all events 0/0 or 0/x say Ingen påmelding. This was wrong as all events that were yet to open got this message. Long ago we fixed the serializer so activation time is sent with the event.

This will(should) add a better tooltip under the event. I think I've thought of all scenarios that can happen to an event.

screenshot 2019-01-15 at 21 57 20

### DIFF
--- a/app/routes/overview/components/EventItem.js
+++ b/app/routes/overview/components/EventItem.js
@@ -9,6 +9,7 @@ import { Link } from 'react-router';
 import { Flex } from 'app/components/Layout';
 import { colorForEvent } from 'app/routes/events/utils';
 import styles from './EventItem.css';
+import moment from 'moment-timezone';
 
 type Props = {
   item: Event | Article,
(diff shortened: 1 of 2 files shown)


app/routes/overview/components/EventItem.js
```diff
@@ -20,7 +21,16 @@ class EventItem extends Component<Props, *> {
   render() {
     const { item, url, meta } = this.props;
     const TITLE_MAX_LENGTH = 50;
-    const { registrationCount, totalCapacity } = item;
+    const { registrationCount, totalCapacity, activationTime } = item;
+
+    // This value will less then 0 if activationTime has yet to come
+    let future = moment().diff(activationTime) < 0;
```
odinuge: Maybe `const` this?

SmithPeder: ups 💯

PR #1397, opened 2019-01-17
Next event

SmithPeder:
Created a component that looks like this. Planning on putting it on the frontpage. 
The picture is a little outdated, but it will show the time remaining before an event opens for admission.

- It will show zero, one or two events
- It will only show events that have admission the next three days
- It will show events that have admission this day, even when they are open.

screenshot 2019-01-15 at 20 50 00


app/routes/overview/components/NextEvent.js
```diff
@@ -0,0 +1,126 @@
...
+import { orderBy } from 'lodash';
+import moment from 'moment-timezone';
+import Icon from 'app/components/Icon';
+import Tooltip from 'app/components/Tooltip';
+
+type Props = {
+  event: Event
+};
+
+class EventItem extends React.Component<Props, *> {
```
odinuge: `type State = {
...`

app/routes/overview/components/NextEvent.js
```diff
@@ -0,0 +1,126 @@
...
+  state = {
+    time: ''
+  };
+  interval: IntervalID;
+
+  setTime(props) {
+    moment.locale('no');
+    let now = moment();
+    let start = moment(props.event && props.event.activationTime);
```
odinuge: `const`

app/routes/overview/components/NextEvent.js
```diff
@@ -0,0 +1,126 @@
...
+
+class EventItem extends React.Component<Props, *> {
+  state = {
+    time: ''
+  };
+  interval: IntervalID;
+
+  setTime(props) {
+    moment.locale('no');
+    let now = moment();
```
odinuge: `const`

app/routes/overview/components/NextEvent.js
```diff
@@ -0,0 +1,126 @@
...
+    let start = moment(props.event && props.event.activationTime);
+    let time = now.to(start);
```
odinuge: `const`

app/routes/overview/components/NextEvent.js
```diff
@@ -0,0 +1,126 @@
...
+};
+
+class EventItem extends React.Component<Props, *> {
+  state = {
+    time: ''
+  };
+  interval: IntervalID;
+
+  setTime(props) {
+    moment.locale('no');
```
odinuge: I don't think you need this (but not 110% sure), I think we have a global one somewhere.

PR #1398, opened 2019-01-17
Redo frontpage to be more responsive

LudvigHz:
Rewrote frontpage to use css grid on mobile and a more responsive flexbox layout on full size screens. Also changed so only one weekly article is shown.


odinuge:
@LudvigHz @SmithPeder pls fix https://sentry.abakus.no/webkom/lego-webapp/issues/6209/ :computer:

PR #1401, opened 2019-01-18
Send announcement from a group

### LINKED ISSUE #1322: Impotent message author
Admins should always write `important messages` but we need we should be able to write author on the message

LudvigHz:
Adds the option to send an announcement from a group. If specified, the selected group replaces the user as the sender. The user who creates the announcements is still shown in the timeline.
Resolve https://github.com/webkom/lego/issues/1389

### DIFF
--- a/app/models.js
+++ b/app/models.js
@@ -183,6 +183,11 @@ export type Announcement = {
   groups: Array<Object>,
   events: Array<Object>,
   meetings: Array<Object>,
+  fromGroup: {
+    id: number,
+    name: string,
+    type: string
+  },
   sent?: boolean
 };
 
--- a/app/models.js
+++ b/app/models.js
@@ -183,11 +183,7 @@ export type Announcement = {
   groups: Array<Object>,
   events: Array<Object>,
   meetings: Array<Object>,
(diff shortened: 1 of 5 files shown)


app/models.js
```diff
@@ -180,6 +180,11 @@ export type Announcement = {
   groups: Array<Object>,
   events: Array<Object>,
   meetings: Array<Object>,
+  fromGroup: {
```
odinuge: Maybe use the `Group` entity here?

app/routes/announcements/components/AnnouncementsCreate.js
```diff
@@ -86,6 +87,13 @@ const AnnouncementsCreate = ({
                 component={SelectInput.AutocompleteField}
               />
             </Flex>
+            <span className={styles.formHeaders}>Sendt fra:</span>
```
odinuge: `Send på vegne av` maybe?

finnss:
Anything missing here? @LudvigHz

LudvigHz:
This depends on https://github.com/webkom/lego/pull/1348

PR #1406, opened 2019-01-21
Update copyright year

edvardvb:
This was quite out of date. 🤓

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014-2017 Abakus Webkom
+Copyright (c) 2014-2019 Abakus Webkom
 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014-2017 Abakus Webkom
```
edvardvb: Notice how this says 2017 instead of 2018.

SmithPeder:
Have you considered using `momentjs` for this to make it dynamic and responsive for years to come?

edvardvb:
I think that would be a great idea for future enhancements, and it should be added to the task list, but we need this fix implemented asap to continue development.

PR #1408, opened 2019-01-21
Only list future events & joblistings on company detail page

LudvigHz:


app/reducers/companies.js
```diff
@@ -246,7 +250,8 @@ export const selectJoblistingsForCompany = createSelector(
     return joblistings.filter(
       joblisting =>
         joblisting.company &&
-        Number(joblisting.company.id) === Number(companyId)
+        Number(joblisting.company.id) === Number(companyId) &&
+        moment().isBefore(joblisting.deadline)
```
edvardvb: Sometimes users want to see old joblistings, so that they can prepare for an interview. How can we handle this?

app/reducers/companies.js
```diff
@@ -233,7 +234,10 @@ export const selectEventsForCompany = createSelector(
   (events, companyId) => {
     if (!companyId || !events) return [];
     return events.filter(
-      event => event.company && Number(event.company.id) === Number(companyId)
+      event =>
+        event.company &&
+        Number(event.company.id) === Number(companyId) &&
+        moment().isBefore(event.startTime)
```
edvardvb: I know the suggestion from Bedkom is to only show upcoming events, but I feel it might be useful to show recent events as well. Thoughts?

Also, are we sure that the backend will give us the most recent events? Check Bekk in prod, it only shows events from 2010 and earlier. We might run into some pagination and sorting issues here!

PR #1422, opened 2019-01-28
Add function for creating event from template

SmithPeder:
Solves https://github.com/webkom/lego/issues/1355

This allows the user to create a new event based on another event (use an old event as template)
screenshot 2019-01-28 at 21 36 30

This is done via query params.
screenshot 2019-01-28 at 21 36 42

### DIFF
--- a/app/routes/events/EventCreateRoute.js
+++ b/app/routes/events/EventCreateRoute.js
@@ -26,6 +26,49 @@ const mapStateToProps = (state, props) => {
     eventId
   });
 
+  const selectedValues = {
+    event: {
+      addFee: valueSelector(state, 'addFee'),
+      isPriced: valueSelector(state, 'isPriced'),
+      eventType: valueSelector(state, 'eventType'),
+      priceMember: valueSelector(state, 'priceMember'),
+      startTime: valueSelector(state, 'startTime')
+    },
+    pools: valueSelector(state, 'pools')
+  };
+
+  const initialCreateValues = {
+    initialValues: {
+      title: '',
+      startTime: time({ hours: 17, minutes: 15 }),
+      endTime: time({ hours: 20, minutes: 15 }),
+      description: '',
+      text: '',
+      eventType: '',
+      company: null,
+      responsibleGroup: null,
+      location: 'TBA',
+      isPriced: false,
+      useStripe: true,
+      priceMember: 0,
+      paymentDueDate: time({ days: 7, hours: 17, minutes: 15 }),
+      mergeTime: time({ hours: 12 }),
+      useCaptcha: true,
+      heedPenalties: true,
+      isAbakomOnly: false,
+      feedbackDescription: 'Melding til arrangører',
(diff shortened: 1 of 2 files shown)


app/routes/events/EventCreateRoute.js
```diff
@@ -1,41 +1,86 @@
-// @flow
```
odinuge: :trollface:

app/routes/events/EventCreateRoute.js
```diff
@@ -55,8 +100,10 @@ const mapDispatchToProps = {
 
 export default compose(
   replaceUnlessLoggedIn(LoginPage),
+  prepare(({ location }, dispatch) => dispatch(fetchEvent(location.query.id))),
```
odinuge: What if this is a `create`, and `query.id` is `undefined`?

app/routes/events/EventCreateRoute.js
```diff
@@ -1,41 +1,86 @@
...
+    isAbakomOnly: false,
+    feedbackDescription: 'Melding til arrangører',
+    pools: [],
+    unregistrationDeadline: time({ hours: 12 })
+  };
+
+  if (isEmpty(eventTemplate) && eventId) {
+    return { initialValues: null };
+  }
+
```
odinuge: Maybe rewrite the return logic, to make it a bit more sensible?

```js
if (!eventId) return {initialValues: initialValuesForCreate};
if (isEmpty(eventTemplate)) return {initialValues: null}
return {
    ...stuff
};
```

odinuge:
@ingridkindem :arrow_up:

app/routes/events/EventCreateRoute.js
SmithPeder: Hmm, I thought about that when when returning initialValues, but forgot to check the dispatch. Good catch

app/routes/events/EventCreateRoute.js
SmithPeder: This won't work! You really got me confused with this one 👎Been looking at _Cannot read 'eventType' of undefined for an hour now_.

app/routes/events/EventCreateRoute.js
```diff
@@ -36,6 +49,51 @@ const mapStateToProps = (state, props) => {
...
     },
+    event: {},
+    pools: [],
+    actionGrant: []
+  };
+
+  /* If there is no eventId in the params then we use the
+   * normal initalValues for a new event
+   */
+  if (!eventId) return initialCreateValues;
```
odinuge: You will need the
```
event: {
     addFee: valueSelector(state, 'addFee'),
etc....
``` here.

otherwise it won't work :/

SmithPeder: I can't just return an empty object?

SmithPeder: It looks to working fine in the browser

Issue #1266, opened 2018-08-25
Captcha should be active earlier when registering for event


estensen:
The captcha when registering for an event pops up 1 min before the registering opens.
If you are really unfortunate you get a difficult captcha and won't get a spot if it's a popular event.

Suggestion: increase the captcha popup to 2 min.

PR #1425, opened 2019-01-31
Change time for captcha to show from 1 minute before event opens to 90 seconds

asmundh:
Fikser dette captcha timeren?? Hvis ja, det er litt kjipt når captchaen tar over ett minutt å løse og den først vises ett minutt før påmelding. Hvis feil, please send help

### DIFF
--- a/app/routes/events/components/JoinEventFormCountdownProvider.js
+++ b/app/routes/events/components/JoinEventFormCountdownProvider.js
@@ -30,7 +30,7 @@ const COUNTDOWN_INTERVAL = 1000;
 // Must be sorted lo->hi
 const TICK_ACTIONS: Array<[number, Action]> = [
   [0, 'REGISTRATION_AVAILABLE'],
-  [3 * 60 * 1000, '90_SECONDS_LEFT'],
+  [1.5 * 60 * 1000, '90_SECONDS_LEFT'],
   [60 * 10000, '10_MINUTE_LEFT'],
   [Infinity, 'STILL_WAITING']
 ];
--- a/app/routes/events/components/JoinEventFormCountdownProvider.js
+++ b/app/routes/events/components/JoinEventFormCountdownProvider.js
@@ -6,7 +6,7 @@ import type { Dateish, Event, EventRegistration } from 'app/models';
 
 type Action =
   | 'REGISTRATION_AVAILABLE'
-  | '3_MINUTE_LEFT'
+  | '90_SECONDS_LEFT'
   | '10_MINUTE_LEFT'
   | 'STILL_WAITING'
   | 'REGISTERED_OR_REGISTRATION_ALREADY_OPENED'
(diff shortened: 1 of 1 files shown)


app/routes/events/components/JoinEventFormCountdownProvider.js
```diff
@@ -30,7 +30,7 @@ const COUNTDOWN_INTERVAL = 1000;
 // Must be sorted lo->hi
 const TICK_ACTIONS: Array<[number, Action]> = [
   [0, 'REGISTRATION_AVAILABLE'],
-  [60 * 1000, '1_MINUTE_LEFT'],
+  [180 * 1000, '3_MINUTE_LEFT'],
```
edvardvb: 3 * 60 * 1000 for maximum readability? 😛

asmundh: Fair 8-)

hanswilw:
I think the verification time is (or atleast was) 2 minutes when we made this. Meaning it potentially would invalidate before you press register. Should check that out before merging this!

odinuge:
cc: #1266 & #540

asmundh:
3 minutes wouldn't work then. 90 seconds might help prevent a user from not having enough time to solve the captcha. 
Do you know if the captcha reset timer starts when a user confirms their humanity? If so, 2 minutes could also work, chances are no humans will solve the captcha in <1 second. This will however give the user a smaller time window from an event opening until they're too late in registering.

odinuge:
I have read quite a few stuff, but I haven't found a proper "black-on-white" explanation. I think that the "timer" starts when the captcha is verified/turns green.

The problem I see with 2 minutes is if the users captcha is verifyed without having to "find buses", it will reset 1-2 secs after the registration opens, and then the user has to click into the captcha again. Using 90 sec will leave a time window of approx 30sec after the registration opens, and that should be enough.

odinuge:
@asmundh If you change this to 90 sec, we can do some testing, and get this into master :taco: :tada:

asmundh:
Should I merge into a different branch than master?

app/routes/events/components/JoinEventFormCountdownProvider.js
```diff
@@ -30,7 +30,7 @@ const COUNTDOWN_INTERVAL = 1000;
 // Must be sorted lo->hi
 const TICK_ACTIONS: Array<[number, Action]> = [
   [0, 'REGISTRATION_AVAILABLE'],
-  [3 * 60 * 1000, '3_MINUTE_LEFT'],
+  [3 * 60 * 1000, '90_SECONDS_LEFT'],
```
JonasBak: Shouldn't the value also be changed to `90 * 1000` or `1.5 * 60 * 1000`?

asmundh: Oops, yes.

Issue #1259, opened 2018-08-16
Youtube video as article cover


odinuge:
It should be possible to add a youtube-url, and us it as the cover on an article. Normal cover would still be required becuse of sharing etc.

PR #1428, opened 2019-02-05
Youtube player as cover in events, articles and joblistings

mrfarstad:
Implement embedded Youtube player as cover in the detail view of articles. This will make it possible to post a link with attributes which will become the properties of the Youtube video. Works with ssr 🎉 

The list of available parameters is listed here: [https://developers.google.com/youtube/player_parameters](url)

What parameters we ought to use is discussable, but the basic video id and start time will be implemented in the first run :skra:

Need to store these links in the API, and this needs to be merged before merging this 🍰 

https://github.com/webkom/lego/pull/1455

#1259

### DIFF
--- a/app/components/Content/Content.js
+++ b/app/components/Content/Content.js
@@ -8,12 +8,12 @@ import { Image } from 'app/components/Image';
 import Youtube from 'react-youtube';
 import Flex from '../Layout/Flex';
 import LoadingIndicator from 'app/components/LoadingIndicator';
-import type { YoutubePlayer } from 'app/models';
 import { isEmpty } from 'lodash';
+import getParamsFromUrl from 'app/utils/getParamsFromUrl';
 
 type Props = {
   banner?: string,
-  youtubeParams?: YoutubePlayer,
+  youtubeUrl?: string,
   className?: string,
   children: Node
 };
@@ -35,9 +35,10 @@ type Props = {
  * ```
  */
 
-function Content({ banner, youtubeParams, children, className }: Props) {
+function Content({ banner, youtubeUrl, children, className }: Props) {
   const [isClicked, click] = useState(false);
   const [isLoading, setIsLoading] = useState(true);
+  const youtubeParams = youtubeUrl && getParamsFromUrl(youtubeUrl);
 
   return (
     <Flex column alignItems="center">
--- a/app/components/Content/Content.js
+++ b/app/components/Content/Content.js
@@ -6,6 +6,7 @@ import styles from './Content.css';
 import { Image } from 'app/components/Image';
 import Youtube from 'react-youtube';
 import Flex from '../Layout/Flex';
+import LoadingIndicator from 'app/components/LoadingIndicator';
 
 type Props = {
   banner?: string,
@@ -33,22 +34,32 @@ type Props = {
 
 function Content({ banner, youtubeParams, children, className }: Props) {
   const [isClicked, click] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
 
   return (
     <Flex column alignItems="center">
       {youtubeParams ? (
-        <Flex
-          justifyContent={'center'}
-          className={styles.youtubeFrame}
-          style={{ paddingBottom: isClicked ? '56.25%' : '30%' }}
-        >
-          <Youtube
-            videoId={youtubeParams.v}
-            opts={{ playerVars: { start: youtubeParams.t } }}
-            onPlay={() => click(true)}
-            style={{ width: '100%' }}
+        <div>
+          <LoadingIndicator
+            loading={isLoading}
+            className={isLoading ? {} : styles.hidden}
           />
-        </Flex>
+          <Flex
+            justifyContent={'center'}
+            style={{
+              maxHeight: isLoading ? '0' : isClicked ? '619px' : '358px'
+            }}
+            className={styles.youtubeFrame}
+          >
+            <Youtube
+              videoId={youtubeParams.v}
+              opts={{ playerVars: { start: youtubeParams.t } }}
+              onStateChange={() => click(true)}
+              onReady={() => setIsLoading(false)}
+              className={isLoading ? styles.hidden : {}}
+            />
+          </Flex>
+        </div>
       ) : (
         banner && (
           <div className={cx(styles.cover, className)}>
--- a/app/components/Content/Content.js
+++ b/app/components/Content/Content.js
@@ -1,16 +1,15 @@
 // @flow
 
-import React, { type Node } from 'react';
+import React, { type Node, useState } from 'react';
 import cx from 'classnames';
 import styles from './Content.css';
 import { Image } from 'app/components/Image';
 import Youtube from 'react-youtube';
 import Flex from '../Layout/Flex';
-import { FlexColumn } from '../FlexBox';
 
 type Props = {
   banner?: string,
-  youtubeUrl?: string,
+  youtubeParams?: Object,
   className?: string,
   children: Node
 };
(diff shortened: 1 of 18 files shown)


app/components/Content/Content.js
```diff
@@ -27,24 +35,48 @@ type Props = {
...
  */
-function Content({ banner, children, className }: Props) {
-  return (
-    <div>
-      {banner && (
-        <div className={cx(styles.cover, className)}>
-          <Image src={banner} />
+
+class Content extends React.Component<Props, State> {
+  state = {
```
mrfarstad: A bit unsure if I want to change the `<Content>` component  like this or just change `<ArticleDetail>` instead. This works for now, but I think it's a bit risky to change the `<Content>` component

app/routes/articles/components/ArticleDetail.js
```diff
@@ -19,7 +19,8 @@ type Props = {
   comments: Array<CommentEntity>,
   loggedIn: boolean,
   author: UserEntity,
-  currentUser: UserEntity
+  currentUser: UserEntity,
+  youtubeParams: Object
```
odinuge: youtubeParams should be optional, and maybe it should be a part of the ArticleEntity instead?

app/routes/articles/components/ArticleEditor.js
```diff
@@ -130,10 +137,17 @@ const onSubmit = (
...
+  const youtubeParams = {};
+  for (const [key, value] of params.entries()) {
+    youtubeParams[key] = value;
+  }
+
   const body = {
     ...(isNew ? {} : { id: articleId }),
     ...(data.cover ? { cover: data.cover } : {}),
     ...normalizeObjectPermissions(data),
+    youtubeParams,
```
odinuge: Isn't it better to save the url, instead of parsing it before sending? You should also remember to make it support initialValues into the form.

app/routes/articles/components/ArticleEditor.js
```diff
@@ -64,6 +64,13 @@ const ArticleEditor = ({
           aspectRatio={20 / 6}
           img={article && article.cover}
         />
+        <Field
+          name="youtubeUrl"
+          label="YouTube cover url"
+          placeholder="https://www.youtube.com/watch?v=VvFj1zdT9so&t=20"
+          component={TextInput.Field}
+          aspectRatio={20 / 6}
```
odinuge: Why aspectRatio here? :thinking:

app/components/Content/Content.js
odinuge: Hmm. Imo its ok, because we can then reuse it in events.

Adding state to the Content component is another case tho.

app/components/Content/Content.js
```diff
@@ -4,13 +4,21 @@ import React, { type Node } from 'react';
 import cx from 'classnames';
 import styles from './Content.css';
 import { Image } from 'app/components/Image';
+import Youtube from 'react-youtube';
+import Flex from '../Layout/Flex';
+import { FlexColumn } from '../FlexBox';
```
odinuge: This is deprecated. Or atleast it should be

app/components/Content/Content.js
```diff
@@ -27,24 +35,48 @@ type Props = {
...
+class Content extends React.Component<Props, State> {
+  state = {
+    isClicked: false
+  };
+
+  render() {
+    const { banner, youtubeParams, children, className } = this.props;
+    const { isClicked } = this.state;
+    return (
+      <FlexColumn alignItems={'center'}>
```
odinuge: `alignItems="center"`

app/components/Content/Content.js
mrfarstad: True, but I think I will need state if I want the component to expand if clicked 🤔

app/routes/articles/components/ArticleEditor.js
mrfarstad: True, I can fix so that the parsing happens on receiving the url from the API

app/routes/articles/components/ArticleEditor.js
mrfarstad: le copy pasta :skra:

mrfarstad:
Needs more validation on the URL. Need to verify that the link is from Youtube with the correct parameters

SmithPeder:
You define the params `youtubeParams: Object`. Then use vars like v and t.
```
<Youtube
  videoId={youtubeParams.v}
  opts={{ playerVars: { start: youtubeParams.t } }}
  onPlay={() => click(true)}
  style={{ width: '100%' }}
/>
```

Maybe type this up with a type in `models.js`?  I see you're pulling the params from the searchParams from the youtubeUrl. I guess this is the video and time tags for a video like `https://www.youtube.com/watch?v=_h1ooyyFkF0`? 

Hard to read code with something that has `.v` and `.s` from an object :) 
```
const params = new URL(article.youtubeUrl).searchParams;
    for (const [key, value] of params.entries()) {
      youtubeParams[key] = value;
    }
```

app/components/Content/Content.js
```diff
@@ -1,12 +1,15 @@
...
+import React, { type Node, useState } from 'react';
 import cx from 'classnames';
 import styles from './Content.css';
 import { Image } from 'app/components/Image';
+import Youtube from 'react-youtube';
+import Flex from '../Layout/Flex';
 
 type Props = {
   banner?: string,
+  youtubeParams?: Object,
```
SmithPeder: Type YoutubeParams

app/routes/articles/components/ArticleDetail.js
SmithPeder: And Type YoutubeParams

app/routes/articles/ArticleDetailRoute.js
```diff
@@ -22,12 +23,19 @@ const mapStateToProps = (state, props) => {
   const comments = selectCommentsForArticle(state, { articleId });
   const author = selectUserById(state, { userId: article.author });
 
+  let youtubeParams = {};
+  const { youtubeUrl } = article;
+  if (youtubeUrl) {
+    youtubeParams = getParamsFromUrl(youtubeUrl);
```
odinuge: Isn't it better to do this inside the `Content` component?

app/routes/articles/components/ArticleDetail.js
```diff
@@ -33,9 +35,10 @@ const ArticleDetail = ({
   loggedIn,
   currentUser,
   comments,
-  deleteComment
+  deleteComment,
+  youtubeParams
 }: Props) => (
-  <Content banner={article.cover}>
+  <Content banner={article.cover} youtubeParams={youtubeParams}>
```
odinuge: And pass url here

app/components/Content/Content.js
```diff
@@ -27,13 +34,40 @@ type Props = {
...
+  const [isClicked, click] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+
   return (
-    <div>
-      {banner && (
-        <div className={cx(styles.cover, className)}>
-          <Image src={banner} />
+    <Flex column alignItems="center">
+      {!isEmpty(youtubeParams) ? (
```
odinuge: Why `!isEmpty`? You save one byte by switching the logic to `isEmpty`

app/components/Content/Content.js
```diff
@@ -27,13 +34,40 @@ type Props = {
...
+        <div>
+          <LoadingIndicator
+            loading={isLoading}
+            className={isLoading ? {} : styles.hidden}
```
odinuge: Hmmm.
Why not 

```jsx
<LoadingIndicator loading={isLoading}>  
          <Flex
            justifyContent={'center'}
            style={{
              maxHeight: isLoading ? '0' : isClicked ? '619px' : '358px'
            }}
            className={cx(styles.youtubeFrame, isLoading ? styles.hidden : {})}
          >
            <Youtube
              videoId={youtubeParams.v}
              opts={{ playerVars: { start: youtubeParams.t } }}
              onStateChange={() => click(true)}
              onReady={() => setIsLoading(false)}
            />
          </Flex>
</LoadingIndicator>
```

app/components/Content/Content.js
```diff
@@ -27,13 +34,40 @@ type Props = {
...
+          />
+          <Flex
+            justifyContent={'center'}
+            style={{
+              maxHeight: isLoading ? '0' : isClicked ? '619px' : '358px'
```
odinuge: Double trouble

app/components/Content/Content.js
```diff
@@ -27,13 +34,40 @@ type Props = {
...
+            }}
+            className={cx(styles.youtubeFrame, isLoading ? styles.hidden : {})}
```
odinuge: Isn't it possible to just do `isLoading && styles.hidden`

app/components/Content/Content.js
mrfarstad: Because the `<Youtube>` component will not load, and therefore the `onReady` function will never run. As a result, the loading indicator will not disapear 😨

app/routes/joblistings/JoblistingDetailedRoute.js
```diff
@@ -14,12 +14,14 @@ const mapStateToProps = (state, props) => {
   const joblisting = selectJoblistingById(state, { joblistingId });
   const { fetching } = state.joblistings;
   const actionGrant = (joblisting && joblisting.actionGrant) || [];
+  const youtubeUrl = joblisting && joblisting.youtubeUrl;
```
odinuge: Do you have to do this here?

PR #1431, opened 2019-02-08
Company detail events pagination

LudvigHz:
Add pagination and sorting to events on company detail route.

### DIFF
--- a/app/routes/company/CompanyDetailRoute.js
+++ b/app/routes/company/CompanyDetailRoute.js
@@ -15,17 +15,35 @@ import {
   selectEventsForCompany,
   selectJoblistingsForCompany
 } from 'app/reducers/companies';
+import { selectPagination } from '../../reducers/selectors';
+import createQueryString from 'app/utils/createQueryString';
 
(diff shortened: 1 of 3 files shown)


app/routes/company/CompanyDetailRoute.js
```diff
@@ -15,17 +15,25 @@ import {
...
+import { selectPagination } from '../../reducers/selectors';
 
-const fetchData = (props, dispatch) =>
-  Promise.all([
-    dispatch(fetchCompany(props.params.companyId)),
-    dispatch(fetchEventsForCompany(props.params.companyId)),
-    dispatch(fetchJoblistingsForCompany(props.params.companyId))
+const fetchData = (props, dispatch) => {
+  const { companyId } = props.params;
+  const queryString = `?company=${companyId}&ordering=-start-time`;
```
orhanhenrik: I would make this a function that accepts either props or companyId. You are defining the same queryString in three  different places in this file.  
I also think you need to use `start_time` with an underscore instead of a dash.

PR #1440, opened 2019-02-16
Add bundle analyzer

odinuge:
`BUNDLE_ANALYZER=true yarn build`

Fun with flags :100: 
maxresdefault @mrfarstad 

screenshot from 2019-02-16 21-03-57


SmithPeder:
Jesus. I said we should have used 2 flag icons...

odinuge:
Yeah. Makes no sense to bundle all the flags in the world into our webapp, when we use two of them :upside_down_face:

PR #1441, opened 2019-02-16
Random evnets editor fixes

odinuge:


app/components/Form/ImageUploadField.js
```diff
@@ -28,12 +29,13 @@ class ImageUploadField extends Component<Props> {
   };
 
   render() {
-    const { className, style, ...props } = this.props;
+    const { className, name, style, ...props } = this.props;
 
     return (
       <div
         className={cx(styles.base, styles.coverImage, className && className)}
```
SmithPeder: In the `EditorField` file the className can be passed right in, but here we check `classname && className`. The props is optional in both component, so it this unnessesery to have here?
```sh
# Form/EditorField.js
<Editor
className={cx(styles.input, className)}        
{...props}
{...props.input}
{...props.meta}
/>
```

odinuge: Wuut. Yeah

`className && className` is the same as just `className`

PR #1442, opened 2019-02-18
Webpack 4 & babel 7 :fire:

odinuge:
:fire:


SmithPeder:
`yarn test` finds no tests now
```js
~/code/lego-webapp on  webpack-4 yarn test                                        7:14:18
yarn run
$ NODE_ENV=test jest
No tests found
In /Users/smith/code/lego-webapp
  5 files checked.
  testMatch:  - 5 matches
  testPathIgnorePatterns: /node_modules/ - 5 matches
  testRegex: \.spec\.(js|jsx)$ - 0 matches
Pattern:  - 0 matches
✨  Done in 0.88
```

SmithPeder:
resolves #1405

package.json
```diff
@@ -92,65 +93,72 @@
...
+    "enzyme-adapter-react-16": "^1.0.2",
+    "eslint": "^5.14.1",
+    "eslint-config-prettier": "^4.1.0",
+    "eslint-config-react-app": "^3.0.7",
+    "eslint-import-resolver-webpack": "^0.11.0",
+    "eslint-plugin-flowtype": "^3.4.2",
     "eslint-plugin-import": "^2.8.0",
-    "eslint-plugin-jsx-a11y": "5.1.1",
-    "eslint-plugin-react": "^7.4.0",
+    "eslint-plugin-jsx-a11y": "^6.2.1",
```
SmithPeder: Bump to `6.2.0` to so we can get this aswell https://github.com/webkom/lego-webapp/pull/1415

odinuge:
This should be more or less good to go now! @SmithPeder 

There is quite a bit of duplication between the two webpack configs, so a lot of that should be fixed at a later stage. Haven't tested sourcemaps on sentry, but it should work as before. HMR is also kinda hard to test, but it looks good (and quick) so far.

(pls get this in quick, so we don't end up with a million merge conflicts :upside_down_face:)

odinuge:
Current chunk map
screenshot from 2019-03-05 02-32-18

package.json
```diff
@@ -6,14 +6,14 @@
...
-    "start": "webpack --config config/webpack.server.js",
+    "start": "webpack --config config/webpack.server.js --mode=development",
     "start:staging": "API_URL=https://lego-staging.abakus.no/api/v1 BASE_URL=https://lego-staging.abakus.no WS_URL=wss://ws-staging.abakus.no CAPTCHA_KEY=6LfS1TQUAAAAAIcxgj4EfkM7lwWeG9efLvS37tCW yarn start",
     "ssr:staging": "API_URL=https://lego-staging.abakus.no/api/v1 BASE_URL=https://lego-staging.abakus.no WS_URL=wss://ws-staging.abakus.no CAPTCHA_KEY=6LfS1TQUAAAAAIcxgj4EfkM7lwWeG9efLvS37tCW yarn ssr",
     "ssr": "echo \"Remember that SSR does not have hot reloading, build with \"yarn build\"\" && node ./dist/server.js",
     "build": "concurrently \"yarn run build:server\" \"yarn run build:client\"",
     "build:all": "concurrently \"yarn build:server\" \"yarn build:client\" \"yarn styleguide:build\"",
-    "build:client": "NODE_ENV=production webpack --config config/webpack.client.js",
-    "build:server": "NODE_ENV=production webpack --config config/webpack.server.js -p",
+    "build:client": "webpack --config config/webpack.client.js --mode=production",
```
SmithPeder: You've tested this? My understanding was that `--mode=production` was the new way of setting `-p`. Only the `build: server` used to have this flag. Maybe this was intentional(for some reason), as the rational thing would be to build for production.

package.json
```diff
@@ -92,65 +93,72 @@
...
+    "@babel/preset-stage-2": "^7.0.0",
     "assets-webpack-plugin": "^3.5.1",
-    "babel-core": "^6.23.1",
-    "babel-eslint": "^8.0.1",
-    "babel-jest": "^21.0.2",
-    "babel-loader": "^7.1.1",
+    "babel-eslint": "^10.0.1",
+    "babel-jest": "^24.1.0",
+    "babel-loader": "^8.0.5",
     "babel-plugin-lodash": "^3.2.11",
```
SmithPeder: 2 years old. `3.3.4` exists, maybe we should bump this with the rest of babel upgrades?

package.json
```diff
@@ -92,65 +93,72 @@
...
     "babel-plugin-transform-import-to-require": "^1.0.0",
     "babel-polyfill": "^6.23.0",
```
SmithPeder: 2 years old. `6.26.0` and `7.0.0-beta3` exist 🔥 🔥 #GonneGoSoWell

package.json
```diff
@@ -92,65 +93,72 @@
...
     "babel-preset-latest": "^6.22.0",
-    "babel-preset-react": "^6.23.0",
+    "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "^6.22.0",
```
SmithPeder: With the new `.babelrc` I think these three can be tossed?
```
    "babel-preset-latest": "^6.22.0",
    "babel-preset-react": "^6.24.1",
    "babel-preset-stage-2": "^6.22.0"
```

All the new presets are `@babel/` presets for flow, env and react, so if these 3 presets aren't used anywhere else I think they serve no purpose.

I also think this plugin is unused, as it's no longer used in the `.babelrc`
```
"babel-plugin-transform-import-to-require": "^1.0.0",
```

package.json
odinuge: Ahh, yee, it is more or less the same as `-p`, but with a few more optimizations. The new mode has a lot more optimizations that we previously enabled by using `NODE_ENV=production` (or `development`). More info here: <https://webpack.js.org/concepts/mode/>

I have no idea why the client didn't use `-p` before tho.

package.json
odinuge: :fire: :ok_hand:

package.json
odinuge: Yee, I guess so :ok_hand:

package.json
SmithPeder: 👌

app/routes/users/components/Penalties.js
```diff
@@ -43,9 +43,9 @@ function Penalties({
                     {penalty.weight} {word}
                   </strong>
                 </div>
-                <div>
+                <>
```
orhanhenrik: What is this?

odinuge: React fragments :heart_eyes: :heart: :heart_eyes_cat:

odinuge: https://reactjs.org/docs/fragments.html

Only available now with babel7

SmithPeder: Agreed! Did you just pop one in to test? 👍

odinuge: Hahah, yee. Nice to have when you don't want to return a div, just a set of components.

odinuge:
:fire:

PR #1443, opened 2019-02-18
Route for validating that a user exists

JonasBak:
A (very) simple version of Abacard that only checks if a user exists. Can be extended later to support more filters such as groups. Useful when you have events that doesn't have pools and registrations.


app/components/UserValidator/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+import * as React from 'react';
+import _ from 'lodash';
+import cx from 'classnames';
+import SearchPage from 'app/components/Search/SearchPage';
+import type { SearchResult } from 'app/reducers/search';
+import styles from './Validator.css';
+
+// $FlowFixMe
+import goodSound from '../../assets/good-sound.mp3';
```
SmithPeder: `app/assets/good-sound.mp3` 💯

app/components/UserValidator/index.js
```diff
@@ -0,0 +1,94 @@
...
+type State = {
+  showCompleted: boolean
+};
+type Props = {
```
SmithPeder: `location, onQueryChanged, results and searching` are not used in the `<Validator />` component. (i might be wrong here)
If this is the case, why isn't ESLint catching this?

app/routes/userValidator/ValidatorRoute.js
```diff
@@ -0,0 +1,59 @@
...
+      if (query) {
+        dispatch(autocomplete(query, searchTypes));
+      }
+    }, 300)
+  };
+};
+
+const WrappedValidator = props => (
+  <Content>
+    <Validator {...props} />{' '}
```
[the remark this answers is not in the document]
JonasBak: Nope, that's some good ALE-fixing right there :ok_hand:

app/components/UserValidator/index.js
JonasBak: This and the `cx` are here because I'm lazy and just moved code from Abacard out to its own component without checking too much. I don't know why the linter don't catch these.

SmithPeder: Yehh i saw you made a `quick ctrl-c-v` on that cx(), but it was ugly and should be removed :)

app/components/UserValidator/index.js
JonasBak: Self documenting :cool:

app/components/UserValidator/index.js
JonasBak: These props are given to the `SearchPage` component with `...this.props`.

SmithPeder: Yes the pris are passed on but I don't think the types are. So they should be typed in the next component?

PR #1447, opened 2019-02-20
Fix global field placeholder style

mrfarstad:
The redux-form placeholder text has different color than react-select form placeholders


app/components/Form/TextInput.css
```diff
@@ -16,6 +16,10 @@
   border-color: #aaa;
 }
 
+.input::placeholder {
+  color: #aaa;
```
SmithPeder: 10/10

PR #1448, opened 2019-02-21
Fix frontpage width issue on smaller screens

orhanhenrik:
Use word-break: break-word on event titles so that the browser breaks words that cannot fit on one line.  
Add a temporary truncate fix on compactEvent titles for small screens.

### DIFF
--- a/app/routes/overview/components/Overview.js
+++ b/app/routes/overview/components/Overview.js
@@ -91,6 +91,7 @@ class Overview extends Component<Props, State> {
       <CompactEvents
         events={frontpage.filter(isEvent)}
         className={styles.compactEvents}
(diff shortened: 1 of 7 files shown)


app/routes/overview/components/Overview.js
```diff
@@ -96,10 +96,17 @@ class Overview extends Component<Props, State> {
     const { frontpage, loadingFrontpage, readmes } = this.props;
     const pinned = frontpage[0];
     const { isMobile } = this.state;
+
+    // TODO: Figure out a more responsive way to truncate titles
+    let titleMaxLength = 25;
+    if (window.innerWidth < 350) {
```
odinuge: This will break the server side renderer :cry:

PR #1452, opened 2019-02-25
Distinguish between upcoming events based on whether you are in the w…

tmsolber:
…aiting list or not. Need to merge backend first #1467
Fix on issue: #1399

### DIFF
--- a/app/models.js
+++ b/app/models.js
@@ -63,7 +63,8 @@ export type Event = EventBase & {
   company: Company,
   comments: Array<Comment>,
   pools: Array<EventPool>,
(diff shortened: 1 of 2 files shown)


app/models.js
```diff
@@ -63,7 +63,8 @@ export type Event = EventBase & {
   company: Company,
   comments: Array<Comment>,
   pools: Array<EventPool>,
-  survey: ?ID
+  survey: ?ID,
+  userReg: Object
```
odinuge: You can use `userReg: EventRegistration` here instead. (see line `100` in this file).

PR #1453, opened 2019-02-25
Fix url issues connected to Om Abakus

cherrybeans:
Fix broken urls around abakus.no and fix routing issues when saving or creating article.

https://github.com/webkom/lego/issues/1461

### DIFF
--- a/app/reducers/search.js
+++ b/app/reducers/search.js
@@ -66,7 +66,7 @@ const searchMapping = {
     title: 'title',
     type: 'Side',
     color: '#E8953A',
-    path: '/pages/info/',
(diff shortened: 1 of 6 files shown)


app/reducers/search.js
```diff
@@ -66,7 +66,7 @@ const searchMapping = {
     title: 'title',
     type: 'Side',
     color: '#E8953A',
-    path: '/pages/info/',
+    path: '/pages/generelt/',
```
odinuge: You should update the search serializer for flatpages to include the `category`, and redirect to the right url.

odinuge: @cherrybeans

PR #1454, opened 2019-02-26
Add ability to delete comment & mobile improvements

LudvigHz:
Adds an option to delete a comment. Also improves comment styling on mobile.
Resolve https://github.com/webkom/lego/issues/1465
Resolve https://github.com/webkom/lego/issues/1398

### DIFF
--- a/app/components/Comments/CommentView.js
+++ b/app/components/Comments/CommentView.js
@@ -7,9 +7,11 @@ import CommentForm from 'app/components/CommentForm';
 import { Flex } from 'app/components/Layout';
 import type { UserEntity } from 'app/reducers/users';
 import CommentTree from './CommentTree';
+import type { CommentEntity } from 'app/reducers/comments';
+import type { ID } from 'app/models';
 
 type Props = {
-  comments: Array<Object>,
+  comments: Array<CommentEntity>,
   formDisabled?: boolean,
   commentTarget: string,
   user: UserEntity,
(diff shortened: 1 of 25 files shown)


app/components/Comments/CommentView.js
```diff
@@ -32,10 +32,29 @@ const CommentView = (props: Props) => {
...
+    (obj, comment) => ({ ...obj, [comment.id]: true }),
+    { null: true }
+  );
+  const undefinedParents = new Set(
+    comments
+      .filter(comment => !commentIds[comment.parent])
+      .map(comment => comment.parent)
+  );
+
+  const emptyComments = [...undefinedParents].map(id => ({
```
orhanhenrik: Maybe this should be implemented in the backend instead.. If the deleted parent was a reply, you lose that tree structure, since all emptyComments have no parent, even if the original comment did.
Thoughts?

LudvigHz: Yeah, the only issue is when you have a tree with three branches and the middle one is deleted. You could probably implement it in backend, but since the object is just flagged as deleted, you would have to update the keys of the related comments. Also, there is the issue with the user field being empty/null. So I'm not sure how to handle this, but I can take a look!

orhanhenrik:
Looks good to me, but I haven't checked styling on mobile.
Needs to be merged after webkom/lego#1475

PR #1456, opened 2019-02-26
Replace react-world-flags

mrfarstad:
Replace the package `react-world-flags` because it was a huge mistake :skra:


app/routes/companyInterest/components/CompanyInterest.css
```diff
@@ -169,3 +169,9 @@
 .companyInterestListMobile td {
   border: none;
 }
+
+.flag {
+  height: 3em;
+  width: 5em;
```
SmithPeder: Are you going to start using `em` now too???

mrfarstad: yes 😢

app/routes/companyInterest/components/CompanyInterestPage.js
```diff
@@ -137,9 +138,13 @@ const OtherBox = ({ fields, language }: FieldProps) => (
   </Flex>
 );
 
-const LanguageFlag = ({ isEnglish }: { isEnglish: boolean }) => {
-  return <Flag code={isEnglish ? 'nor' : 'gbr'} style={{ height: '30px' }} />;
-};
+const LanguageFlag = ({ isEnglish }: { isEnglish: boolean }) => (
```
odinuge: What about:
```
const LanguageFlag = ({ flag }: { flag: String }) => 
// or
const LanguageFlag = ({ language }: { language: String }) => 
```

Taking a boolean in a case like this is kinda strange

PR #1461, opened 2019-03-04
Add visible to column props in <Table />

odinuge:


app/routes/events/components/EventAdministrate/AttendeeElements.js
```diff
@@ -115,23 +114,21 @@ export const Unregister = ({
   fetching,
   handleUnregister,
   id,
-  clickedUnregister,
-  showUnregister
+  clickedUnregister
 }: UnregisterProps) => {
   return (
     <div>
-      {showUnregister &&
```
SmithPeder: Why is this check no longer needed?

odinuge: The column will now disappear instead of being empty.

PR #1464, opened 2019-03-04
Issi about fixes

iselinee:
Fix the layout of the comittee members, and move the Change and New button to the right side of the page.


cherrybeans:
Get this to pass CI so we can merge it in :smile:

PR #1469, opened 2019-03-05
Babel why???

odinuge:
First commit fixes https://sentry.abakus.no/webkom/lego-webapp/issues/6413/?query=is:unresolved (and sourcemaps still works, yay!), but i cannot understand why... Worked well in babel 6.

### DIFF
--- a/app/components/AnnouncementInLine/index.js
+++ b/app/components/AnnouncementInLine/index.js
@@ -30,7 +30,6 @@ type State = {
 };
 
 class AnnouncementInLine extends Component<Props, State> {
-  props: Props;
   state = {
     hidden: true,
     sent: false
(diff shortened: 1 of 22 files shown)


odinuge:
Yee. We already write ` Component<Props, State>`.

SmithPeder:
I bet you did a lot of searching for that buggy code but don't you think this has something with this component being a  `PureCompoent`?
> React.PureComponent is similar to React.Component. The difference between them is that React.Component doesn’t implement shouldComponentUpdate(), but React.PureComponent implements it with a shallow prop and state comparison.

Maybe it has something to do with this shallow prop and state comparsion #noclue

odinuge:
Hmmm. That may be the problem; i havent read the react docs that well. Will take a look at the code generated by babel to see why it crashed :thinking:

PR #1480, opened 2019-03-07
Change how following is checked

JonasBak:
Dependent on https://github.com/webkom/lego/pull/1485

### DIFF
--- a/app/actions/EventActions.js
+++ b/app/actions/EventActions.js
@@ -386,13 +386,13 @@ export function unfollow(followId: number, eventId: number): Thunk<Promise<*>> {
     );
 }
 
-export function isUserFollowing(eventId: number, userId: number) {
+export function isUserFollowing(eventId: number) {
   return callAPI({
     types: Event.IS_USER_FOLLOWING,
-    endpoint: `/followers-event/?target=${eventId}&follower=${userId}`,
+    endpoint: `/followers-event/?target=${eventId}`,
     method: 'GET',
     meta: {
-      errorMessage: 'Henting av hendelse følgere feilet'
+      errorMessage: 'Henting av interesse feilet'
     }
   });
 }
--- a/app/routes/events/EventDetailRoute.js
+++ b/app/routes/events/EventDetailRoute.js
@@ -125,13 +125,11 @@ const mapDispatchToProps = {
   updateUser
 };
 
-const loadData = ({ params: { eventId }, currentUser, loggedIn }, dispatch) => {
-  const userId = currentUser.id;
-  return dispatch(fetchEvent(eventId)).then(
-    () => loggedIn && dispatch(isUserFollowing(eventId, userId))
-  );
-};
-
+const loadData = ({ params: { eventId }, loggedIn }, dispatch) =>
+  Promise.all([
+    dispatch(fetchEvent(eventId)),
+    loggedIn && dispatch(isUserFollowing(eventId))
+  ]);
 const propertyGenerator = (props, config) => {
   if (!props.event) return;
   const tags = (props.event.tags || []).map(content => ({


app/actions/EventActions.js
```diff
@@ -386,13 +386,21 @@ export function unfollow(followId: number, eventId: number): Thunk<Promise<*>> {
...
-    types: Event.IS_USER_FOLLOWING,
-    endpoint: `/followers-event/?target=${eventId}&follower=${userId}`,
-    method: 'GET',
-    meta: {
-      errorMessage: 'Henting av hendelse følgere feilet'
-    }
-  });
+export function isUserFollowing(
+  eventId: number,
+  userId: ?number
```
odinuge: Why do we still need `userId` here? The action is only used once, without sending `userId`

JonasBak:
Fixes https://github.com/webkom/lego/issues/1481

app/actions/EventActions.js
JonasBak: Nice, I just assumed it was used elsewhere :upside_down_face:

app/actions/EventActions.js
```diff
@@ -386,13 +386,17 @@ export function unfollow(followId: number, eventId: number): Thunk<Promise<*>> {
...
-    meta: {
-      errorMessage: 'Henting av hendelse følgere feilet'
-    }
-  });
+export function isUserFollowing(eventId: number): Thunk<Promise<*>> {
+  return (dispatch, getState) =>
+    dispatch(
+      callAPI({
+        types: Event.IS_USER_FOLLOWING,
+        endpoint: `/followers-event/?target=${eventId}&follower=${getState()
```
odinuge: My intuition still says that this potentially _can_ crash while in frontend. :fire:

The `state.auth.id` is used when selecting the user from the state inside the `EventDetailRoute`:
https://github.com/webkom/lego-webapp/blob/33c35bb32086df956818e675e74eca92ccda4a55/app/routes/app/AppRoute.js#L133

This code (also the old one) works flawlessly in SSR because of the waterfall-like execution of the actions. This is where the user is fetched in SSR: https://github.com/webkom/lego-webapp/blob/33c35bb32086df956818e675e74eca92ccda4a55/app/routes/app/AppRoute.js#L144

and this is where it is fetched during client side rendering: https://github.com/webkom/lego-webapp/blob/33c35bb32086df956818e675e74eca92ccda4a55/app/index.js#L54

When running on the client however, it will however fetch the actions inside `EventDetailRoute` right after this action (without depending on `User.FETCH`):

https://github.com/webkom/lego-webapp/blob/33c35bb32086df956818e675e74eca92ccda4a55/app/routes/app/AppRoute.js#L164


As seen here :arrow_down:, `User.FETCH.SUCESS` can potentially be dipatched after `Event.IS_USER_FOLLOWING.SUCCESS`; and `state.auth.id` isn't reduced into the state before `User.FETCH.SUCCESS`.
screenshot from 2019-03-07 19-57-50

The reason the fetch doesn't crash here, is because `/followers-event/?target=${eventId}&follower=` is a valid request.

**BUT**: The token is a jwt, so a hacky way is to run `jwtDecode(getState().auth.token).user_id` :fire: :fireworks: :fire_engine: :man_firefighter: :woman_firefighter: (or something like that). The best would however be a separate endpoint, or just `?follower=me`.

orhanhenrik:
I'm not really sure this fixes the underlying issue.  
The implementation here depends on the user being fetched faster than the event is fetched. If the event is fetched first, `isUserFollowing` should still fail. The reason that the user usually finishes first is because it gets a head start because of rendering/setup times.
If the user request is slow, you will load `?target=123&follower=`, you can test it by delaying the `fetchUser` method.

```
export function fetchUser(username: string = 'me') {
  return (dispatch) => {
      const delay = new Promise(resolve => setTimeout(resolve, 3000));
      return delay.then(() =>
          dispatch(callAPI({............................}))
      )
  };
}
```

app/actions/EventActions.js
```diff
@@ -386,13 +386,17 @@ export function unfollow(followId: number, eventId: number): Thunk<Promise<*>> {
...
+          .auth.id || ''}`,
+        method: 'GET',
+        meta: {
+          errorMessage: 'Henting av hendelse følgere feilet'
```
SmithPeder: I don't really like the phrasing of this error message

SmithPeder: I know it was there before your edit, but this is a prime time to change it :)

JonasBak: Yeah, I noticed when testing after I committed, sounds like a quick copy-paste-adapt gone a bit wrong

app/actions/EventActions.js
```diff
@@ -386,13 +386,16 @@ export function unfollow(followId: number, eventId: number): Thunk<Promise<*>> {
...
-  return callAPI({
-    types: Event.IS_USER_FOLLOWING,
-    endpoint: `/followers-event/?target=${eventId}&follower=${userId}`,
-    method: 'GET',
-    meta: {
-      errorMessage: 'Henting av hendelse følgere feilet'
-    }
-  });
+export function isUserFollowing(eventId: number): Thunk<Promise<*>> {
+  return (dispatch, getState) =>
```
odinuge: Guess this doesn't need to be a Thunk any more?

app/routes/events/EventDetailRoute.js
```diff
@@ -101,7 +101,7 @@ const mapStateToProps = (state, props) => {
   return {
     comments,
     actionGrant,
-    notLoading: !state.events.fetching,
+    notLoading: !state.events.fetching && !state.users.fetching,
```
odinuge: What is the point of this?

app/routes/events/EventDetailRoute.js
```diff
@@ -125,10 +125,9 @@ const mapDispatchToProps = {
   updateUser
 };
 
-const loadData = ({ params: { eventId }, currentUser, loggedIn }, dispatch) => {
-  const userId = currentUser.id;
+const loadData = ({ params: { eventId }, loggedIn }, dispatch) => {
   return dispatch(fetchEvent(eventId)).then(
```
odinuge: Make these two concurrent while fixing stuff here (and remember to handle errors). No need to wait for fetchEvent

app/actions/EventActions.js
odinuge: Way better to write 3 comments instead of fixing it?? :thinking:

app/actions/EventActions.js
JonasBak: Nope, just didn't change everything after deciding to try to fix this in backend instead. Didn't have that much free time during hælga :upside_down_face:

app/routes/events/EventDetailRoute.js
JonasBak: Just thought I wanted to wait with drawing everything until after your user is fetched, didn't really do anything though, hehe (TODO remove)

PR #1485, opened 2019-03-08
Performance tweaks

odinuge:
screenshot from 2019-03-08 16-42-31


config/webpack.client.js
```diff
@@ -101,7 +103,13 @@ module.exports = (env, argv) => {
       })
     ]),
     resolve: {
-      modules: [root, 'node_modules']
+      modules: [root, 'node_modules'],
+      alias: {
+        lodash: 'node_modules/lodash-es',
```
SmithPeder: Does this ensure that every time someone imports something from `lodash` the request gets routed to `loadash-es`? That's very clever 💯

package.json
```diff
@@ -50,10 +50,10 @@
     "js-cookie": "^2.1.4",
     "jsdom": "^11.6.2",
     "jwt-decode": "2.2.0",
-    "lodash": "^4.17.11",
-    "medium-draft": "beta",
+    "lodash-es": "^4.17.11",
+    "medium-draft": "^0.6.0-beta1",
```
SmithPeder: Is this new beta1 `editor` flawless 👍 hehe

config/webpack.client.js
odinuge: Yes! And then only the stuff we use is added to the bundle!

package.json
odinuge: Hahahah, yeee. I checked for an update, but we were running the latest version :/ (as you can see in yarn.lock)

PR #1490, opened 2019-03-11
Remove duplicates of badges on profile

ingraso:
fixes https://github.com/webkom/lego/issues/1360

### DIFF
--- a/app/routes/users/components/UserProfile.js
+++ b/app/routes/users/components/UserProfile.js
@@ -21,7 +21,7 @@ import cx from 'classnames';
 import EventItem from 'app/components/EventItem';
 import EmptyState from 'app/components/EmptyState';
 import moment from 'moment-timezone';
-import type { Dateish } from 'app/';
+import type { Dateish } from 'app/models';
 import { Image } from 'app/components/Image';
 import frame from 'app/assets/frame.png';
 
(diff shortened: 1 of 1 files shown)


app/routes/users/components/UserProfile.js
```diff
@@ -92,38 +93,47 @@ const BadgeTooltip = ({
...
-  start,
-  end
-}: {
-  group: Group,
-  start: Dateish,
-  end?: ?Dateish
-}) => {
+const GroupBadge = ({ memberships }: { memberships: Array<Object> }) => {
+  const activeMemberships = memberships.find(membership => membership.isActive);
+  const [{ abakusGroup }] = memberships;
```
orhanhenrik:
```
const abakusGroup = memberships[0].abakusGroup;
```
is more readable to me

app/routes/users/components/UserProfile.js
```diff
@@ -92,38 +93,47 @@ const BadgeTooltip = ({
...
+  // $FlowFixMe
+  const sorted = orderBy(memberships, membership =>
+    moment(membership.startDate || membership.createdAt)
+  );
+  const [firstMembership] = sorted;
+  const [lastMembership] = sorted.reverse();
```
orhanhenrik: You don't need to reverse the whole list to get the last element:
```
const firstMembership = sorted[0];
const lastMembership = sorted[sorted.length - 1];
```
I would also rename `sorted` to something like `sortedMemberships`.

app/routes/users/components/UserProfile.js
orhanhenrik: You could do
```
if (!abakusGroup.showBadge) return null;
```
here to reduce the needed logic further down

PR #1494, opened 2019-03-12
Polls hotfixes

LudvigHz:
Add total votes to the component and move percentage text after bar if it cant fit inside.
Screenshot from 2019-03-12 01-12-10

Fixes https://github.com/webkom/lego/issues/1493

### DIFF
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -78,8 +78,15 @@ class Poll extends React.Component<Props, State> {
             <Icon name="stats" />
             <span className={styles.pollHeader}>{title}</span>
           </Link>
-          <Tooltip content="Avstemningen er anonym.">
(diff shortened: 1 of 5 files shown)


app/components/Poll/index.js
```diff
@@ -138,16 +147,18 @@ class Poll extends React.Component<Props, State> {
...
               <Icon
                 onClick={this.toggleTruncate}
                 className={styles.arrow}
                 size={20}
                 name={expanded ? 'arrow-up' : 'arrow-down'}
               />
-            </div>
-          )}
+            )}
+            <span style={{ width: '40%' }} />
```
SmithPeder: Nah m8, empty span with a width?? By default, span element is an inline element, and it ignores width. You can display it as `inline-block` if you really want this effect.

PR #1496, opened 2019-03-12
Add better sourcempas in dev

odinuge:
Not that slow anymore. Waaaay better sourcempas when debugging.

We can test this and see how it works. :smile:


odinuge:
Kjör

PR #1501, opened 2019-03-14
Switch back to slug

odinuge:


config/webpack.client.js
```diff
@@ -62,6 +62,7 @@ module.exports = (env, argv) => {
       new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
       isProduction && new DuplicatePackageCheckerPlugin(),
       new webpack.IgnorePlugin(/^jsdom$/),
+      new webpack.IgnorePlugin(/unicode\/category\/So/),
```
odinuge: This 2mb file with unicode chars can safely be ignored: https://yarnpkg.com/en/package/slug

orhanhenrik:
The module is published with ES6 Code which is not supported in all browsers

PR #1502, opened 2019-03-15
Change wait-list icon color and add a separate icon for admin registrations in the attendance list

kotlarz:
- Add the admin registration count to the attendance list header, as seen below.

- Add orange color to the registrations in the waiting list, and also add a seperate icon for admin registrations (with message if you hover over the icon), as seen below.

### DIFF
--- a/app/models.js
+++ b/app/models.js
@@ -102,6 +102,7 @@ export type EventRegistrationChargeStatus = 'manual' | 'succeeded' | 'failed';
 export type EventRegistration = {
   id: number,
   user: User,
+  adminRegistrationReason: string,
   registrationDate: Dateish,
   unregistrationDate: Dateish,
   pool: number,
--- a/app/routes/events/components/EventAdministrate/RegistrationTables.js
+++ b/app/routes/events/components/EventAdministrate/RegistrationTables.js
@@ -67,7 +67,7 @@ export class RegisteredTable extends Component<Props> {
             iconClass={
               pool
                 ? cx('fa fa-check-circle', styles.greenIcon)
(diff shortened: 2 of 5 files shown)


app/models.js
```diff
@@ -102,6 +102,8 @@ export type EventRegistrationChargeStatus = 'manual' | 'succeeded' | 'failed';
 export type EventRegistration = {
   id: number,
   user: User,
+  adminRegistrationReason: string,
+  company: Company,
```
odinuge: company? Did I miss out on something?

kotlarz: I have honestly no idea... good eye 💯

app/routes/events/components/EventAdministrate/RegistrationTables.js
```diff
@@ -61,13 +61,21 @@ export class RegisteredTable extends Component<Props> {
...
-        render: pool => (
+        render: (pool, registration) => (
           <TooltipIcon
-            content={pool ? 'Påmeldt' : 'Venteliste'}
+            content={
+              registration.adminRegistrationReason !== ''
+                ? 'Adminpåmeldt: ' + registration.adminRegistrationReason
+                : pool
+                ? 'Påmeldt'
+                : 'Venteliste'
```
orhanhenrik: Nested ternary is not very readable 😞

kotlarz: True, but couldn't be bothered as the linter wasn't complaining. #todo? 

We should add `no-nested-ternary` to our eslint config if we don't want that in the future :shrug:

PR #1508, opened 2019-03-18
change time format in events list

amalieeh:
fix https://github.com/webkom/lego/issues/1506

before:
image

after:
image


orhanhenrik:
#1505 tried to fix the same thing, I don't know which is better.

amalieeh:
Oh, lol, we should go with yours

PR #1516, opened 2019-03-21
Cypress end to end testing

orhanhenrik:
Depends on: https://github.com/webkom/lego/pull/1516  

Need to do some investigation into why chromium crashes sometimes on CI, it seems to be an issue of not having enough memory.


wait-for-it.sh
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env bash
+#   Use this script to test if a given TCP host/port are available
```
SmithPeder: Do you need this script in both the front-end and the back-end?

orhanhenrik: Yep, because the process that uses this doesn't have access to the backend code. It is needed so that we don't start testing before both frontend _and_  backend are up and running.  
It could also be downloaded every time, but I don't think that's a better solution than having it in the repo(s).

PR #1518, opened 2019-03-25
Add scroll-to-top/scroll restoration on route change.

cherrybeans:
* Fixes https://github.com/webkom/lego/issues/1520

Quick fix for v3, but when upgrading to v4 this won't work anymore. But the people at React Router has provided an alternative so no biggie. 

This is the package used (plug and play): https://github.com/taion/react-router-scroll

Further reading for v4 solutions and reasoning for removing it (tldr: browsers are starting to implement it):
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md
https://reacttraining.com/react-router/web/guides/scroll-restoration

Note: Removed global `scroll-behavior: smooth` css, since it makes the screen jump when navigating, and while this property is nice for admin pages like lego form and searching in those pages (groups, email etc), I think regular users should be prioritised. So I'm removing this, but in the future we might want to look into alternatives. `scrollIntoView` as used in legoForm has a property where you can specify smooth scrolling, but I did not change this as it does not seem to work on my pc at least. Hopefully we can find other solutions for smooth scrolling where it would be nice to have it. https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

### DIFF
--- a/app/Root.js
+++ b/app/Root.js
@@ -2,11 +2,12 @@
 
 import React from 'react';
 import { setStatusCode } from 'app/actions/RoutingActions';
-import { Router } from 'react-router';
+import { Router, applyRouterMiddleware } from 'react-router';
 import { Provider } from 'react-redux';
 import type { Store } from 'app/types';
 import { connect } from 'react-redux';
 import ErrorBoundary from 'app/components/ErrorBoundary';
+import { useScroll } from 'react-router-scroll';
 
 type Props = {
   store: Store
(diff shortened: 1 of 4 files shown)


cherrybeans:
Edit: "Do not apply the useScroll middleware when rendering on a server. You may use <ScrollContainer> in server-rendered components; it will do nothing when rendering on a server.". DOH!

PR #1520, opened 2019-03-25
Tool tip fix

erlingfn:
Modify tooltip compenent in order to make it look good regardless of where it is placed without the need for inline styling, as well as make it easier to customise how it is displayed.
Demonstration of being able to choose direction for the tooltip to render.
Before:
Screenshot from 2019-03-25 21-30-28
After:
Screenshot from 2019-03-25 21-28-18
Tooltips are now centered by default:
Screenshot from 2019-03-25 21-31-46
Able to specify the start of tooltips as well:
Screenshot from 2019-03-25 22-18-00

### DIFF
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -78,7 +78,7 @@ class Poll extends React.Component<Props, State> {
             <Icon name="stats" />
             <span className={styles.pollHeader}>{title}</span>
           </Link>
-          <Tooltip content="Avstemningen er anonym." renderToThe="left">
+          <Tooltip content="Avstemningen er anonym." renderDirection="left">
             <Icon
               name="information-circle-outline"
               size={20}
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -78,14 +78,11 @@ class Poll extends React.Component<Props, State> {
             <Icon name="stats" />
             <span className={styles.pollHeader}>{title}</span>
           </Link>
-          <Tooltip
-            content="Avstemningen er anonym."
-            style={{ marginLeft: '-180px' }}
-          >
+          <Tooltip content="Avstemningen er anonym." renderToThe={'left'}>
             <Icon
               name="information-circle-outline"
               size={20}
-              style={{ marginLeft: '10px', cursor: 'pointer' }}
+              style={{ cursor: 'pointer' }}
             />
           </Tooltip>
         </Flex>
--- a/app/components/Tooltip/index.js
+++ b/app/components/Tooltip/index.js
@@ -11,8 +11,8 @@ type Props = {
   onClick?: () => void,
   style?: Object,
   list?: boolean,
-  renderToThe?: string,
-  pointerToThe?: string
+  renderDirection?: string,
+  pointerPosition?: string
 };
 
 type State = {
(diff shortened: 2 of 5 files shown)


app/components/Tooltip/index.js
```diff
@@ -38,18 +64,66 @@ export default class Tooltip extends Component<Props, State> {
...
+        case 'left':
+          renderToTheClass = styles.renderToTheLeft;
+          break;
+        case 'right':
+          renderToTheClass = styles.renderToTheRight;
+          break;
+      }
+    }
+    let startPointChildren = 2;
+    if (!list) {
```
LudvigHz: Can be combined into one if `if(!list) {...}`

mrfarstad:
Can this change potentially break some of the existing usages of the `Tooltip` component? Do we need to remove custom styles somewhere?

orhanhenrik:
Did you check if event registrations look correct? When you hover over user icons the tooltip shows the name of the user.

app/components/Poll/index.js
```diff
@@ -78,14 +78,11 @@ class Poll extends React.Component<Props, State> {
             <Icon name="stats" />
             <span className={styles.pollHeader}>{title}</span>
           </Link>
-          <Tooltip
-            content="Avstemningen er anonym."
-            style={{ marginLeft: '-180px' }}
-          >
+          <Tooltip content="Avstemningen er anonym." renderToThe={'left'}>
```
orhanhenrik: You can write `renderToThe="left"` instead

app/components/Tooltip/index.js
```diff
@@ -10,20 +10,46 @@ type Props = {
...
+  childrenContainerWidth: number
 };
 
+/**
+ * A tooltip that appears when you hover over the component placed within.
+ * The tooltip will by default be centered, however it supports a 'renderToThe'
+ * prop that will make it render either to the left. The pointer
+ * (the small arrow that points towards the component within the tooltip) will
+ * also default to center, and it can be adjusted with the 'pointerToThe' prop.
+ * Both props can be set as either 'left' or 'right'.
```
orhanhenrik: How do the different combinations look? If you renderToThe left and pointerToThe right, what happens?

erlingfn: Screenshot from 2019-03-28 15-42-52
Here is an example of that. The renderToThe prop selects which way the component will render compared to the pointer. The pointerToThe prop selects where the pointer will be on the component within the tooltip. So in this example the pointer is to the right on label within the component, and from there the tooltip renders to the left. If you were to set pointerToThe left the tooltip would render to the left starting at the left end of the label.

app/components/Poll/index.js
erlingfn: Thanks, I didn't know that!

erlingfn:
> Can this change potentially break some of the existing usages of the `Tooltip` component? Do we need to remove custom styles somewhere?
While it won't break anything, as none of the new props are required, the fact that the tooltip is now centered by default can cause it to look a bit weird some places. However, as far as I have checked, there are no places where it looks worse than before. It might also need some custom styling some places depending on the margin where is it used.
I haven't found any places where custom styling conflicts with the new changes, but there might be some more places like polls where the new props solve problems better than custom styling.

erlingfn:
> Did you check if event registrations look correct? When you hover over user icons the tooltip shows the name of the user.

Added some custom styling to make it hover a tiny bit above the user icon

app/components/Tooltip/index.js
orhanhenrik: So "renderToThe" is relative to the pointer, and the pointer is relative to the underlying element? Maybe add a small description about that.

I would also rename the props, maybe something like `pointerPosition` and `tooltipPosition`, or whatever makes sense.

erlingfn: Explained that the tooltip is adjusted relative to the pointer, and renamed the props to renderDirection and pointerPosition. Do these names work?

orhanhenrik: Perfect ✅

PR #1522, opened 2019-03-26
[WIP] Add kubernetes limits and requests to drone tasks

orhanhenrik:
Kubernetes limits and requests should be used to ensure test pods aren't spawned on near capacity nodes. It will also make sure pods behave well and don't take too many resources.

With v1.0.0 of drone it is currently not feasible, since all pods have affinity to the same node, and the initial node is picked at random (where ever the drone-job pod is scheduled). If drone-job is scheduled on a node with almost no leftover resources, most pods will just stay in "pending" state until the node has more resources available. In this case the drone build will just hang until killed manually.

https://discourse.drone.io/t/drone-services-mount-source-code/4108/5?u=orhanhenrik

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,6 +11,13 @@ steps:
     CYPRESS_INSTALL_BINARY: 0
   commands:
     - yarn --frozen-lockfile
+  resources:
+    limits:
+      cpu: 1
+      memory: 400Mi
+    requests:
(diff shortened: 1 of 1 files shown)


LudvigHz:
@orhanhenrik I'm assuming this isn't relevant anymore, if so, could you close this?

PR #1527, opened 2019-03-27
[WIP] Payment redesign, stripe v3 & wip-google pay

odinuge:
We can get some inspiration here: https://stripe.dev/elements-examples/

Should be more info about payments after the user has paid too.
Screenshot from 2019-03-27 23-31-51

### DIFF
--- a/app/reducers/registrations.js
+++ b/app/reducers/registrations.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { union } from 'lodash';
+import { union, get } from 'lodash';
 import { Event } from '../actions/ActionTypes';
 import createEntityReducer from 'app/utils/createEntityReducer';
 import { normalize } from 'normalizr';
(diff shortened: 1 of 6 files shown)


odinuge:
This will make payment a lot simpler (still in work) for users to understand. We should look at https://developers.google.com/pay/api/web/guides/brand-guidelines (required if we will use google pay, but they are generally nice). We (HS/Arrkom, @JonasBak) should also write some information about how refunds works (should work). We dont need to implement refunds via lego-webapp, but the user should understand the terms of their payment. Like what happen if you unregister after you have paid.

We should also a add a `You have paid 123,-`-info, together with the date/time and payment method.

orhanhenrik:
What happened to the "avregistrer" button styling? Was it like that before?

I agree that there should be more information around terms and feedback on what happened. Since this is regarding payments/money, users will probably be extra concerned as well.

odinuge:
Haha, nop. Double WIP-PR with some quick global styling for testing :tm:

LudvigHz:
I'm thinking we should wait with this until https://github.com/webkom/lego/issues/1631 is done, and we have upgraded the Stripe API. I'm going to have to switch to stripe elements in the PR i'm working on now to get the new API working, so not sure what you want to do with this?

odinuge:
Most of this has been cherry-picked or re-implemented in #1590, so ill just close this.

PR #1532, opened 2019-03-30
Fix all the image things 🤷 ⛑

odinuge:
Make gallery upload 3 files concurrently (not pooled), and add a sweet `3min` timeout on the file upload. Uploading large jpegs (10M+) takes quite some time on slow connections.

### DIFF
--- a/app/actions/GalleryPictureActions.js
+++ b/app/actions/GalleryPictureActions.js
@@ -3,7 +3,7 @@
 import { GalleryPicture, Gallery } from './ActionTypes';
 import { galleryPictureSchema } from 'app/reducers';
 import { uploadFile } from './FileActions';
-import { chunk } from 'lodash';
+import PromisePool from 'es6-promise-pool';
 import { type GalleryPictureEntity } from 'app/reducers/galleryPictures';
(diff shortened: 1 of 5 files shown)


app/actions/GalleryPictureActions.js
```diff
@@ -113,38 +113,43 @@ export function CreateGalleryPicture(galleryPicture: { galleryId: number }) {
...
+        }
+      });
+    });
+
+  const promiseProducer = function*() {
+    for (const file of files) {
+      yield uploadPictureWithErrorhandler(file);
+    }
+  };
+  const _data = promiseProducer();
```
odinuge: Promise pool should support generator functions, but the reflection done inside the lib doesn't work properly after terser.

orhanhenrik: What happens and why doesn't it work? There is an example in their github using a generator function.
https://github.com/timdp/es6-promise-pool#iterator

odinuge: Yee. It works well without Terser. I think Terser does some shady stuff that changes the type of something. The code crashes (when Terser is enabled) when trying to take `.then` on an Object inside the lib.

PR #1534, opened 2019-04-01
[WIP] Upgrade redux-form to 7.4.2

mrfarstad:
"As of v7, redux-form officially supports Flow for static type checking" 🎉 

v8 is the newest version, but this requires react-redux v6, which breaks many packages we use :rip:


odinuge:
What about v8 @mrfarstad? :disappointed:

mrfarstad:
> What about v8 @mrfarstad? 😞

Will try to do this incrementally, but currently v8 breaks too many of our packages

PR #1535, opened 2019-04-01
Correctly handle cases where activation time is null

kotlarz:
`event.activationTime` is `null` if a user is not a member of the groups that are in a pool for an event.

This caused issues with displaying the attendance count correctly.

This fixes webkom/lego#1378 and fixes webkom/lego#1381.

### Event where `event_status_type` is `NORMAL` _Vanlig påmelding_ or `INFINITE` _Åpen (med påmelding)_:
**Before:** 
Selection_128

**After:**
Selection_130

**Before:** 
Selection_132

**After:**
Selection_131

### Event where `event_status_type` is `OPEN` _Åpen (uten påmelding)_:
**Before:**
Selection_133

**After:** 
Selection_134

### DIFF
--- a/app/utils/eventStatus.js
+++ b/app/utils/eventStatus.js
@@ -5,8 +5,11 @@ import moment from 'moment-timezone';
 
 // Calculate diplay message for an event based on
 // eventStatusType, activationTime, capacity and totalCapacity
-
-const eventStatus = (event: Event, pill: boolean = false) => {
+const eventStatus = (
+  event: Event,
+  loggedIn: boolean = false,
+  isPill: boolean = false
+) => {
   const {
     registrationCount,
     totalCapacity,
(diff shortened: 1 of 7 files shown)


app/utils/eventStatus.js
```diff
@@ -23,14 +22,36 @@ const eventStatus = (event: Event) => {
     case 'OPEN':
       return 'Åpent arrangement';
     case 'INFINITE':
+      if (activationTime === null) {
+        return 'Ingen påmeldingsrett';
+      }
       return 'Åpent med påmelding';
     case 'NORMAL':
+      if (activationTime === null) {
```
odinuge: What will happen if the user has registered, and `activationTime` is (then) null?

SmithPeder:
Did https://github.com/webkom/lego/pull/1550 solve 
> "What will happen if the user has registered, and activationTime is (then) null?"

 or do we need to make some front-end changes?

kotlarz:
> Did [webkom/lego#1550](https://github.com/webkom/lego/pull/1550) solve
> 
> > "What will happen if the user has registered, and activationTime is (then) null?"
> 
> or do we need to make some front-end changes?

We need to make some front-end changes, I've already started on that. I'll do a PR before the weekend.

kotlarz:
@SmithPeder @odinuge: New case (_yay!_)

On the event list page (`/events`), what should we do when a visitor (i.e. not logged in) vists the page? Should we show a message that they need to log in to register or display nothing?

i.e.
Selection_152

Or display nothing?:
Selection_153

SmithPeder:
I think the first one looks great. Then it clearly shows that you are not logged in.

app/utils/eventStatus.js
kotlarz: Added logic in 6ff1375.

kotlarz:
The admittance logic has been added, but also logic to check if the user is logged in or not (and display an appropriate message if they are not).

SmithPeder:
Nice work. `lint` & `flow` tho :shock:

kotlarz:
> Nice work. `lint` & `flow` tho :shock:

_Should_ be good now, god damn local flow lied to me when I ran it earlier.

PR #1536, opened 2019-04-01
Add support for polling results adding up to 100% for an arbitrary number of poll options.

HoboKristian:
In accordance with: https://stackoverflow.com/questions/13483430/how-to-make-rounded-percentages-add-up-to-100


app/components/Poll/index.js
```diff
@@ -98,15 +120,14 @@ class Poll extends React.Component<Props, State> {
           <Flex column className={styles.optionWrapper}>
             <table className={styles.pollTable}>
               <tbody>
-                {optionsToShow.map(option => {
-                  const ratio = (options.length > 2 ? Math.floor : Math.round)(
-                    (option.votes / totalVotes) * 100
-                  );
+                {this.perfectRatios(this.optionsWithRatios(options, optionsToShow)).map(({id, name, votes, ratio}) => {
```
HoboKristian: I would like this line to be:
this.optionsWithRatios(options, optionsToShow).perfectRatios().map(…

but haven't been able to figure out how this is done in Javascript (or if it is possible).

app/components/Poll/index.js
```diff
@@ -64,17 +64,39 @@ class Poll extends React.Component<Props, State> {
...
+    return options
+        .sort((a, b) => {
+          const aMantissa = a.ratio - Math.floor(a.ratio);
+          const bMantissa = b.ratio - Math.floor(b.ratio);
+          return aMantissa < bMantissa;
+        })
+        .map((option, index) => {
+          return {...option, "ratio": Math.floor(option.ratio) + (index < off ? 1 : 0)}
+        })
+        .sort((a, b) => b.ratio - a.ratio)
```
HoboKristian: The end result is sorted on ratio with the biggest ratio being the uppermost option in the poll. Maybe you would prefer to sort on something else?

app/components/Poll/index.js
```diff
@@ -64,17 +64,39 @@ class Poll extends React.Component<Props, State> {
       : this.setState({ optionsToShow: options, expanded: true });
   };
 
+  optionsWithRatios = (options, optionsToShow) => {
+    const totalVotes = options.reduce((a, option) => a + option.votes, 0);
```
HoboKristian: This could be an argument to the function as we do have this value precomputed. However, I believe this results in cleaner code. I would have loved to remove one of the arguments but both is needed to replicate the existing code.

Why would you only show some options but still have their ratios calculated based on the total number of votes? It would make the poll look even weirder.

app/components/Poll/index.js
```diff
@@ -64,17 +64,39 @@ class Poll extends React.Component<Props, State> {
...
+    return optionsToShow.map(option => {
+      return {...option, "ratio": option.votes / totalVotes * 100}
+    });
+  };
+
+  // As described in: https://stackoverflow.com/questions/13483430/how-to-make-rounded-percentages-add-up-to-100
+  perfectRatios = (options) => {
+    const off = 100 - options.reduce((a, option) => a + Math.floor(option.ratio), 0);
+    return options
+        .sort((a, b) => {
```
odinuge: Since good old `Array.prototype.sort()` is in place, this can mutate the redux-state. You can use `lodash` `sortBy` (https://lodash.com/docs/4.17.11#sortBy), or just use `slice()` like this to take a shallow copy of the array:

```js
return options.slice().sort(...).map(...).sort(...)
```

app/components/Poll/index.js
LudvigHz: Thats not possible as far as i know, but I think it would be better to just have one function which calculates the perfect ratio, since the float version only is used for calculating the ratios that are shown.
So just have `this.perfectRatios(options, optionsToShow).map(...)` and just add the ratio property in the one function.

app/components/Poll/index.js
LudvigHz: Seems logical to me

app/components/Poll/index.js
```diff
@@ -5,6 +5,8 @@ import Button from 'app/components/Button';
 import styles from './Poll.css';
 import type { PollEntity, OptionEntity } from 'app/reducers/polls';
 import { Link } from 'react-router';
+import { _, sortBy } from 'lodash';
```
LudvigHz: Remove `_` please, since its unused.

orhanhenrik:
Currently our Ci sadly doesn't run external PRs, so sorry for the inconvenience. If you run `yarn flow` you will get a couple of errors in your code, as you need to add type annotations on the `optionsWithPerfectRatios` and `perfectRatios` methods.

HoboKristian:
@orhanhenrik It should be fixed now. At least it works on my machine™.

orhanhenrik:
Good job 👍 

https://abakus.no/polls/3

Before: (sum 98%)
image

After: (sum 100%)
image

PR #1541, opened 2019-04-03
Add support for showing previous events which a user attended

HoboKristian:
This PR is dependent on https://github.com/webkom/lego/pull/1553 passing.


HoboKristian:
There is a lot of DRY in this PR..

app/routes/users/components/UserProfile.js
```diff
@@ -57,6 +57,7 @@ type Props = {
   feed: Object,
   isMe: boolean,
   loading: boolean,
+  previousEvents: Array<Event>,
   upcomingEvents: Array<Event>,
```
HoboKristian: Should these two use the types defined below? And also maybe merge the two types into 1?

app/reducers/events.js
```diff
@@ -218,7 +234,7 @@ const mutate = joinReducers(mutateComments('events'), mutateEvent);
 export default createEntityReducer({
   key: 'events',
   types: {
-    fetch: [Event.FETCH, Event.FETCH_UPCOMING]
+    fetch: [Event.FETCH, Event.FETCH_PREVIOUS, Event.FETCH_UPCOMING]
```
HoboKristian: Is this setting up a listener for the events defined? At least that is what I assumed. If so, I believe we can remove Event.FETCH as it is never caught by the mutateEvent()-switch?

app/reducers/events.js
```diff
@@ -248,6 +264,11 @@ export const selectEvents = createSelector(
   (eventsById, eventIds) => eventIds.map(id => transformEvent(eventsById[id]))
 );
 
+export const selectPreviousEvents = createSelector(
+    selectEvents,
+    events => events.filter(event => !event.isUsersUpcoming)
```
odinuge: Since all (other) events will have `isUserUpcoming: undefined`, so this could just use `events.filter(event => event.isUsersUpcoming===false)`

app/routes/users/components/UserProfile.js
```diff
@@ -363,6 +364,15 @@ export default class UserProfile extends Component<Props, UpcomingEventsProps> {
             ) : (
               <LoadingIndicator loading />
             )}
+            <h3>Dine tidligere arrangementer ({previousEvents.length})</h3>
+            {loading ? (
+              <LoadingIndicator margin={'20px auto'} loading />
+            ) : (
+              <ListEvents
+                events={previousEvents.filter(e => e.userReg.pool !== null)}
```
edvardvb: Continuing the discussion in https://github.com/webkom/lego/pull/1553

You could also filter on `userReg.presence` to see if the user actually showed up for the event. However, that depends on whether or not the event used Abacard (and penalties, indirectly). I'm not sure if we can guarantee that right now, so might be better to just show all events for now 😄

odinuge: Yee. If you were registered and attendance isn't set, you get the message `Oppmøte ble ikke sjekket`.
Screenshot from 2019-04-03 17-43-57

edvardvb: In the spirit of keeping things simple: You could simply exclude registrations where the user is marked `not attending`. That would at least remove events where we know for a fact that the user didn't show up, without having to guess if Abacard was used or not 🙃

app/routes/users/components/UserProfile.js
```diff
@@ -363,6 +362,18 @@ export default class UserProfile extends Component<Props, UpcomingEventsProps> {
...
               <LoadingIndicator loading />
             )}
+            <h3>Dine tidligere arrangementer ({previousEvents.length})</h3>
+            {loading ? (
+              <LoadingIndicator margin={'20px auto'} loading />
+            ) : (
+              <ListEvents
+                events={previousEvents
+                  .filter(e => e.userReg.pool !== null)
+                  .filter(e => e.userReg.presence !== 'NOT_PRESENT')
```
HoboKristian: @edvardvb is this what you were thinking of?

edvardvb: Yup!

HoboKristian:
@orhanhenrik can you test this one?

orhanhenrik:
Yep, everything is looking good as of commit f05b6e0 👍

HoboKristian:
Can you merge it then? I am unable to do so

PR #1547, opened 2019-04-06
Switch arg to token

SmithPeder:
Less confusion as we use the sentry auth token system and not the legacy key system.


SmithPeder:
yes, (#woking on it), so not yet. I just need the right token from sentry

PR #1550, opened 2019-04-09
Bump react-router and all redux dependencies

LudvigHz:
- Bump redux, redux-form, redux-thunk & react-redux
- Switch from react-router-redux to connected-react-router
- Bump react-router to `5.0.0`

Current issue is getting react-router to render anything with the current setup of AppRoute being a top-level component.


LudvigHz:
https://github.com/webkom/lego-webapp/pull/1550/commits/0a439f6f3a1cf2f66db6ff5f6bc0569e8094fea8 is a first draft that shows a possible configuration with react-router v5. All routes in the new config works as they should (from my limited testing).

### Things that work
- routing
- client-side rendering w/ async fetching
- redux and redux-form seem to work fine, no fixes required

### Things that don't work
- The `events/calendar/:year/:month` routes.  I could not figure out how to make react-router and react-frontload to play nice, And i've not looked at how react-router v5 handles multiple routes rendering the same component.
- Async loading. I very briefly tried using `resolveAsyncRoute()`. This could maybe be fixed though.
- Server-side-rendering: Have not really tested this much

It would be cool if anyone has any ideas on improvements or thoughts about how this should be done before I continue to do the whole app this way.
@odinuge @JonasBak @orhanhenrik @erlingfn @SmithPeder @cherrybeans @edvardvb

app/routes/events/EventCreateRoute.js
```diff
@@ -19,9 +19,10 @@ import loadingIndicator from 'app/utils/loadingIndicator';
 import moment from 'moment-timezone';
 
 const mapStateToProps = (state, props) => {
+  console.log(props);
```
odinuge: pls

app/routes/overview/components/Overview.js
```diff
@@ -99,6 +99,7 @@ class Overview extends Component<Props, State> {
       votePoll
     } = this.props;
     const pinned = frontpage[0];
+    console.log(this.props);
```
odinuge: pls

app/routes/app/AppRoute.js
```diff
@@ -135,25 +150,25 @@ function mapStateToProps(state) {
...
   fetchNotificationFeed,
   markAllNotifications,
   fetchNotificationData,
-  setStatusCode
+  setStatusCode,
+  loginAutomaticallyIfPossible
+};
+
+const fetchInitialOnServer = props => {
+  props.loginAutomaticallyIfPossible();
```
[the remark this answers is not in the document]
erlingfn: Would it work as an async function rather than a promise?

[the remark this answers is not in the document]
LudvigHz: Yes, it would do the same. An async function would return a Promise.

server/ssr.js
```diff
@@ -27,8 +27,8 @@ class TimeoutError {
 const isTimeoutError = (error: Error) => error instanceof TimeoutError;
 const isReactHooksError = (error: Object) =>
   typeof error === 'object' &&
-  error.name === 'Invariant Violation' &&
-  error.stack.includes('https://fb.me/react-invalid-hook-call');
+  error.name === 'Error' &&
```
odinuge: :heart_eyes:

app/routes/admin/groups/components/AddGroupMember.js
```diff
@@ -1,14 +1,16 @@
...
 import SelectInput from 'app/components/Form/SelectInput';
 import { createValidator, required } from 'app/utils/validation';
 import { ROLES } from 'app/utils/constants';
+import type { role } from 'app/models';
 
-type Props = FieldProps & {
-  group: Object
+type Props = FormProps & {
+  group: Object,
+  addMember: ({ role: role, userId: Number, groupId: Number }) => Promise<*>
```
odinuge:
```suggestion
  addMember: ({ role: $Keys<typeof ROLES>, userId: Number, groupId: Number }) => Promise<*>
```

.flowconfig
```diff
@@ -4,6 +4,7 @@
 .*/node_modules/react-motion*
 .*/node_modules/jss/*
 .*/node_modules/findup/*
+.*/node_modules/redux-form/lib/index.js.flow*
```
odinuge: FYI
This will "disable" the typings on everything related to redux-form, since all types from redux-form now will become "Object", and `type A = Object & {...stuff}` will match everything.. :/

app/routes/admin/groups/components/AddGroupMember.js
odinuge: The `role` type is the same as `$Values<typeof ROLES>` and not the key (the value we send to the api).

.flowconfig
odinuge: @erlingfn

`yarn add -D flow-bin@0.77.0 && base64 -d <<< "ZGlmZiAtLWdpdCBhL2FwcC9hY3Rpb25zL1VzZXJBY3Rpb25zLmpzIGIvYXBwL2FjdGlvbnMvVXNlckFjdGlvbnMuanMKaW5kZXggYWYwMDc3MDYuLjc3ZTUzNGY3IDEwMDY0NAotLS0gYS9hcHAvYWN0aW9ucy9Vc2VyQWN0aW9ucy5qcworKysgYi9hcHAvYWN0aW9ucy9Vc2VyQWN0aW9ucy5qcwpAQCAtMjU0LDcgKzI1NCw3IEBAIGV4cG9ydCBmdW5jdGlvbiBmZXRjaFVzZXIodXNlcm5hbWU6IHN0cmluZyA9ICdtZScpIHsKICAgfSk7CiB9CiAKLWV4cG9ydCBmdW5jdGlvbiByZWZyZXNoVG9rZW4odG9rZW46IEVuY29kZWRUb2tlbikgeworZXhwb3J0IGZ1bmN0aW9uIHJlZnJlc2hUb2tlbih0b2tlbjogRW5jb2RlZFRva2VuKTogVGh1bms8YW55PiB7CiAgIHJldHVybiBjYWxsQVBJKHsKICAgICB0eXBlczogVXNlci5SRUZSRVNIX1RPS0VOLAogICAgIGVuZHBvaW50OiAnLy9hdXRob3JpemF0aW9uL3Rva2VuLWF1dGgvcmVmcmVzaC8nLAo=" | git apply && yarn flow`

:angel:

erlingfn: Thanks a lot for the help. Might be a pleb question, but why is it necessary to apply the change to fetchuser as a command ant not simply suggest the change?

package.json
```diff
@@ -68,17 +69,17 @@
     "react-cropper": "^1.3.0",
     "react-dom": "npm:@hot-loader/react-dom@^16.8.6",
     "react-dropzone": "^3.10.0",
+    "react-frontload": "^1.0.7",
```
LudvigHz: This isn't needed anymore

erlingfn: *yeet*

.flowconfig
odinuge: > but why is it necessary to apply the change to fetchuser as a command and not simply suggest the change?

Not necessary at all, but fun (and quick) :rocket:

app/routes/events/components/EventEditor/index.js
```diff
@@ -56,7 +56,8 @@ type Props = {
   setCoverPhoto: (number, string) => void,
   deleteEvent: (eventId: ID) => Promise<*>,
   submitting: boolean,
-  pristine: boolean
+  pristine: boolean,
+  initialized: boolean
```
LudvigHz: Where does this prop come from? And what does it do?

erlingfn: I was going to make a comment about, I knew I forgot something. The reason for this prop is an [issue](https://github.com/redux-form/redux-form/issues/621) in redux form that causes all fields to be initially rendered with an empty string as value, due to the form not being initialized. Due to the editorstate being immutable, the editorfield does not update once it is passed the correct initialvalue on the second render. This props "solves" the issue by enabling the editorfield to only render once the form has been initialized. The initialized prop is passed to the form by the redux-form HoC, see [this](https://redux-form.com/8.3.0/docs/api/props.md/) page for details. The "solution" is a hack, yet i could find no better way to solve this.

app/routes/admin/groups/components/AddGroupMember.js
erlingfn: Thanks! Always nice to get feedback from people who know what they are doing.

app/routes/events/components/EventEditor/index.js
LudvigHz: Ah, I see, it's not our bad code :1st_place_medal:. You should add a comment in the `EditorField` component explaining this then.

erlingfn: Will do

mrfarstad:
What is the status of this PR? @LudvigHz @erlingfn 🌮

LudvigHz:
> What is the status of this PR? @LudvigHz @erlingfn taco

Other than the conflicts, I believe this is "ready". It just needs to be tested some more, and then we can get this in asap.

erlingfn:
Will have to rebase it again as it conflicts with almost all other PR's, but as far as I know, it should be ready for review

erlingfn:
If you could test this it would be much appreciated @mrfarstad

mrfarstad:
> If you could test this it would be much appreciated @mrfarstad

Can give it a try 💯

erlingfn:
> Awesome! I have tested this a bunch now, and I couldn't find any _noticible_ issues 
> However, I also get the `react prepare timeout when server side rendering` when rendering the frontpage and an `undefined render error` when rendering `/events/{id}`. If the react-prepare timeout has no real effect, I think we should just fix that later. If you know what the events issue is, I would like to know what causes it.
> 
> Other than that, this is super smooth, and the only visible issue I found was some weird background color on the pool fields in the event form (not important!). Great work!

Regarding the the prepare timeout, this _does_ have an effect, namely that the page is not server side rendered. You can check this by viewing the page source. The reason for this is that we simply have a hard coded timeout on 4 seconds for ssr, and this is not enough time to ssr the frontpage when running locally. I, at least, get a prepare timeout when trying to ssr the frontpage in master as well. As far as I am aware, this does seem like simply being due to the ssr timeout being designed for a server.

erlingfn:
The error in /events/<id>/ seems to be an old error, as it is currently present in prod as well, causing that page to not be server side rendered

LudvigHz:
> The error in /events/{id}/ seems to be an old error, as it is currently present in prod as well, causing that page to not be server side rendered

Is this the `invalid hook call` error?

erlingfn:
> > The error in /events/{id}/ seems to be an old error, as it is currently present in prod as well, causing that page to not be server side rendered
> 
> Is this the `invalid hook call` error?

No. The error is thrown by the editor since it cannot be server side rendered. As far as I know this is the intended behavior

app/Root.js
```diff
@@ -1,65 +1,27 @@
...
-  null,
-  mapDispatchToProps
-)(
-  ({
-    setStatusCode,
-    ...restProps
-  }: {
-    setStatusCode: (statusCode: ?number) => void
-  }) => (
-    <ErrorBoundary openReportDialog>
```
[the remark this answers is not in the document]
erlingfn: Good catch! I have added it back in

mrfarstad:
Seems to work fine on my end! 🚀 

However, when rendering the front page, I get a warning that is not present in master:
`Warning: Unknown event handler property 'onlyActiveOnIndex'. It will be ignored.`
Any thoughts on this?

erlingfn:
> Seems to work fine on my end! 
> 
> However, when rendering the front page, I get a warning that is not present in master:
> `Warning: Unknown event handler property 'onlyActiveOnIndex'. It will be ignored.`
> Any thoughts on this?

Looks like it was a leftover from react router v3. It should have been removed now

mrfarstad:
> > Seems to work fine on my end!
> > However, when rendering the front page, I get a warning that is not present in master:
> > `Warning: Unknown event handler property 'onlyActiveOnIndex'. It will be ignored.`
> > Any thoughts on this?
> 
> Looks like it was a leftover from react router v3. It should have been removed now

Ahh, I see. Cool! Then I think we should merge this PR if no one argues against it 🚀

PR #1552, opened 2019-04-15
E2E tests for basic routing around the webapp

SmithPeder:
This will press all links in the navbar, and then press something within
the page they land on. This will also press on the extended menu and
check every(none admin) page.

Screenshot 2019-04-16 at 00 29 15

### DIFF
--- a/cypress/integration/router_spec.js
+++ b/cypress/integration/router_spec.js
@@ -1,0 +1,278 @@
+import { c, a } from '../support/utils.js';
+
+describe('Create event', () => {
+  beforeEach(() => {
+    cy.resetDb();
+    cy.cachedLogin();
+  });
+
+  const openMenu = () => {
+    cy.get(c('buttonGroup')).within(() => {
+      cy.get(c('searchIcon')).click();
+    });
+  };
+
+  it('should be able to navigate to events', () => {
+    cy.visit('/');
+    cy.get(c('navigation')).within(() => {
+      cy.get(a('/events')).click();
+    });
+    cy.url().should('contain', '/events');
+    cy.contains('Denne uken');
+    cy.contains('Liste');
+  });
+
+  it('should be able to navigate to joblistings', () => {
+    cy.visit('/');
+    cy.get(c('navigation')).within(() => {
+      cy.get(a('/joblistings')).click();
+    });
+    cy.url().should('contain', '/joblistings');
+    cy.contains('Jobbannonser');
+    cy.contains('Søknadsfrist');
+  });
+
+  it('should be able to navigate to about-page', () => {
+    cy.visit('/');
+    cy.get(c('navigation')).within(() => {
+      cy.get(a('/pages/info-om-abakus')).click();
+    });
+    cy.url().should('contain', '/pages/info-om-abakus');
+    cy.contains('Generelt');
+    cy.contains('Komiteer');
+  });
+
+  it('should be able to navigate to users profile', () => {
+    cy.visit('/');
+    cy.get(c('buttonGroup')).within(() => {
+      cy.get('img[alt="user"]').click();
+    });
+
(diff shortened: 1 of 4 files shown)


cypress/integration/router_spec.js
```diff
@@ -0,0 +1,278 @@
...
+    });
+    cy.url().should('contain', '/events');
+    cy.contains('Denne uken');
+    cy.contains('Liste');
+  });
+
+  it('should be able to navigate to joblistings', () => {
+    cy.visit('/');
+    cy.get(c('navigation')).within(() => {
+      cy.get(a('/joblistings')).click();
```
orhanhenrik: Maybe it's better to select the link by its text content rather than its URL? That's how a user would find the right link to click. And you already assert the URL further down.

If you agree, the same applies to the rest of the tests.

cypress/integration/router_spec.js
```diff
@@ -0,0 +1,278 @@
...
+
+  it('should be able to logg out', () => {
+    cy.visit('/');
+    cy.get(c('buttonGroup')).within(() => {
+      cy.get('img[alt="user"]').click();
+    });
+
+    // Logg out
+    cy.get(c('Dropdown')).within(() => {
+      cy.get(c('Button')).click();
```
orhanhenrik: Is there a better way to select this button? It looks very generic and could easily fail if we add more options to the dropdown

SmithPeder: Yeah, that sucked 🙃  I'ma switch that out with `.contains('Logg ut')`

cypress/integration/router_spec.js
SmithPeder: I was dabbling with the same thought, what is the best way to select an element, using `.contians` on the text or `.get` on some property. I thought about what was least likely to change over time. It was just an effort to keep the test viable in future rewrites.

But you are right about the _"That's how a user would find the right link"_, so maybe the `.contains` way is better.

orhanhenrik: I think if you change "big" things like how navigation works or what things are named, you should expect to have to fix some tests. I'm not sure exactly what is best though, it's a bit hard to choose which selectors to use. 

In their best practices they have some examples of what cypress thinks you should do: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

image

orhanhenrik: In the case of these tests, I think the wording in each link is also an important thing to test, as it is essential for the use of navigation. If the URLs are correct but the text is wrong, the tests should ideally fail

SmithPeder: You got a good point there

cypress/integration/router_spec.js
```diff
@@ -0,0 +1,229 @@
+import { c, a } from '../support/utils.js';
+
+describe('Create event', () => {
+  beforeEach(() => {
+    cy.resetDb();
+    cy.cachedLogin();
+  });
+
+  // Open the hambuermenu and select by name, then assert by path
+  const openMenuAndSelect = (name, path) => {
```
SmithPeder: It now selects every menu link by name and asserts it by its path

cypress/integration/router_spec.js
```diff
@@ -0,0 +1,229 @@
...
+
+  it('should be able to navigate to users profile', () => {
+    cy.visit('/');
+    cy.get(c('buttonGroup')).within(() => {
+      cy.get('img[alt="user"]').click();
+    });
+
+    // Go to profile
+    cy.get(c('Dropdown')).within(() => {
+      cy.get(a('/users/me')).click();
```
SmithPeder: I don't know how to switch out this with a contains, as the name is "webkom" or whatever, so I think it has to stay like this :/

PR #1553, opened 2019-04-16
[VERY WIP] 🔒🔒Typescript  🔒🔒 

SmithPeder:
image **Switch from flow to TypeScript** 

> _NOTE: This rewrite is in an early (read VERY EARLY) stage, so there is no reason to review_ 

<hr>

### I'm going to use the `R2-D2` account to steal A LOT of these lines.

<hr>

This is just to show the (very large) community that we are working on this.

**Progress:**
`Disclaimer: It might not work perfectly just because it's check of`
- [x] Switch file names
- [x] tsconf
- [x] babelconf
- [x] eslintconf
- [x] Switch to interfaces in components
- [x] Create new API interfaces per endpoint 
- [ ] Switch to interfaces in routes
- [ ] Find all types for all dependencies
- [ ] Get Webpack working with new babel conf and loaders
- [ ] Get coverage report to work
- [ ] Get ~100% coverage

- [ ] A lot of duplicate and wrong entities and types everywhere

- 💯more 

**Problems:**
- One devDep per dep with `@types/`
- TypeScript coverage `tscov`
- Css-loader with types `webpack`
- Everything else with `webpack`
- Old types for react-redux, react-router, react-select
- Sealed object notation in Flow `({||} vs {})`.
- Untyped files
- Lacking types in files
- Wrong/lacking models and entities
- 💯more


odinuge:
I think this requires a rebase, or what do you think @SmithPeder? :thinking:

PR #1557, opened 2019-04-23
Start using immer in reducers :shipit:

orhanhenrik:
This simplifies the code a lot, especially when you update deeply nested objects

### DIFF
--- a/app/actions/NotificationSettingsActions.js
+++ b/app/actions/NotificationSettingsActions.js
@@ -19,7 +19,7 @@ export function fetchNotificationSettings() {
     types: NotificationSettings.FETCH,
     endpoint: '/notification-settings/',
     meta: {
-      errorMessage: 'Henting av varslingsintillinger feilet'
+      errorMessage: 'Henting av varslingsinntillinger feilet'
     },
     propagateError: true
   });
@@ -39,7 +39,8 @@ export function updateNotificationSetting(
       channels
     },
     meta: {
-      errorMessage: 'Oppdatering av varslingsintillinger feilet'
+      errorMessage: 'Oppdatering av varslingsinntillinger feilet',
+      successMessage: 'Varslingsinnstillinger oppdatert'
     }
   });
 }
--- a/app/actions/NotificationSettingsActions.js
+++ b/app/actions/NotificationSettingsActions.js
@@ -19,7 +19,7 @@ export function fetchNotificationSettings() {
     types: NotificationSettings.FETCH,
     endpoint: '/notification-settings/',
     meta: {
-      errorMessage: 'Henting av varslingsinntillinger feilet'
+      errorMessage: 'Henting av varslingsinnstillinger feilet'
     },
     propagateError: true
   });
@@ -39,7 +39,7 @@ export function updateNotificationSetting(
       channels
     },
     meta: {
-      errorMessage: 'Oppdatering av varslingsinntillinger feilet',
+      errorMessage: 'Oppdatering av varslingsinnstillinger feilet',
       successMessage: 'Varslingsinnstillinger oppdatert'
     }
   });
--- a/app/reducers/announcements.js
+++ b/app/reducers/announcements.js
@@ -4,6 +4,7 @@ import { createSelector } from 'reselect';
 import { Announcements } from '../actions/ActionTypes';
 import createEntityReducer from 'app/utils/createEntityReducer';
 import moment from 'moment-timezone';
+import { pull } from 'lodash';
 import produce from 'immer';
 
 type State = any;
@@ -21,6 +22,12 @@ export default createEntityReducer({
         case Announcements.SEND.SUCCESS:
           newState.byId[action.meta.announcementId].sent = moment();
           break;
+<<<<<<< HEAD
+=======
+        case Announcements.DELETE.SUCCESS:
+          pull(newState.items, action.meta.announcementId);
+          break;
+>>>>>>> Use lodash pull and remove instead of filter in reducers
       }
     }
   )
--- a/app/reducers/announcements.js
+++ b/app/reducers/announcements.js
@@ -4,7 +4,6 @@ import { createSelector } from 'reselect';
 import { Announcements } from '../actions/ActionTypes';
 import createEntityReducer from 'app/utils/createEntityReducer';
 import moment from 'moment-timezone';
-import { without } from 'lodash';
 import produce from 'immer';
 
 type State = any;
@@ -22,12 +21,6 @@ export default createEntityReducer({
         case Announcements.SEND.SUCCESS:
           newState.byId[action.meta.announcementId].sent = moment();
           break;
-<<<<<<< HEAD
-=======
-        case Announcements.DELETE.SUCCESS:
-          newState.items = without(newState.items, action.meta.announcementId);
-          break;
->>>>>>> Use lodash pull and remove instead of filter in reducers
       }
     }
   )
--- a/app/reducers/announcements.js
+++ b/app/reducers/announcements.js
@@ -15,14 +15,15 @@ export default createEntityReducer({
     mutate: Announcements.CREATE,
     delete: Announcements.DELETE
   },
-  mutate: (state, action) =>
-    produce(state, newState => {
+  mutate: produce(
+    (newState: State, action: any): void => {
       switch (action.type) {
         case Announcements.SEND.SUCCESS:
           newState.byId[action.meta.announcementId].sent = moment();
           break;
       }
-    )
+    }
+  )
 });
 
 export const selectAnnouncements = createSelector(
--- a/app/reducers/announcements.js
+++ b/app/reducers/announcements.js
@@ -4,6 +4,7 @@ import { createSelector } from 'reselect';
 import { Announcements } from '../actions/ActionTypes';
 import createEntityReducer from 'app/utils/createEntityReducer';
 import moment from 'moment-timezone';
+import produce from 'immer';
 
 export default createEntityReducer({
   key: 'announcements',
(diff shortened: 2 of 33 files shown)


SmithPeder:
Looks really nice and clean 🎉  👍

app/reducers/announcements.js
```diff
@@ -4,37 +4,32 @@ import { createSelector } from 'reselect';
...
-          byId: {
-            ...state.byId,
-            [announcementId]: {
-              ...state.byId[announcementId],
-              sent: moment()
-            }
-          }
-        };
+  mutate: (state: State, action): State =>
+    produce(
```
SmithPeder: Will `produce` return the state if none of the cases trigger, since you don't write any default case?

app/reducers/comments.js
```diff
@@ -15,62 +16,45 @@ export type CommentEntity = {
...
-          byId: {
-            ...state.byId,
-            [targetId]: {
-              ...state.byId[targetId],
-              comments: [
-                ...state.byId[targetId].comments,
-                action.payload.result
-              ]
+  return (state: State, action: any) => {
+    return produce(
```
SmithPeder: Maybe an implicit return here? To avoid the double return.

app/reducers/comments.js
```diff
@@ -15,62 +16,45 @@ export type CommentEntity = {
...
-          }
-        }
-      };
-    }
-    default: {
-      return state;
+type CommentState = any;
+
+function mutate(state: CommentState, action: any): CommentState {
+  return produce(
```
SmithPeder: Maybe `const mutate..... => produce()` ? I know you didn't write it, but when you're here, refactoring, we might as well make it nice (Like you did with mutateArticle)

app/reducers/companies.js
```diff
@@ -74,99 +74,69 @@ export type CompanyContactEntity = {
...
-      const semesterStatuses = state.byId[companyId].semesterStatuses.map(
-        status => (status.id === semesterStatusId ? action.payload : status)
-      );
-      return mergeObjects(state, {
-        byId: {
-          [companyId]: { semesterStatuses }
+type State = any;
+
+function mutateCompanies(state: State, action) {
+  return produce(
```
SmithPeder: (Like you did with mutateArticle)

app/reducers/events.js
```diff
@@ -10,226 +10,139 @@ import { normalize } from 'normalizr';
...
-      const events = action.payload.result.reduce(
-        (total, id) => ({
-          ...total,
-          [id]: {
-            ...action.payload.entities.events[id],
-            isUsersUpcoming: true
+type State = any;
+
+function mutateEvent(state: State, action: any): State {
+  return produce(
```
SmithPeder: (Like you did with mutateArticle)

app/reducers/announcements.js
orhanhenrik: Yes, it returns newState, no matter how/if you mutate it.

orhanhenrik:
Need to do some manual testing, as our current tests don't cover all of the changes made..

SmithPeder:
In regards to the testing bit, maybe we should write some reducer tests? Seams like something smart to have tests on, and should be very simple to write with assertion tests.

app/actions/NotificationSettingsActions.js
```diff
@@ -19,7 +19,7 @@ export function fetchNotificationSettings() {
     types: NotificationSettings.FETCH,
     endpoint: '/notification-settings/',
     meta: {
-      errorMessage: 'Henting av varslingsintillinger feilet'
+      errorMessage: 'Henting av varslingsinntillinger feilet'
```
LudvigHz: Intillinger 😛

orhanhenrik:
Yes, some unit tests for reducers would be nice. That would be easier if we had proper types everywhere as well 😂  
We can make tests in a separate PR and then rebase this branch on the test branch to see that everything passes both before and after changes.  
Don't have much time until next week, but I will start on it then unless somebody else wants to.

app/reducers/companySemesters.js
```diff
@@ -12,26 +13,25 @@ export type CompanySemesterEntity = {
...
-      case Company.ADD_SEMESTER.SUCCESS: {
-        return {
-          ...state,
-          byId: {
-            ...state.byId,
-            [action.payload.id]: action.payload
-          },
-          items: state.items.concat(action.payload.id)
-        };
+  // TODO: I think this can be removed by using { types: mutate: Company.ADD_SEMESTER} above
```
odinuge: Mutate is for dealing with optimistic stuff. If the result is in the response, and the schema is set; createEntityReducer should do this by default i think.

app/reducers/toasts.js
```diff
@@ -17,26 +17,19 @@ type State = {
...
-          if (toast.id === action.payload.id) {
-            return { ...toast, removed: true };
-          }
-          return toast;
-        }): Array<Toast>)
-      };
-
-    default:
-      return state;
+      newState.items.forEach(toast => {
```
odinuge: Hmm. Changing an object inside a forEach... :thinking: Not my favorite thing to do even tho immer allows it...

app/reducers/routing.js
```diff
@@ -1,19 +1,18 @@
+// @flow
 import { Routing } from 'app/actions/ActionTypes';
+import produce from 'immer';
 
 const initialState = {
   statusCode: null
 };
 
-export default function routing(state = initialState, action) {
+type State = typeof initialState;
```
odinuge: :biohazard:

app/reducers/articles.js
```diff
@@ -24,19 +26,17 @@ export type ArticleEntity = {
...
-    case Article.DELETE.SUCCESS: {
-      const { articleId } = action.meta;
-      return {
-        ...state,
-        items: state.items.filter(id => id !== articleId)
-      };
+type State = any;
+
+const mutateArticle = produce(
+  (newState: State, action: any): void => {
```
odinuge: Why use any for the action instead of typing it properly?

app/reducers/companies.js
```diff
@@ -74,100 +75,65 @@ export type CompanyContactEntity = {
...
-          [companyId]: { semesterStatuses }
-        }
-      });
+type State = any;
+
+const mutateCompanies = produce(
+  (newState: State, action: any): void => {
+    switch (action.type) {
+      case Company.DELETE.SUCCESS:
+        pull(newState.items, action.meta.id);
```
odinuge: Hmm... Unsure about what I think about using utils doing in place edits on objects like this, since it makes my inner js-developer think that it doesn't change the object - even tho it does.

The alternative is
```
newState.items = without(newState.items, action.meta.id);
```

But I am not 100% sure about what i liked best.

app/reducers/events.js
```diff
@@ -10,227 +10,142 @@ import { normalize } from 'normalizr';
...
-        byId: {
-          ...state.byId,
-          [action.meta.id]: {
-            ...state.byId[action.meta.id],
-            loading: true
+        );
+        break;
+
+      case Event.FETCH_UPCOMING.SUCCESS:
+        Object.entries(action.payloads.entities.events).forEach(
```
odinuge: Holy moly :scream:

app/reducers/routing.js
orhanhenrik: This is used a few other places, is there something wrong with it?

https://github.com/webkom/lego-webapp/blob/60aae52f29b89f1f7bd23150d952df69b38ac090/app/reducers/meetingsToken.js#L12
https://github.com/webkom/lego-webapp/blob/60aae52f29b89f1f7bd23150d952df69b38ac090/app/reducers/notificationsFeed.js#L11
https://github.com/webkom/lego-webapp/blob/963d4dd22231e59c796661ed61076b89687c7c5a/app/reducers/search.js#L119

odinuge: Yee, typing is kinda bad. I guess it works, but it isn't that expressive. And when `statusCode` is `null`, what is then the type? :man_shrugging:

app/reducers/articles.js
orhanhenrik: I figured the effort to type actions properly would be a bit wasted since we are doing a ts rewrite.  
I tried using https://github.com/webkom/lego-webapp/blob/master/app/types.js#L74 but it yielded lots of errors when trying to access parts of action.payload

odinuge: Ahh

Payload is however optional, aka. can be null. Our reducers mostly doesn't care tho, so using the type can help avoid nullptr exceptions

app/reducers/companies.js
orhanhenrik: Yes, we can use without instead to be a bit more expressive.

app/reducers/events.js
orhanhenrik: Haha I guess this would be nicer as a simple for loop

app/reducers/toasts.js
orhanhenrik: I can change this to find -> mutate

app/reducers/routing.js
orhanhenrik: Maybe todo for typescript? I can add better types here, but it will have to be changed soon anyways, and I don't know if we want to do too much in one PR.

app/reducers/articles.js
orhanhenrik: What payload is/can be is really dependent on action.type. Not sure if this can be expressed with flow or typescript. 

It looks like typescript can handle this really well
https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions
We will need to define a type for each action type though, but I think that makes sense if we are to get any benefit from typing.

SmithPeder: Yeah, I looked into this as well, there seem to be well-suited methods for typing reducers.
https://www.matthewgerstman.com/tech/ts-tricks-type-safe-reducers/

PR #1559, opened 2019-04-27
CreateEntityReducer delete

odinuge:
This will remove ish. `50%` of our reducers; aka. remove quite a lot of duplicate code. Most of the stuff we do is CRUD anyways.

PS: DO NOT MERGE INTO MASTER NOW. This should be merged into the `immer` branch when done (after we agree about the design of it :smile:)

Also some small adjustements to CER.

### DIFF
--- a/app/actions/PageActions.js
+++ b/app/actions/PageActions.js
@@ -22,7 +22,7 @@ export function deletePage(pageSlug: string) {
     endpoint: `/pages/${pageSlug}/`,
     method: 'DELETE',
     meta: {
-      pageSlug,
+      id: pageSlug,
       errorMessage: 'Sletting av side feilet'
     }
   });
--- a/app/reducers/events.js
+++ b/app/reducers/events.js
@@ -51,12 +51,6 @@ function mutateEvent(state: any, action: any) {
         byId: mergeObjects(state.byId, events)
       };
     }
-    case Event.DELETE.SUCCESS: {
-      return {
-        ...state,
-        items: state.items.filter(id => id !== action.meta.id)
-      };
-    }
     case Event.SOCKET_EVENT_UPDATED: {
       const events = normalize(action.payload, eventSchema).entities.events;
       return {
@@ -237,7 +231,8 @@ const mutate = joinReducers(mutateComments('events'), mutateEvent);
 export default createEntityReducer({
   key: 'events',
   types: {
-    fetch: [Event.FETCH, Event.FETCH_PREVIOUS, Event.FETCH_UPCOMING]
+    fetch: [Event.FETCH, Event.FETCH_PREVIOUS, Event.FETCH_UPCOMING],
+    delete: Event.DELETE
   },
   mutate
 });
--- a/app/utils/__tests__/createEntityReducer.spec.js
+++ b/app/utils/__tests__/createEntityReducer.spec.js
@@ -1,6 +1,6 @@
 import createEntityReducer, {
   fetching,
-  entities
+  updateEntities
 } from '../createEntityReducer';
 import joinReducers from '../joinReducers';
 import { eventSchema } from 'app/reducers';
@@ -282,7 +282,7 @@ describe('entities()', () => {
       items: []
     };
 
-    const reducer = entities('users');
+    const reducer = updateEntities({}, 'users');
 
     const user = {
       id: 1,
@@ -311,7 +311,7 @@ describe('entities()', () => {
   });
 
   it('should merge carefully', () => {
-    const reducer = entities('events');
+    const reducer = updateEntities({}, 'events');
     const events = [
       {
         id: 1,
--- a/app/utils/__tests__/createEntityReducer.spec.js
+++ b/app/utils/__tests__/createEntityReducer.spec.js
@@ -1,16 +1,24 @@
 import createEntityReducer, {
   fetching,
-  updateEntities
+  updateEntities,
+  deleteEntities
 } from '../createEntityReducer';
 import joinReducers from '../joinReducers';
 import { eventSchema } from 'app/reducers';
 import { normalize } from 'normalizr';
 
 const FETCH = {
-  BEGIN: 'FETCH_BEGIN',
-  SUCCESS: 'FETCH_SUCCESS',
-  FAILURE: 'FETCH_FAILURE'
+  BEGIN: 'FETCH.BEGIN',
+  SUCCESS: 'FETCH.SUCCESS',
+  FAILURE: 'FETCH.FAILURE'
 };
+const DELETE = {
+  BEGIN: 'DELETE.BEGIN',
+  SUCCESS: 'DELETE.SUCCESS',
+  FAILURE: 'DELETE.FAILURE'
+};
+
+const SMASH = 'SMASH';
 
 const reducer = createEntityReducer({
   key: 'events',
@@ -21,7 +29,7 @@ const reducer = createEntityReducer({
     smashed: false
   },
   mutate: (state, action) => {
-    if (action.type === 'SMASH') {
+    if (action.type === SMASH) {
       return { ...state, smashed: true };
     }
     return state;
@@ -29,9 +37,15 @@ const reducer = createEntityReducer({
 });
 
 const FETCH_OTHER = {
-  BEGIN: 'FETCH_OTHER_BEGIN',
-  SUCCESS: 'FETCH_OTHER_SUCCESS',
-  FAILURE: 'FETCH_OTHER_FAILURE'
+  BEGIN: 'FETCH_OTHER.BEGIN',
+  SUCCESS: 'FETCH_OTHER.SUCCESS',
+  FAILURE: 'FETCH_OTHER.FAILURE'
+};
+
+const DELETE_OTHER = {
+  BEGIN: 'DELETE_OTHER.BEGIN',
+  SUCCESS: 'DELETE_OTHER.SUCCESS',
+  FAILURE: 'DELETE_OTHER.FAILURE'
 };
 
 const otherReducer = createEntityReducer({
@@ -43,7 +57,7 @@ const otherReducer = createEntityReducer({
     smashed: false
   },
   mutate: (state, action) => {
-    if (action.type === 'SMASH') {
+    if (action.type === SMASH) {
       return { ...state, smashed: true };
     }
     return state;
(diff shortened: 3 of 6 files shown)


app/utils/createEntityReducer.js
```diff
@@ -99,7 +99,26 @@ export function entities(key: string, fetchType?: ?AsyncActionType) {
   };
 }
 
-export function optimistic(mutateType?: ?AsyncActionType) {
+/* This part of CER will handle deletion of values of a given entity.
+ * For actions of the type given into `types.delete` when setting up CER will be deleted.
+ *
+ * Make sure to set `meta.id` in the action.
+ */
+export function deleteEntities(key: string, deleteType: AsyncActionType) {
```
SmithPeder: No need for a key here?

app/utils/createEntityReducer.js
```diff
@@ -119,7 +138,7 @@ export function optimistic(mutateType?: ?AsyncActionType) {
   };
 }
 
-export function paginationReducer(key: string, fetchType?: ?AsyncActionType) {
+export function paginationReducer(key: string, fetchType: AsyncActionType) {
```
SmithPeder: Maybe no need for a key here either? `insertEntities` uses the key in an assertion with the schema key, but here the pagination key is just retrieved.

app/utils/createEntityReducer.js
```diff
@@ -173,16 +190,23 @@ export default function createEntityReducer({
     ...initialState
   };
 
-  const fetchTypes = isArray(fetch) ? fetch : [fetch];
+  const toArray = <T>(value: T | Array<T>) =>
```
SmithPeder: I think this function should be able to take the value `undefined` and return an empty array `[]`. That would remove the need to check and pass empty arrays when one of the types is undefined ( ` || [] ` ).

app/utils/createEntityReducer.js
```diff
@@ -173,16 +190,23 @@ export default function createEntityReducer({
...
     ...fetchTypes.map(fetchType =>
       joinReducers(
         fetching(fetchType),
-        entities(key, fetchType),
+        insertEntities(key, fetchType),
         paginationReducer(key, fetchType)
       )
     ),
-    optimistic(mutateType),
+    ...deleteTypes.map(deleteType => deleteEntities(key, deleteType)),
```
SmithPeder: nokey

app/utils/createEntityReducer.js
```diff
@@ -173,16 +190,23 @@ export default function createEntityReducer({
...
+  const deleteTypes = toArray(types.delete || []);
+  const mutateTypes = toArray(types.mutate || []);
+
   const reduce = joinReducers(
     ...fetchTypes.map(fetchType =>
       joinReducers(
         fetching(fetchType),
-        entities(key, fetchType),
+        insertEntities(key, fetchType),
         paginationReducer(key, fetchType)
```
SmithPeder: nokey

orhanhenrik:
Having CreateEntityReducer handle deletes will definitely be nice! Lots of repetition in reducers at the moment, I noticed that as well while refactoring reducers..

app/actions/PageActions.js
```diff
@@ -22,6 +22,7 @@ export function deletePage(pageSlug: string) {
     endpoint: `/pages/${pageSlug}/`,
     method: 'DELETE',
     meta: {
+      id: pageSlug,
       pageSlug,
```
orhanhenrik: Should only need id after rewrite, can delete pageSlug

app/utils/createEntityReducer.js
```diff
@@ -99,7 +99,26 @@ export function entities(key: string, fetchType?: ?AsyncActionType) {
...
+/* This part of CER will handle deletion of values of a given entity.
+ * For actions of the type given into `types.delete` when setting up CER will be deleted.
+ *
+ * Make sure to set `meta.id` in the action.
+ */
+export function deleteEntities(key: string, deleteType: AsyncActionType) {
+  return (state: any = defaultState, action: any) => {
+    const resultId = action.meta && action.meta.id;
+
+    if (!resultId || action.type !== get(deleteType, 'SUCCESS')) return state;
```
orhanhenrik: Do we need to use `get` when deleteType is not optional?

app/reducers/events.js
```diff
@@ -237,7 +231,8 @@ const mutate = joinReducers(mutateComments('events'), mutateEvent);
 export default createEntityReducer({
   key: 'events',
   types: {
-    fetch: [Event.FETCH, Event.FETCH_PREVIOUS, Event.FETCH_UPCOMING]
+    fetch: [Event.FETCH, Event.FETCH_PREVIOUS, Event.FETCH_UPCOMING],
+    delete: [Event.DELETE]
```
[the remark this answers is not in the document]
odinuge: Proof of concept :tm:

app/utils/createEntityReducer.js
odinuge: :+1:

app/utils/createEntityReducer.js
odinuge: Sounds good :+1:

app/utils/createEntityReducer.js
odinuge: Yeah, good catch. No need to have optional types here

app/utils/createEntityReducer.js
odinuge: :+1:

orhanhenrik:
Also need to add tests for delete

app/utils/createEntityReducer.js
```diff
@@ -99,11 +106,38 @@ export function entities(key: string, fetchType?: ?AsyncActionType) {
...
+    ) {
+      return state;
+    }
+    const resultId = action.meta && action.meta.id;
+
+    if (!resultId) return state;
+
+    return {
+      ...state,
+      byId: omit(state.byId, resultId),
```
odinuge: Should we do this, or just keep the element in the map? Most reducers only removed the entity from the "items" list. We have no track on the relations, so it is impossible to know if the pk is refrenced in another object... :thinking:

odinuge:
Do you want this merged into your `immer` branch or into master (so you have to rebase) @orhanhenrik ?

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -382,3 +408,168 @@ describe('entities()', () => {
...
+          title: 'Second Event',
+          comments: []
+        }
+      },
+      items: [1, 2],
+      pagination: {}
+    };
+
+    const action = {
+      type: 'DELETE_SUCCESS',
```
orhanhenrik: Use DELETE.SUCCESS?

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -382,3 +408,168 @@ describe('entities()', () => {
...
```
orhanhenrik: same here

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -382,3 +408,168 @@ describe('entities()', () => {
...
+          title: 'Second Event',
+          comments: []
+        }
+      },
+      items: [1, 2],
+      pagination: {}
+    };
+
+    const action = {
+      type: 'DELETE_FAILED',
```
orhanhenrik: same

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -382,3 +408,168 @@ describe('entities()', () => {
...
+          id: 2,
+          title: 'Second Event',
+          comments: []
+        }
+      },
+      items: [2],
+      pagination: {}
+    });
+  });
+  it('should handle deleteTypes as array', () => {
```
orhanhenrik: Maybe also check that DELETE_OTHER works?

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -382,3 +408,168 @@ describe('entities()', () => {
...
+          title: 'Second Event',
+          comments: []
+        }
+      },
+      items: [1, 2],
+      pagination: {}
+    };
+
+    const action = {
+      type: 'DELETE_OTHER_SUCCESS',
```
orhanhenrik: same

app/utils/createEntityReducer.js
orhanhenrik: Good question. Not sure if there is any reason for the current behaviour. I think deleting it everywhere makes sense. If there is UI using the deleted object it should probably crash and let us know instead of keep working.

app/utils/__tests__/createEntityReducer.spec.js
odinuge: Just kept the same as in the other tests, but I guess we can switch to `DELETE.SUCCESS`

orhanhenrik:
We can merge it into either immer or master, I'm fine with either. I can change the rest of the reducers in immer so we don't get too many conflicts.

orhanhenrik:
Actually you can change all the reducers if you want. I'm writing reducer tests atm and it would be nice to get rid of all the delete code.

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -272,9 +284,23 @@ describe('fetching()', () => {
...
+
+    expect(reducer({ fetching: false }, { type: DELETE.SUCCESS })).toEqual({
+      fetching: false
+    });
+  });
 });
 
-describe('entities()', () => {
+describe('updateEntities()', () => {
   it('store entities', () => {
```
SmithPeder: throw in a `should` here, as it's the only test with no should 🔥

app/utils/__tests__/createEntityReducer.spec.js
```diff
@@ -382,3 +408,168 @@ describe('entities()', () => {
...
+          id: 2,
+          title: 'Second Event',
+          comments: []
+        }
+      },
+      items: [1, 2],
+      pagination: {}
+    });
+  });
+  it('Should only delete on matching type', () => {
```
SmithPeder: small `s`

odinuge: :bigshaq:

app/utils/createEntityReducer.js
odinuge: :+1: Ill just keep it like this this. Agree that this makes more sense than the current behavior

PR #1561, opened 2019-04-29
tests on frontpage and joblistings

amalieeh:
Have to make tests for browsing joblistings also


orhanhenrik:
Should we merge this @amalieeh ? You can add more tests in a separate PR if you are not working on this at the moment.

PR #1564, opened 2019-05-08
Stop using dispatched

odinuge:
Stop using dispatched as it isn't supported with the new `react-redux` connector. :smile:

### DIFF
--- a/app/routes/events/EventAbacardRoute.js
+++ b/app/routes/events/EventAbacardRoute.js
@@ -3,7 +3,7 @@ import { compose } from 'redux';
 import { connect } from 'react-redux';
 import { push } from 'react-router-redux';
 import { debounce } from 'lodash';
-import { dispatched } from '@webkom/react-prepare';
+import prepare from 'app/utils/prepare';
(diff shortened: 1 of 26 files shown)


app/routes/events/EventAbacardRoute.js
```diff
@@ -16,10 +16,10 @@ import { getRegistrationGroups } from 'app/reducers/events';
 
 const searchTypes = ['users.user'];
 
-const loadData = (props, dispatch) => {
+const loadData = async (props, dispatch) => {
   const query = props.location.query.q;
   if (query) {
-    dispatch(autocomplete(query, searchTypes));
+    await dispatch(autocomplete(query, searchTypes));
```
LudvigHz: Why is this async? Wouldn't returning the promise from `dispatch` be sufficient?

odinuge: That works too, but then a `return Promise.resolve()` is needed in the end (if `!!query === false`)

LudvigHz: Ah, right. I guess async/await is cleaner then

PR #1567, opened 2019-05-11
Cleanup server side renderer

odinuge:
This makes server part of webpack build way faster during development,
~10x.

Benchmark (my xps):
- OLD: 19.35s
- NEW:  1.97s

Also:
- Update react-hot-loader 
- Use render instead of hydrate when not using ssr
- Bump webpack deps
- Ignore react-invalid-hook-call on SSR (see commit for more info, https://sentry.abakus.no/webkom/lego-webapp/issues/6808/?query=is:unresolved)
    - https://ci.abakus.no/webkom/lego-webapp/5397/16 vs. https://ci.abakus.no/webkom/lego-webapp/5398/16
    - Currently there is no official (nor unofficial) implementation of `react-with-hooks` for node. The `renderToString` does work, but no other low-level stuff :/
- Remove unused dependency parse5


odinuge:
Screenshot from 2019-05-11 19-50-19

server/render.js
```diff
@@ -1,238 +1,4 @@
...
-        ${dllPlugin}
-        ${scripts}
-      </body>
-    </html>
-   `;
-}
-
-export default render;
+const noSSR = (req, res) => res.send(pageRenderer());
+export default (__DEV__ ? noSSR : require('./ssr.js').default);
```
odinuge: Since this doesn't import './ssr.js' in development, it doesn't have to compile all the millions of lines with js (since we don't use the SSR in dev, heh)

app/render.js
```diff
@@ -2,20 +2,32 @@
...
+import { hydrate, render } from 'react-dom';
+import { setConfig } from 'react-hot-loader';
 import { match } from 'react-router';
 import Root from './Root';
-import { AppContainer } from 'react-hot-loader';
 import routes from 'app/routes';
 import type { Store } from 'app/types';
+setConfig({
+  //pureSFC: true
+  ignoreSFC: true, // RHL will be __completely__ disabled for SFC
```
orhanhenrik: Maybe write what RHL and SFC stands for in the comment for readability

server/pageRenderer.js
```diff
@@ -0,0 +1,120 @@
...
+
+  !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t,e){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=e};analytics.SNIPPET_VERSION="4.1.0";
+          analytics.load("${config.segmentWriteKey}");
+          }}();
+      </script>`
+            : ''
+        }
+        <script async src="https://js.stripe.com/v2/"></script>
+        <script async src="https://js.stripe.com/v3/"></script>
```
orhanhenrik: Not really related to this PR, but do we still need both stripe versions?

odinuge: Yee. :/

We still use the old checkout-api (v2) and the new `react-stripe-elements`. We can remove v2 when this (not the GPay part) is done https://github.com/webkom/lego-webapp/pull/1527. We can also use the new checkout api (just a few days old i think) https://stripe.com/docs/payments/checkout for users without google or apple pay. :smile:

app/render.js
odinuge: Nice catch! Almost forgot this..

Did some testing with this and https://github.com/mzgoddard/hard-source-webpack-plugin/. hard-source-webpack-plugin makes `yarn start` take ~7-8.sec, and I had to tweak react-hot-loader to make it work. Not 110% stable tho, more like 60-70%, so we can look into that later.

From what I remember, we may as well just remove this setConfig altogether. 

Some background info/kok
https://github.com/gaearon/react-hot-loader/issues/1088#issuecomment-452904866

odinuge: Yee. Works well without it, so I removed it now.

PR #1570, opened 2019-06-09
Add button to download survey as csv

JonasBak:
I could not get this to work with a simple <a> tag as the endpoint requires an auth token in the header, so you have to first click a button to make the request with the header and get a blob, and then press a link to actually save the file. It should work just fine, but let me know if you have a better solution.

### DIFF
--- a/app/routes/surveys/SubmissionsRoute.js
+++ b/app/routes/surveys/SubmissionsRoute.js
@@ -1,3 +1,4 @@
+import 'isomorphic-fetch';
 import { connect } from 'react-redux';
 import prepare from 'app/utils/prepare';
 import {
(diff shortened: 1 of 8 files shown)


app/routes/surveys/SubmissionsRoute.js
```diff
@@ -16,11 +21,14 @@ import { push } from 'react-router-redux';
 import loadingIndicator from 'app/utils/loadingIndicator';
 import { LoginPage } from 'app/components/LoginForm';
 import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
+import config from 'app/config';
+
+const getCsvUrl = surveyId => `${config.serverUrl}/surveys/${surveyId}/csv/`;
```
SmithPeder: Considered putting this in the `utils` folder?

app/routes/surveys/components/AdminSideBar.js
```diff
@@ -15,16 +15,19 @@ type Props = {
...
   shareSurvey: number => Promise<*>,
-  hideSurvey: number => Promise<*>
+  hideSurvey: number => Promise<*>,
+  exportSurvey?: number => Promise<*>
 };
 
 type State = {
-  copied: boolean
+  copied: boolean,
+  exported: ?{ url: string, filename: string }
```
SmithPeder: Maybe a shitty name for an object that can be `undef` 🥇🥇??  Sound like a bool 🙃. Maybe call it something like **generatedCVC**? Makes more sense when it's used like `generatedCVC.filename` and `generatedCVC.url`.

PR #1576, opened 2019-08-11
༼ つ ◕_◕ ༽つ

SmithPeder:
### lodash

Fix lodash-es security alert by bumping to `4.17.15`

<hr>

### validator

Resolves https://github.com/webkom/lego-webapp/pull/1569

This is unused?? I can't find one import using this, and everything seems to be working good without it. Removed the dependency.

<hr>

### react-swipeable

Resolves https://github.com/webkom/lego-webapp/pull/1531

Bump to v5. Some small migrations with prop changes. The latest version is 5.4.0, but anything above 5.0.0 created some unwanted behavior(aka no behavior at all, cant swipe). I could not figure out the reason behind this, as all the breaking changes stated in the changelog was from v4 to v5, and not from 5.0.0 to 5.1.0... 😢

### DIFF
--- a/app/routes/events/utils.js
+++ b/app/routes/events/utils.js
@@ -8,7 +8,7 @@ export type eventTypes = $Values<typeof EVENT_CONSTANTS>;
 
 // Current eventTypes
 export const EVENT_CONSTANTS: eventTypes = {
-  company_presentation: 'Bedprespresentasjon',
+  company_presentation: 'Bedriftspresentasjon',
   lunch_presentation: 'Lunsjpresentasjon',
   alternative_presentation: 'Alternativt bedpres',
   course: 'Kurs',
--- a/app/routes/photos/components/GalleryPictureModal.js
+++ b/app/routes/photos/components/GalleryPictureModal.js
@@ -13,7 +13,7 @@ import { Link } from 'react-router';
 import CommentView from 'app/components/Comments/CommentView';
 import Modal from 'app/components/Modal';
 import styles from './GalleryPictureModal.css';
-import Swipeable from 'react-swipeable';
+import { Swipeable, RIGHT, LEFT } from 'react-swipeable';
 import type { EntityID } from 'app/types';
 import type { ID } from 'app/models';
 import Button from 'app/components/Button';
@@ -181,12 +181,9 @@ export default class GalleryPictureModal extends Component<Props, State> {
     }
   };
 
-  handleSwipeRight = () => {
-    this.previousGalleryPicture();
-  };
-
-  handleSwipeLeft = () => {
-    this.nextGalleryPicture();
+  handleSwipe = ({ dir }: { dir: RIGHT | LEFT }) => {
+    dir === RIGHT && this.previousGalleryPicture();
+    dir === LEFT && this.nextGalleryPicture();
   };
 
   render() {
@@ -204,10 +201,7 @@ export default class GalleryPictureModal extends Component<Props, State> {
     const { showMore } = this.state;
 
     return (
-      <Swipeable
-        onSwipingLeft={this.handleSwipeLeft}
-        onSwipingRight={this.handleSwipeRight}
-      >
+      <Swipeable onSwiping={this.handleSwipe}>
         <Modal
           onHide={() => push(`/photos/${gallery.id}`)}
           backdropClassName={styles.backdrop}
--- a/package.json
+++ b/package.json
@@ -97,7 +97,6 @@
     "serialize-javascript": "^1.4.0",
     "slugify": "^1.3.4",
     "source-map-support": "^0.5.8",
-    "validator": "^9.2.0",
     "victory": "^0.25.7",
     "websocket.js": "^0.1.7"
   },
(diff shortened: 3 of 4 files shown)


app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -204,10 +201,7 @@ export default class GalleryPictureModal extends Component<Props, State> {
     const { showMore } = this.state;
 
     return (
-      <Swipeable
-        onSwipingLeft={this.handleSwipeLeft}
-        onSwipingRight={this.handleSwipeRight}
-      >
+      <Swipeable onSwiping={event => this.handleSwipe(event)}>
```
odinuge: `onSwiping={this.handleSwipe}`?

app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -181,12 +181,9 @@ export default class GalleryPictureModal extends Component<Props, State> {
...
   };
 
-  handleSwipeRight = () => {
-    this.previousGalleryPicture();
-  };
-
-  handleSwipeLeft = () => {
-    this.nextGalleryPicture();
+  handleSwipe = ({ dir }: RIGHT | LEFT) => {
+    dir == RIGHT && this.previousGalleryPicture();
```
odinuge: ===

app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -181,12 +181,9 @@ export default class GalleryPictureModal extends Component<Props, State> {
     }
   };
 
-  handleSwipeRight = () => {
-    this.previousGalleryPicture();
-  };
-
-  handleSwipeLeft = () => {
-    this.nextGalleryPicture();
+  handleSwipe = ({ dir }: RIGHT | LEFT) => {
```
odinuge: Does this make sense? `dir` is inside an object og type ` RIGHT | LEFT`, and then compared to `RIGHT` or `LEFT` 🤔

app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -13,7 +13,7 @@ import { Link } from 'react-router';
 import CommentView from 'app/components/Comments/CommentView';
 import Modal from 'app/components/Modal';
 import styles from './GalleryPictureModal.css';
-import Swipeable from 'react-swipeable';
+import { Swipeable, RIGHT, LEFT } from 'react-swipeable';
```
odinuge: Should these be imported as types, or?

app/routes/photos/components/GalleryPictureModal.js
SmithPeder: Ahh, that seems to do the same 🔥 I'll go with that

app/routes/photos/components/GalleryPictureModal.js
SmithPeder: 👍

app/routes/photos/components/GalleryPictureModal.js
SmithPeder: Is there a way to get these types in flow?
image

or should one just `import type {RIGHT, LEFT} from 'react-swipeable'`?
image

SmithPeder:
I blame the migration guide for half my mistakes 🙃 🔥 
image

app/routes/photos/components/GalleryPictureModal.js
SmithPeder: Yeah, that's wrong 🙃

SmithPeder:
And the output from `yarn audit` is absolutely terrifying
```
16075 vulnerabilities found - Packages audited: 885242
Severity: 14 Low | 10 Moderate | 16050 High | 1 Critical
✨  Done in 13.38s. 
```

app/routes/photos/components/GalleryPictureModal.js
odinuge: Ahh. Then it makes some more sense! Replacing `RIGHT | LEFT` with `EventData` (if the flow type exists), it should be ok. (or even `{ dir } : { dir: RIGHT | LEFT | UP | DOWN }`)

app/routes/photos/components/GalleryPictureModal.js
odinuge: The current import looks correct, since the values are ja exports, not types :+1:

app/routes/photos/components/GalleryPictureModal.js
SmithPeder: The flow type didn't exist, so I went with the latter approach.

app/routes/events/utils.js
```diff
@@ -8,7 +8,7 @@ export type eventTypes = $Values<typeof EVENT_CONSTANTS>;
 
 // Current eventTypes
 export const EVENT_CONSTANTS: eventTypes = {
-  company_presentation: 'Bedprespresentasjon',
+  company_presentation: 'Bedriftspresentasjon',
```
odinuge: Sneaky 😍

PR #1578, opened 2019-08-17
Remove option

SmithPeder:
This has been fully agreed on by both sides.


fridasolheim:
Approved

fridasolheim:
> This has been fully agreed on by both sides.

Approved

mrfarstad:
Maybe we should remove the option from the api as well?

SmithPeder:
Hmm, yeah I think we could do that as well.

PR #1583, opened 2019-08-24
Replace SoundCloud token with a currently working token

kotlarz:
Replace the token and re-enable the listing of podcasts.

This closes https://github.com/webkom/lego/issues/1653


kotlarz:
Will be fixed by https://github.com/webkom/lego-webapp/pull/1589

PR #1584, opened 2019-08-25
Change unregistration deadline to 24 hours before event

regineru:
Want to have the unregistration deadline 24 hours before event instead of 12


JonasBak:
This actually just sets the unregistration to 00:00 the day after the event, as the time is hour of day, not time relative to the event:
image

PR #1589, opened 2019-09-02
Add Reactions to Articles, and rename comment_target to content_target

kotlarz:
**https://github.com/webkom/lego/pull/1659 needs to be merged first.**

This closes https://github.com/webkom/lego/issues/1448 and https://github.com/webkom/lego/issues/1647 together with https://github.com/webkom/lego/pull/1659.
And closes webkom/lego#1653, so https://github.com/webkom/lego-webapp/pull/1583 can be deleted.

**NB!** The following browsers need to be tested before merging:
- [x] Test with Android Chrome
- [x] Test with Android Firefox
- [ ] Test with iOS Safari
- [x] Test with Chrome on Linux
- [x] Test with Firefox on Linux
- [x] Test with Chrome on MacOS
- [x] Test with Firefox on MacOS
- [x] Test with Safari on MacOS

This adds a `Reactions` component, which consists of `Reaction` components and a `ReactionPicker`, as seen below.

## Reactions
A typical `Article` with reactions looks like:
Selection_260

## Add/Remove reactions
A user can add/remove reactions directly:

## ReactionPicker
The `ReactionPicker` consits of different categories, each with their own list of emojis.
Selection_261

## ReactionPicker search
The `ReactionPicker` has a search input field that searches through all the emojis. The search looks through the `shortCode` of each emoji (i.e. `:grin:`), but also looks through the keywords for each emoji (i.e. the `:grin:` emoji contains the following keywords: `["face", "happy", "smile", "joy", "kawaii"]`.
image

## Reaction limit
There is also an limit of how many reactions each user can have per `content_target` (i.e. an Article). The current limit is *10* reactions.

## Article with 0 reactions
An `Article` with no reactions looks like this:
Selection_263


kotlarz:
I have added an `NotoColorEmoji` font and used it in a new font-family, as Chrome (on Linux) didn't initially display Unicode emojis correctly.

Here was Chrome vs Firefox (on Linux) before adding the font:
image (2)

Chrome (on Linux) currently looks like:
image

app/components/Reactions/ReactionPicker.js
```diff
@@ -0,0 +1,107 @@
...
+  isLoading,
+  emojis,
+  addReaction,
+  deleteReaction,
+  contentTarget
+}: Props) => {
+  const [activeCategory, setActiveCategory] = useState('people');
+  const [searchString, setSearchString] = useState(null);
+
+  const categories = useMemo(() => {
```
odinuge: :heart_eyes:

app/components/Emoji/index.js
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+import styles from './index.css';
+
+type Props = {
+  unicodeString: string,
+  shortCode?: string,
+  size?: string
+};
+
+//       <g-emoji fallback-src="t-rex.png" alias="T-Rex">{unicodeString}</g-emoji>
```
kotlarz: Delete

app/components/Reactions/ReactionPickerCategory.js
```diff
@@ -0,0 +1,60 @@
...
+    default:
+      return name.charAt(0).toUpperCase();
+  }
+};
+
+class ReactionPickerCategory extends React.Component<Props> {
+  render() {
+    const { isActive, name, onCategoryClick } = this.props;
+
+    let containerClasses = [styles.reactionPickerCategoryContainer];
```
odinuge: `const`a

app/components/Reactions/Readme.md
```diff
@@ -0,0 +1,15 @@
+Basic tag list:
```
kotlarz: Delete file

app/components/Reactions/index.js
```diff
@@ -0,0 +1,131 @@
...
+  state = {
+    hovered: false,
+    reactionPickerOpen: false,
+    fetchedEmojis: false
+  };
+
+  constructor() {
+    super();
+
+    this.handleOutsideClick = this.handleOutsideClick.bind(this);
```
odinuge: Maybe remove this? :zap:

app/routes/podcasts/components/Podcast.js
```diff
@@ -33,7 +33,7 @@ class Podcast extends Component<Props, State> {
   };
 
   render() {
-    const CLIENT_ID = 'E8IqLGTYxHll6SyaM7LKrMzKveWkcrjg';
+    const CLIENT_ID = 'WOwjzZGoYrSrOOAILojG4miJtS4pZjSg';
```
kotlarz: This came in due to https://github.com/webkom/lego-webapp/pull/1583

I'm removing it as it actually works, we can close the PR above instead.

app/components/Reactions/ReactionPickerCategory.js
```diff
@@ -0,0 +1,60 @@
...
+    case 'travel_and_places':
+      return '✈';
+    case 'objects':
+      return '💡';
+    case 'symbols':
+      return '❤';
+    case 'flags':
+      return '🇵🇱 ';
+    default:
+      return name.charAt(0).toUpperCase();
```
LudvigHz: What do you think about using the categories to set the icons? f.ex. the first emoji in the category, or maybe modifying the list of emojis in backend to set a  __main__ emoji in a category there?

app/reducers/articles.js
```diff
@@ -18,6 +20,8 @@ export type ArticleEntity = {
   startTime: string,
   text: string,
   tags: Array<string>,
+  reactionsGrouped: Array<Object>,
```
LudvigHz: Pls fix, no Object

app/reducers/emojis.js
```diff
@@ -0,0 +1,39 @@
...
+import { createSelector } from 'reselect';
+
+export type EmojiEntity = {
+  shortCode: string,
+  keywords: Array<string>,
+  unicodeString: string,
+  fitzpatrickScale: boolean,
+  category: string,
+  hasReacted?: boolean,
+  reactionId?: number
```
LudvigHz: type ID

app/components/Reactions/ReactionPicker.js
```diff
@@ -0,0 +1,107 @@
...
+          <ReactionPickerHeader
+            activeCategory={activeCategory}
+            categories={Object.keys(categories)}
+            onCategoryClick={onCategoryClick}
+          />
+          <ReactionPickerContent
+            emojis={
+              categories[activeCategory]
+                ? categories[activeCategory].emojis
+                : []
```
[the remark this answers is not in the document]
kotlarz: This is due to categories not being set on the initial render.
In addition all categories will have emojis, as the `categories` variable is mapped by looking through the categories in `emojis`.

app/components/Reactions/ReactionPickerCategory.js
kotlarz: This is a good idea, but this will be a #TODO for a future PR, where we add a new model for emojis (i.e. `EmojiCategory`) where we can have a `emoji` field.

SmithPeder:
When posting a reaction i just get 401, and 403 when checking the API? What kind of permissions is this?
`[django.channels.server] HTTP GET /api/v1/reactions/ 401 [0.09, 127.0.0.1:50002]`

image

app/actions/ReactionActions.js
```diff
@@ -0,0 +1,70 @@
...
+        meta: {
+          emoji,
+          contentTarget,
+          unicodeString
+        },
+        schema: reactionSchema
+      })
+    ).catch(action => {
+      const status = action.payload.response.status;
+      let errorMessage = 'Reaksjon feilen';
```
SmithPeder: Spelling

app/components/Reactions/ReactionPicker.js
```diff
@@ -0,0 +1,110 @@
...
+    });
+    return mappedCategories;
+  }, [emojis]);
+
+  const searchResults = useMemo(() => {
+    if (searchString === null || searchString === '') {
+      return null;
+    }
+    return emojis.filter(emoji => {
+      const idMatches = emoji.shortCode.toLowerCase().includes(searchString);
```
SmithPeder: fuzzy

kotlarz: I've been thinking a bit about this an believe I have a proper solution:

`searchString`: `cat`

1. Look up exact matches for the `shortCode`:
  - `:cat:`
2. Look up word matches (i.e. look up if the `shortCode` starts/ends with the searchString and such) :
  - `:cat_crying:`
  - `:mad_cat:`
  - `:hungry_cat_sad:`
  - `:cat2:`
3. Perform a fuzzy search for the `shortCode`:
  - `:capitcal:`
  - `:cactus:`
  - `:bar_char:`

Append each result to a list, and pop the results from the original emoji list for each result.

I have not decided what to do with `keywords` yet, maybe just look them up if there are exact matches for the keywords, and insert this a separate step before the fuzzy search.

kotlarz:
@odinuge @SmithPeder @LudvigHz @orhanhenrik 
I'd like some feedback on the latest [commit](https://github.com/webkom/lego-webapp/pull/1589/commits/e6f3c5ceaa97e827f45f39a6bc87f91cfeb7bc1d).
If approved the only thing left to test is Safari on iOS.

app/components/Reactions/ReactionPicker.js
```diff
@@ -21,6 +22,120 @@ type Props = {
...
+   *   - ":taxi:" (match = "vehicle", keywords = ["uber", "vehicle", "cars", "transportation"])
+   *   - ":tractor:" (match = "vehicle", keywords = ["vehicle", "car", "farming", "argiculture"])
+   *   - ":oncoming_automobile:" (match = "vehicle", keywords = ["car", "vehicle", "transportation"])
+   */
+  currentEmojis = currentEmojis.filter(emoji => {
+    if (
+      emoji.keywords.find(keyword => keyword.toLowerCase() == searchString) !==
+      undefined
+    ) {
+      matchingEmojis.push(emoji);
```
LudvigHz: return false here

app/components/Reactions/ReactionPicker.js
```diff
@@ -21,6 +22,120 @@ type Props = {
...
+  /*
+   * STEP 4
+   * Perform a fuzzy search at the end for the remaining emojis.
+   *
+   * Examples (searchString = "car"):
+   *   - ":woman_cart_wheeling:" (match = ":woman_<car>t_wheeling:")
+   *   - ":costa_rica:" (match = ":<c>ost<a>_<r>ica")
+   *   - ":christmas_tree:" (match = ":<c>hristm<a>s_t<r>ee")
+   */
+  var results = fuzzy.filter(searchString, currentEmojis, {
```
LudvigHz: aaah, `var`. This can be `const` or no?

app/components/Reactions/ReactionPicker.js
```diff
@@ -21,6 +22,120 @@ type Props = {
...
+    pre: '<',
+    post: '>',
+    extract: emoji => {
+      return emoji.shortCode;
+    }
+  });
+
+  currentEmojis = results.forEach(result => {
+    matchingEmojis.push(result.original);
+  });
```
LudvigHz: It seems a bit unnecessary to modify `currentEmojis`, can't you instead do something like:
```js
return [...results.map(result => result.original), ...matchingEmojis]
```
I'm not sure what the performance difference is though.

PR #1606, opened 2019-09-10
Add possibility for additional email fields in email lists

erlingfn:
Make it possible to add emails for people outside of abakus to email lists manually

### DIFF
--- a/app/routes/admin/email/components/EmailListEditor.js
+++ b/app/routes/admin/email/components/EmailListEditor.js
@@ -100,7 +100,7 @@ export default legoForm({
       groupRoles: (data.groupRoles || []).map(groupRole => groupRole.value),
       groups: (data.groups || []).map(group => group.value),
       users: (data.users || []).map(user => user.value),
-      additionalEmails: (data.additionalEmails || []).map(email => email.value)
+      additionalEmails: data.additionalEmails
     }).then(({ payload }) => {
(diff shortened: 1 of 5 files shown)


mrfarstad:
Nice work! Is the dependency upgrade related to this PR? Maybe move the upgrading to another PR. That way it would be easier to revert the PR if the upgrading causes issues. 😁

erlingfn:
Thanks for the feedback, I believe I have made the changes you requested now!

app/routes/admin/email/components/EmailListEditor.js
```diff
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import Button from 'app/components/Button';
-import { createValidator, required } from 'app/utils/validation';
+import { createValidator, required, EMAIL_REGEX } from 'app/utils/validation';
```
JonasBak: There is a `isEmail` function here you can import instead of the "raw" regex

erlingfn: There is, however I could not find a way to use it to validate an array of emails

odinuge: :+1: Should make some logic for validating arrayfields

odinuge: Out of scope here tho.

PR #1610, opened 2019-09-14
Switch to lego-editor :fire: :fire:

LudvigHz:
Finally, everything works as it should! I have not done super extensive testing, but I have tested with a couple articles and events, and everything was smooth.

image


app/components/Form/Field.js
```diff
@@ -128,6 +131,33 @@ export function createField(Component: ComponentType<*>) {
         )}
         {hasWarning && <RenderWarningMessage warning={meta.warning} />}
       </div>
+    ) : (
```
odinuge: Duplicate code much, or?

app/components/Form/Field.js
```diff
@@ -76,7 +76,10 @@ export type FieldProps = {
  *
  * http://redux-form.com/6.0.5/docs/api/Field.md/
  */
-export function createField(Component: ComponentType<*>) {
+export function createField(
+  Component: ComponentType<*>,
+  useLabel?: boolean = true
```
odinuge: I prefer an object over a simple boolean. Also, it would be nice with some docs about what `useLabel` does.

app/components/Form/EditorField.js
```diff
@@ -10,31 +16,47 @@ type Props = {
...
-}
+};
+
+const EditorField = connect(
+  null,
+  mapDispatchToProps
+)(EditorFieldComponent);
+
+// $FlowFixMe
+EditorField.Field = connect(
```
odinuge: `compose` would make this a bit cleaner

app/components/Form/EditorField.js
```diff
@@ -10,31 +16,47 @@ type Props = {
...
+  null,
+  mapDispatchToProps
```
odinuge: Also, this seems a bit fishy. Connecting inside a random component.

app/components/Form/EditorField.js
```diff
@@ -10,31 +16,47 @@ type Props = {
...
+    uploadFile: async file => {
+      const response = await dispatch(uploadFile({ file, isPublic: true }));
+      return { fileKey: response.meta.fileKey };
+    }
+  };
+};
 
-function EditorField({ className, name, ...props }: Props) {
-  if (!__CLIENT__) {
-    throw new NoSSRError('Cannot SSR editor');
```
odinuge: Are you sure this is ok to remove?

app/components/Comments/__tests__/CommentTree.spec.js
```diff
@@ -6,6 +6,9 @@ import comments from './fixtures/comments';
 import { generateTreeStructure } from '../../../utils';
 
 describe('<CommentTree />', () => {
+  beforeAll(() => {
+    window.getSelection = () => {};
```
odinuge: Make a comment here so folks from the "gemene hop" can understand what this is.

Also, does it make sense to have it here?

app/components/DisplayContent/index.js
```diff
@@ -1,7 +1,11 @@
 //@flow
 
 import React from 'react';
-import Editor from 'app/components/Editor';
+import Editor from '@webkom/lego-editor';
+import '@webkom/lego-editor/dist/Editor.css';
```
odinuge: Is there a reason to include these twice, and not just directly in `index.js`?

odinuge:
Adding WIP to this since it hasn't been touched in two weeks, and no one has answered my questsions and looked at the review :cry: 

Also conflicting file...

app/components/Form/Field.js
LudvigHz: agreed!

app/components/Form/EditorField.js
LudvigHz: It's fishy yes, but I'm pretty sure its the right way to do it, as otherwise you would have to connect the route, which I think is just more confusing, as you would have to do this everywhere the editor is used and pass the props through a lot of components, many of which are abstract or HOC.

app/components/DisplayContent/index.js
LudvigHz: I just think it makes more sense to have them beside the component.

app/components/Form/EditorField.js
LudvigHz: I put it back just in case now. You may be right in that it will throw errors still...

LudvigHz:
Since this PR opened, I have updated lego-editor to version 1.1.0, see https://github.com/webkom/lego-editor/releases for more info. This release fixes all of the small things holding this PR back, so I'm confident we're ready to get this in now!

I appreciate if people will test this out a bit, so we can find any possible bugs before getting this in.

erlingfn:
The editor looks good! Some parts are a little rough around the edges, yet it is in my opinion superior to the current editor. Therefore I suggest me merge it, as it does not need to be perfect, only better than the current one. We should, however, inform our users that we make this change and encourage them to send feedback.

LudvigHz:
image
Aaand, just almost a year later, I finally made the cypress tests work (and some other stufff) :shipit: :1st_place_medal:

erlingfn:
Does not look like H4's are working as intended. They lose their formatting upon saving the form, and disappear when you try to edit the field again. Other than that the editor works well!

LudvigHz:
> Does not look like H4's are working as intended. They lose their formatting upon saving the form, and disappear when you try to edit the field again. Other than that the editor works well!

The H4 issue seems to be an issue in https://github.com/webkom/lego and not here or the editor.

The editor seems stable enough to merge though, so I think we should get it in and do some more testing in staging to catch any possible major issues there.

PR #1621, opened 2019-09-18
Fix checkmark in bdb

theodorsm:
Fix oversized checkmark in bdb, and redesign bdb dropdown.
Fixes webkom/lego/issues/1678

Before: 
Screenshot from 2019-09-18 20-35-06

After: 
Screenshot from 2019-09-18 20-59-44

### DIFF
--- a/app/routes/bdb/components/SemesterStatusContent.js
+++ b/app/routes/bdb/components/SemesterStatusContent.js
@@ -53,17 +53,19 @@ export default class SemesterStatusContent extends Component<Props, State> {
     const dropDownItems = (
       <Dropdown.List>
         {statusCodes.map((statusString, j) => (
-          <Button flat key={j} onClick={e => editFunction(statusString)}>
-            <Dropdown.ListItem className={styles.dropDownItem}>
+          <Dropdown.ListItem key={j} className={styles.dropDownItem}>
+            <Button flat onClick={e => editFunction(statusString)}>
               <div>
                 {semesterStatus.contactedStatus.indexOf(statusString) !== -1 ? (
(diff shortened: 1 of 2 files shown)


finnss:
Good work! I've just got a few comments.

First, if we're fixing the design on these anyway, would you mind fixing the margin around the clickable area?

Skjermbilde 2019-09-18 kl  20 27 16

The gray area of the second button is what I highlighted (the clickable part), and the white area around it (and all the buttons) is kinda ugly.

Second, clicking one of these buttons makes the page jump to the top. We don't want it to work like that, preferably. Could you take a quick look to see if that's easily fixable? Let's finish the styling part in this PR and merge it, then look at the jumping behaviour in a new PR.

theodorsm:
> Good work! I've just got a few comments.
> 
> First, if we're fixing the design on these anyway, would you mind fixing the margin around the clickable area?
> 
> Skjermbilde 2019-09-18 kl 20 27 16
> 
> The gray area of the second button is what I highlighted (the clickable part), and the white area around it (and all the buttons) is kinda ugly.
> 
> Second, clicking one of these buttons makes the page jump to the top. We don't want it to work like that, preferably. Could you take a quick look to see if that's easily fixable? Let's finish the styling part in this PR and merge it, then look at the jumping behaviour in a new PR.

Fixed the styling issue, see the updated screenshot. The jumping behavior does not seem easily fixable to me.

finnss:
Good stuff 👍 The jumping behaviour is pretty low prio, you can just keep it in mind if you want something else to fiddle with at some point. Merging this

PR #1625, opened 2019-09-18
Remove autofocus in menu for mobile

ingraso:


app/components/Header/index.js
```diff
@@ -262,6 +262,7 @@ class Header extends Component<Props, State> {
             onHide={this.props.toggleSearch}
             backdropClassName={styles.backdrop}
             backdrop
+            autoFocus={false}
```
kotlarz: This one is funny, the `Modal` component is a part of `react-overlays`, which states:

> When true The modal will automatically shift focus to itself when it opens, and replace it to the last focused element when it closes. This also works correctly with any Modal children that have the autoFocus prop.
> **Generally this should never be set to false** as it makes the Modal less accessible to assistive technologies, like screen readers.
https://react-bootstrap.github.io/react-overlays/#modals

However, the page will scroll to the bottom/footer if `autoFocus is set to `true` and you open the modal on a mobile devices 🙃

PR #1629, opened 2019-09-25
Remove atom reference

odinuge:


odinuge:
Nice demo of the cypress integrations :tada:

PR #1632, opened 2019-09-25
Add cypress tests for company interest create, delete and edit

mrfarstad:


odinuge:
:tada:

orhanhenrik:
We should probably also test this page without being signed in, since most users of this page will not have an abakus user

mrfarstad:
> We should probably also test this page without being signed in, since most users of this page will not have an abakus user

Yep, we definitely should! :)

However, the tests seem to be non-deterministic, so I think I will close this PR

app/routes/companyInterest/components/CompanyInterestPage.js
```diff
@@ -363,5 +363,6 @@ const validate = createValidator({
 export default reduxForm({
   form: 'CompanyInterestForm',
   validate,
-  enableReinitialize: true
+  enableReinitialize: true,
+  keepDirtyOnReinitialize: true
```
odinuge: Dirty stuff :fire:

PR #1635, opened 2019-09-26
[WIP] Initial attempt at adding GitHub Actions

kotlarz:


.github/workflows/lego.yml
```diff
@@ -0,0 +1,142 @@
...
+
+    - name: Set up Node 11.x
+      uses: actions/setup-node@v1
+      with:
+        node-version: '11.x'
+
+    - name: Install dependencies
+      run: yarn --forzen-lockfile
+
+    - name: Setup MinIO
```
orhanhenrik: Can't minio run as a service like postgres and redis?

odinuge: Looks like it was for debugging purposes. Should be ok to move it into a service now.

Is it possible to look at the logs for the services? :thinking: Can't find a way to do it right now. :/

kotlarz: I'm trying different solutions. I need to run certain commands within the container, and it's not possible to send/run commands there. That's why you have to create an Action for that specific service (or a new image that run the commands).

In addition, there are no caching options (i.e. for node_modules) currently. The only options is to upload and download artifacts, but I tested this and it was too slow for node_modules (50k+ files!). Cache is supposed to be implemented in mid-November, https://github.community/t5/GitHub-Actions/Caching-files-between-GitHub-Action-executions/m-p/30974/highlight/true#M630.

I am still testing this and different ways to implement Actions, the documentation isn't always that detailed.

Regarding logs for services, I haven't found a way to look at them yet... Will test some more when I get some time.

kotlarz:
Going to wait for GitHub to add cache for the workflow. ETA. mid-November.

PR #1637, opened 2019-09-30
Add reducer tests

orhanhenrik:


app/reducers/__tests__/events.spec.js
```diff
@@ -0,0 +1,571 @@
...
+        }
+      });
+    });
+
+    it('Event.CLEAR', () => {
+      const prevState = baseState;
+      const action = {
+        type: Event.CLEAR
+      };
+      // TODO: Is this expected behavior?
```
mrfarstad: This should be correct, so remove this comment 💯

PR #1639, opened 2019-09-30
Frontpage Quotes

finnss:
Aint she a beauty

Screenshot 2019-10-28 at 18 43 46


app/reducers/index.js
```diff
@@ -167,7 +169,11 @@ export const galleryPictureSchema = new schema.Entity('galleryPictures', {
 export const gallerySchema = new schema.Entity('galleries');
 
 export const quoteSchema = new schema.Entity('quotes', {
-  comments: [commentSchema]
+  reactions: [reactionSchema]
```
finnss: do I need this?

app/components/RandomQuote/RandomQuote.css
```diff
@@ -1,29 +1,27 @@
-.heading {
-  position: relative;
-  display: flex;
-  align-items: center;
-  justify-content: center;
+/*
+.quoteSource,
```
orhanhenrik: Remove comment

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+    contentTarget: string
+  }) => Promise<*>,
+  deleteReaction: ({ reactionId: ID, contentTarget: string }) => Promise<*>,
+  fetchEmojis: () => Promise<*>,
+  fetchingEmojis: boolean,
+  emojis: Array<EmojiEntity>,
+  currentQuote: QuoteEntity
+};
+
+class RandomQuote extends Component<Props> {
```
orhanhenrik: This can be a functional component instead

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+      emojis,
+      fetchEmojis,
+      fetchingEmojis,
+      currentQuote
+    } = this.props;
+
+    let mappedEmojis = [];
+    if (!fetchingEmojis) {
+      mappedEmojis = emojis.map(emoji => {
+        const foundReaction = !currentQuote.reactionsGrouped
```
orhanhenrik: This would be a bit more readable like this
```
const foundReaction = currentQuote.reactionsGrouped && currentQuote.reactionsGrouped.find(...)
```

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+          ? undefined
+          : currentQuote.reactionsGrouped.find(
+              reaction =>
+                emoji.shortCode == reaction.emoji && reaction.hasReacted
+            );
+        if (foundReaction !== undefined) {
+          emoji.hasReacted = true;
```
orhanhenrik: This mutates the `emoji` object which comes from props. The mutation could cause ugly side effects or bugs if other code is using the same object. You can either copy the object in the beginning of this function, e.g. `const newEmoji = {...emoji}` or you can return from this statement, e.g. `return {...emoji, hasReacted: true, reactionId:...}`

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+          emoji.reactionId = foundReaction.reactionId;
+        } else {
+          emoji.hasReacted = false;
+          emoji.reactionId = -1;
+        }
+        return emoji;
+      });
+    }
+
+    return loggedIn ? (
```
orhanhenrik: For readability I would instead write something like this:
```
if (!loggedIn) return <div>Logg inn pls</div>

return <div>logged in content</div>
```
Also you could move this if-statement to the beginning of the function so you know from the beginning that the user will be logged in

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+      <div className={className ? className : ''}>
+        <Flex row justifyContent={'space-between'} alignItems={'flex-start'}>
```
orhanhenrik: You could write `justifyContent="space-between"` instead of `justifyContent={'space-between'}`, but no big deal

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+            <div className={styles.quoteText}>{currentQuote.text}</div>
+            <div className={styles.quoteSource}>-{currentQuote.source}</div>
+          </Flex>
+
+          <Flex
+            column
+            justifyContent={'space-between'}
+            className={styles.actions}
+          >
+            <Button flat onClick={() => this.props.fetchRandomQuote()}>
```
orhanhenrik: Is there a loading indicator when you click this?

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+              <i className="fa fa-refresh" />
+            </Button>
+            <Link to={'/quotes/add'} className={styles.add}>
```
orhanhenrik: Nice way to encourage adding quotes 👍

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+        <div className={styles.quoteReactions}>
+          <Reactions
+            emojis={mappedEmojis}
+            fetchEmojis={fetchEmojis}
+            fetchingEmojis={fetchingEmojis}
+            addReaction={addReaction}
+            deleteReaction={deleteReaction}
+            contentTarget={currentQuote.contentTarget}
+          >
+            {currentQuote.reactionsGrouped.map(reaction => {
```
orhanhenrik: On line 45 your code says `currentQuote.reactionsGrouped` might be undefined. Is it safe to map over it here?

app/components/Footer/index.js
```diff
@@ -77,7 +71,10 @@ const Footer = (props: Props) => (
             <p>7491 Trondheim</p>
           </div>
         </Flex>
-        <Image className={styles.cooperator} src={netcompany} />
+        <Image
+          className={cx(styles.cooperator, hiddenOnDesktop)}
```
SmithPeder: What's the thought here? The other logo is inside a `hiddenOnMobile` div and this image gets the `hiddenOnDesktop` class. Doesn't that mean the image is shown neither on phone or on desktop?

app/components/RandomQuote/RandomQuote.css
SmithPeder: +1

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,117 @@
...
+
+import React, { Component } from 'react';
+import styles from './RandomQuote.css';
+import Button from '../Button';
+import type { QuoteEntity } from 'app/reducers/quotes';
+import Reaction from 'app/components/Reactions/Reaction';
+import Reactions from 'app/components/Reactions';
+import type { EmojiEntity } from 'app/reducers/emojis';
+import type { ID } from 'app/models';
+import { Link } from 'react-router';
```
SmithPeder: Use the LEGO Link component for this?

app/components/RandomQuote/RandomQuote.js
SmithPeder: This ^, and similar at the other place this is done (QuotesRoute).

app/components/RandomQuote/RandomQuote.js
SmithPeder: The same could be done with `alignItems`
 And `row` does nothing there, it's not a valid prop in `<Flex />` as `row` is the default setting.

app/components/RandomQuote/RandomQuote.js
finnss: I totally agree – not that it's really an excuse, but I just copied this code from Article Reactions. It feels like something that should be placed in reducers or something, and after a short discussion with @kotlarz I decided to just copy what he had for now instead of optimizing something we were planning to replace anyway. But it's exactly the kind of thing that noone ever gets around to fixing so I guess we might as well do it in this PR. 

In order to remove the need for this kind of code snippet every time someone wants to use reactions, where would you put this logic?

app/components/RandomQuote/RandomQuote.js
finnss: Not yet, good idea

app/components/RandomQuote/RandomQuote.js
finnss: This is another side effect of the code replication thing. I think ideally, reactionsGrouped should always exist at this point in the code and be an empty array worst case. We should review this once a better solution to the whole reaction-emoji mapping thing has been written

app/components/RandomQuote/RandomQuote.js
SmithPeder: If it has to run every time somebody adds/removes a reaction I also feel like it can be placed in some reducer or action.

orhanhenrik: Yeah, putting it in reducer code works fine. 
You could also create a new `<LegoReactions>` (naming is hard, don't judge me) component that does everything between line 42-60 and 85-108. Then you can just pass the `reactionsGrouped` object into that component and the component does the mapping and rendering logic.

I think this makes the most sense if the `reactionsGrouped` looks the same across different entities (articles, quotes, etc.). It's probably also a bit simpler than trying to change the reducers a lot, and also a bit more flexible I guess.

app/reducers/index.js
```diff
@@ -167,9 +169,11 @@ export const galleryPictureSchema = new schema.Entity('galleryPictures', {
...
 });
 
+export const randomQuoteSchema = new schema.Entity('randomQuote');
```
mrfarstad: Adding this as a separate schema entity seems a little overkill. Maybe just add `randomQuoteSchema` as a key in the Quote entity reducer instead? 😄

app/components/Footer/index.js
erlingfn: I believe the thought is to remove the logo in the bottom and instead have it in place of the random quoute on desktop. Due to the location of the logo being different, the components have to be in different places.

app/components/RandomQuote/RandomQuote.js
finnss: ?

app/reducers/index.js
finnss: Agreed, done 👍

app/reducers/index.js
erlingfn: I have not done extensive testing, yet removing it does not seem to do anything. I believe this is due to there not being a reaction field in the payload, it instead being named reactionsgrouped

app/components/Search/utils.js
```diff
@@ -61,7 +61,7 @@ const LINKS: Array<Link> = [
   },
   {
     key: 'quotes',
-    title: 'Sitater',
+    title: 'AbaQuote',
```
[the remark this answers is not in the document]
finnss: I made a poll on the frontpage of prod abakus.no asking what people think the name should be, let's see!

app/components/RandomQuote/RandomQuote.js
```diff
@@ -0,0 +1,80 @@
...
+    addReaction,
+    deleteReaction,
+    emojis,
+    fetchEmojis,
+    fetchingEmojis,
+    currentQuote
+  } = props;
+
+  if (!loggedIn) {
+    return <div>Logg inn for å se sitater.</div>;
```
LudvigHz: Is this needed if you need to be logged in to see the frontpage anyway?

PR #1643, opened 2019-10-02
Add type safe event constants

odinuge:


app/routes/events/utils.js
```diff
@@ -37,8 +34,8 @@ export const COLOR_CONSTANTS: eventTypes = {
   other: '#111111'
 };
 
-// Returns the color code of an eventType
-export const colorForEvent = (eventType: eventTypes) => {
+// Returns the color code of an EventType
+export const colorForEvent = (eventType: EventType) => {
   return COLOR_CONSTANTS[eventType] || COLOR_CONSTANTS['other'];
```
orhanhenrik: This doesn't really need the `|| COLOR_CONSTANTS['other']` if the input eventType is type checked, right?

odinuge: In theory no (but this is js :trollface:), but when people add new values we don't catch that when doing `.json()`.

orhanhenrik: True..

PR #1647, opened 2019-10-03
Add check for https:// in company website link from backend.

asmundh:
If I've done this correctly it should prevent https:// company links from being prepended with http://. Alternatively import httpCheck from bdb utils?


asmundh:
Also found ResolveLink component. Could maybe use that?

orhanhenrik:
This was a good catch, thank you!

PR #1648, opened 2019-10-04
Fix loading indicator in email routes

JonasBak:
The admin/email route bugs out after a previous bugfix


orhanhenrik:
Will this still show a loading indicator when loading more when scrolling to bottom?

JonasBak:
> Will this still show a loading indicator when loading more when scrolling to bottom?

Yes, and it makes the user list and the "lists-list" behave the same, which it doesn't do in prod now

JonasBak:
> Will this still show a loading indicator when loading more when scrolling to bottom?

This is just the initial loading-indicator for the route

LudvigHz:
Shouldn't the loading indicator in the paginator do this as well? I.e having a loading indicator on the route is unnecessary.

mrfarstad:
LGTM 💯

PR #1662, opened 2019-10-09
Bump css-loader to v1.0.0

erlingfn:
Seems to work, fixes issues in some of it's sub dependencies.
I have no idea why adding post-css in the webpack server config fixes the issue, but somehow it does :man_shrugging:

### DIFF
--- a/config/webpack.server.js
+++ b/config/webpack.server.js
@@ -99,7 +99,26 @@ module.exports = (env, argv) => {
                   loader: 'css-loader',
                   options: {
                     modules: true,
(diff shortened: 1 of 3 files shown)


mrfarstad:
`Seems is relative` - Erling 2019

erlingfn:
> `Seems is relative` - Erling 2019

Worked on my machine :man_shrugging:

erlingfn:
There must be a better way to make this work than adding something that should be unrelated in the webpack server config

PR #1675, opened 2019-10-10
Speedup building in CI

odinuge:
- Stop sending sourcemaps for client to Sentry (they are public)
- Don't build and ship styleguide (we don't use it)
- Only run `docker` on the prod branch (saves hours of CI time)


odinuge:
Quick 8 vs 18 min

PR #1686, opened 2019-10-23
Require all poll options be visible before being able to vote

AdrianAndersen:
image

With this change, users have to click before giving their vote. A blur effect is applied when the user hovers over the poll. 

Solves https://github.com/webkom/lego/issues/1712 (Needs to be marked as closed manually)

### DIFF
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -101,7 +101,7 @@ class Poll extends React.Component<Props, State> {
       .sort((a, b) => b.ratio - a.ratio);
   };
 
-  shuffle = array => {
+  shuffle = (array: Array<OptionEntityRatio>) => {
     let counter = array.length;
 
     while (counter > 0) {
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -21,10 +21,7 @@ type Props = {
 
 type State = {
   truncateOptions: boolean,
-  allOptions: Array<OptionEntityRatio>,
-  optionsToShow: Array<OptionEntityRatio>,
-  shuffledAllOptions: Array<OptionEntityRatio>,
-  shuffledOptionsToShow: Array<OptionEntityRatio>,
+  shuffledOptions: Array<OptionEntityRatio>,
   expanded: boolean
 };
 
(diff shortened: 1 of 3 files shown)


JonasBak:
I suggest blurring by default on mobile, as hovering is a bit tricky :male_detective:

app/components/Poll/index.js
```diff
@@ -163,11 +175,47 @@ class Poll extends React.Component<Props, State> {
...
+              <Flex
+                style={{ display: isHovering ? 'flex' : 'none' }}
+                className={styles.blurContainer}
+                onClick={this.toggleTruncate}
+              >
+                <p
+                  style={{
+                    marginTop: '25px'
+                  }}
+                  className={styles.blurOverlay}
```
LudvigHz: Better if you just use `className` here. If you want to apply custom styling for this element, you can use [classnames](https://github.com/JedWatson/classnames) to apply multiple classnames. See f.ex [here](https://github.com/webkom/lego-webapp/blob/60aae52f29b89f1f7bd23150d952df69b38ac090/app/components/Form/RadioButton.js#L43) for an example.
(In general, try to avoid inline styling)

[PS :mega:](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords)

app/components/Poll/index.js
```diff
@@ -163,11 +175,47 @@ class Poll extends React.Component<Props, State> {
...
+                >
+                  Klikk her for å stemme.
+                </p>
+                <Icon
+                  style={{
+                    marginTop: '40px'
+                  }}
+                  className={styles.blurOverlay}
```
LudvigHz: Same here with the css.

app/components/Poll/index.js
```diff
@@ -163,11 +175,47 @@ class Poll extends React.Component<Props, State> {
...
+
             {options &&
               optionsToShow.map(option => (
                 <Flex style={{ justifyContent: 'center' }} key={option.id}>
                   <Button
+                    style={
+                      !expanded && isHovering
+                        ? { filter: 'blur(3px)', pointerEvents: 'none' }
+                        : {}
+                    }
```
LudvigHz: Aaand here :stuck_out_tongue:

app/components/Poll/index.js
```diff
@@ -163,11 +175,47 @@ class Poll extends React.Component<Props, State> {
           </Flex>
         )}
         {!hasAnswered && (
-          <Flex column className={styles.optionWrapper}>
+          <Flex
+            column
+            className={styles.optionWrapper}
+            onMouseEnter={this.toggleHover}
+            onMouseLeave={this.toggleHover}
+            style={{}}
```
LudvigHz: Hmmmmm :thinking:

app/components/Poll/index.js
```diff
@@ -98,11 +101,25 @@ class Poll extends React.Component<Props, State> {
       .sort((a, b) => b.ratio - a.ratio);
   };
 
+  shuffle = array => {
+    let counter = array.length;
+
+    while (counter > 0) {
```
LudvigHz: Maybe have this as a for loop?

app/components/Poll/index.js
```diff
@@ -98,11 +101,25 @@ class Poll extends React.Component<Props, State> {
...
+      let index = Math.floor(Math.random() * counter);
+      counter--;
+      let temp = array[counter];
+      array[counter] = array[index];
+      array[index] = temp;
```
LudvigHz: `const` pls.

app/components/Poll/index.js
```diff
@@ -98,11 +101,25 @@ class Poll extends React.Component<Props, State> {
...
+
+    return array;
+  };
+
   render() {
     const { poll, handleVote, backgroundLight, details } = this.props;
     const { truncateOptions, optionsToShow, expanded } = this.state;
     const { id, title, description, options, hasAnswered, totalVotes } = poll;
-
+    const shuffledOptionsToShow = this.shuffle(optionsToShow);
```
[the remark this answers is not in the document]
finnss: Also, performing this shuffle inside the `render()` function causes it to re-shuffle a lot. In particular, when you expand the list to show all options, the order is re-arranged. Rather annoying. Slightly less importantly, it shuffles twice when you first load the page. 

There are several ways you could combat this. My initial suggestion would be to use a state for it, setting it in `componentDidMount` only if it's empty (initial state) or the same as ` props.options`. Untested pseudocode:

```
state = {
    shuffled: []
}

componentDidMount() {
    // Should only run the first time. if not, you could check the previous 
    // value of this.state.shuffled and only shuffle if it's empty or the same as props.options
    this.setState({ shuffled: shuffle(this.props.options) }); 
}
```

There might be a nice way to do this without state, don't take my word for it. It shouldn't re-shuffle when the list is expanded or collapsed, though.

PR #1688, opened 2019-10-23
Fix sentry error in company interest list

mrfarstad:
Solves https://github.com/webkom/lego/issues/1713.

This should solve the following sentry error. This error is not visible to the user, so it is a bit hard to reproduce.
https://sentry.abakus.no/webkom/lego-webapp/issues/7654/?referrer=slack


JonasBak:
Should probably wait for https://github.com/webkom/lego-webapp/pull/1687 :ear_of_rice:

PR #1689, opened 2019-10-23
Add template name for surveys

ingraso:
Add template name to inform what template you are using.

image

### DIFF
--- a/app/routes/surveys/components/SurveyEditor/SurveyEditor.js
+++ b/app/routes/surveys/components/SurveyEditor/SurveyEditor.js
@@ -29,6 +29,7 @@ type Props = FieldProps & {
   submitFunction: (SurveyEntity, ?number) => Promise<*>,
   push: string => void,
   template?: Object,
+  selectedTemplateType?: string,
   destroy: () => void,
   initialize: () => void,
   activeFrom: string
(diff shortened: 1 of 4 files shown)


app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -119,31 +121,37 @@ class SurveyEditor extends Component<Props, State> {
...
-              </Button>
-            }
-          >
-            <TemplateTypeDropdownItems
-              survey={survey}
-              push={push}
-              destroy={destroy}
-            />
-          </Dropdown>
+          <div className={styles.questionTypeSelectorBox}>
```
finnss: questionType? Should be templateType, shouldn't it?

app/routes/surveys/components/surveys.css
```diff
@@ -109,6 +109,12 @@
   border: 1px solid #ccc;
 }
 
+.questionTypeSelectorBox span {
+  display: block;
```
finnss: If you want it to be a block, it should be a `<div>`. That's the exact difference between divs and spans; spans are "inline" while divs "break" away from the content around it – implemented through `display: block`.

PR #1692, opened 2019-10-28
Ensure the Freshness™ of random quotes

finnss:
Depends on the backend in https://github.com/webkom/lego/pull/1715.

Adds a local state to Random Quote (through hooks 😎) that keeps track of what random quotes the user already has seen. This prevents getting duplicates when repeatedly clicking ♻️to browse multiple random quotes, which of course is a behaviour we want to support.

When all quotes have been seen, the backend falls back to providing "stale" quotes.

### DIFF
--- a/app/actions/QuoteActions.js
+++ b/app/actions/QuoteActions.js
@@ -7,6 +7,7 @@ import callAPI from 'app/actions/callAPI';
 import { Quote } from './ActionTypes';
 import { addToast } from 'app/actions/ToastActions';
 import type { Thunk } from 'app/types';
+import type { ID } from 'app/models';
 
 const getEndpoint = (state, loadNextPage, queryString) => {
   const pagination = state.quotes.pagination;
@@ -79,12 +80,14 @@ export function fetchQuote(quoteId: number) {
   });
 }
 
-export function fetchRandomQuote() {
+export function fetchRandomQuote(seenQuotes?: Array<ID> = []) {
+  const queryString = `?seen=[${String(seenQuotes)}]`;
   return callAPI({
     types: Quote.FETCH_RANDOM,
-    endpoint: '/quotes/random/',
+    endpoint: '/quotes/random/' + queryString,
     method: 'GET',
     meta: {
+      queryString,
       errorMessage: 'Henting av tilfeldig quote feilet'
     },
     useCache: false,
--- a/app/components/RandomQuote/RandomQuote.js
+++ b/app/components/RandomQuote/RandomQuote.js
@@ -1,5 +1,6 @@
 // @flow
 
+// $FlowFixMe
 import React, { useState, useEffect } from 'react';
 import styles from './RandomQuote.css';
 import Button from '../Button';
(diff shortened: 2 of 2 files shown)


app/actions/QuoteActions.js
```diff
@@ -79,12 +80,14 @@ export function fetchQuote(quoteId: number) {
   });
 }
 
-export function fetchRandomQuote() {
+export function fetchRandomQuote(seenQuotes?: Array<ID> = []) {
+  const queryString = `?seen=[${String(seenQuotes)}]`;
```
mrfarstad: Why do you want to send seenQuotes from the client? Isn't it easier to store seenQuotes in the DB when the random quote ID is generated in the backend? 😸The query strings can get quite large

finnss: You sure? This feels like a front-end thing to me. The query string can indeed get quite large, but storing it backend would mean incurring multiple extra database queries for each request. I can't imagine a long query string worse performance-wise than that. 

Further, which quotes you've "seen" seems like a per-session feature, right? I'm not sure it's desired functionality for that sort of memory to persist between sessions. There could be months between those sessions, which means "resetting" makes sense to me. I'm not saying your point is invalid, thank you for the input! If we want long-term persistence then I agree we can store it back-end, it would just be a different design decision.

app/components/RandomQuote/RandomQuote.js
```diff
@@ -35,6 +35,15 @@ const RandomQuote = (props: Props) => {
     currentQuote
   } = props;
 
+  const [seenQuotes, setSeenQuotes] = useState([]);
+
+  useEffect(() => {
+    const quoteId = props.currentQuote.id;
+    if (!seenQuotes.includes(quoteId)) {
+      setSeenQuotes([...seenQuotes, quoteId]);
```
odinuge: Since this will result in a re-render (and we need no re-render), using `useRef` may be more suitable. 

https://stackoverflow.com/questions/56455887/react-usestate-or-useref

app/actions/QuoteActions.js
```diff
@@ -79,12 +80,14 @@ export function fetchQuote(quoteId: number) {
...
   return callAPI({
     types: Quote.FETCH_RANDOM,
-    endpoint: '/quotes/random/',
+    endpoint: '/quotes/random/' + queryString,
```
odinuge:
```suggestion
    endpoint: `/quotes/random/${queryString}`,
```

app/components/RandomQuote/RandomQuote.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 
+// $FlowFixMe
```
odinuge: Ahh, yee, someone should look into bumping flow to get these types, since it looks like TS has ETA `inf`. Or @SmithPeder?

SmithPeder: ETA > `inf` 🙃 Have too much server/wiki things with more Prio lately 😢

finnss:
I missed your last review before merging somehow @odinuge, I'll get it done in another PR

app/actions/QuoteActions.js
mrfarstad: I see your point so we can stick to queries 💯

PR #1694, opened 2019-11-01
WIP stuff to make it work

odinuge:


erlingfn:
Seems to remove the errors and work locally, what additional fixes are necessary in order to merge it into redux-bumps?

odinuge:
Not sure about why it was changed, but this (with the PR) is the same as it is in `master`

erlingfn:
Does not seem to be any reason for this to be changed to begin with, should this be merged?

PR #1700, opened 2019-11-08
Add link to event-survey from bdb-page

iselinee:
Related to: https://github.com/webkom/lego/issues/1388

### DIFF
--- a/app/routes/bdb/components/BdbDetail.js
+++ b/app/routes/bdb/components/BdbDetail.js
@@ -242,7 +242,7 @@ export default class BdbDetail extends Component<Props, State> {
             <td>
               {event.survey && (
                 <Tooltip content="Spørreundersøkelse">
-                  <Link to={`surveys/${event.survey}`}>
+                  <Link to={`/surveys/${event.survey}`}>
                     <i
                       className={cx(
                         'fa fa-bar-chart',
--- a/app/routes/bdb/components/BdbDetail.js
+++ b/app/routes/bdb/components/BdbDetail.js
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { Component } from 'react';
+import cx from 'classnames';
 import styles from './bdb.css';
 import { Content } from 'app/components/Content';
 import {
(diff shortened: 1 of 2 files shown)


iselinee:
Skjermbilde 2019-11-08 kl  16 36 24

app/routes/bdb/components/BdbDetail.js
```diff
@@ -404,6 +421,7 @@ export default class BdbDetail extends Component<Props, State> {
                     <th>Når</th>
                     <th>Hvor</th>
                     <th>Hva</th>
+                    <th />
```
erlingfn: Would it make sense to write 'spørreundersøkelser' in order to clarify explicitly where the link goes?

iselinee: Concluded that 'spørreundersøkelse' became a bit much as a title for a simple icon. Instead added a tooltip hoovering effect. 
Skjermbilde 2019-11-08 kl  17 17 13

app/routes/bdb/components/BdbDetail.js
```diff
@@ -237,6 +239,21 @@ export default class BdbDetail extends Component<Props, State> {
             </td>
             <td>{truncateString(event.location, 50)}</td>
             <td>{truncateString(event.description, 70)}</td>
+            <td>
+              {event.survey && (
+                <Tooltip content="Spørreundersøkelse">
+                  <Link to={`surveys/${event.survey}`}>
```
finnss: This becomes a relative link, instead of an absolute one.

By following link:
Screenshot 2019-11-11 at 12 29 00
Screenshot 2019-11-11 at 12 29 03


Actual survey, found by navigating there manually:
Screenshot 2019-11-11 at 12 29 09
Screenshot 2019-11-11 at 12 29 13

I believe this is fixed by just changing the link to `/surveys/...`, ie adding an initial slash.

kotlarz: Good catch!

PR #1702, opened 2019-11-14
Fix helment null ptr exception

odinuge:
This fixes https://github.com/webkom/lego/issues/1720

### DIFF
--- a/app/utils/helmet.js
+++ b/app/utils/helmet.js
@@ -9,33 +9,41 @@ import { type ComponentType, createElement } from 'react';
  * A higher order component that wraps the given component in
  * LoadingIndicator while `props[loadingProp]` is being fetched.
  */
-
-type PropertyGenerator = (
-  props: Object,
-  config?: Object
-) => Array<{
+type Property = {
   property?: string,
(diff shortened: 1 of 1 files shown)


app/utils/helmet.js
```diff
@@ -9,33 +9,41 @@ import { type ComponentType, createElement } from 'react';
...
-          ({ element, children, ...props }, index) =>
-            createElement(element || 'meta', { key: index, ...props }, children)
+    propertyGenerator: ?PropertyGenerator
+  }) => {
+    const properties: ?Array<Property> =
+      propertyGenerator && propertyGenerator(props, config);
+
+    return (
+      <>
+        {properties && (
```
SmithPeder: Was this the fix?

odinuge: Correct

SmithPeder: Nice :)

odinuge: Aka. same as doing (ish) `(propertyGenerator(props, config) || []).map` instead of the old `propertyGenerator(props, config).map`

PR #1715, opened 2019-12-06
Fix double/triple fetching of random quote

odinuge:
Since the quote is fetched during ssr, it doesn't have to be fetched on
the client. Only fetch using prepare when there is no quote. The "new
random quote" button will still work.

This fixes the annoying flicker when the random quote change during
first client render.


SmithPeder:
We ( @finnss and me ) looked at this bug for a long time without finding the reason for it, of course its the SSR :( It's always the SSR :(

PR #1727, opened 2020-01-07
Dispatch isUserFollowing after fetchEvent is completed

erlingfn:
Fix bug with isuserfollowing causing error when going directly to an event


mrfarstad:
Great work! =)

PR #1736, opened 2020-01-14
Fix a shitload of gallery scrolling issues

odinuge:
Changes:

91db5ec1 (Odin Ugedal, 75 seconds ago)
   Fix scrolling to the botton when viewing photos

   onKeyDown is shitty

393f0d25 (Odin Ugedal, 2 minutes ago)
   Add hack that retains the scroll position

   When clicking "Back", it will use history.back in case there is something
   in the hash. This is only the case if the previous page was
   /photos

87a2fdda (Odin Ugedal, 3 minutes ago)
   Don't scroll when navigating inside gallery

### DIFF
--- a/app/routes/photos/components/GalleryPictureModal.js
+++ b/app/routes/photos/components/GalleryPictureModal.js
@@ -18,6 +18,7 @@ import type { EntityID } from 'app/types';
 import type { ID } from 'app/models';
 import Button from 'app/components/Button';
 import { Image } from 'app/components/Image';
+// $FlowFixMe
 import { useEffect } from 'react';
 
 type Props = {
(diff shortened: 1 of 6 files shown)


app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -44,7 +45,16 @@ type State = {
   hasNext: boolean,
   hasPrevious: boolean
 };
-
+const OnKeyDownHandler = ({ handler }: { handler: KeyboardEvent => void }) => (
```
odinuge: :trollface:

SmithPeder: :trollface:

mrfarstad:
very good ol' friend

PR #1742, opened 2020-01-19
Create/Redo weekly item, clean up some code

SmithPeder:
As per https://github.com/webkom/lego/issues/1730

Soooo, the `red` border is very disputed here as **Webkom Design AS**, but we couldn't come up with anything __cooler__ 💯 We wanted to make it **POP**, so people see it or something.

### When there is an Event
Screenshot 2020-01-19 at 21 01 19

### When there is no Event
Screenshot 2020-01-19 at 21 01 52

### DIFF
--- a/app/routes/overview/components/Overview.js
+++ b/app/routes/overview/components/Overview.js
@@ -22,6 +22,7 @@ import NextEvent from './NextEvent';
 import Poll from 'app/components/Poll';
 import type { PollEntity } from 'app/reducers/polls';
 import RandomQuote from 'app/components/RandomQuote';
+import WeeklyItem from './Weekly';
 
 type Props = {
   frontpage: Array<Object>,
@@ -99,7 +100,8 @@ class Overview extends Component<Props, State> {
       poll,
       votePoll
     } = this.props;
-    const pinned = frontpage[0];
+
+    // CompactEvents Event Module
     const compactEvents = (
       <CompactEvents
         events={frontpage.filter(isEvent)}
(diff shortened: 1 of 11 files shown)


SmithPeder:
@Magssch

Magssch:
Awesome! IMO the red border looks good, though I guess whether it's too much of a distraction or not could be up to debate. I think it's appropriate to differentiate it a bit from standard articles though, and considering the header should be identical every time this looks good in my opinion. I also like the new ordering in the right column! I'm not quite sure how the #weekly-module works, but one issue might be that during the summer and in exam periods, when there are no weekly-articles, weekly would still be at the top, even if it contains very outdated information. Could we perhaps, in the future at least, put an "expiration date" on the module? E.g. when two weeks have passed since the last weekly was posted, the module will be hidden as well, as to promote more relevant and newer articles.

Magssch:
Any thoughts @asnestige ?

asnestige:
😍😍😍 meget nydelig Peder

SmithPeder:
> Could we perhaps, in the future at least, put an "expiration date" on the module?

@Magssch Yes, we can remove the module when the `#weekly` is more then 2 weeks old

https://github.com/webkom/lego-webapp/blob/a3926f2637f01b2b7770ebdf6f2180a951cfe5c1/app/routes/overview/components/Overview.js#L164

Magssch:
> > Could we perhaps, in the future at least, put an "expiration date" on the module?
> 
> @Magssch Yes, we can remove the module when the `#weekly` is more then 2 weeks old
> 
> https://github.com/webkom/lego-webapp/blob/a3926f2637f01b2b7770ebdf6f2180a951cfe5c1/app/routes/overview/components/Overview.js#L164

Great! 2 weeks should be a sufficient time before removing the module :)

app/routes/overview/components/Overview.js
```diff
@@ -99,177 +102,178 @@ class Overview extends Component<Props, State> {
...
+      .slice(0, this.state.eventsToShow);
+
+    // Events Module
     const events = (
-      <Flex column className={styles.eventsWrapper}>
-        <Link to="/events">
-          <h3 className="u-ui-heading">Arrangementer</h3>
-        </Link>
+      <React.Fragment>
+        <h3 className="u-ui-heading">Arrangementer</h3>
```
erlingfn: Why remove the link here?

SmithPeder: To many links!

app/routes/overview/components/Overview.js
```diff
@@ -99,177 +102,178 @@ class Overview extends Component<Props, State> {
...
-          </div>
-        )}
-      </LoadingIndicator>
+    // Resolve the pinned item
+    const pinned = frontpage[0];
+
+    // Pinned Module
+    const pinnedComponent = pinned && (
+      <div>
+        <h3 className="u-ui-heading">Festet Oppslag</h3>
```
SmithPeder: So this large `O` killed some E2E tests i suppose

LudvigHz:
@SmithPeder status on this? It seems people like this change, so IMO we should get this merged asap.

LudvigHz:
Closing as this is outdated and we will redo much of the frontpage.

PR #1744, opened 2020-01-21
Fetch groups before memberships on pages

erlingfn:
I know the for loop is not ideal, yet none of the array.prototype was able to execute the dispatches in the correct order. Fixing this sentry error.
https://sentry.abakus.no/webkom/lego-webapp/issues/8198/?referrer=slack

### DIFF
--- a/app/routes/pages/PageDetailRoute.js
+++ b/app/routes/pages/PageDetailRoute.js
@@ -97,7 +97,7 @@ const getSection = sectionName =>
     fetchItemActions: []
   };
 
-const loadData = (props, dispatch) => {
+const loadData = async (props, dispatch) => {
   const { fetchItemActions } = getSection(props.params.section);
   const { pageSlug } = props.params;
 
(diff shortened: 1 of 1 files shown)


app/routes/pages/PageDetailRoute.js
```diff
@@ -109,13 +109,17 @@ const loadData = (props, dispatch) => {
         .concat(dispatch(fetchAllPages()))
     );
   }
+  const itemActions = [];
 
+  for (let i = 0; i < fetchItemActions.length; i++) {
+    itemActions[i] = await dispatch(fetchItemActions[i](pageSlug));
```
orhanhenrik: You can use Promise.all here so that these are fetched concurrently.
```
const itemActions = await Promise.all(  fetchItemActions.map(action => dispatch(action(pageSlug)))  )
```

erlingfn: That is hoq it was before. However it is imortant to fetch them in order as the wrong order will lead to an error

PR #1745, opened 2020-01-21
[WIP] Implement dispatchInOrder

erlingfn:
Create a utility function for dispatching actions one after the other, i. e. dispatching fetchEvent before isUserFollowing

### DIFF
--- a/app/routes/events/EventDetailRoute.js
+++ b/app/routes/events/EventDetailRoute.js
@@ -26,6 +26,7 @@ import {
 import loadingIndicator from 'app/utils/loadingIndicator';
 import helmet from 'app/utils/helmet';
 import { deleteComment } from 'app/actions/CommentActions';
+import dispatchInOrder from 'app/utils/dispatchInOrder';
 
 const mapStateToProps = (state, props) => {
   const {
@@ -124,11 +125,14 @@ const mapDispatchToProps = {
   isUserFollowing,
   deleteComment
 };
-
+/*
 const loadData = async ({ params: { eventId }, loggedIn }, dispatch) => [
   await dispatch(fetchEvent(eventId)),
   loggedIn && (await dispatch(isUserFollowing(eventId)))
 ];
+*/
+const loadData = async ({ params: { eventId }, loggedIn }, dispatch) =>
+  await dispatchInOrder([fetchEvent, isUserFollowing], dispatch, eventId);
 
 const propertyGenerator = (props, config) => {
   if (!props.event) return;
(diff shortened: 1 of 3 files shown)


app/routes/events/EventDetailRoute.js
```diff
@@ -124,11 +125,14 @@ const mapDispatchToProps = {
...
 };
-
+/*
 const loadData = async ({ params: { eventId }, loggedIn }, dispatch) => [
   await dispatch(fetchEvent(eventId)),
   loggedIn && (await dispatch(isUserFollowing(eventId)))
 ];
+*/
+const loadData = async ({ params: { eventId }, loggedIn }, dispatch) =>
+  await dispatchInOrder([fetchEvent, isUserFollowing], dispatch, eventId);
```
orhanhenrik: In this refactoring you are missing the `loggedIn` check before calling `isUserFollowing`. Is that intentional?

app/routes/pages/PageDetailRoute.js
```diff
@@ -109,11 +110,8 @@ const loadData = async (props, dispatch) => {
         .concat(dispatch(fetchAllPages()))
     );
   }
-  const itemActions = [];
+  const itemActions = dispatchInOrder(fetchItemActions, dispatch, pageSlug);
```
orhanhenrik: In this case, do you really need to dispatch them in exact order? Can't all `fetchItem`s be called at the same time?

app/routes/events/EventDetailRoute.js
erlingfn: It is not, nice catch.

app/routes/pages/PageDetailRoute.js
erlingfn: I believe the committee actions have to be dispatched consecutively, as you need a group in the state in order to mutate it's membership.

erlingfn:
I agree that forcing all actions to take the same argument restricts the cases where this abstraction would be useful. The original thought behind it was that I saw two similar errors with very similar solutions, and that an abstraction would be necessary if we encountered more in order to not have to tailor make a solution every time. Does anyone have any thoughts on whether or not we need a general solution for this?

app/routes/pages/PageDetailRoute.js
erlingfn: I believe the committee actions have to be dispatched consecutively, as you need a group in the state in order to mutate it's membership.

erlingfn: I believe the committee actions have to be dispatched consecutively, as you need a group in the state in order to mutate it's membership.

mrfarstad:
Cool suggestion! This will definitely make the code shorter, and I can't say that I vote against it 🌮

mrfarstad:
Have this PR been discussed further? Do we think this abstraction is necessary? ☯️

erlingfn:
I have not found any more places to use this, and since it is not very flexible I recommend that we scrap this PR. If a utility function becomes necessary in the future it is not difficult to implement

PR #1746, opened 2020-01-27
Fixed start-end time issue

AdrianAndersen:
image

Displays an error message when users tries to set start time after end time in events.


theodorsm:
lgtm :+1:

ingraso:
Nice!

PR #1765, opened 2020-02-12
Add reordering of questions

ingraso:
Add field for reordering of questions.

image
Depends on https://github.com/webkom/lego/pull/1780


app/routes/surveys/components/SurveyEditor/Question.js
```diff
@@ -72,7 +75,25 @@ const QuestionTypeValue = (props: Object, b) => (
...
+  relativeIndex,
+  numberOfQuestions,
+  question,
+  questionData,
+  deleteQuestion,
+  updateRelativeIndexes,
+  fields
+}: Props) => {
+  const indexOptions = questionIndexMappings([
+    ...Array(numberOfQuestions).keys()
```
SmithPeder: Real fancy

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -89,6 +89,39 @@ class SurveyEditor extends Component<Props, State> {
     templateTypeSelected: ''
   };
 
+  updateRelativeIndexes = (oldIndex, newIndex, fields) => {
+    fields.move(oldIndex, newIndex);
```
SmithPeder: Nice use of the redux-form function, actually creating a new array. 👍

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -89,6 +89,39 @@ class SurveyEditor extends Component<Props, State> {
...
+  };
+
+  renderQuestions = ({ fields, meta: { touched, error } }) => {
+    return [
```
LudvigHz: Is there a reason why this return an array?

ingraso:
Yeah, we discussed 1-indexing vs. 0-indexing without any good conclusion. But I agree that 1-indexing looks better.

ingraso:
Changed to "Spørsmålsplassering" instead of "Rekkefølge", since this makes more sense to me. If anyone has a better suggestion, please do not hesitate to say so.

image

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
finnss: This `renderQuestions` is a prop to redux-form's FieldArray, and I think that just expects an array.

app/routes/surveys/components/SurveyEditor/Question.js
```diff
@@ -116,6 +137,22 @@ const Question = ({ index, question, questionData, deleteQuestion }: Props) => {
           />
         </div>
 
+        <div className={styles.reorderQuestion}>
+          <SelectInput
+            value={{
```
finnss: You should be able to add `optionComponent={QuestionTypeOption}` and `valueComponent={QuestionTypeValue}` to this `<SelectInput />` in order to style it the same way we do all selects. See https://github.com/webkom/lego-webapp/pull/1771/files#diff-8b2d27bca1f613218ed08ae6d905ae0fR72

You might need to wait for that PR to be merged / steal the code temporarily if you want certain Icons.

PR #1778, opened 2020-02-24
Don't send UPPER ROUND char to backend

SmithPeder:
Per the mail regarding image upload issues. The problem was that the filename has `Å` in it (on mac). This was parsed as **JS** as as ["a", "<030a>", "a", "<030a>"], which is a `offset upper rounded` char. This will cause the backend to reject the image.

Example, running split on a string "ååå.png" gives:
Screenshot 2020-02-24 at 23 53 32

To solve this we can swap the the `regex` so it removes all chars we dont want.


SmithPeder:
Plz Merge real faast :)

PR #1789, opened 2020-03-02
Reduce perms when summing for all groups

odinuge:
Only show the subset of perms that matters, since wider permissions
implicitly include all the stricter ones.

This is only done for the sum in order to avoid ambiguity.

Screenshot from 2020-03-02 13-59-48
Screenshot from 2020-03-02 13-59-02


odinuge:
Yee.

All the permission stuff should be moved from `app/routes/` to `app/{utils,components}` instead.

PR #1795, opened 2020-03-11
Add buttons to remove profile picture

theodorsm:
Needed for issue: https://github.com/webkom/lego/issues/1801

fjerne_bilde
fjerne_bilde_2

Backend: https://github.com/webkom/lego/pull/1807

### DIFF
--- a/app/routes/users/components/RemovePicture.js
+++ b/app/routes/users/components/RemovePicture.js
@@ -1,0 +1,69 @@
+// @flow
+
+import React, { Component } from 'react';
+import Button from 'app/components/Button';
+import Flex from 'app/components/Layout/Flex';
+
+import styles from './GroupChange.css';
+
+type Props = {
+  removePicture: string => Promise<*>,
+  username: string
+};
+
+class RemovePicture extends Component<Props, State> {
+  constructor() {
+    super();
+    this.state = {
+      selected: false
+    };
+  }
+
+  handleOnClick = () => {
+    if (this.state.selected)
+      this.props
+        .removePicture(this.props.username)
+        .then(() => this.setState({ selected: false }));
+  };
+
+  handleChange = () => {
+    this.setState({ selected: true });
+  };
+
+  handleReset = () => {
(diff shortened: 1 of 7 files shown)


app/routes/users/components/RemovePicture.js
```diff
@@ -0,0 +1,71 @@
...
+        .then(() => this.setState({ selected: false }));
+  };
+
+  handleChange = () => {
+    this.setState({ selected: true });
+  };
+
+  handleReset = () => {
+    this.setState({ selected: false });
+  };
```
SmithPeder: A little picky, but this can just be one function
```js
toggleSelected = () => this.setState( { selected: !this.state.selected } )
```
Since the two buttons that use the functions are mutually exclusive there will never be the option of what button to press, meaning a choice never occurs. In these cases, the two functions can be merged.

app/routes/users/components/RemovePicture.js
```diff
@@ -0,0 +1,71 @@
...
+          >
+            Slett Bilde
+          </Button>
+        )}
+
+        {this.state.selected && (
+          <div>
+            <Button className={styles.saveButton} onClick={this.handleReset}>
+              Angre
+            </Button>
```
SmithPeder: If you merge the functions into one toggle function, you can also simplify this, since it will be the same button, with different text :) 
```js
<Button className={styles.saveButton} onClick={this.toggleSelected}>
    {this.state.selected ? "Angre" : "Slett Bilde"}
</Button>
```

app/routes/users/components/UserProfile.js
```diff
@@ -180,6 +180,12 @@ export default class UserProfile extends Component<Props, EventsProps> {
     return <ul>{tags}</ul>;
   }
 
+  handleClick = () => {
+    /* eslint-disable no-console */
+    console.log('halla');
+    /* eslint-enable no-console */
+  };
+
```
SmithPeder: Remove

PR #1796, opened 2020-03-11
Fix styling of survey result page

iselinee:


iselinee:
When merging the updates for SurveyResult page, noticed AdminSidebar div got pushed to the bottom.

image


image






Fixed it to look like this:

image


On tablet:


image



And on mobile:

image


image

PR #1801, opened 2020-03-17
New Sentry

LudvigHz:
Removes legacy raven and raven-js libraries and changes to the new sdks.

I don't really know what the `UniversalRaven` thing does, it seems like it just sets a context that does not actually change anything. If someone knows why it was there please let me know!

I tested this out with sentry.io, see https://sentry.io/organizations/abakus/issues/?project=5018548 for examples. (the plan is to swith to sentry.io from our self-hosted)


server/ssr.js
```diff
@@ -94,7 +93,7 @@ const createServerSideRenderer = (
         // $FlowFixMe
         const err = error.error ? error.payload : error;
         log.error(err, 'render_error');
-        raven.captureException(err);
+        SentryInstance.captureException(err);
```
LudvigHz: :open_mouth:

PR #1829, opened 2020-04-01
Fix: Improve client/ssr-side checks.

AdrianAndersen:
Discovered that the checks needed to include document as well, in order to load the darkMode-modifided logos properly. Should be merged asap, as the logos will sometimes be loaded as their light-mode variant without this fix.
image

### DIFF
--- a/app/utils/themeUtils.js
+++ b/app/utils/themeUtils.js
@@ -3,15 +3,17 @@ import imageLogoDarkMode from 'app/assets/logo.png';
 import bannerLightMode from 'app/assets/om-abakus-banner.png';
 import bannerDarkMode from 'app/assets/om-abakus-banner-dark-mode.png';
 
+const isClientSide = () => typeof window !== 'undefined' && document;
+
 export const applySelectedTheme = theme => {
-  if (typeof window !== 'undefined') {
+  if (isClientSide()) {
     document.documentElement.setAttribute('data-theme', theme);
     global.dispatchEvent(new Event('themeChange'));
   }
 };
 
 export const getTheme = () =>
-  typeof window !== 'undefined'
+  isClientSide()
     ? document.documentElement.getAttribute('data-theme')
     : 'light';
 


app/utils/themeUtils.js
```diff
@@ -3,15 +3,17 @@ import imageLogoDarkMode from 'app/assets/logo.png';
 import bannerLightMode from 'app/assets/om-abakus-banner.png';
 import bannerDarkMode from 'app/assets/om-abakus-banner-dark-mode.png';
 
+const isClientSide = () => typeof window !== 'undefined' && document;
+
 export const applySelectedTheme = theme => {
-  if (typeof window !== 'undefined') {
+  if (isClientSide()) {
```
odinuge, line 9: Just use `__CLIENT__` to avoid hardcoding this "hack". :smile:

```js
if (__CLIENT__) {
    [...]
```

instead.

See this for client:

https://github.com/webkom/lego-webapp/blob/611c13b6a6cb7b945a4af462f0cef47a25b01404/config/webpack.client.js#L84

and this for server

https://github.com/webkom/lego-webapp/blob/611c13b6a6cb7b945a4af462f0cef47a25b01404/config/webpack.server.js#L47

app/utils/themeUtils.js
```diff
@@ -3,15 +3,17 @@ import imageLogoDarkMode from 'app/assets/logo.png';
...
     document.documentElement.setAttribute('data-theme', theme);
```
odinuge, line 10: To make this work with ssr you should add this attr during ssr too

Here:
https://github.com/webkom/lego-webapp/blob/2c57fdfef169e641f491f8ae3e5cb1af3e7d2ac5/server/pageRenderer.js#L60

app/utils/themeUtils.js
```diff
@@ -3,15 +3,17 @@ import imageLogoDarkMode from 'app/assets/logo.png';
...
     global.dispatchEvent(new Event('themeChange'));
   }
 };
 
 export const getTheme = () =>
-  typeof window !== 'undefined'
+  isClientSide()
     ? document.documentElement.getAttribute('data-theme')
```
odinuge, line 17: This will mess up the img/colors used in ssr. Should be possible to use a react-context to store the value https://reactjs.org/docs/context.html, or just use simple css to show/hide the correct images based on the `data-theme` attr in the html tag.

app/utils/themeUtils.js
AdrianAndersen: Thanks for your feedback regarding this issue. A PR with the changes you suggested is now open.

PR #1831, opened 2020-04-04
Improve SSR/Client-side checks + change how theme changed images are …

AdrianAndersen:
Changed SSR-check to use webpack configs __CLIENT__. 
Changed how images changed by darkMode is loaded.
Added theme-tag to ssr html code
Tested with ssr:staging with no errors (Sometimes timeouts, but I assume this is because of my laptop.)

### DIFF
--- a/server/pageRenderer.js
+++ b/server/pageRenderer.js
@@ -58,7 +58,7 @@ export default function pageRenderer({
   const isSSR = body === '' ? 'false' : 'true';
   return `
     <!DOCTYPE html>
-    <html>
+    <html data-theme="light">
       <head>
         <meta charset="utf-8">
         ${helmet ? helmet.title.toString() : ''}
--- a/server/pageRenderer.js
+++ b/server/pageRenderer.js
@@ -5,6 +5,8 @@ import serialize from 'serialize-javascript';
 import config from '../config/env';
 import webpackClient from '../config/webpack.client.js';
 import type { State } from '../app/types';
+import { selectCurrentUser } from 'app/reducers/auth';
+import { isEmpty } from 'lodash';
 
 import 'source-map-support/register';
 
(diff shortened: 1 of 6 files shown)


server/pageRenderer.js
```diff
@@ -58,7 +58,7 @@ export default function pageRenderer({
   const isSSR = body === '' ? 'false' : 'true';
   return `
     <!DOCTYPE html>
-    <html>
+    <html data-theme="light">
```
odinuge, line 61:
```suggestion
    <html data-theme=${serialize(selectCurrentUser(state).selectedTheme || "light")}>
```

odinuge: This will SSR with the correct value from the user :tm:

odinuge: With 

```js
import { selectCurrentUser } from 'app/reducers/auth';
```

AdrianAndersen: @odinuge Hm, when I tested your change in ssr:staging, it breaks ssr upon first load, because selectCurrentUser is passed undefined. I tried to fix this with an if test, which worked, but got flow errors:

image

It seemed to work really well tho, when the page is allready loaded, and you refresh the page. It removes the white flash that occurs before the methods in ThemeUtils are called. Any suggestings on how this can be solved? :thinking:

odinuge: Ahh.

You should be able to use:

```js
import {isEmpty} from 'lodash';

const getIt = () =>{
  if (!isEmpty(state)){
    return selectCurrentUser(state).selectedTheme;
  }
}
```

And use it like

```
serialize(getIt() || 'light');
```

AdrianAndersen: Thank you for your continued follow up! Tested this on ssr, and it worked really well.

AdrianAndersen: I thought about having an auto option, which would be great for dark mode power users. It is a nice-to-have feature, and I will probably implement this soon. This was not included in v1, because I wanted to keep it as simple as possible for the first release. :dark_sunglasses:

PR #1838, opened 2020-04-13
Remove all instances of location.query and fix some other bugs

erlingfn:
Change some of the old react router v3 code that slipped through the cracks. Should fix this [issue](https://github.com/webkom/lego/issues/1853)

### DIFF
--- a/app/routes/surveys/index.js
+++ b/app/routes/surveys/index.js
@@ -44,7 +44,6 @@ const surveysRoute = ({ match }: { match: { path: string } }) => (
           Component={SurveyDetailRoute}
         />
         <RouteWrapper
-          exact
           path={`${match.path}/:surveyId/edit`}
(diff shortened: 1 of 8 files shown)


app/routes/surveys/index.js
```diff
@@ -44,7 +44,6 @@ const surveysRoute = ({ match }: { match: { path: string } }) => (
           Component={SurveyDetailRoute}
         />
         <RouteWrapper
-          exact
```
LudvigHz, line 47: Why did you remove `exact` here? If the idea is that the above route matches first, I still like the explicit handling to avoid unexpected behaviour.

erlingfn: Thought it had something to do with urls with query params not being matched, but it turned out to be the templates not being in the backend fixtures. Will add it back in

PR #1851, opened 2020-05-10
Fix crash when accessing a picture url directly

erlingfn:
Makes website not crash when going directly to a picture url, as in not going via the gallery route. To test simply to a specific picture and hard refresh. Fixes [this](https://sentry.io/organizations/abakus/issues/1657741862/?project=5018548&referrer=slack) sentry error.

### DIFF
--- a/app/routes/photos/GalleryPictureRoute.js
+++ b/app/routes/photos/GalleryPictureRoute.js
@@ -14,7 +14,7 @@ import {
   selectCommentsForGalleryPicture
 } from 'app/reducers/galleryPictures';
 import { deletePicture } from 'app/actions/GalleryPictureActions';
-import { updateGalleryCover } from 'app/actions/GalleryActions';
+import { updateGalleryCover, fetchGallery } from 'app/actions/GalleryActions';
 import { push } from 'connected-react-router';
 import { deleteComment } from 'app/actions/CommentActions';
 import { selectGalleryById } from 'app/reducers/galleries';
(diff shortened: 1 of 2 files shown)


app/routes/photos/GalleryPictureRoute.js
```diff
@@ -97,8 +99,12 @@ export default compose(
     mapStateToProps,
     mapDispatchToProps
   ),
-  prepare(({ match: { params } }, dispatch) =>
-    dispatch(fetchGalleryPicture(params.galleryId, params.pictureId))
+  prepare(
+    ({ match: { params } }, dispatch) =>
+      Promise.all[
+        (dispatch(fetchGalleryPicture(params.galleryId, params.pictureId)),
+        dispatch(fetch(params.galleryId)).catch())
```
LudvigHz, line 106: Why the empty `catch`? What happens if we get an error?

LudvigHz, line 106: Fetching the whole gallery when accessing a single photo seems a bit excessive, but might be a good thing :thinking: ?

erlingfn: Was a leftover from a copy paste, will remove it

LudvigHz:
Also, why didn't drone run on the commit :open_mouth:?

app/routes/photos/GalleryPictureRoute.js
erlingfn: Yeah, you can do it with just fetching the singe gallery, got those two confused

erlingfn:
> Also, why didn't drone run on the commit ?

I do not know :shrug:, but i seems to have ran when I forced pushed.

PR #1869, opened 2020-06-11
Fix some survey bugs

LudvigHz:
Fixes https://github.com/webkom/lego/issues/1880

Also fixes an issue where the public results route had the option to change chart type, but that crashes the page since the user does not have edit permissions. For now I just removed the options, but there are some issues where the user is logged out when navigating there which should be fixed. (Super minor issue without side-effects).

### DIFF
--- a/app/routes/surveys/components/Submissions/Results.js
+++ b/app/routes/surveys/components/Submissions/Results.js
@@ -27,7 +27,7 @@ type Props = {
   graphData: Object,
   numberOfSubmissions: number,
   generateTextAnswers: (QuestionEntity) => any,
-  editSurvey: (Object) => Promise<*>,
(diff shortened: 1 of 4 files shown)


LudvigHz:
@erlingfn is this the way to fix this issue? :thinking: I believe we had an issue with 404 routes catching more than it is supposed to before

erlingfn:
> @erlingfn is this the way to fix this issue?  I believe we had an issue with 404 routes catching more than it is supposed to before

Yeah, the PageNotFound route component matches all routes, and as such must be wrapped within a switch component in order for it to only be used when no other routes match. In cases such as these where a switch component would mess with the use of React.Children, removing it is in my opinion the best course of action

PR #1889, opened 2020-06-26
Drone caching

LudvigHz:
Testing adding yarn caching to drone. This is such a simple way to improve build time, I don't know why we haven't done it before :upside_down_face: 

> Tested with very high load on the server! (all cores 90-100%)

**Before:**
image

**After:**
image

The time difference won't be as dramatic under more forgiving circumstances, but it should improve total time by ~1 min I think.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -138,10 +138,13 @@ steps:
       - thumbor
       - redis
       - api
+    volumes:
+      - name: yarn_cache
+        path: /tmp/cache
     environment:
       CYPRESS_INSTALL_BINARY: 0
     commands:
-      - yarn --frozen-lockfile
+      - yarn --frozen-lockfile --cache-folder /tmp/cache
 
   - name: build_server
     image: node:11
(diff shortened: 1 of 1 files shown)


LudvigHz:
> Btw. yarn doesn't support multiple instances running at the same time, with the same cache. It _might_ just work, but it _might_ also just corrupt the cache. Did you try running multiple pipelines at once?
> 
> There is a flag to make it _safe_, but I am not sure about the performance hit of using it. https://classic.yarnpkg.com/en/docs/cli/#toc-concurrency-and-mutex

Ooh, I did not think of that :open_mouth:. It has been fine until now though, with several concurrent builds. According to the docs they do support a shared cache, https://yarnpkg.com/configuration/yarnrc#cacheFolder. But I'm guessing that it's not valid for 1.x versions. 

We can let it run for a bit, and if there are any problems, we can try the `mutex` flag or just remove the cache.

PR #1894, opened 2020-07-14
Update editor to 1.2.0 & fix events ssr

LudvigHz:
- Updated editor to v1.2.0. This update allows the editor to be server-side rendered when disabled (aka. fixes ssr currently broken in staging). The fix here is just to use jsdom to serialize the string value to HTML if it is rendering on node.

- Finally fixes ssr on event detail when logged in :100:! No more `undefined_render_error` from sentry! (or much less at least).

The fix is :1st_place_medal: imo. We just render the comment form in the `disabled` state, and use whatever events we can to re-render it in the `editable` state.

Tested on mobile and desktop. (on mobile it should trigger on `scrollEvent` 99% of the time as the comments are always at the bottom of the page, but I added an `onPointerDown` for safety. So if someone has a tablet or some large touch-screen feel free to test it. It should be unnoticeable.

Fixes https://github.com/webkom/lego/issues/1459

### DIFF
--- a/app/components/CommentForm/index.js
+++ b/app/components/CommentForm/index.js
@@ -12,6 +12,7 @@ import { ProfilePicture } from 'app/components/Image';
 import { addComment } from 'app/actions/CommentActions';
 import type { CommentEntity } from 'app/actions/CommentActions';
 import styles from './CommentForm.css';
+import DisplayContent from 'app/components/DisplayContent';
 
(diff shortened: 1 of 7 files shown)


LudvigHz:
Done some more testing locally. Looks good other than a few random crashes on Firefox for android. But I'm not able to reproduce so maybe someone wants to test in staging.
:shipit:

PR #1590, opened 2019-09-08
Switch to Stripe Elements and Intents API

LudvigHz:
Depends on https://github.com/webkom/lego/pull/1669
This PR:
- Switches to [Stripe Elements](https://stripe.com/docs/payments/payment-intents/web#elements)
- Implements the new [Intents API](https://stripe.com/docs/payments/payment-intents)

What's missing:
- [x] Handle the payment requests server response. [docs](https://stripe.com/docs/payments/payment-intents/migration#elements-step-3)
- [x] Testing with a newer API version
- [x] Some styling and finishing touches
- [x] Error handling

~~The current plan is to use the [synchronous flow](https://stripe.com/docs/payments/payment-intents/migration#elements) as Stripe calls it, which is the easiest method of handling the payments.~~ Stripe recommends using the [automatic confirmation](https://stripe.com/docs/payments/payment-intents/migration/automatic-confirmation#elements) method though. ~~I'm thinking we should go the easier way since our payments are so simple, and we wouldn't need [webhooks](https://stripe.com/docs/payments/payment-intents/verifying-status#webhooks), need feedback on this!~~

~~I've migrated backend to the synchronous flow automatic confirmation method, but need to test and make sure we go this way before making a PR.~~

~~I am not able to test this until we upgrade our Stripe API version though. Should be safe to upgrade in a few days!~~

PR #1897, opened 2020-07-22
Add cypress tests for registrations and payments

LudvigHz:
Requires #1590 

Add cypress E2E tests with stripe. Hopefully this will increase confidence in merging #1590. Also this is really something we should have proper tests for.

**TODO:**
- [x] Add tests for weird user behaviour. (refreshing in the middle of payments, clicking away etc.)
- [x] Add tests for various 3d secure cases (https://stripe.com/docs/testing#regulatory-cards)
- [x] Investigate if it is possible to test paymentRequests


LudvigHz:
This spec will not run in CI, as it currently would take waaay to much time and is sensitive to high server load (due to stripe js loading). We could add a nightly task in master f.ex where this runs, so we actually run it automatically.

Would be nice with some thoughts on this!

PR #1911, opened 2020-08-20
Fix issue with missing user info in event routes

LudvigHz:
Fixes an issue where user information was missing from event list and calendar routes.

This caused the "you must be logged in register" text showing up even though the user is registered
and the add to calendar missing from the bottom of both pages.


erlingfn (approved):
Looks like an oversight from redux-bumps. Great work!

PR #1937, opened 2020-09-02
Include joblistings where workplaces is [] when filtering for "Annet"

jakob-lj:
When a joblisting has workplaces: [], they are not included when filtering workplaces=Annet.
Before:
Skjermbilde 2020-09-03 kl  00 37 19

After:
Skjermbilde 2020-09-03 kl  00 37 33


LudvigHz (approved):
Nice! I agree that this makes sense. Thanks for the contribution 🎉

PR #1939, opened 2020-09-04
Fix bug in editor datepicker

erlingfn:
Fix a bug where the datepicker field is not loaded with the correct initialvalue. This bug would result in the datepicker being initialized to the current date when editing an old meeting, which is rather annoying. Also remove an error with a field intended as read only not being set as read only


SmithPeder (approved):
Very nice

LudvigHz (approved):
Finally 😍

PR #1943, opened 2020-09-07
Don't push route on modal

SmithPeder:


LudvigHz (approved):
Hmm, weird how this got in here :thinking:

SmithPeder:
I got a theory, maybe they stole it from here :D So I guess you can't delete Companies in the `BDB` either 🙃 
https://github.com/webkom/lego-webapp/blob/c87589654355bd0fe85bcadaa2287b8b540e94a7/app/routes/bdb/utils.js#L185-L193

SmithPeder:
@LudvigHz One could not delete companies either, so this fixes this as well. looked at all the other `ConfirmModalWithParant` and these two were the only ones that used the `NavigationLink`

LudvigHz:
Nice :ok_hand:

erlingfn:
Great work!

PR #1948, opened 2020-09-13
[Ignore for now]Add cookie pop-up

theodorsm:
Add annoying cookie consent pop-up. 

image

### DIFF
--- a/app/components/CookieConsenter/index.js
+++ b/app/components/CookieConsenter/index.js
@@ -1,0 +1,35 @@
+// @flow
+
+import React from 'react';
+import styles from './index.css';
+import NavigationLink from 'app/components/NavigationTab/NavigationLink';
+import CookieConsent from 'react-cookie-consent';
+
+type Props = {
+  content: string,
+  buttonText?: string,
+  link?: string,
+};
+
+const CookieConsenter = ({ content, buttonText, link }: Props) => {
+  return (
+    <CookieConsent
+      style={{ background: '#bc1818' }}
+      buttonStyle={{ background: 'white', borderRadius: '5px' }}
+      buttonText={buttonText}
+    >
+      {content + ' '}
+      {link ? (
+        <NavigationLink to={'/' + link}>
+          <span>
+            Les mer <u>her</u>.
+          </span>
+        </NavigationLink>
+      ) : (
+        ''
+      )}
+    </CookieConsent>
+  );
+};
+
+export default CookieConsenter;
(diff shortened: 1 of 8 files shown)


app/routes/app/AppRoute.js
```diff
@@ -141,6 +142,12 @@ class App extends PureComponent<AppProps> {
 
           <PhotoUploadStatus />
 
+          <CookieConsent
+            content="Når du besøker abakus.no installerer vi cookies som er nødvendig for at du skal kunne benytte deg av nettsidene våre."
```
AdrianAndersen, line 146:
```suggestion
            content="Når du besøker abakus.no, installerer vi cookies som er nødvendig for at du skal kunne benytte deg av nettsidene våre."
```

app/components/CookieConsenter/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+const CookieConsenter = ({ content, buttonText, link }: Props) => {
+  return (
+    <CookieConsent
+      style={{ background: '#bc1818' }}
+      buttonStyle={{ background: 'white', borderRadius: '5px' }}
+      buttonText={buttonText}
+    >
+      {content + ' '}
+      {link ? (
```
LudvigHz, line 22: `link && (...)`?

LudvigHz (approved):
lgtm

AdrianAndersen (commented):
Looking good. I would recommend to use --css-variables on your colors to make it more uniform. (For instance **var(--lego-red)** instead of **#bc1818**) Make sure to work out the whether the styles sheet should be used or not. I also added a grammar-suggestion. GDPR-compliance here we come. :smile:

app/routes/app/AppRoute.js
theodorsm: Thanks for the review. I will correct the grammar, but it is not easy to use css-variables as the CookieConset component needs inline styling, so it's not possible to use css classes.

AdrianAndersen: Ait, no worries :+1:

AdrianAndersen (approved):

odinuge:
Can we avoid using cookies for un-authed users instead of this? And instead store the answer for each user in the db so that folks only need to agree once.

theodorsm:
Need to migrate stripe: https://github.com/stripe/react-stripe-js/blob/master/docs/migrating.md

theodorsm:
We are thinking about making a privacy page for users later.

theodorsm:
Screenshots: 
image

image

theodorsm:
The website is not unsuable for users without consenting for cookies. If you take remove your consent, the banner will show again.
image

LudvigHz:
Closing this in favor of #1985

PR #1964, opened 2020-09-22
fix refetcherror in interestgroup

PeterJFB:
Interestgroup did not refetch when navigating from one interestgroup to another


erlingfn (approved):
Nice catch, it looks good!

LudvigHz (approved):
lgtm

PR #1965, opened 2020-09-22
Remove sparse checkout

LudvigHz:


SmithPeder (approved):

theodorsm (approved):

LudvigHz:
Don't merge. Forgot to update drone.yml hash

odinuge (commented):
Why do we need sparse checkout here at all? This makes everything waaaaay more complex and hard to reason about

SmithPeder:
It kinda made sense when we only wanted one folder, but with `shared` we might as well clone the whole repo 🤷

LudvigHz:
> Why do we need sparse checkout here at all? This makes everything waaaaay more complex and hard to reason about

I think the original thought was to limit what is cloned. But yeah, it probably doesn't really make a lot of sense.

SmithPeder:
> But yeah, it probably doesn't really make a lot of sense.

It did make some sense, and the intention was good. If you had cloned the whole repo from the beginning one could have made the same argument in reverse with  _"Why do we clone the whole repo and not just use a sparse checkout"_

So, I vote that we scrap the sparse checkout because it trades complexity for efficiency and scope, and not because it was a bad idea 👍

odinuge:
> I think the original thought was to limit what is cloned.

Afaik. sparse checkout only limits what files is checked out, and git will still clone the whole repo.  And i guess the overhead of checking out all the files is probably equal to 0. If "cloning less" for eg. bandwith and speed is the issue we can still use something like `--depth 1`.

LudvigHz:
> > I think the original thought was to limit what is cloned.
> 
> Afaik. sparse checkout only limits what files is checked out, and git will still clone the whole repo. And i guess the overhead of checking out all the files is probably equal to 0. If "cloning less" for eg. bandwith and speed is the issue we can still use something like `--depth 1`.

Ah, good to know :+1:. The issue I was trying to solve was to limit what secrets and configuration files are cloned. +1 for removing it though.

.drone.yml
```diff
@@ -304,15 +304,7 @@ steps:
       - touch /root/.ssh/known_hosts
       - chmod 600 /root/.ssh/known_hosts
       - ssh-keyscan -t rsa -H github.com > /etc/ssh/ssh_known_hosts 2> /dev/null
-      - cd /infra
```
odinuge, line 307: nit: use `/infrastructure` here instead in order to avoid the ambiguity we discussed yesterday.

odinuge (approved):
lgtm

SmithPeder (approved):
Approved with Odins suggestion of infra -> infrastructure

SmithPeder (approved):

PR #1966, opened 2020-09-23
Sort interest groups in alphabetical order

asmundh:
Sorts ÆØÅ incorrectly :upside_down_face:

### DIFF
--- a/app/routes/interestgroups/components/InterestGroupList.js
+++ b/app/routes/interestgroups/components/InterestGroupList.js
@@ -37,7 +37,8 @@ const InterestGroupList = ({ actionGrant, interestGroups }: Props) => {
         </div>
       </div>
       <div className="groups">
-        {interestGroups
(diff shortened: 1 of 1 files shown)


asmundh:
About sorting ÆØÅ https://stackoverflow.com/questions/20520888/sort-danish-alphabetically-using-javascript

SmithPeder:
@asmundh So the issue https://github.com/webkom/lego/issues/1936 is not marked as backend, so if you fix the prettier error we can merge this.
```
$ prettier '**/*.{js,css,md}' --check
--
7 | Checking formatting...
8 | app/routes/interestgroups/components/InterestGroupList.js
```

LudvigHz:
Nice one! Also, you should not sort the list in-place (`array.prototype.sort` is in-place) as this could cause some weird results. Lodash is not in-place but does not have the option to specify a predicate yet, so I recommend creaiting a copy first. Something like `interestGroups.slice().sort(...)` or `[...interestGroups].sort(...)` should work. (although it's a bit ugly. A comment about this would also be awesome!

LudvigHz (approved):

PR #1974, opened 2020-09-29
Disable reaction buttons when not logged in

Eik1234:
Closes webkom/lego#1861
Passes down loggedIn to Reaction component and disable clicking if not logged in.

### DIFF
--- a/app/components/Reactions/Reaction.js
+++ b/app/components/Reactions/Reaction.js
@@ -41,7 +41,7 @@ class Reaction extends React.Component<Props> {
     }: Props = this.props;
     const classes = [
       className ? className : styles.reaction,
-      ...(canReact ? [styles.clickable] : []),
+      canReact && styles.clickable,
     ];
 
     if (hasReacted) {
--- a/app/components/Reactions/Reaction.js
+++ b/app/components/Reactions/Reaction.js
@@ -39,7 +39,10 @@ class Reaction extends React.Component<Props> {
       reactionId,
       contentTarget,
     }: Props = this.props;
-    const classes = [className ? className : styles.reaction, ...(canReact ? [styles.clickable] : [])];
+    const classes = [
+      className ? className : styles.reaction,
+      ...(canReact ? [styles.clickable] : []),
+    ];
 
     if (hasReacted) {
       classes.push(styles.reacted);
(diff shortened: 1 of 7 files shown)


app/components/Reactions/Reaction.js
```diff
@@ -34,10 +35,11 @@ class Reaction extends React.Component<Props> {
       addReaction,
       deleteReaction,
       hasReacted,
+      canReact,
       reactionId,
       contentTarget,
     }: Props = this.props;
-    const classes = [className ? className : styles.reaction];
+    const classes = [className ? className : styles.reaction + (canReact ? " "+styles.clickable : "")];
```
LudvigHz, line 42: We use `classNames` here, hence the array. So it's better to add the classname as an element in the array.

A suggestion is to do something like this. (or something cleaner if you can :smile:)
```js
const classes = [className ? className : styles.reaction, canReact && styles.clickable].filter(Boolean) 
```

LudvigHz (approved):
Nice one! :100:
Users cannot actually access quotes when not logged in, but no harm in having the check there.

app/components/Reactions/Reaction.js
```diff
@@ -34,10 +35,14 @@ class Reaction extends React.Component<Props> {
...
       deleteReaction,
       hasReacted,
+      canReact,
       reactionId,
       contentTarget,
     }: Props = this.props;
-    const classes = [className ? className : styles.reaction];
+    const classes = [
+      className ? className : styles.reaction,
+      ...(canReact ? [styles.clickable] : []),
```
SmithPeder, line 44: `classnames` can take false values https://www.npmjs.com/package/classnames, so you can actually don't need to pass the empty list if they cannot react. So you can actually do `canReact && styles.clickable`. No need for checks 👍

SmithPeder:
```js
// other falsy values are just ignored
classNames(null, false, 'bar', undefined, 0, 1, { baz: null }, ''); // => 'bar 1'
```

Eik1234: Ah, cool, i'll do that

SmithPeder (approved):
⭐

PR #1978, opened 2020-10-01
Add support for filtering on date published

odinuge:
Should we use `createdAt` as default instead of the current `deadline`?

### DIFF
--- a/app/routes/joblistings/JoblistingRoute.js
+++ b/app/routes/joblistings/JoblistingRoute.js
@@ -36,16 +36,26 @@ function filterJoblistings(joblistings, grades, jobTypes, workplaces) {
   });
 }
 
-const dateSort = (a, b) => {
-  const date1 = moment(a.deadline);
-  const date2 = moment(b.deadline);
-  return date1.isAfter(date2) ? 1 : -1;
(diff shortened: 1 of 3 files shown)


SmithPeder:
> Should we use createdAt as default instead of the current deadline?

Hmm, I see how it could be nice to have `createdAt` as the default, but I feel like it's more natural to have `deadline`. Getting a listing that was just created but has a deadline on 1.mai at the top is kinda strange.

odinuge:
> Hmm, I see how it could be nice to have createdAt as the default, but I feel like it's more natural to have deadline. Getting a listing that was just created but has a deadline on 1.mai at the top is kinda strange.

Yeah, fair point. But then it is easier to spot new stuff tho., but it definitely a tradeoff. Guess we can use the users grade to show most relevant stuff on top as well

liseph:
I agree with Peder, `deadline` feels more natural. We could consider pinning new ones at the top, but that would either have to be static, like pinned for 24h, or dynamic for each user (new for you). I suppose the last option is better, but that would require some more work...  :no_good:

liseph (approved):
lgtm :woman_shrugging:

PR #1981, opened 2020-10-04
Bump node to v14-alpine

LudvigHz:
Seems to be no issues :shipit:


odinuge (approved):
lgtm

The reason we didn't use alpine in k8s was becuase of some strange dns ndots issues, but that should be no problem now.

PR #1987, opened 2020-10-07
EventList handle update

collinlokken:
Checks if state is null to avoid setting state as null


collinlokken:
`TypeError: Cannot read property 'field' of null at i.value (EventList.js:113)` når man trykker backspace i arrangement filteret.

Video: https://drive.google.com/file/d/12hgDxQ5FTN8rGj5q8HCTVkti7n3quB9n/view?usp=sharing

PR #1989, opened 2020-10-08
Added backspaceRemoves and deleteRemoves props to EventList Select

collinlokken:
to avoid clearing the field so that the change handler does not set state to null

video : https://drive.google.com/file/d/12hgDxQ5FTN8rGj5q8HCTVkti7n3quB9n/view?usp=sharing


SmithPeder (approved):
Nice 👍

LudvigHz (approved):
Nice one!

PR #1998, opened 2020-10-13
Add force fetch of readme-as-a-function on ssr

odinuge:
CORS is broken, so this is currently the only way to fetch it.


LudvigHz:
readme CORS is good now, so I don't think we need this anymore :upside_down_face.

PR #2002, opened 2020-10-14
Speedup ssr by a s***ton

odinuge:
Stop waterfall fetching stuff in ssr. This cleans up a bit, but require
the fixes in @webkom/react-prepare to properly speedup stuff...

My simple testing gives when running ssr locally (compared to abakus.no):
- `/` (authenticated as me) 0.80s - 1.20s vs 3-4-5sec (and constant timeouts, since 4sec is ssr timeout).

This also fixes some crazy double fetching of quotes...

(might be some rough edges, so I guess we can try it on staging first...)

### DIFF
--- a/app/routes/app/AppRoute.js
+++ b/app/routes/app/AppRoute.js
@@ -190,7 +190,7 @@ export default compose(
   prepare((_, dispatch) => dispatch(fetchMeta()), [], {
     componentDidMount: false,
     componentWillReceiveProps: false,
-    hasSsrDataDeps: false,
+    awaitOnSsr: false,
   }),
   prepare(
     (props, dispatch) =>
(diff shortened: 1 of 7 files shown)


odinuge:
Changes in `@webkom/react-prepare` here: https://github.com/webkom/react-prepare/pull/6

app/routes/app/AppRoute.js
```diff
@@ -182,33 +182,33 @@ const mapDispatchToProps = {
...
-      return dispatch(
-        fetchMeetings({
-          dateAfter: moment().format('YYYY-MM-DD'),
-        })
-      );
-    })
+  prepare(
+    (props, dispatch) =>
+      Promise.all([
+        fetchNotificationData(),
```
SmithPeder, line 198: I don't understand everything here, but why does this no longer need a `dispatch`?

odinuge: ops, good catch!!! :100:

LudvigHz (approved):
lgtm

PR #2004, opened 2020-10-15
Cleaup user fetching on /user/xyz

odinuge:


app/routes/users/UserProfileRoute.js
```diff
@@ -29,21 +29,20 @@ const loadData = (
...
-  // .then(action =>
-  //   dispatch(fetchUserFeed(action.payload.result))
-  //  );
-};
+) =>
+  Promise.all([
+    dispatch(fetchAllWithType('klasse')),
+    isMe && dispatch(fetchPrevious()),
+    isMe && dispatch(fetchUpcoming()),
+    isMe || dispatch(fetchUser(username)),
```
LudvigHz, line 40: Why can we ignore fetching the user data?

odinuge: Since it is the same endpoint as we use when fetching the user info for the logged in user. Might be na issue when the user changes some settings tho., so will have to look at that.

odinuge: Yeah, the USER.UPDATE.SUCCESS payload is reduced correctly into the state, so this works as intended. :smile:

LudvigHz: Nice! :100:

LudvigHz (approved):

PR #2005, opened 2020-10-16
Check if event is using penalties when displaying penalty notice

erlingfn:
Only show penalty notice when event is actually using penalties. By doing this we avoid saying that the registration has been postponed for the user, even though that is not the case.


LudvigHz (approved):
LGTM :shipit:

PR #2006, opened 2020-10-16
Don't fetch memberships when not necessary

odinuge:


app/routes/interestgroups/components/InterestGroup.js
```diff
@@ -15,15 +15,14 @@ type Props = {
...
-  const { memberships = [] } = group;
   return (
     <Flex className={styles.listItem}>
       <Flex column className={styles.listItemContent} style={{ flex: '1' }}>
         <Link to={`/interestgroups/${group.id}`} className={styles.link}>
           <h2>{group.name}</h2>
         </Link>
         <div>{group.description}</div>
-        <div>{memberships.length} medlemmer</div>
+        <div>{group.numberOfUsers} medlemmer</div>
```
odinuge, line 25: This results in `1` request on `abakus.no/interestgroups`instead of `30` today. 10/10 chance of ssr timeout...

app/routes/admin/groups/components/GroupMembers.js
```diff
@@ -42,6 +42,9 @@ export const GroupMembers = ({
   fetch,
 }: Props) => (
   <div className={styles.groupMembers}>
+    {showDescendants || (
+      <>Antall medlemmer: {groupsById[groupId.toString()].numberOfUsers}</>
```
odinuge, line 46: Makes it easy to see how many users are in a given group (only current group, not child groups..)

Screenshot from 2020-10-16 14-53-17

app/actions/InterestGroupActions.js
```diff
@@ -8,48 +8,6 @@ import { push } from 'connected-react-router';
 import type { Thunk } from 'app/types';
 import createQueryString from 'app/utils/createQueryString';
```
odinuge, line 10: This file can be nuked all together, I just didn't care to do it in this PR.

app/reducers/groups.js
```diff
@@ -31,6 +31,11 @@ export default createEntityReducer({
         newState.byId[action.meta.groupId].memberships.push(
           action.payload.result
         );
+        if (
+          typeof newState.byId[action.meta.groupId].numberOfUsers === 'number'
```
odinuge, line 35: This will btw. never be 100% correct, since the count is for all groups below as well, but this is a ish. heuristic.

LudvigHz: Why don't we just refetch the group when adding a membership?

app/routes/admin/groups/components/GroupTree.js
```diff
@@ -16,23 +16,36 @@ function getUrl(group: Object, pathname: string) {
...
-        </TreeView>
+  return groups
+    .slice()
+    .sort(({ name: nameA }, { name: nameB }) => nameA.localeCompare(nameB))
+    .map((group) => {
+      const href = getUrl(group, pathname);
+      const link = (
+        <Link to={href}>
+          {group.name}
+          {typeof group.numberOfUsers === 'number' && (
```
LudvigHz, line 27: When is this not a number??

app/routes/interestgroups/InterestGroupListRoute.js
```diff
@@ -15,9 +12,7 @@ const mapStateToProps = (state) => ({
   actionGrant: state.groups.actionGrant,
 });
 
-const mapDispatchToProps = { fetchAll, createInterestGroup };
-
 export default compose(
-  prepare((props, dispatch) => dispatch(fetchAll())),
-  connect(mapStateToProps, mapDispatchToProps)
+  prepare((props, dispatch) => dispatch(fetchAllWithType('interesse'))),
```
LudvigHz, line 16: Can we have this as a constant?

app/routes/interestgroups/components/InterestGroup.js
LudvigHz: :heart_eyes:

app/routes/admin/groups/components/GroupTree.js
odinuge: If the object is lacking the attribute, and we cannot do `group.numberOfUsers &&` since `0` is a valid number

app/routes/interestgroups/InterestGroupListRoute.js
odinuge: Ouf, yeah, I guess so. Would be nice with importing types and constants from the backend tho.

app/reducers/groups.js
odinuge: We would then need to refetch all groups, since this is the members in the group and all its children. I think this is ok for now.

app/routes/interestgroups/InterestGroupListRoute.js
odinuge: Did that nao

app/reducers/groups.js
LudvigHz: hmm, I see. I guess it's fine then

LudvigHz (approved):

PR #2008, opened 2020-10-17
Add proper following support

odinuge:
With fancy optimistic loading, since it looks shitty without it...

### DIFF
--- a/app/reducers/index.js
+++ b/app/reducers/index.js
@@ -3,7 +3,6 @@
 import { schema } from 'normalizr';
 import { combineReducers } from 'redux';
 import { connectRouter } from 'connected-react-router';
-import { Event } from 'app/actions/ActionTypes';
 import routing from './routing';
 import allowed from './allowed';
 import form from './forms';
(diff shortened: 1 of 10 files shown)


app/reducers/index.js
```diff
@@ -229,3 +183,93 @@ export const surveySubmissionSchema = new schema.Entity('surveySubmissions', {
...
+);
+
+export const [
+  followersCompanySchema,
+  followersCompany,
+] = followersSchemaGenerator('Company');
+
+export const [followersUserSchema, followersUser] = followersSchemaGenerator(
+  'User'
+);
```
LudvigHz, line 222: Maybe a bit random to have this here? I think it's a bit cleaner to have this with the followers selectors.

app/reducers/index.js
```diff
@@ -229,3 +183,93 @@ export const surveySubmissionSchema = new schema.Entity('surveySubmissions', {
...
+) => {
+  const key = `followers${title}`;
+  return [
+    new schema.Entity(key, {
+      follower: userSchema,
+    }),
+    createEntityReducer({
+      key,
+      types,
+    }),
```
LudvigHz, line 203: How does mutating the follower states work now? I'm not familiar with the default handling in `createEntityReducer` :upside_down_face:.

LudvigHz (commented):
Awesome! It's a bit complicated, but I'm assuming it makes sense.

Is there any reason to have a separate schema for each target on the followers. The only difference I can see is the target type, so why not just have the target type on the schema and have all the followers in one state? (Sort of like comments)

app/reducers/index.js
odinuge: It woks as expected (normal CRUD) as long as the actions define the correct schema.

app/reducers/index.js
odinuge: Guess we can do that, but then we end with twice as much code, and need to defined the various following types twice (bigger change for errors), but guess we can move them.

LudvigHz (commented):
This looks great now! I think a couple of tests is all that's missing before we can get this merged :rocket:

PR #2014, opened 2020-10-19
Add curl to dockerimage

odinuge:
Appanently healthchecks run inside the container using curl :upside_down_face:


odinuge (commented):
lgtm

PR #2017, opened 2020-10-20
improve galleryeditor

juniwbjerde:
image
fixes https://github.com/webkom/lego/issues/1528

### DIFF
--- a/app/routes/photos/components/GalleryEditorActions.js
+++ b/app/routes/photos/components/GalleryEditorActions.js
@@ -2,10 +2,10 @@
 
 import React from 'react';
 import { Flex } from 'app/components/Layout';
-import Icon from 'app/components/Icon';
 import { Collapse } from 'react-collapse';
 import Sticky from 'react-stickynode';
 import styles from './Overview.css';
+import Button from 'app/components/Button';
 
 type Props = {
   selectedCount: number,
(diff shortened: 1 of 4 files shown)


erlingfn:
Nice work! Could you add some before and after pictures to highlight the changes made?

app/routes/photos/components/GalleryEditorActions.js
```diff
@@ -26,37 +26,31 @@ const GalleryEditorActions = ({
...
-            <Icon
-              className={styles.deselectIcon}
-              onClick={onDeselect}
-              name="close"
-            />
-            {selectedCount} Selected
-          </span>
-          <div className={styles.actionBar}>
+      <Flex className={styles.actionsContainer}>
+        <Flex className={styles.actionBar}>
```
LudvigHz, line 30: Does this do anyting? I.e the inner `Flex` (actionBar) does not have any flex properties. I think you need a `FlexItem` between here, but I'm not 100% sure though.

LudvigHz (approved):
LGTM
Nice work! This is alot better :smile:

LudvigHz:
> Nice work! Could you add some before and after pictures to highlight the changes made?

@erlingfn You can look at the image in the issue: https://github.com/webkom/lego/issues/1528

app/routes/photos/components/GalleryEditorActions.js
SmithPeder: +1 The double flex will have an effect, due to the classNames used, but they will override each other.

I recommend that you remove the inner Flex (actionBar),  and delete that css-class, then make some changes to the actionContainer instead

Then make the required changes to `actionContainer` to make it look how you like.

Changing the justify-content to space-between will make it look almoast the same. Maybe change some padding aswell
```sh
```

SmithPeder (commented):
Nice, good job 👍

AdrianAndersen:
Lgtm, good job! :smile: :100:

AdrianAndersen (approved):

PR #2018, opened 2020-10-21
[DISCUSSION] Rename prod -> docker

SmithPeder:
## The Problem

Yesterday @odinuge and I talked about **prod**, and that prod can be a little misleading branch name, as it doesn't actually mean that anything will go into production. The real action is that a **docker build** will occur, and the result will be an **image**

This has always been difficult to explain to new members, as we say things like:

> 1. Push the master to prod??
> 2. Then push the prod to staging??
> 3. Then push the prod to prod??

It's obvious for us what it means, but it doesn't really make sense when you try to talk about it, and/or explain it.

## The Proposed Solution

Rename the prod branch to **docker** as this is what actually happens. Then we can more easily explain the process like:

> 1. Push master to the docker branch in order to trigger a new build
> 2. Deploy this new build staging to check that your changes work
> 3. Then deploy the new build to production in order for it to go live

### TODO for this change

- [x] Rename prod to docker in `.drone.yml`

- [x] Decide on a name for the `tags`. Today the tag is equal to the **branch-sha**, and **docker-fn371ghf** is kind of a dumb name. We can hardcode prod-{SHA}, but i also feel like the prod part of the image tag is unnessesary, as the image will always be a prod image. I suggest using **version-{DRONE_SHA:0:7}**, og just the **sha**.
https://github.com/webkom/lego-webapp/blob/3958d760999346dfecc472e2dbdf6f7b19e0b098/.drone.yml#L277

- [ ] Make the same changes in **LEGO**

- [x] Required changes in **infrastructure** are allready done, since we use `registry.webkom.dev/webkom/lego:{{ lego_version }}` and `registry.webkom.dev/webkom/lego-webapp:{{ webapp_version }}`

- [ ] Update docs where we say push to prod

- [ ] Anywhere else this breaks?

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -274,10 +274,10 @@ steps:
       password:
         from_secret: docker_password
       tags:
-        - ${DRONE_BRANCH}-${DRONE_COMMIT_SHA:0:7}
+        - ${DRONE_CALVER_SHORT}-${DRONE_COMMIT_SHA:0:7}
         - latest
       build_args:
-        - RELEASE=${DRONE_BRANCH}-${DRONE_COMMIT_SHA:0:7}
+        - RELEASE=${DRONE_CALVER_SHORT}-${DRONE_COMMIT_SHA:0:7}
         - COMMIT_SHA=${DRONE_COMMIT_SHA}
         - SENTRY_ORG=abakus
         - SENTRY_PROJECT=lego-webapp
@@ -324,7 +324,7 @@ steps:
       tags:
         - webapp
       limit: ${DRONE_DEPLOY_TO}
-      extra_vars: webapp_version=${DRONE_BRANCH}-${DRONE_COMMIT_SHA:0:7}
+      extra_vars: webapp_version=${DRONE_CALVER_SHORT}-${DRONE_COMMIT_SHA:0:7}
       inventory: /infrastructure/inventory
       vault_password:
         from_secret: ansible_vault_password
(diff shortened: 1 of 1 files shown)


odinuge (approved):
lgtm

Lets merge this now and fix the other stuff later.

When it comes to what to call the tag, the objectively best is to name it after `git describe`.

```
> g describe --always --tags
v1.0.0-2506-g6830d280
```

odinuge:
You gotta do the drone sign stuff tho.

SmithPeder:
> You gotta do the drone sign stuff tho.

Yeah, I didn't bother since a change in the "tags" section would result in a new hash

.drone.yml
```diff
@@ -324,7 +324,7 @@ steps:
       tags:
         - webapp
       limit: ${DRONE_DEPLOY_TO}
-      extra_vars: webapp_version=${DRONE_BRANCH}-${DRONE_COMMIT_SHA:0:7}
+      extra_vars: webapp_version=${DRONE_CALVER_SHORT}-${DRONE_COMMIT_SHA:0:7}
```
odinuge, line 327: :+1:

LudvigHz (commented):
I think the idea is good, but I don't know about the `docker` name. I don't really feel like that really describes what it is. To me, it sounds like it's a branch that's docker specific?? That might be sort of correct, but I think its very confusing too.

LudvigHz:
What about `build` instead? Or something similar

SmithPeder:
I agree that `docker` can be a little confusing too, as you might believe that it actually sends master to docker or something.

`build` might be a better name. Maybe we should add something like `ci` to the name, to self-describe that this branch is a `CI` specific branch? Or just `build`?

odinuge:
:shipit:

PR #2019, opened 2020-10-21
Stop relying on bad reducer stuff

odinuge:


app/reducers/memberships.js
```diff
@@ -18,16 +17,26 @@ export const selectMembershipsForGroup = createSelector(
...
+  (
+    descendants,
+    groupId,
+    pagination,
+    membershipsById,
+    membershipsItems,
+    groupsById,
+    users
+  ) => {
+    if (!pagination && descendants) {
```
SmithPeder, line 33: 👍

SmithPeder (approved):
So we no longer push the memberships, only the `numberOfUsers`?

odinuge:
Yeah! When selecting memberships we select from `state.memberships` insetead. That is simpler and less prone to errors.

PR #2031, opened 2020-11-03
Add profile frame for charity auction winner (me!)

AdrianAndersen:


LudvigHz (approved):
lgtm :shipit: 

We should also add the frame to the profile picture on events and meetings.(The profile picture component in general. (Can do this in another PR).

PR #2033, opened 2020-11-03
Move loading indicator

PeterJFB:
Moved the front pages' loading icon from pinned article to website logo.

Before:
image

After:
image

### DIFF
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -23,6 +23,7 @@ import { Image } from 'app/components/Image';
 import type { UserEntity } from 'app/reducers/users';
 import logoLightMode from 'app/assets/logo-dark.png';
 import logoDarkMode from 'app/assets/logo.png';
+import LoadingIndicator from 'app/components/LoadingIndicator';
 
 type Props = {
   searchOpen: boolean,
@@ -37,6 +38,7 @@ type Props = {
   markAllNotifications: () => Promise<void>,
   fetchNotificationData: () => Promise<void>,
   upcomingMeeting: string,
+  loading: boolean,
 };
 
 type State = {
@@ -115,7 +117,7 @@ class Header extends Component<Props, State> {
   };
 
   render() {
-    const { loggedIn, currentUser } = this.props;
+    const { loggedIn, currentUser, loading } = this.props;
     const isLogin = this.state.mode === 'login';
     let title, form;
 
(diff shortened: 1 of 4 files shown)


ivarn01 (approved):
Looks great!

app/routes/overview/IndexRoute.js
```diff
@@ -19,7 +19,7 @@ import { selectPinnedPolls } from 'app/reducers/polls';
 import { votePoll } from 'app/actions/PollActions';
 
 const mapStateToProps = (state) => ({
-  loadingFrontpage: state.frontpage.fetching,
+  loadingFrontpage: !state.frontpage.fetching,
```
AdrianAndersen, line 22: Why do you do this?

AdrianAndersen (changes requested):
Good job! :100:  You should squash your commits into 1 with `git rebase -i HEAD~5
`

app/routes/overview/IndexRoute.js
PeterJFB: Followed example from EventDetailRoute.js
https://github.com/webkom/lego-webapp/blob/0417f452d48ffebbb5f21e019702b7c9bdcb0e40/app/routes/events/EventDetailRoute.js#L116

AdrianAndersen (approved):
lgtm :+1:

app/components/Header/index.js
```diff
@@ -115,7 +117,7 @@ class Header extends Component<Props, State> {
   };
 
   render() {
-    const { loggedIn, currentUser } = this.props;
+    const { loggedIn, currentUser, loadingFrontpage } = this.props;
```
LudvigHz, line 120: Since this is a general component, it should be agnostic to the frontpage. So you should call this prop something more generic, f.ex. `loading`.

app/routes/overview/IndexRoute.js
LudvigHz: In `eventDetailRoute` the prop is opposite from here, i.e. `notLoading` is `true` whenever we are **not** loading :upside_down_face:. That's why it's negated.

LudvigHz: Also, if `loadingFrontpage` is not used in the `Overview` component anymore, you should remove it.

LudvigHz (changes requested):

LudvigHz (approved):

PR #2041, opened 2020-11-06
Bump ESLint, prettier and stylelint

AdrianAndersen:
Bump ESLint and conform code to the new linting standards. 

**Exceptions:**
- Some legacy code: Suppressed with eslint-ignore-next-line. 
- Non-string literal property keys are not supported in flow yet, and thus I had to apply some FlowFixMe in two cases


LudvigHz (approved):
Nice! I can't believe that eslint didn't enforce strict equals before :3rd_place_medal:.

PR #2055, opened 2020-11-11
Revert "Bump ESLint, prettier and stylelint"

odinuge:
Reverts webkom/lego-webapp#2041


theodorsm (approved):
Yup this breaks some stuff.

PR #2056, opened 2020-11-11
[WIP] Revert "Revert "Bump ESLint, prettier and stylelint""

odinuge:
Reverts webkom/lego-webapp#2055

EDIT: This is solved by: #2084


app/components/Reactions/ReactionPicker.js
```diff
@@ -175,13 +173,13 @@ const ReactionPicker = ({
     return searchEmojis(emojis, searchString);
   }, [searchString, emojis]);
 
-  const onCategoryClick = useCallback((category) => {
+  const onCategoryClick = (category) => {
```
odinuge, line 176: What was the warning/reason for this? I guess we should declare an empty array as deps, but not sure why it warns.

app/components/Table/index.js
```diff
@@ -273,7 +273,7 @@ export default class Table extends Component<Props, State> {
 
     const match = Object.keys(this.state.filters).filter((key) => {
       const { inlineFiltering = true, filterMapping = (val) => val } =
-        this.props.columns.find((col) => col.dataIndex == key) || {};
+        this.props.columns.find((col) => col.dataIndex === key) || {};
```
odinuge, line 276: Sure this logic stays 100% similar and works if the dataIndex is a number?

app/reducers/events.js
```diff
@@ -327,7 +330,7 @@ export const selectAllRegistrationsForEvent = createSelector(
           updatedBy,
         });
       })
-      .filter((reg) => reg.event == eventId)
+      .filter((reg) => reg.event === eventId)
```
odinuge, line 333: :shipit:

app/routes/admin/groups/components/GroupMembers.js
```diff
@@ -95,6 +95,7 @@ function loadData({ query, match: { params }, location }, dispatch) {
 
 function mapStateToProps(state, props) {
   const { pathname, search } = state.router.location;
+  // eslint-disable-next-line no-restricted-globals
```
odinuge, line 98: What about fixing this instead of ignoring it, since the warning is legit?

app/routes/admin/groups/components/GroupMembersList.js
```diff
@@ -44,9 +44,12 @@ const GroupMembersList = ({
   filters,
 }: Props) => {
   const GroupMembersListColumns = (fullName, membership) => {
+    console.log(groupsById);
```
odinuge, line 47: :1st_place_medal:

app/routes/admin/groups/components/GroupMembersList.js
```diff
@@ -44,9 +44,12 @@ const GroupMembersList = ({
...
     const { user, abakusGroup } = membership;
     const performRemove = () =>
+      // eslint-disable-next-line no-restricted-globals
       confirm(
+        // eslint-disable-next-line react/prop-types
```
odinuge, line 52: What is this warning, and how do we fix it?

app/routes/admin/groups/components/GroupMembersList.js
```diff
@@ -68,6 +71,7 @@ const GroupMembersList = ({
   const GroupLinkRender = (abakusGroup) =>
     true && (
       <Link to={`/admin/groups/${abakusGroup}/members?descendants=false`}>
+        {/* eslint-disable-next-line react/prop-types */}
```
odinuge, line 74: What is this warning, and how do we fix it?

app/routes/bdb/utils.js
```diff
@@ -64,8 +64,11 @@ export const selectMostProminentStatus = (
 
 export const semesterNameOf = (index: number) => {
   const indexToSemesterName = {
-    '0': 'spring',
-    '1': 'autumn',
+    // Non-string literal property keys is not supported in flow yet
+    // $FlowFixMe
+    0: 'spring',
```
odinuge, line 69: Why are you changing this? Objects can only have strings as keys, so this doesn't change anything?

app/routes/bdb/utils.js
```diff
@@ -155,6 +158,7 @@ export const getContactedStatuses = (
     ) {
       contacted.splice(contacted.indexOf(status), 1);
     }
+    return null;
```
odinuge, line 161: Is there a reason to have a `map` and not a `forEach` here?

app/routes/errors/HTTPError.js
```diff
@@ -6,11 +6,17 @@ import { Link } from 'react-router-dom';
 import renderAbakus from './renderAbakus';
 
 const HTTPMapping = {
-  '400': 'Noe gikk galt med forespørselen',
-  '401': 'Du er ikke logget inn',
-  '403': 'Denne siden har du ikke tilgang på',
-  '404': 'Denne siden finnes ikke',
-  '500': 'Noe gikk veldig galt, Webkom er på saken!',
+  // Non-string literal property keys is not supported in flow yet
```
odinuge, line 9: Same here, is there a reason?

app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -53,10 +53,12 @@ const OnKeyDownHandler = ({
 }) => (
   useEffect(
     () => (
+      // eslint-disable-next-line no-sequences
```
odinuge, line 56: When we have to have this warning, isn't it better to just rewrite the code?

app/routes/surveys/components/SubmissionEditor/SubmissionEditor.js
```diff
@@ -158,6 +158,7 @@ const validateMandatory = (inputAnswers: Array<Object>, props) => {
     if (question.mandatory && !answeredQuestionIds.includes(question.id)) {
       errors.questions[question.id] = 'Dette feltet er obligatorisk';
     }
+    return null;
```
odinuge, line 161: `map`?

app/utils/createEntityReducer.js
```diff
@@ -158,6 +160,7 @@ export function deleteEntities(deleteTypes: ?EntityReducerTypes) {
               ? [Number(resultId), resultId.toString()]
               : [resultId])
           ),
+          // eslint-disable-next-line no-sequences
```
odinuge, line 163: Maybe rewrite this instead?

odinuge (commented):
nice!

Would be nice if you could split the various things into separate commits, so that we can review flow changes separate from logical changes in the code.

app/routes/admin/groups/components/GroupMembersList.js
erlingfn: This looks like a bug in eslint. They closed the [issue](https://github.com/yannickcr/eslint-plugin-react/issues/2330) but even though we have the newest version it still does not look like it has been fixed.

app/routes/admin/groups/components/GroupMembersList.js
erlingfn: This looks like a bug in eslint. They closed the issue but even though we have the newest version it still does not look like it has been fixed.

app/components/Reactions/ReactionPicker.js
AdrianAndersen: This was caused by: ```warning  React Hook useCallback does nothing when called with only one argument. Did you forget to pass an array of dependencies?  react-hooks/exhaustive-deps` eslint-plugin-react-hooks```
Fixed this by passing an empty array as you suggested. :100:

app/components/Table/index.js
AdrianAndersen: As far as i can tell, dataIndex is always a string (In our codebase). If the dataIndex is a "number", it will probably just be typeof string anyways. :shrug:

app/reducers/events.js
AdrianAndersen: Rip `int` and `string` comparison. Fixed by wrapping `eventId` in `parseInt()`

app/routes/admin/groups/components/GroupMembers.js
AdrianAndersen: You're right. I fixed this with `window.location` instead of `location` and `ConfirmModalWithParent` instead of `confirm()`

app/routes/admin/groups/components/GroupMembersList.js
AdrianAndersen: Ripperino :sweat_smile:

app/routes/bdb/utils.js
AdrianAndersen: Yeah, I agree. Moved these to forEach.

app/routes/errors/HTTPError.js
AdrianAndersen: Not an issue in the latest version of eslint.

app/routes/photos/components/GalleryPictureModal.js
AdrianAndersen: I fixed these, but I am unsure of the () => **(func1(), func2())** syntax meaning. Does this first execute func1, and then return func2? Hopefully you can review my suggested solution in my recently created PR. :slightly_smiling_face:

app/routes/surveys/components/SubmissionEditor/SubmissionEditor.js
AdrianAndersen: Yup, moved to forEach. :smile:

app/utils/createEntityReducer.js
AdrianAndersen: Also unsure of the syntax here :slightly_smiling_face:, but I think the rewrite should be about right. :upside_down_face:

app/routes/admin/groups/components/GroupMembersList.js
AdrianAndersen: Not an issue as we moved to ConfirmModalWithParent instead of confirm()

app/routes/admin/groups/components/GroupMembersList.js
AdrianAndersen: Not an issue since we moved location to window.location

app/routes/bdb/utils.js
AdrianAndersen: eslint were changing these by rule. Not an issue in the latest version.

AdrianAndersen:
This PR should be resolved in #2084

PR #2058, opened 2020-11-11
Replace @babel/polyfill with corejs directly

LudvigHz:
`@babel/polyfill` is deprecated. The latest version of `@babel/polyfill` also uses `corejs2`, so this should fix the `(...).flat` is undefined errors we are getting:tm:. As `flat` and `flatMap` are stable as of corejs3 (is my understanding at lest)


odinuge (approved):
:shipit:

PR #2059, opened 2020-11-12
Fix darkmode logo in header

PeterJFB:
Changes in #2033 was not properly tested in dark mode
fix #2033
Before
image

After
image


LudvigHz (approved):
LGTM :shipit:

PR #2060, opened 2020-11-12
//$FlowFixMe

odinuge:
HOCs, reselect and react-redux are still 10/10 broken tho.


app/routes/bdb/utils.js
```diff
@@ -187,7 +188,7 @@ export const DetailNavigation = ({
       message="Er du sikker på at du vil slette denne bedriften?"
       onConfirm={() => deleteFunction(companyId)}
     >
-      <NavigationLink>Slett</NavigationLink>
+      <NavigationLink to="#">Slett</NavigationLink>
```
LudvigHz, line 191: 🥇

app/routes/photos/components/GalleryPictureModal.js
```diff
@@ -194,7 +193,7 @@ export default class GalleryPictureModal extends Component<Props, State> {
     }
   };
 
-  handleSwipe = ({ dir }: { dir: RIGHT | LEFT }) => {
+  handleSwipe = ({ dir }: { dir: typeof RIGHT | typeof LEFT }) => {
```
LudvigHz, line 196: Hmm, this seems a bit weird 🤔 . Can you not treat this as an enum anymore? I'm guessing this now is identical to `string | string`.

app/utils/replaceUnlessLoggedIn.js
```diff
@@ -4,11 +4,11 @@ import React, { PureComponent, type ComponentType } from 'react';
...
-  ReplacementComponent: ComponentType<Props>
-): (ActualComponent: ComponentType<Props>) => ComponentType<Props> {
+export default function replaceUnlessLoggedIn(
+  ReplacementComponent: ComponentType<any>
+): (ActualComponent: ComponentType<any>) => ComponentType<any> {
   return (ActualComponent) => {
-    class Replacement extends PureComponent<Props & LoginProps> {
+    class Replacement extends PureComponent<any & LoginProps> {
       render() {
         const { loggedIn, ...props } = this.props;
```
LudvigHz, line 13: Could you leave a comment here that we need to fix this? It shouldn't really be that big of a deal as far as I can see, at least if we get TS. Just something about `replaceUnlessLoggedIn` should have a type parameter with the props of the wrapped component would be nice.

LudvigHz (approved):
I looked over most of this now, and it definitely looks like an improvement, even though there are alot of `object`s and _alot_ left to fix.

Nice work 💯

app/index.js
```diff
@@ -69,7 +69,6 @@ Sentry.init({
   dsn: config.sentryDSN,
   release: config.release,
   environment: config.environment,
-  normalizeDepth: 10,
```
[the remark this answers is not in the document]
odinuge: Ahh, flow said that this was a bad attr., but it is just missing in the flow type. Added it again with a new `fixme`. :sweat_smile:

app/reducers/readme.js
```diff
@@ -2,7 +2,7 @@
 import { Readme } from 'app/actions/ActionTypes';
 import produce from 'immer';
 
-const mutateReadme = produce((newState: any, action: any): void => {
+const mutateReadme = produce<any>((newState: any, action: any): void => {
```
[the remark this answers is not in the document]
odinuge: Another PR i guess. Can extract types from graphql or just hardcode them. :sweat_smile:

app/routes/photos/components/GalleryPictureModal.js
odinuge: Yeah, types can be a bit strange, but I think this is the best way to do it (and it works).

Screenshot from 2021-01-06 10-41-41

(197 works, 198 breaks)

app/utils/replaceUnlessLoggedIn.js
odinuge: Yeah.  Will add comment.

Adding proper typings to it will make all routes using it break... :laughing:

app/routes/photos/components/GalleryPictureModal.js
LudvigHz: Ah. I guess I am mistaken then. Awesome

app/utils/replaceUnlessLoggedIn.js
LudvigHz: Yeah. It definitely is out of scope for this PR 🙃

app/routes/photos/components/GalleryPictureModal.js
odinuge: lol. You were right. This is too early I guess.... Changed to string now :sweat_smile:

PR #2064, opened 2020-11-19
Add some more checks to verify fetchHistory works as expected

odinuge:
Fixes `/tags` etc.

### DIFF
--- a/app/reducers/fetchHistory.js
+++ b/app/reducers/fetchHistory.js
@@ -1,26 +1,59 @@
 // @flow
 
 import { FetchHistory, User } from 'app/actions/ActionTypes';
+import { omit } from 'lodash';
 
 type State = {};
 
 const initialState = {};
+export const fetchHistoryEntryKey = (meta: {
+  paginationKey: string,
+  cursor: string,
+  endpoint: string,
+}) =>
+  // Add random (but stable) string before endpoint so that we don't mix actions with paginationKey and those without
+  meta &&
+  (meta.paginationKey
+    ? 'paginationKeyAndCursor:' + meta.paginationKey + '&cursor=' + meta.cursor
+    : meta.endpoint);
 
 export default function fetchHistory(state: State = initialState, action: any) {
   const success = action.meta && action.meta.success;
   switch (action.type) {
     case success: {
+      const key = fetchHistoryEntryKey(action.meta);
       // We only want to cache on the server-side, to avoid having to redo all
       // the requests during the initial render. Caching on the client is sort
       // of annoying, as it creates a lot of weird issues.
-      if (!success || action.cached || __CLIENT__) return state;
+      if (!success || typeof key !== 'string') return state;
+      if (__CLIENT__) {
+        if (state[key]) {
+          const newState = {
+            ...state,
+            [key]: {
+              ...state[key],
+              cacheCounter: state[key].cacheCounter + 1,
+            },
+          };
+
+          // Evict cache if it is used 3 or more times
+          // This is due to some behavior that runs the same action multiple times during first render on the client
+          if (newState[key].cacheCounter >= 3) {
+            return omit(state, key);
+          }
+          return newState;
+        }
+        return state;
+      }
+      if (action.cached) return state;
       return {
         ...state,
-        [action.meta.endpoint]: {
+        [key]: {
           action: {
             ...action,
             cached: true,
           },
+          cacheCounter: 0,
           timestamp: Date.now(),
         },
       };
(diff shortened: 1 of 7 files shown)


LudvigHz (approved):
Awesome!

I'm not all too familiar with pagination and caching, but it seems like the tags actions with pagination is still mixed with actions without, so that it tries using the cache on the wrong action. So in this case the max 3 cache usage seems to fix the problem nicely. However I'm guessing you added the new key in the fetchHistory to avoid this exact problem? I briefly looked around, but this seems to be a problem in `createEntityReducer` and or `callAPI` and not directly related to the caching.

This is really another issue though, but if you have some insight, I would love to know!

erlingfn (approved):
Looks like a great improvement!
However, after having a look at `/tags/`, it looks like we can fix this specific problem by making the fetchAll tags action use pagination in the more generic way enabled by `createEntityReducer`. This means that it does not look at the `/tags/` key when fetching with pagination, yet it also does not use the cached first 30 or so values that it gets from the ssr. Tries debugging a bit, but I sadly lack the time to get a grip on the problem atm, as it seems to be located in `createEntityReducer`. Did you try this while implementing this fix?

odinuge:
> I'm not all too familiar with pagination and caching, but it seems like the tags actions with pagination is still mixed with actions without, so that it tries using the cache on the wrong action.

Yeah, kind of, but that is not that is not the main problem, but thought I could fix that while at it. Since cached actions are used only for a few sec., it wouldn't really happen today unless you keep navigating around _very quick_ after ssr (but then it ofc. results in quite strange behavior). All stuff should move to `paginationNext` instead (and then we can rename that to pagination) to make caching and handling in redux actually make sense. We "can" fix the caching stuff for the old pagination, but I am a bit scared that some places depends on this behavior, so not 100% sure.

But yeah, guess we can add the cursor to the "key of the cache index" for non-paginationNext as well, didn't think of that.


> So in this case the max 3 cache usage seems to fix the problem nicely. However I'm guessing you added the new key in the fetchHistory to avoid this exact problem

Yeah, but that doesn't solve the underlying problem that we might end up in an endless loop.

The optimal should be to just say cache is only used once during rehydration, but since we for eg. infinite scrolling run the action both in `prepare` and in the table/list, 3 makes stuff work more or less predictable at least, while still avoiding the endless loop. I guess we should make sure it is only used in `prepare` during rehydrate, but I am not really sure if that is possible. (eg. if you mutate an entity via a POST, getting the cached requet for GET afterwards results in quite weird behavior.)

> However, after having a look at /tags/, it looks like we can fix this specific problem by making the fetchAll tags action use pagination in the more generic way enabled by createEntityReducer.

Yeah, we have 3 ways of doing pagination right now, and tags use the "ok" one (and not the "good" (paginationNext with the paginationKey) or "bad" one). We can (and should) switch to that, but evicting cache after 3 times makes sure we never end up in an endless loop. This applies to quite a few routes, so fixing it like this is a "better start" imo.

> This means that it does not look at the /tags/ key when fetching with pagination, yet it also does not use the cached first 30 or so values that it gets from the ssr. Tries debugging a bit, but I sadly lack the time to get a grip on the problem atm, as it seems to be located in createEntityReducer. Did you try this while implementing this fix?

Yeah, this is kinda tricky, but currently boils down to the fact that when fetching "more" tags, it use the cache and dispatches `Tag.FETCH.SUCCESS`, and then the infinite scroller see that we can fetch more tags; then in fetches more tags, get the _same_ data from the cache, and keeps on like this until the stack overflows (since this will happen in sync). The best way is to look at what actions are dispatched when going to https://abakus.no/tags

It is also pretty easy to see the breadcrumbs here: https://sentry.io/organizations/abakus/issues/2038878056/?project=5018548&referrer=github_integration

(if this wall of text made it any simpler :sweat_smile: )

app/reducers/fetchHistory.js
```diff
@@ -1,26 +1,59 @@
...
+export const fetchHistoryEntryKey = (meta: {
+  paginationKey: string,
+  cursor: string,
+  endpoint: string,
+}) =>
+  // Add random (but stable) string before endpoint so that we don't mix actions with paginationKey and those without
+  meta &&
+  (meta.paginationKey
+    ? 'paginationKeyAndCursor:' + meta.paginationKey + '&cursor=' + meta.cursor
+    : meta.endpoint);
```
odinuge, line 18: Thought about adding query to the key when using `meta.endpoint` as well, but this cursor is from the internal pagination (paginationNext), and with the old pagination stuff it is hard to calculate the query since it is used in a few different ways.

So yeah, this "works" for all kinds of pagination, but only works "perfect" for the new style.

LudvigHz: I feel like this is the "correct" way of doing it, since we shouldn't need to fix this for legacy stuff. It will just get more confusing. There's no point in making it work with actions not using `paginationKey`, and it should give some more incentive to migrate old stuff so this actually works with it as well :sweat_smile:.

PR #2084, opened 2020-12-30
Bump Style Rules and Linting

AdrianAndersen:
## Changelog

- Bump stylelint and prettier (No drama :wink: )
- Bump eslint to newest version (aka. a lot of new rules:)
  - "==" and "!=" must be moved to "===" and "!=="
    - This caused issues in the [previous PR](https://github.com/webkom/lego-webapp/pull/2041) :skull:, but should be working now.
  - All switch statements needs a `default:`-tag (ez fix)
  - Flow type definitions needed to be sorted a bit
    - 2 places had circular defined types (aka. two types using eachother) I added this to eslint-ignore as this will be fixed when we move to Typescript.
  - `allowed.hasOwnProperty(key)` => `Object.prototype.hasOwnProperty.call(allowed, key)` (improves secuirty)
  - Some places where the return value of `map()` were unused needed to be moved to `forEach()`
  - `location` and `confirm()` are restricted globals, and had to be moved to `window.location` and `ConfirmModalWithParent`
  - Some places used `() => (func1(), func2())` syntax, which had to be rewritten. I am a bit unsure about these rewrites, since I don't fully understand the aforementioned syntax. Hopefully someone can check that these are correct :slightly_smiling_face: 
- Bumped eslint-config-react-app
  - Added eslint-plugin-react-hooks (required dependency to the new version)
  - Non named exports are no longer allowed, created a default export name instead
  - Had to pass a dependency array to `useCallback()`, passed [] to fix eslint warning
- Bumps to other plugins (Not mentioned since these required no other changes)

## Future changes
As I worked on bumping and configuring eslint, I discovered that we have a quite poorly configured .eslintrc. There are many eslint plugins currently not in use, which limits eslint's benefits. In a future PR, I will configure some sane defaults for eslint, as well as enable default rules for the plugins we currently don't use. This will help improving the code quality of our codebase.


app/reducers/events.js
```diff
@@ -327,7 +330,7 @@ export const selectAllRegistrationsForEvent = createSelector(
           updatedBy,
         });
       })
-      .filter((reg) => reg.event == eventId)
+      .filter((reg) => reg.event === parseInt(eventId))
```
odinuge, line 333: `parseInt(eventId, 10)` or `Number(eventId)` (otherwise strange things may happen :sweat_smile: )

app/routes/admin/groups/components/GroupMembers.js
```diff
@@ -89,13 +89,13 @@ export const GroupMembers = ({
 );
 
 function loadData({ query, match: { params }, location }, dispatch) {
-  const showDescendants = location.search.includes('descendants=true');
+  const showDescendants = window.location.search.includes('descendants=true');
```
odinuge, line 92: Hmmm, isn't this using the location param from above? Using `window` will break ssr. Or am I missing something?

app/routes/admin/groups/components/GroupMembers.js
```diff
@@ -89,13 +89,13 @@ export const GroupMembers = ({
...
   return dispatch(fetchMemberships(params.groupId, showDescendants, query));
 }
 
 function mapStateToProps(state, props) {
   const { pathname, search } = state.router.location;
-  const showDescendants = location.search.includes('descendants=true');
+  const showDescendants = window.location.search.includes('descendants=true');
```
odinuge, line 98: This should just be `search` from line `97`. Nice that eslint caught this bug tho! :100:

app/routes/admin/groups/components/GroupPermissions.js
```diff
@@ -90,11 +90,17 @@ const PermissionList = ({
...
-
+              <ConfirmModalWithParent
+                title={'Bekreft fjerning av rettighet'}
+                message={`Er du sikker på at du vil fjerne tilgangen ${permission}?`}
+                closeOnConfirm={true}
+                onConfirm={() => removePermission(permission, group, editGroup)}
+              >
+                <i
+                  className={`fa fa-times ${styles.removeIcon}`}
+                  onClick={() => removePermission(permission, group, editGroup)}
```
odinuge, line 101: Remove this `onClick` maybe?

app/routes/admin/groups/components/GroupMembersList.js
```diff
@@ -45,18 +46,16 @@ const GroupMembersList = ({
...
     return (
       true && (
         <>
-          <i
-            key="icon"
-            className={`fa fa-times ${styles.removeIcon}`}
-            onClick={performRemove}
-          />
+          <ConfirmModalWithParent
+            title={'Bekreft utmelding'}
```
odinuge, line 53: `title="str"` instead.

app/routes/admin/groups/components/GroupPermissions.js
```diff
@@ -90,11 +90,17 @@ const PermissionList = ({
         {permissions.length ? (
           permissions.map((permission) => (
             <li key={permission}>
-              <i
-                className={`fa fa-times ${styles.removeIcon}`}
-                onClick={() => removePermission(permission, group, editGroup)}
-              />
-
+              <ConfirmModalWithParent
+                title={'Bekreft fjerning av rettighet'}
```
odinuge, line 94: `title="str"` instead

app/routes/events/components/EventAdministrate/Attendees.js
```diff
@@ -144,7 +144,7 @@ export default class Attendees extends Component<Props, State> {
             </Link>
           </h2>
           {event.useContactTracing &&
-            currentUser.id == event.createdBy &&
+            currentUser.id === event.createdBy &&
```
odinuge, line 147: Btw. I think `createdBy` can be an object with an user some time. Maybe outside the scope of this PR tho.

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
```diff
@@ -28,15 +28,13 @@ const SubmissionPublicResultsPage = ({
...
-    Object.keys(results[question.id])
-      .map((name) => {
-        if (name !== 'questionType') {
-          texts = results[question.id][name].map((answer, i) => (
-            <li key={i}>{answer}</li>
-          ));
-        }
-      })
-      .filter(Boolean);
+    Object.keys(results[question.id]).forEach((name) => {
```
odinuge, line 31: Hahaha, wut, I don't really understand what this func is doing.. :sweat_smile: But this doesn't change the logic, so I guess it's ok!

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -150,37 +144,52 @@ class SurveyEditor extends Component<Props, State> {
...
-          ) : (
-            <ListNavigation title={titleField} />
-          )}
+        {survey && survey.id ? (
+          <DetailNavigation title={titleField} surveyId={Number(survey.id)} />
+        ) : (
+          <ListNavigation title={titleField} />
+        )}
+        <ConfirmModalWithParent
+          title={'Bekreft bruk av mal'}
```
odinuge, line 153: `title="str"` instead

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -66,15 +67,8 @@ function TemplateTypeDropdownItems({
             onClick={(e) => {
               e.preventDefault();
               e.stopPropagation();
-              if (
```
odinuge, line 69: Maybe add a wrapper around this as well, since `confirm` is removed?

odinuge (commented):
Awesome! Just a few nits, otherwise goood to go!

(Splitting of commits was awesome!)

app/routes/admin/groups/components/GroupMembers.js
odinuge:
```suggestion
  const showDescendants = search.includes('descendants=true');
```

app/reducers/events.js
AdrianAndersen: Oh, ok. Will fix. :+1:

app/routes/admin/groups/components/GroupPermissions.js
AdrianAndersen: :+1:

app/routes/admin/groups/components/GroupMembersList.js
AdrianAndersen: Hm, thought prettier would fix this automagicaly. :thinking:

app/routes/events/components/EventAdministrate/Attendees.js
AdrianAndersen: Hmm, I could just do ```(currentUser.id === event.createdBy || currentUser.id ===.event.createdBy.id)``` to cover both cases. 
I think it is a bit out of scope to make event.createdBy to always be either a `User` or a number. :thinking:

app/routes/surveys/components/Submissions/SubmissionPublicResults.js
AdrianAndersen: It creates a list of ```<li>``` elements, but the function is a bit obscure. The .filter() did nothing, since we don't use the return type, so I just removed it. :shrug:

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
AdrianAndersen: :+1:

app/routes/admin/groups/components/GroupPermissions.js
AdrianAndersen: :+1:

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
AdrianAndersen: Hm, so I initially tried to wrap the list elements with a `ConfirmModalWithParent`, but discovered that those elements gets destroyed when you click on one of them, thereby no showing the Modal. Therefore, i moved the confirmation to the button opening the template-menu. ("bruk mal"). It is a bit less smooth, but it does its purpose ig. :thinking:

app/routes/admin/groups/components/GroupMembers.js
AdrianAndersen: Hm, eslint does not complain about the use of only location in this case. Maybe it is okay to just leave it to ```const showDescendants = location.search.includes('descendants=true');```javascript here, and just refer to search below?

app/routes/admin/groups/components/GroupMembers.js
AdrianAndersen: Will add this to my amend commit to 81cb617

app/routes/admin/groups/components/GroupMembersList.js
AdrianAndersen: Actually, prettier is changing " to ' in this case. I guess this is okay then?

odinuge: Hmm. Sure u tried `title="bar` nad not `title={"bar"}`?

AdrianAndersen: Oh, I see. Forgot to remove the curly brackets { }

odinuge (approved):
lgtm
:shipit:

PR #2092, opened 2021-01-19
Add autocomplete fields to help browsers autofill

odinuge:


AdrianAndersen (approved):
Nice! Tested this with Bitwarden, but it still does not understand the difference between new and old password in password reset. (Guessing this is a BW issue tho)

erlingfn:
Looks good!

erlingfn (approved):

erlingfn:
I think I have yet to find a site where bitwarden understands which field is for the new password, so I think it is a bitwarden issue

PR #2097, opened 2021-01-24
[WIP] Webpack 5

LudvigHz:
_Try number 2_

Both SSR and dev-server works without issues now.

There seems to be some changes to how the manifest file is generated, which needs to be fixed, as well as some warnings and deprecations.


odinuge (approved):
lgtm, awesome!

I see some `export 'default' (imported as 'styles') was not found in './GroupTree.css' (possible exports: )` type stuff in build logs, but not sure if they are important.

:shipit:

LudvigHz:
> I see some `export 'default' (imported as 'styles') was not found in './GroupTree.css' (possible exports: )` type stuff in build logs, but not sure if they are important.

That seems to be unrelated to this indeed. I see also that it's set as global style so there are no modules there :upside_down_face:

odinuge:
> That seems to be unrelated to this indeed. I see also that it's set as global style so there are no modules there upside_down_face

Yeah, they should probably just be imported as `import "file.css"` instead.

PR #2100, opened 2021-01-26
Add support for meeting desc

odinuge:


app/routes/meetings/components/MeetingDetail.js
```diff
@@ -192,6 +192,9 @@ class MeetingDetails extends Component<Props> {
 
           <ContentSection>
             <ContentMain>
+              {meeting.description && (
+                <DisplayContent content={meeting.description} />
```
LudvigHz, line 196: I don't think we should use `DisplayContent` for non-rich-text. I'm not sure if we have some component to render plain text, but as discussed(https://github.com/webkom/lego/pull/2035#discussion_r564351182) we should maybe create it.

odinuge: Ahh, sure. Guess we can just use it as plain text in a `<p />` or something

package.json
```diff
@@ -96,6 +96,7 @@
     "reselect": "4.0.0",
     "serialize-javascript": "^3.0.0",
     "slugify": "^1.3.6",
+    "validator": "^13.5.2",
```
odinuge, line 99: We "can" do this manually, but I don't want to deal with that... :sweat_smile:

app/routes/meetings/components/MeetingEditor.js
```diff
@@ -165,6 +170,12 @@ function MeetingEditor({
           </div>
         )}
 
+        <button
```
ingraso, line 173: What does this button do?

odinuge: Ops. Found it here: https://stackoverflow.com/a/51507806 (because now enter in a field submits the form and saves the meeting)

Maybe clicking enter inside the textArea works without it. Need to test.

ingraso: Aha, ok!

LudvigHz (approved):

ingraso (approved):

PR #2085, opened 2021-01-05
add movement to background nodes

PeterJFB:
Add movement to gray nodes in background

Light theme

https://user-images.githubusercontent.com/42469466/103674506-21b6c880-4f7f-11eb-9361-189b0ca609f4.mp4

Dark theme

https://user-images.githubusercontent.com/42469466/103674538-28ddd680-4f7f-11eb-9621-1702febfed85.mp4

Reloading page, resizing and mobile

https://user-images.githubusercontent.com/42469466/103674673-5165d080-4f7f-11eb-878a-cdf1e14c85b7.mp4

PR #2118, opened 2021-02-16
Revert "add movement to background nodes"

PeterJFB:
Reverts webkom/lego-webapp#2085

Animation led to poor performance on phones and older computers.


LudvigHz (approved):

LudvigHz:
If you want to implement this again, you will have to do so with (pure) CSS animations. Doing this in JS is _not_ an option, as we have seen :sweat_smile:

PR #2146, opened 2021-03-22
Change language and casing on user settings

tirilsys:


SmithPeder (approved):
Good work!!!!!!

LudvigHz (approved):
:rocket:

SmithPeder:
@tirilsys There are some E2E test cases that look for these strings when navigating the page 😢 

Here the string "Change password" is hard-coded
https://github.com/webkom/lego-webapp/blob/master/cypress/integration/password_change_spec.js

Here the string "Submit" is hard-coded
https://github.com/webkom/lego-webapp/blob/master/cypress/integration/profile_spec.js#L123

SmithPeder (changes requested):
Fix `E2E cypress` tests, then we can merge 👍🏻

LudvigHz:
This now targets the heading, since it contains "Endre passord" as well as the button.
We need to change the selector used to get the button element i cypress for this test to be fixed.

PR #2150, opened 2021-03-23
Fix radio buttons again

LudvigHz:
Fixes https://github.com/webkom/lego/issues/1483 (hopefully)


SmithPeder (approved):
I have tested this is Safari, works great!! Good work 🌮

PR #2173, opened 2021-04-20
Show multiple leaders of interest groups

amalieeh:
Fix webkom/lego#2152 
Now will all leaders of an interest group be shown, also if there are multiple leaders.

E.g. if there are 5 leaders in an interest group:
Before:
image

After:
image

### DIFF
--- a/app/routes/interestgroups/components/InterestGroupDetail.js
+++ b/app/routes/interestgroups/components/InterestGroupDetail.js
@@ -94,8 +94,20 @@ const Contact = ({ group }: { group: Group }) => {
     );
   }
 
-  const leader = leaders[0];
+  if (leaders.length > 1) {
+    return (
+      <Flex column>
+        <h4>Ledere</h4>
+        <ul>
+          {leaders.map((leader) => (
+            <li key={leader.user.username}>{leader.user.fullName}</li>
+          ))}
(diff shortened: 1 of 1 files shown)


amalieeh:
Also I got a sudden lint error, so I changed `==` into `===` in the cypress/supprt/index.js

PeterJFB (approved):
lgtm!^o^

cypress/support/index.js
```diff
@@ -2,7 +2,7 @@ import './commands';
 
 Cypress.on('uncaught:exception', (err, runnable, promise) => {
   console.log(err.payload.response.status);
-  if (promise && err.payload && err.payload.response.status == 404) {
+  if (promise && err.payload && err.payload.response.status === 404) {
```
erlingfn, line 5: Are we sure this still works? If status is a string this change will break the comparison

amalieeh: Not really, I just found it weird that `yarn flow` reported it so Thought maybe to change it 🤷 
I'll change it back and if it comes up again we can do more research :)

LudvigHz (approved):

LudvigHz:
@amalieeh I think you may need to rebase this :thinking:

PR #2190, opened 2021-05-23
Enable hiding number of votes cast for individual poll options

LarsSelbekk:
As requested by Glorious Leader @AdrianAndersen, this PR will enable a poll to hide the number of votes cast for each option if the corresponding backend-PR is accepted. See PR [webkom/lego#2196](https://github.com/webkom/lego/pull/2196) for more details.

### DIFF
--- a/app/components/Poll/Poll.css
+++ b/app/components/Poll/Poll.css
@@ -34,7 +34,6 @@
   composes: withShadow from '~app/styles/utilities.css';
   background: var(--lego-card-color);
   padding: 15px 20px 8px;
-  min-height: 195px;
 }
 
 .pollLight {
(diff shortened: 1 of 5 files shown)


LarsSelbekk:
Updated with CSS fix, and correct initial value for resultsHidden in the editing menu.

LarsSelbekk:
With changes in the backend-PR this isn't necessary to actually hide the tallies since they're never sent, however this is necessary for the display UI not to freak out, for creating a poll with hidden results, and for showing the results to those with editing permission.

LarsSelbekk:
> With changes in the backend-PR this isn't necessary to actually hide the tallies since they're never sent, however this is necessary for the display UI not to freak out, for creating a poll with hidden results, and **for showing the results to those with editing permission**.  

It seems I forgot to include this part in the PR, fixed in #2193.

PR #2191, opened 2021-05-23
Auto theme option

Mathiazom:
Solves https://github.com/webkom/lego/issues/2000

See https://github.com/webkom/lego/pull/2197 for backend PR

Added option for auto theme where the theme is chosen based on device settings. This is now the default theme option (does not affect existing user preferences) and is applied when not logged in.

'Light' and 'Dark' options take precedence and will ignore OS theme.

image

image

image

### DIFF
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -127,17 +127,11 @@ class Header extends Component<Props, State> {
     const isLogin = this.state.mode === 'login';
     let title, form;
 
-    if (loggedIn) {
-      if (
-        currentUser &&
-        ((currentUser.selectedTheme === 'auto' &&
-          getTheme() !== getOSTheme()) ||
-          currentUser.selectedTheme !== getTheme())
-      ) {
-        applySelectedTheme(currentUser.selectedTheme || 'auto');
-      }
-    } else {
-      applySelectedTheme('auto');
+    if (loggedIn && currentUser &&
+      ((currentUser.selectedTheme === 'auto' &&
+        getTheme() !== getOSTheme()) ||
+        currentUser.selectedTheme !== getTheme())) {
+      applySelectedTheme(currentUser.selectedTheme || 'light');
     }
 
     switch (this.state.mode) {
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -127,10 +127,12 @@ class Header extends Component<Props, State> {
     const isLogin = this.state.mode === 'login';
     let title, form;
 
-    if (loggedIn && currentUser &&
-      ((currentUser.selectedTheme === 'auto' &&
-        getTheme() !== getOSTheme()) ||
-        currentUser.selectedTheme !== getTheme())) {
+    if (
+      loggedIn &&
+      currentUser &&
+      ((currentUser.selectedTheme === 'auto' && getTheme() !== getOSTheme()) ||
+        currentUser.selectedTheme !== getTheme())
+    ) {
       applySelectedTheme(currentUser.selectedTheme || 'light');
     }
 
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -18,7 +18,7 @@ import {
 } from 'app/components/LoginForm';
 import { Flex } from 'app/components/Layout';
 import cx from 'classnames';
-import { applySelectedTheme, getTheme } from 'app/utils/themeUtils';
+import { applySelectedTheme, getOSTheme, getTheme } from 'app/utils/themeUtils';
 
 import type { UserEntity } from 'app/reducers/users';
 import logoLightMode from 'app/assets/logo-dark.png';
@@ -127,8 +127,17 @@ class Header extends Component<Props, State> {
     const isLogin = this.state.mode === 'login';
     let title, form;
 
-    if (loggedIn && currentUser && currentUser.selectedTheme !== getTheme()) {
-      applySelectedTheme(currentUser.selectedTheme || 'light');
+    if (loggedIn) {
+      if (
+        currentUser &&
+        ((currentUser.selectedTheme === 'auto' &&
+          getTheme() !== getOSTheme()) ||
+          currentUser.selectedTheme !== getTheme())
+      ) {
+        applySelectedTheme(currentUser.selectedTheme || 'auto');
+      }
+    } else {
+      applySelectedTheme('auto');
     }
 
     switch (this.state.mode) {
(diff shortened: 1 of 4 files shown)


LudvigHz:
Awesome :heart_eyes:!
However, I am very against using the device theme by defualt. I for one have a dark device theme, but don't like using dark themes in the browser. There are also a lot of people that have "dark" themes on their devices, but don't like to use dark themes on webpages. 

(Also I think that Abakus.no looks a lot better with the default theme as it is what the page was designed for, no offense to dark mode :smile:)

Mathiazom:
Good point! So the device theme should not be used for unauthed users, and the default should still be light mode for new user?

odinuge:
Awesome! Thanks!

> Awesome heart_eyes!
However, I am very against using the device theme by defualt. I for one have a dark device theme, but don't like using dark themes in the browser. There are also a lot of people that have "dark" themes on their devices, but don't like to use dark themes on webpages.

I vote for having `auto` the default. The people who don't like auto can set their theme explicitly. Most users wouldn't care to go into settings to change, and the reason for auto is to actually be auto, imo.

I still think we should save theme as a cookie/localstorage instead of per user (since it depends on the device you are using), but that is outside the scope for this.

LudvigHz:
> I vote for having auto the default. The people who don't like auto can set their theme explicitly. Most users wouldn't care to go into settings to change, and the reason for auto is to actually be auto, imo.

Sure, it's just my personal opinion :smile:. But I do still think that for unauthed uses we should not use auto (at least since they cannot change it :upside_down_face:)

odinuge:
I like the way https://www.tek.no/ have a dropdown on the bottom of the page.

> Sure, it's just my personal opinion smile. But I do still think that for unauthed uses we should not use auto (at least since they cannot change it upside_down_face)

Yee, that is an ok compromise I guess. Too bad the old default was/is light theme and not unset/null. So changing the default for authed users to auto is probably a bit edgy, or?

Mathiazom:
So to sum up we should have the default selected theme (in the model) be `auto` for registered users (so revert the revert...), but stick with the light theme when not logged in (since this cannot be changed without logging in)?

odinuge:
> So to sum up we should have the default selected theme (in the model) be auto for registered users (so revert the revert...), but stick with the light theme when not logged in (since this cannot be changed without logging in)?

I vote for that option!

app/utils/themeUtils.js
```diff
@@ -1,6 +1,9 @@
 export const applySelectedTheme = (theme) => {
   if (__CLIENT__) {
-    document.documentElement.setAttribute('data-theme', theme);
+    document.documentElement.setAttribute(
```
odinuge, line 3: This will probably not work that well with the server side renderer :sweat_smile:

odinuge (approved):
Still need some more work to properly work with ssr, but this is a nice first step! :1st_place_medal: Thanks!

lgtm

app/utils/themeUtils.js
LudvigHz: Shouldn't this only run on the client due to the if above?

odinuge: Ahh, yeah, the placement of this comment is kinda awkward, I agree... :sweat_smile: 

SSR theme is set here: https://github.com/webkom/lego-webapp/blob/b78907fa7c575053d6d996588c3cb4cf0e9f3557/server/pageRenderer.js#L67, and to get auto working there will probably require some work. I have not tested this PR, so unsure what happen if that value is set to "auto", other than the fact that auto will not work properly. (eg. will it be all funky until light/dark is set, or will it look like "light" before the correct theme is applied).

odinuge: Screenshot from 2021-05-24 10-40-26

Yeah, it will start loading like this (without abakus-logo), and then switch to correct light or dark theme on first render. It works, but feels a bit flashy.

odinuge: Replacing 'auto' with 'light' in SSR will help mitigate most issues, and only have the flash for users with auto while in dark mode. Then we can fix SSR later.

app/components/Header/index.js
```diff
@@ -127,7 +127,12 @@ class Header extends Component<Props, State> {
     const isLogin = this.state.mode === 'login';
     let title, form;
 
-    if (loggedIn && currentUser && currentUser.selectedTheme !== getTheme()) {
+    if (
+      loggedIn &&
+      currentUser &&
+      ((currentUser.selectedTheme === 'auto' && getTheme() !== getOSTheme()) ||
+        currentUser.selectedTheme !== getTheme())
```
odinuge, line 134: If `currentUser.selectedTheme = "auto"`, `getTheme()` returns `light` and `getOSTheme()` returns true;

This would become `loggedIn && currentUser && (currentUser.selectedTheme !== getTheme())` than will essentially be true on each renderer, or am I just overthinking? Not sure how often this component is rendered tho., but might be an issue

Mathiazom: Oh yea, I could easily fix that

app/utils/themeUtils.js
Mathiazom: I was also thinking that it would be great if the regular light and dark mode (not auto) still had the benefit of SSR. I will look into it

odinuge:
If you replace `auto` with `light` here;

https://github.com/webkom/lego-webapp/blob/b78907fa7c575053d6d996588c3cb4cf0e9f3557/server/pageRenderer.js#L61

This is ready to merge.

Mathiazom:
I'm not sure SSR works correctly with my local backend, as it displays blinking even on `master`. Also, with `ssr:staging` there does not seem to be any blinking with normal dark mode, even without the suggested fix 🤔

odinuge:
> I'm not sure yarn build && yarn ssr works correctly with my local backend, as it displays blinking even on master

Hmm. Sometimes/often SSR times out, but if you get `window.__IS_SSR__ = true;` in the bottom of the HTML, it should have SSRed properly. `yarn ssr` and `yarn ssr:staging` should work the same as https://webapp-staging.abakus.no/ and https://abakus.no

> Also, with ssr:staging there does not seem to be any blinking with normal dark mode, even without the suggested fix thinking

Ahh, no, the blinking would only occur when set to `auto`. Normal dark and light mode work with SSR.

odinuge (approved):

odinuge:
Aweseome! :tada: 

Should be on https://webapp-staging.abakus.no/ in ~half an hour and on https://abakus.no some time later today! If there are some issues, those can be fixed in a follow up PR :smile:

PR #2193, opened 2021-05-26
Allow users with edit access to view hidden poll results

LarsSelbekk:
In #2190 I forgot to show the results of a poll with hidden results to users with edit access, which this PR will amend. It also fixes a minor bug where the expand/contract-arrow of a poll was unnecessarily visible for polls with hidden results.

### DIFF
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -14,6 +14,7 @@ import cx from 'classnames';
 type Props = {
   poll: PollEntity,
   handleVote: (pollId: number, optionId: number) => Promise<*>,
+  allowedToViewHiddenResults: boolean,
   backgroundLight?: boolean,
   truncate?: number,
   details?: boolean,
@@ -95,7 +96,14 @@ class Poll extends Component<Props, State> {
   };
 
   render() {
-    const { poll, handleVote, backgroundLight, details, truncate } = this.props;
+    const {
+      poll,
+      handleVote,
+      backgroundLight,
+      details,
+      truncate,
+      allowedToViewHiddenResults,
+    } = this.props;
     const { truncateOptions, expanded, shuffledOptions } = this.state;
     const {
       id,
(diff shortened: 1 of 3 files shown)


app/components/Poll/index.js
```diff
@@ -14,6 +14,7 @@ import cx from 'classnames';
 type Props = {
   poll: PollEntity,
   handleVote: (pollId: number, optionId: number) => Promise<*>,
+  allowedToViewHiddenResults: boolean,
```
LudvigHz, line 17: Maybe make this optional and/or default `False`?

app/components/Poll/index.js
```diff
@@ -130,8 +138,12 @@ class Poll extends Component<Props, State> {
             <p>{description}</p>
           </div>
         )}
-        {hasAnswered && resultsHidden && <p>Resultatet er skjult</p>}
-        {hasAnswered && !resultsHidden && (
+        {hasAnswered && resultsHidden && !allowedToViewHiddenResults && (
```
LudvigHz, line 141: I think `resultsHidden` and `allowedToViewHiddenResults` can benefit from being merged into one variable. The code is getting a bit hard to read now :upside_down_face:

LudvigHz (approved):
Nice

SmithPeder:
@LarsSelbekk Wanna fix the quick suggestions from Ludvig, then we can merge this 😄

app/components/Poll/index.js
LarsSelbekk: I totally agree that it makes it more complicated, but I was so sure there was a reason it was split into two. I haven't had the chance to test it out yet, but I will very soon.

LarsSelbekk:
The first one is trivial, the second one I need to test a little, and I haven't found the time yet, but I'll take a look over the weekend.

LudvigHz:
@LarsSelbekk I added your commit, with the desired fixes and opened a new PR (#2336) (as I can't push to this branch). So I'll close this and merge the changes in that.

PR #2221, opened 2021-09-06
Change banner to revyopptak

PeterJFB:
Screenshot from 2021-09-06 14-16-11
Text is longer to make sure people notice the change:)
Also removes 'revyBlue' and '--color-blue-revy' variables, as they were only used by the banner

### DIFF
--- a/app/components/Banner/Banner.css
+++ b/app/components/Banner/Banner.css
@@ -37,11 +37,6 @@
   color: var(--color-white);
 }
 
-.revyBlue {
-  background-color: var(--color-blue-revy);
-  color: var(--color-white);
-}
-
 @keyframes wiggle {
   0% {
     transform: rotate(0deg);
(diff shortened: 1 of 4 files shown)


ivarn01 (approved):

SmithPeder (changes requested):
I would like this to be the new Revy-color? Then people will see that it's a new thing :)

juniwbjerde:
Do you mean that you want it red or blue @SmithPeder? Cause the new color for the revy is abakus-red, which is what the banner in the pr is now.

SmithPeder:
Ohh wait the "purple" color is the old Revy color! nvm then :/

PeterJFB:
Red it is. I'll also remove the old color-variable, as it was only used in the banner:)

SmithPeder (approved):
⚡

PR #2224, opened 2021-09-15
Status is ment to show pool information, and not registration status …

SmithPeder:
An issue was pointed out regarding the symbols for the admin `RegistrationTable`. We used to have a separate icon for people on the waiting list, but that was suddenly gone. This is due to a mixup with the name of the column and the `dataIndex`. **Status** is supposed to show "GREEN" for people who are in a pool and "ORANGE" for people who are on the waiting list. As can be seen on line 130 (next line after change), the render method is looking for a pool object to use in its `getRegistartionInfo`, but with the wrong index, it's given a "STATUS_SUCCESS" string.

> Current:
image

> Proposed:
image

### DIFF
--- a/app/routes/events/components/EventAdministrate/RegistrationTables.js
+++ b/app/routes/events/components/EventAdministrate/RegistrationTables.js
@@ -126,7 +126,7 @@ export class RegisteredTable extends Component<Props> {
       {
         title: 'Status',
         center: true,
-        dataIndex: 'status',
+        dataIndex: 'pool',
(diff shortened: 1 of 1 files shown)


pellepingo (approved):

LudvigHz (approved):
lgtm :shipit:

PeterJFB (approved):
lgtm^

EliHaugu (commented):
lgtm

PR #2226, opened 2021-09-19
prevent cypress test from entering usersettings too early

PeterJFB:
cypress attempted to enter usersettings with the "Instillinger"-button before the given profilepage was fully loaded. 
This lead to cypress loading a page containing `... /users/undefined/settings/profile`, which does not exist. The button will now appear after all necessary props are loaded.

### DIFF
--- a/app/routes/users/UserProfileRoute.js
+++ b/app/routes/users/UserProfileRoute.js
@@ -72,7 +72,8 @@ const mapStateToProps = (state, props) => {
   const isMe =
     params.username === 'me' || params.username === state.auth.username;
   const actionGrant = (user && user.actionGrant) || [];
-  const showSettings = isMe || actionGrant.includes('edit');
+  const showSettings =
+    (isMe || actionGrant.includes('edit')) && user && user.username;
   const canChangeGrade = state.allowed.groups;
   const canEditEmailLists = state.allowed.email;
   const canDeletePenalties = state.allowed.penalties;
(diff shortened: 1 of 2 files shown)


app/routes/users/components/UserProfile.js
```diff
@@ -376,7 +376,7 @@ export default class UserProfile extends Component<Props, EventsProps> {
               <h3>Brukerinfo</h3>
               <Card className={styles.infoCard}>
                 {this.renderFields()}
-                {showSettings ? (
+                {showSettings && user.username ? (
```
LudvigHz, line 379: Would it not be better to have this check in the instantiation of `showSettings`?
https://github.com/webkom/lego-webapp/blob/bc050aedfe6605b37281ed2d50ca4d7aa39cdf1c/app/routes/users/UserProfileRoute.js#L75

PeterJFB: Maybe. I regard them as two different checks: `showSettings` checks whether or not user is allowed to see settings, while `user.username` ensures the button is only shown when `username`, is an available prop i.e. relevant prop is loaded. We could replace it with something like 
```py
const showSettings = (isMe || actionGrant.includes('edit')) && user && user.username
```
the variable-name `showSettings` still makes sense and it would make the code cleaner in `UserProfile.js`, but is there a more clean way to make sure `user` is instantiated?:P

LudvigHz: I feel it does essentially the same thing, so it makes sense to have it in the route.

You should be able to do: `user?.username` to make it cleaner.

LudvigHz (approved):
Just fix the flow issue and we're golden!

SmithPeder (approved):

PR #2248, opened 2021-09-28
Set container for table overlays to a div that contains the table

EikHR:
Fix webkom/lego#2322

By setting the container for the overlays to a container div that contains the table, they are no longer hidden by the tables overflow rules.

Before:
Screenshot 2021-09-28 at 19 16 12

After:
Screenshot 2021-09-28 at 19 16 44


LudvigHz (approved):
lgtm

PR #2255, opened 2021-09-28
Hide past interest group memberships with duration under two weeks

EikHR:
Fix webkom/lego#2333

Filters the badges for past memberships in interest groups, to only show interest groups where the user has been a member for more than two weeks.


SmithPeder (approved):
LGTM!

PR #2256, opened 2021-09-28
Fix validation for surveys: Require atleast one option for questions

pellepingo:
# Old
Screenshot from 2021-09-28 23-53-00

# New
Screenshot from 2021-09-28 23-48-27

# More on the change
`createValidator` in `validation.js` now also takes in a `rawValidator` function as a param.

Resilves issue: https://github.com/webkom/lego/issues/2335


app/utils/validation.js
```diff
@@ -48,16 +48,21 @@ export const sameAs = (otherField, message) => (value, context) => [
   message,
 ];
 
-export function createValidator(fieldValidators) {
+export function createValidator(fieldValidators, rawValidator) {
+
   return function validate(input) {
-    return Object.keys(fieldValidators).reduce((errors, field) => {
+  const resultRawValidator = rawValidator(input)
```
SmithPeder, line 54: Maybe call this rawValidatorErrors as that will be result of the function

app/utils/validation.js
```diff
@@ -48,16 +48,21 @@ export const sameAs = (otherField, message) => (value, context) => [
...
+
+    const validatorErrors = Object.keys(fieldValidators).reduce((errors, field) => {
```
SmithPeder, line 56: ...Then call this fieldValidatorErrors, to show that these errors are separate from the rawValidatorErroa

app/utils/validation.js
```diff
@@ -48,16 +48,21 @@ export const sameAs = (otherField, message) => (value, context) => [
...
         fieldValidators[field]
           .map((validator) => validator(input[field], input))
           .filter(([isValid]) => !isValid)
           .map(([, message]) => message || 'not valid') || 0;
       if (fieldErrors.length) errors[field] = fieldErrors;
       return errors;
-    }, {});
+    }, 
+    {});
+    return merge(validatorErrors, resultRawValidator)
```
SmithPeder, line 66: ...then when your merge them, the name's makes it easy to see that the input parameters are used, then merged, into errors

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -263,10 +266,22 @@ export const initialQuestion = {
   ],
 };
 
+export const hasOptions = (data) => {
+  const message = {}
+    message.questions = []
+    data.questions?.forEach((element, i) => {
+      if (!["multiple_choice","single_choice"].includes(element.questionType)) return
```
SmithPeder, line 273: I would like this to be use an ENUM to avoid bugs later on. Not your fault that the ENUM does not exist, as [this](https://github.com/webkom/lego-webapp/blob/79a65ef786cd43dde0aebbd8c3adfa3f05051dd2/app/routes/surveys/utils.js#L13) utility is not exported.

Would be nice if the type are exported as a constant ENUM from this file, and that we use that here 👍

SmithPeder (commented):
Very nice work!! I only have some small nitpicking 😅

SmithPeder:
Also, a small **yarn prettier** is needed here

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
pellepingo: Should I make changes to export `questionStrings` in `utils.js` ?

PR #2263, opened 2021-10-05
Fix radio button alignment

EikHR:
Fixes webkom/lego#2315

Before:
Screenshot 2021-10-05 at 20 30 01
After:
Screenshot 2021-10-05 at 20 27 39

Tested with chrome, firefox and safari on macos, and chrome and firefox on linux.


LudvigHz (approved):
LGTM :100:

juniwbjerde (approved):

PR #2303, opened 2021-10-31
Add temporary link to contact form in footer

PeterJFB:
Contact forms have reportedly been difficult to find, which is why we're adding an additional link in the footer. The link is only visible when a user is logged in, as the form itself can currently only be used by them (see https://github.com/webkom/lego/issues/2345). 

The word "Anonymt" is temporary, and may be removed with the new frontpage or when the form becomes accessible to outside individuals.

image
mobile:
image

### DIFF
--- a/app/components/Footer/index.js
+++ b/app/components/Footer/index.js
@@ -9,12 +9,13 @@ import Octocat from 'app/assets/Octocat.png';
 import Icon from 'app/components/Icon';
 import styles from './Footer.css';
 import moment from 'moment-timezone';
+import { Link } from 'react-router-dom';
 
 type Props = {
   loggedIn: boolean,
 };
 
-const Footer = (props: Props) => (
+const Footer = ({ loggedIn }: Props) => (
   <footer className={styles.footer}>
     <div className={styles.footerContent}>
       <Flex column>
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):
Nice 💯

erlingfn:
Looks nice!

app/components/Footer/index.js
```diff
@@ -64,6 +65,11 @@ const Footer = (props: Props) => (
             />
             <h2>Kontakt</h2>
             <a href="mailto:abakus@abakus.no">abakus@abakus.no</a>
+            {loggedIn && (
+              <p>
+                <Link to="/contact">Anonymt kontaktskjema</Link>
+              </p>
+            )}
```
LudvigHz, line 72: I'm guessing this is just temporary due to the contact form issues?

LudvigHz (approved):

app/components/Footer/index.js
PeterJFB: Assuming you're referring to the `loggedIn` check, yes;) It is now mentioned in the contact form issue, so hopefully we'll remember to remove this check when https://github.com/webkom/lego/issues/2345 is resolved.

PR #2304, opened 2021-10-31
Add button to delete user

theodorsm:
Fixes https://github.com/webkom/lego/issues/2340.
Linked to https://github.com/webkom/lego/pull/2369

- [x] Add action
- [x] Add button-component
- [x] Fix flow?
- [x] Add password auth

It is only possible to delete your **OWN** user, meaning the JWT token + password of a user is used to authenticate the delete action.

Added lot of steps (password auth) to make sure you do not delete your user on accident: 
1.
image
2. 
image
3. 
image

### DIFF
--- a/app/actions/UserActions.js
+++ b/app/actions/UserActions.js
@@ -383,6 +383,19 @@ export function createUser(token: string, user: string): Thunk<*> {
     });
 }
 
+export function deleteUser(username: string): Thunk<any> {
+  return callAPI({
+    types: User.DELETE,
+    endpoint: `/users/${username}/delete_user`,
+    method: 'DELETE',
+    schema: userSchema,
+    meta: {
+      errorMessage: 'Sletting av bruker feilet',
+    },
+    body: {},
+  });
+}
+
 export function sendStudentConfirmationEmail(user: string): Thunk<any> {
   return callAPI({
     types: User.SEND_STUDENT_CONFIRMATION_TOKEN,
--- a/app/actions/UserActions.js
+++ b/app/actions/UserActions.js
@@ -8,6 +8,7 @@ import moment from 'moment-timezone';
 import { push } from 'connected-react-router';
 import { userSchema, penaltySchema } from 'app/reducers';
 import callAPI from 'app/actions/callAPI';
+import { addToast } from 'app/actions/ToastActions';
 import { User, FetchHistory, Penalty } from './ActionTypes';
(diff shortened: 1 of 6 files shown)


app/actions/UserActions.js
```diff
@@ -383,6 +383,19 @@ export function createUser(token: string, user: string): Thunk<*> {
...
+export function deleteUser(username: string): Thunk<any> {
+  return callAPI({
+    types: User.DELETE,
+    endpoint: `/users/${username}/delete_user`,
+    method: 'DELETE',
+    schema: userSchema,
+    meta: {
+      errorMessage: 'Sletting av bruker feilet',
+    },
+    body: {},
```
ivarnakken, line 395: Is there a reason for the empty body key?

theodorsm: Thank you for checking out the code, but it is very WIP. #hacktoberfest

theodorsm: Also yes, it is empty because when issuing a HTTP delete we do not want to send any data in the body.

ivarnakken (commented):
Nicely done 💯 
I like how there are multiple safety nets, but I think we should rather ask the user for their password instead of their username 🔐

app/actions/UserActions.js
```diff
@@ -383,17 +384,22 @@ export function createUser(token: string, user: string): Thunk<*> {
...
+        types: User.DELETE,
+        endpoint: `/users/${username}/`,
+        method: 'DELETE',
+        meta: {
+          id: username,
+          errorMessage: 'Sletting av bruker feilet',
+        },
+      })
+    ).then(() => {
+      dispatch(addToast({ message: 'Deleted' }));
```
LudvigHz, line 400: Success toasts can just be set with `successMessage` in `meta`

LudvigHz: And I think something more descriptive than "deleted" is would be nice.

theodorsm:
There seems to be an issue that Django uses too much time to do the deletion that the promise fails or something. Need to look into that and pushing the correct path depending on the server response.

LudvigHz (approved):

theodorsm:
> There seems to be an issue that Django uses too much time to do the deletion that the promise fails or something. Need to look into that and pushing the correct path depending on the server response.

We have discussed this, and if you are deleting your user, it does not matter if you are pushed to 404. The user gets deleted correctly either way.

PR #2312, opened 2021-11-07
Add active field for groups

SmithPeder:
> Also fix showBadge field that used by the action

This PR makes it possible to mark groups as active/inactive. For now, all I use it for is to differentiate between active and inactive interestgroups.

image


ivarnakken (approved):
Nicely done 😃

LudvigHz (approved):

PR #2317, opened 2021-11-10
Add a more informative loading component when waiting for a pending registration.

LudvigHz:
This component will show between a REGISTRATION.REQUEST_REGISTER_SUCCESS and a
SOCKET_REGISTRATION.SUCCESS or FAILURE. The normal loading component
will still show between the registration initition and the request
register success. Note that since there is no coupling between the
registration and events, this logic is only stored in state and we cannot
determine whether to show the component based on the registration alone.

It's a _bit_ weird to store the `pendingRegistration` status in the form state, but it was where it sort of made the most sense to put it, and it should not affect any other form logic.

Solves https://github.com/webkom/lego/issues/2346

Kooha-11-10-2021-02-00-49

### DIFF
--- a/app/reducers/forms.js
+++ b/app/reducers/forms.js
@@ -76,6 +76,14 @@ export default formReducer.plugin({
           submitting: true,
         };
       }
+      case Event.REQUEST_REGISTER.SUCCESS:
+      case Event.REQUEST_UNREGISTER.SUCCESS: {
+        return {
+          ...state,
+          submitting: false,
+          registrationPending: true,
+        };
+      }
       case Event.REQUEST_REGISTER.FAILURE:
       case Event.REQUEST_UNREGISTER.FAILURE: {
         return {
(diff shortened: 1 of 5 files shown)


ivarnakken (approved):
I like it! It worked like a charm 😍

Arashfa0301 (approved):
Perfectly splendid

app/reducers/forms.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 
+import { createSelector } from 'reselect';
```
PeterJFB, line 3: It looks amazing! Is this method used anywhere specific? Can't seem to spot it...

LudvigHz: Oops. forgot to remove this when I decided not to create a selector.
Nice catch :1st_place_medal:

PeterJFB (approved):
LGTM^^/

SmithPeder (approved):
Looks really godd

PR #2318, opened 2021-11-10
Show event registration and stripe popups for at least 2 seconds

LudvigHz:
This makes the "we are processing your registration" box only pop up if
the regisitration takes longer than 2 seconds AND shows it for at least 2
seconds. Similarly for the stripe success message, it will be shown for
at least 2 seconds. This makes it possible for the user to actually read
the contents of the messages.

Solves https://github.com/webkom/lego/issues/2319


Arashfa0301 (approved):
Looks very good me

PR #2319, opened 2021-11-14
Route new announcement from event to detailed page view

mithunansivakumar:


ivarnakken (commented):
I like this feature a lot! 😄 

However, when you route from an event to /announcements, it would be great if the event field under "Mottakere" contains the very event you came from.

mithunansivakumar:
> I like this feature a lot! 😄
> 
> Perhaps you would like to add the same feature to the meetings page as well?

Yes, i will check it out this evening! ;)

mithunansivakumar:
@ivarnakken would you like to remove "redux stuff from "announcementinline". I have not worked with redux in this way. I guess we could remove everything and only export the function/component.

LudvigHz:
> @ivarnakken would you like to remove "redux stuff from "announcementinline". I have not worked with redux in this way. I guess we could remove everything and only export the function/component.

Since the form is gone, you can remove everything related to `redux-form`, and if there is nothing else dependant on redux in that component (I'm pretty sure there isn't), you can "unconnect" it (remove the `connect`) and export it as a normal component :+1:.

eikhr:
finsished in #2774

PR #2328, opened 2021-11-29
Replace optimize-css-assets-webpack-plugin with css-minimizer-webpack-plugin

LudvigHz:
optimize-css-assets-webpack-plugin is deprecated in favor
of css-minimizer-webpack-plugin.


EikHR (approved):
LGTM

PR #2330, opened 2021-11-30
Codesplitting on all routes using loadable-components

LudvigHz:
Also adds some tree-shaking and webpack optimization tweaks.

~The vendors chunk is still quite large. With zxcvbn using \~700KiB and recharts using \~500KiB. We should find a way to split these out from the main vendor chunk, and load async. (looking at this now).~

In the mean time, this should be alot better.

**Before:**
```
[1] WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
[1] This can impact web performance.
[1] Assets: 
[1]   a10340ceb228c90cbeb6.ttf (8.95 MiB)
[1]   app.20320511.js (6.44 MiB)
[1]   c2bd9cd4b22ff17d91c4.jpg (280 KiB)
[1]   5c3185a787083623adb4.png (348 KiB)
[1]   fa60fac19a5deb240e84.png (819 KiB)
[1]   45b8e04fc7022ab64c11.png (331 KiB)
[1]   vendors.ef812859.js (5.78 MiB)
[1]   styles.c6b02716d66620b130e6.css (272 KiB)
[1]


[1] Entrypoints:
[1]   app (12.6 MiB)
[1]       vendors.ef812859.js
[1]       styles.6df2013d.js
[1]       styles.c6b02716d66620b130e6.css
[1]       app.20320511.js
```
image

**After:**
```
[1] WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
[1] This can impact web performance.
[1] Assets: 
[1]   a10340ceb228c90cbeb6.ttf (8.95 MiB)
[1]   app.3f2c1a49.js (584 KiB)
[1]   c2bd9cd4b22ff17d91c4.jpg (280 KiB)
[1]   5c3185a787083623adb4.png (348 KiB)
[1]   45b8e04fc7022ab64c11.png (331 KiB)
[1]   fa60fac19a5deb240e84.png (819 KiB)
[1]   5677.168c437d.js (1.26 MiB)
[1]   8464.chunk.9dd1bb57.js (366 KiB)
[1]   1840.chunk.ec49e232.js (812 KiB)
[1]   styles.9c103041b67ea59185ad.css (272 KiB)

[1] Entrypoints:
[1]   app (2.16 MiB)
[1]       styles.85f811b8.js
[1]       styles.9c103041b67ea59185ad.css
[1]       5677.168c437d.js
[1]       app.3f2c1a49.js
```

image

PR #2331, opened 2021-11-30
Compress client files with gzip

LudvigHz:
Depends on #2330 

This enables compression using the compression-webpack-plugin. The
default settings should be fine for us. This will compress files that
benefit from it, and express will serve the compressed files if they
exist and the browser can accept the encoding.

Should further help a lot on slower networks and older devices.


erlingfn:
Does this compress all files, or am I wrong? I might very well be wrong, but express-static-gzip states that all files that are to be served must be compressed, but as I understand the webpack-plugin does not compress files that do not benefit from it?

LudvigHz:
> Does this compress all files, or am I wrong? I might very well be wrong, but express-static-gzip states that all files that are to be served must be compressed, but as I understand the webpack-plugin does not compress files that do not benefit from it?

@erlingfn Yeah, the docs are a bit confusing. I assume they mean that _all compressed files must be pre-compressed_. I.e. the module does not do the compression for you. It's just a wrapper around `static-server` that will serve gzip if available. So it will serve regular files if they don't exist in gzipped format. See the [example](https://github.com/tkoenig89/express-static-gzip#example)

erlingfn (approved):
> @erlingfn Yeah, the docs are a bit confusing. I assume they mean that all compressed files must be pre-compressed. I.e. the module does not do the compression for you. It's just a wrapper around static-server that will serve gzip if available. So it will serve regular files if they don't exist in gzipped format. See the example

Ait, then this looks good to me.

PR #2332, opened 2021-12-01
Fix an issue where hydration failed

LudvigHz:
This fixes an issue where, most of the time, hydration would fail, and
the entire app would re-render on the client. This was caused by an
incorrect order of calls to the loadable chunk extractor when rendering
the page on the server.


ivarnakken (approved):
Nice catch! 😄

PR #2333, opened 2021-12-01
Fix development build

LudvigHz:
The introduction of loadable-components and chunk-splitting broke the
development server. This fixes it by ignoring loadable-components
SSR, limiting chunk count to 1, and manually processing stats.

In the process of fixing SSR and the static build, I broke the dev build and HMR. Sadly,
loadable-components is not compatible with HMR, but we can just force webpack to only use one bundle
and we essentially bypass loadable-components and HMR works as intended.

### DIFF
--- a/config/webpack.client.js
+++ b/config/webpack.client.js
@@ -115,6 +115,11 @@ module.exports = (env, argv) => {
 
       isProduction && new CompressionPlugin(),
 
+      !isProduction &&
+        new webpack.optimize.LimitChunkCountPlugin({
+          maxChunks: 1,
+        }),
+
       new AssetsPlugin({
         path: outputPath,
       }),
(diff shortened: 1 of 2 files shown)


config/webpack.client.js
```diff
@@ -150,7 +155,10 @@ module.exports = (env, argv) => {
         },
       },
       minimize: true,
-      minimizer: [new CssMinimizerPlugin(), new TerserPlugin()],
+      minimizer: [
+        new CssMinimizerPlugin(),
+        isProduction && new TerserPlugin(),
```
ivarnakken, line 160: Is there really no other workaround than to limit the JavaScript minimizer to only production? :( 

By the way, since we're on the topic of speed improvement, have you looked into the use of [multi-process parallel](https://webpack.js.org/plugins/terser-webpack-plugin/#parallel)? 🚅

LudvigHz: We could probably enable this for development (not the issue here), but it won't matter much tbh. And I/most prefer faster build time while developing rather than optimizing.

LudvigHz: Have not looked into the parallel option :thinking:. Could try it out and see if it makes a difference.

PR #2335, opened 2021-12-02
Add jsconfig.json & logging to HMR

LudvigHz:


ivarnakken (approved):
LGTM! 😄 I can't believe we got a jsconfig only now ...

PR #2336, opened 2021-12-02
Show hidden poll results

LudvigHz:
Fixed the issues from #2193 and added some styling fixes, which are included here.


ivarnakken (approved):
We can finally merge this 😍

PR #2351, opened 2021-12-03
Remove immutable-js dependency

erlingfn:
Remove the unused immutable-js dependency
As far as I know, this has never been used, but it is definitively not used now.

### DIFF
--- a/config/webpack.client.js
+++ b/config/webpack.client.js
@@ -134,7 +134,6 @@ module.exports = (env, argv) => {
         lodash: 'node_modules/lodash-es',
         'moment-timezone':
           'moment-timezone/builds/moment-timezone-with-data-2012-2022.min',
-        immutable: 'node_modules/immutable',
       },
       fallback: {
         util: require.resolve('util/'),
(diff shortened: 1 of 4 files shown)


SmithPeder:
This
https://github.com/webkom/lego-webapp/blob/d914e582d8d67878ccb93ada913ea476fa5ce574/config/webpack.server.js#L67
and this
https://github.com/webkom/lego-webapp/blob/07a6a65a893a402b7e66a49b4bff7815f2e23a02/config/webpack.client.js#L137
will surly complain then? But if it's not used anywhere else we can probably remove it.

SmithPeder:
But immutable is still a dependency of `connected-react-router`, so it will still be present in the node_modules, that's why nothing is complaining after you removed it.
https://github.com/webkom/lego-webapp/blob/5027228c1a45523a4b2728ee73111dab08ba86e4/yarn.lock#L6956

erlingfn:
> But immutable is still a dependency of connected-react-router, so it will still be present in the node_modules, that's why nothing is complaining after you removed it.

Well then we might as well remove it. No use having unused direct dependencies lying around.

LudvigHz:
`immutable` was added in https://github.com/webkom/lego-webapp/pull/401/files as a direct dependency. I couldn't see it used there anywhere either :thinking:, so I'm sure it's safe to remove.

LudvigHz (approved):
Run `yarn` to get the latest prettier, and we're good!

SmithPeder (approved):

PR #2376, opened 2021-12-16
Add all static files as side effects

LudvigHz:
Tree shaking and minimizer in webpack removed these imports from the build


ivarnakken (approved):
Nice! 💯 I've been trying to figure this out for way too many hours ...

PR #2378, opened 2021-12-18
Redesign footer

ivarnakken:
# Here is my proposition for a ✨redesigned✨ footer

I would love some feedback!

## Things to notice
- The colors utilized are taken directly from [abakus.no/brand](https://abakus.no/brand)
- The anonymous contact form is only shown when logged in, temporarily though, as it relies on issue [!2345](https://github.com/webkom/lego/issues/2345)
- The link to Abakus' slack is also only shown when logged in, as it is not "needed" for outside-people, even though we have nothing to hide there (yes, I made sure to ask our glorious leader @erlingfn)

## Comparison

**Before**

before

**After (when logged in)**

after


app/components/Footer/index.js
```diff
@@ -16,80 +17,105 @@ type Props = {
 };
 
 const Footer = ({ loggedIn }: Props) => (
-  <footer className={styles.footer}>
-    <div className={styles.footerContent}>
-      <Flex column>
-        <Flex>
-          <div className={cx(styles.section, utilityStyles.hiddenOnMobile)}>
+  <div>
```
SmithPeder, line 20: The whole `<footer>` is wrapped in a `<div>`, has this any `css` effect, or can it be removed?

SmithPeder (approved):
Very good job!

app/components/Footer/index.js
ivarnakken: Nice catch, it can definitely be removed! 😍 I believe this was left from back when I was experimenting and trying to figure out how css `composes` actually works ...

app/components/Footer/Footer.css
```diff
@@ -1,61 +1,155 @@
 @import '~app/styles/variables.css';
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&display=swap');
```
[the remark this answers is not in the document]
ivarnakken: I totally agree! I placed it along with the other font imports in `pageRenderer`.

PeterJFB (approved):
Love the transitions :heart_eyes: , great work!

erlingfn (approved):
Looks good!

SmithPeder:
I am not a fan of things that are 100% black or 100% white. How about if we make the headers dark red?
image

ivarnakken:
> I am not a fan of things that are 100% black or 100% white. How about if we make the headers dark red?

Good point, and I think a dark red looks good. However, in accordance with the principles of universal design, I'm not sure whether the contrast between the text and its background is great enough. This goes especially for the text at the bottom with the dark red background. Perhaps we can change the color of this, and let the rest be a dark red 🤔

SmithPeder:
yeah, I agree that the "cookies" is way off, that just happened when I applied some CSS elsewhere. 🤫 
The main elements are LEG, GithubLogo,  KONTAKT OSS, Facebook, Istagram etx. Those really stand out when they are black. I think the rest can be white, or "white" like #eee or somthing

ivarnakken:
I made some color updates. What do you think, @SmithPeder? 

result

SmithPeder:
A major improvement!!

What is the plan for this? Merge and release right away?!!

ivarnakken:
> What is the plan for this? Merge and release right away?!!

Since abakus.no don't really get much traffic during the winter break anyways, I'm actually fine with releasing it.

SmithPeder:
Nice!

SmithPeder:
That's very agile of you, if I must say so myself

PR #2383, opened 2021-12-29
Add mazemap embed to events and meetings

ollfkaih:
Adds a map to the sidebar of events and meetings if the event or meeting has mazemapPoi set,  
and adds fields to the edit pages to be able to search for  rooms and set the mazemapPoi.

Solved the jest/test issues by bundling mazemap as a node module instead of in pageRenderer, based on [this](https://api.mazemap.com/js/v2.0.63/docs/#ex-jsapi-module-import-react-project).

Minor issue:
 - <del> The saved room is not shown in the field, but is shown in the dropdown, because I have not figured out how to fetch the name of the room when EventEditRoute is rendering. </del>

^ Fixed by using location saved on server

Oh, and this is also based on Tirils prior work in pr #1680 

Screenshot for anyone curious  
Screenshot


app/components/Search/mazemapAutocomplete.js
```diff
@@ -0,0 +1,113 @@
...
+
+function mazemapAutocomplete<Props>({
+  WrappedComponent,
+}: {
+  WrappedComponent: ComponentType<Props>,
+}) {
+  const displayName =
+    WrappedComponent.displayName || WrappedComponent.name || 'Unknown';
+
+  const mySearch = new Mazemap.Search.SearchController({
```
SmithPeder, line 27: Maybe call this something else 😄 Something descriptive that fits with the params or to indicate that it's a mazemap search

app/components/MazemapEmbed/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+/** A component that shows a mazemap map of a given poi (e.g. room),
+ * largely based on https://api.mazemap.com/js/v2.0.12/docs/#ex-data-poi
+ */
+
+export const MazemapEmbed = ({ sharepoi, ...props }: Props) => {
+  //useEffect to initialize only once, sharepoi will probably not change
+  useEffect(() => {
+    // $FlowFixMe Mazemap is defined in script tag, has no types
+    const myMap = new Mazemap.Map({
```
SmithPeder, line 19: Very minor, but I don't like variable names like `myMap` and `mySearch` 🙃 Maybe something that describes where this map is centered at, or what it aims to show?

app/components/MazemapEmbed/index.js
```diff
@@ -0,0 +1,90 @@
...
+      <div style={{ height: props.height || 400 }} id="mazemap-embed" />
+      <a
+        href={
+          'https://use.mazemap.com/#v=1&sharepoitype=poi&campusid=1&sharepoi=' +
+          sharepoi
+        }
+        rel="noreferrer noopener"
+        target="_blank"
+      >
+        Åpne kart i ny fane
```
SmithPeder, line 86: Smart 💯

app/components/Search/mazemapAutocomplete.js
```diff
@@ -0,0 +1,113 @@
...
+      this._isMounted = true;
+      if (this.props.meta.initial) {
+        Mazemap.Data.getPoi(this.props.meta.initial.value).then((poi) =>
+          this.setState({
+            result: [
+              {
+                label:
+                  poi.properties.title + ', ' + poi.properties.buildingName,
+                value: this.props.meta.initial.value,
+              },
```
SmithPeder, line 57: I would create a helper method that creates this result `{object}`. It can take the title, name, and value as params and return this. Then it can be used within `mySearch.search()` _(which should be called something else)_.

app/components/Search/mazemapAutocomplete.js
```diff
@@ -0,0 +1,113 @@
...
+              searching: false,
+            });
+          }
+        })
+        .catch(() => {
+          if (this._isMounted) {
+            this.setState({ searching: false });
+          }
+        });
+    };
```
SmithPeder, line 98: I would just both of these in a finally.
```
.finally() => {
          if (this._isMounted) {
            this.setState({ searching: false });
          }
}
```
It makes little difference, but I think it's more clear that way as it shows the reader that a mounted component must set `searching: false` in all cases.

app/components/MazemapEmbed/index.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+    myMap.on('load', () => {
+      // Initialize a Highlighter for POIs
+      // Storing the object on the map just makes it easy to access for other things
+      myMap.highlighter = new Mazemap.Highlighter(myMap, {
+        showOutline: true, // optional
+        showFill: true, // optional
+        outlineColor: Mazemap.Util.Colors.MazeColors.MazeBlue, // optional
+        fillColor: Mazemap.Util.Colors.MazeColors.MazeBlue, // optional
+      });
```
SmithPeder, line 40: Maybe we don't need the `// optional` docs from `mazemap` here?

package.json
```diff
@@ -106,7 +106,8 @@
     "validator": "^13.7.0",
     "vm-browserify": "^1.1.2",
     "websocket.js": "^0.1.7",
-    "zxcvbn": "^4.4.2"
+    "zxcvbn": "^4.4.2",
+    "mazemap": "file:./vendor/mazemap"
```
SmithPeder, line 110: I agree on the choice of a package.json `file:./vender` for this plugin. I looked at your `[WIP]` and was going to suggest you do it this way! Nice that you figured it out 🥇

SmithPeder (commented):
Very good work.

A couple of comments. _(I guess some of them apply to the previous work done by Tiril, but yeah, you'll get the credits and the blame) 😆_ 

1. (Major) When we select a `MazeMapRoom`, it's kinda wired that we also have to fill in the normal `Location`. I get that sometimes one wants to write something in both, but if the event is in R1, then the `mazemapPoi` field will be translated to **R1, Realfagbygget** so writing anything in the **Location** is wired. This is kind of a business-logic problem, but something that should be addressed. I think the solution is to make the fields co-required, meaning at least one has to be filled in.
2. (Minor) When selecting a room in a create or edit view the map is not visible. I can see arguments in favor of hiding a map in this view. But, when trying to select kinky rooms it could be nice to see that the correct room is selected. We could add the _Åpne kart i ny fane_ label under, so the creator can open the map to check that the correct room was selected.

app/components/MazemapEmbed/index.js
ollfkaih: Yes, I guess they're _optional_.

app/actions/MeetingActions.js
```diff
@@ -136,6 +137,7 @@ export function createMeeting({
       endTime: moment(endTime).toISOString(),
       startTime: moment(startTime).toISOString(),
       reportAuthor: reportAuthor && reportAuthor.id,
+      mazemapPoi: mazemapPoi.value > 0 ? mazemapPoi.value : null,
```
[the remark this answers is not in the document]
ollfkaih: The positive check is indeed unnecessary, but if mazemapPoi.value is undefined, it isn't sent in the request to the server, which means the field can be cleared, but the meetings' mazemapPoi will not be reset. Null is a change, however, and is added to the request.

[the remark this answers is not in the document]
SmithPeder: Ohh, you're right 💯  Didn't think about that.

app/components/Search/mazemapAutocomplete.js
```diff
@@ -0,0 +1,113 @@
+// @flow
+
+import React, { Component, type ComponentType } from 'react';
```
ivarnakken, line 3: You don't need to import React 😄

ollfkaih:
I share your sentiment on having only one location field at a time. Perhaps a checkbox ("[x] Bruk mazemap") to toggle whether you type in the freetext location, or if you search for a room and then location gets set to that rooms displayname, would be a good solution.

I skipped adding a map to the event view because I wanted to keep the scope down for this PR so that i could finish it, but did originally consider displaying search results on the map. It would be easy, however, to display the room you've selected (not all results). This would make the sidebar even longer, unless the map is displayed below the description, but that would be a weird place to have the map when the search field is in the sidebar... A link seems like a decent solution.

Also, I'm no UX designer, but I think the colors should be more _Abakus_ and less jarring (blue polygon and purple marker). I'll change them to something more red, but some feedback is appreciated. 
embed-red

SmithPeder:
> the colors should be more Abakus and less jarring (blue polygon and purple marker). I'll change them to something more red, but some feedback is appreciated.

I think the red looks great 👍🏻  Maybe @ivarnakken has some nit-picking 😆 

> This would make the sidebar even longer, unless the map is displayed below the description, but that would be a weird place to have the map when the search field is in the sidebar... A link seems like a decent solution.

yeah, I agree that showing the embedded map can be a bit much. I was thinking a link, or a "Open in new window icon" like this:
Screen Recording 2022-01-08 at 14 30 04

ollfkaih:
> a "Open in new window icon" like this

That is an elegant solution. 
Also, there are some oddities with the map's zLevelControl when changing rooms (it is not removed) due to how mapbox handles its container, so it would require some changes to the embed to display changing mazemapPois on the same map (one could of course create a completely new div on each change, but that seems wasteful).

SmithPeder:
> I share your sentiment on having only one location field at a time. Perhaps a checkbox ("[x] Bruk mazemap") to toggle whether you type in the freetext location, or if you search for a room and then location gets set to that rooms displayname, would be a good solution.

yeah, something like that. It's still nice to have the location field set, because the location is used in multiple places, like here
image

ollfkaih:
I've just added a toggle that switches between mazemap/location fields. Someone more familiar with redux forms might be able to simplify this code, but it seems to work correctly.

app/routes/meetings/components/MeetingEditor.js
```diff
@@ -128,11 +138,43 @@ function MeetingEditor({
...
+              component={SelectInput.MazemapAutocomplete}
+              fieldClassName={styles.metaField}
+              placeholder="R1, Abakus, Kjel4"
+            />
+            {
+              // TODO: Bug; meeting is not updated when the field above is changed, so the link
+              // will go to whatever room mazemapPoi was set to when the editor was opened
+              /*meeting.mazemapPoi > 0 && (
+              <MazemapLink mazemapPoi={meeting.mazemapPoi} linkText="↗️" />
+            )*/
```
ollfkaih, line 174: In the editor for events, I've also added a small icon to link externally to mazemap for the given poi, but I couldn't make this work for meetings due to `meeting` not updating on every change to the field, like `event` does in the event edtior. I've left this as a comment for now.

.drone.yml
```diff
@@ -133,7 +133,7 @@ steps:
       - echo "server stopped"
 
   - name: setup
-    image: node:14-alpine
+    image: node:14
```
ivarnakken, line 136: Do we still need `node:14` now that `curl` is being installed manually?

eikhr: We do not. I've fixed this in https://github.com/webkom/lego-webapp/tree/ollfkaih/master. @ollfkaih do you want to update your branch with my newest changes?

ollfkaih: Tror det ble riktig nå, evt. dobbeltsjekk hashen i drone.yml

SmithPeder (approved):
LGTM 👍🏻

PeterJFB (approved):
Looks good to me! :heart_eyes:

app/routes/events/components/EventEditor/index.js
```diff
@@ -79,6 +81,13 @@ function EventEditor({
   initialized,
 }: Props) {
   const isEditPage = eventId !== undefined;
+
+  const [useMazemap, setUseMazemap] = useState<boolean>(false);
+
+  useEffect(() => {
+    setUseMazemap(event.mazemapPoi && event.mazemapPoi?.value > 0);
+  }, [initialized]);
+
```
LudvigHz, line 90: You could have done this like we have with the `separateDeadlines` for consistency, but seeing as this is the more "modern" method and we avoid some redux I think this probably is better :+1:

app/routes/meetings/components/MeetingEditor.js
LudvigHz: The reason for this is that the state of the form (both the event and the meeting, is stored in the `form` state in redux. This is extracted for each of the editors, so that we can use the current form values in the components again.

If you look in the `EventEditRoute`, in the `mapStateToProps`, you can see that the `event` prop consists of the current `event` **and** all the values from the event form overriding any properties. This is the `valueSelector` that is used.

You can also see in the `MeetingEditRoute` that the valueSelector is used to get the `invitingUsers`. So if you want to use the current `mazeMapPoi` form value, just override that prop on the `meeting` object with the one from the valueSelector just as you have done with events. Then this should work great :smile:

app/components/MazemapEmbed/index.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }, [mazemapPoi]);
+
+  return (
+    <>
+      <div style={{ height: props.height || 400 }} id="mazemap-embed" />
+      <MazemapLink mazemapPoi={mazemapPoi} linkText={props.linkText} />
+    </>
+  );
+};
```
LudvigHz, line 82: Is this tested with the server-side renderer? I have a slight suspicion it might break.

It can be run with `yarn:ssr` and if there are any errors in the console or/and you can see the page loading (some js doing something) it's broken.

I don't have time to test this myself, but I'm sure someone else (@webkom/aktive) can do this real quick and make sure it's a-okay.  If not, we need to do some changes to make sure it doesn't break on the SSR (conditionally loading on the client f.ex.)

LudvigHz (changes requested):
Nice work! This is a really great feature :-1:. Just fixup the comment/TODO in the meeting editor and make sure SSR is good then we can merge this.

app/routes/events/components/EventEditor/index.js
ollfkaih: :)

app/components/MazemapEmbed/index.js
ollfkaih: I think your suspicions were correct. Map does load, but the console dislikes mazemap.min.js :/
```js
loadable-components: failed to synchronously load component, which expected to be available {  
  fileName: 85469,
  chunkName: 'meetings',
  error: 'Element is not defined'
}
08:15:38.559Z ERROR lego-webapp: render_error
    ReferenceError: Element is not defined  
        at eval (webpack://Mazemap/./src/Util/Polyfills.js?:21:5)
        at Object.5516 (/home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:99125)
        at __webpack_require__ (/home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:1911363)
        at eval (webpack://Mazemap/./src/mazemap.js_+_124_modules?:412:17)
        at Module.2061 (/home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:104806)
        at __webpack_require__ (/home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:1911363)
        at /home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:1911965
        at /home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:1912019
        at /home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:81
        at Object.<anonymous> (/home/debian/lego-webapp/node_modules/mazemap/mazemap.min.js:1:195)
```

LudvigHz: Yeah, that's what I feared. I think the best way to do this, is to not render the mazemap component on first render, for then to conditionally render it on the client after hydration. I.e. Allocate some "placeholder" wrapper where the map should be and only render that (so that the page content does not jump around, with the exact dimensions before the map is loaded).

You can then, on the client, render the component after the app is hydrated and the wrapper component is mounted. The proper way to do this is simply with a `useEffect` hook in the wrapper (since these don't run on render, only on mount). [This guide](https://tech.willhaben.at/how-to-shoot-yourself-in-the-foot-in-react-401e7dbba720)[^1] covers it pretty good how you should do this.

[^1]: He also links to [this article](https://www.joshwcomeau.com/react/the-perils-of-rehydration/) in the guide, which goes more in-depth into the problem if you are interested

ollfkaih: So I've fixed it for events by dynamically importing mazemap (seems the offending code is in the mazemap module), but meetings still have the issue for some reason.

Lazyloading seemed like a solution (```const RawMazemap = lazy(() => import('./RawMazemap'));```) where `RawMazemap` is the real mazemap component, and we wrap it in MazemapEmbed.  However, `Suspense` is experimental or not supported with ReactDOM, so I discarded that. 

I don't understand why only meetings get this error now, it seems to load `MazemapEmbed` in the exact same way as events. Maybe a wrapper that dynamically imports another component, which imports `mazemap` works, or maybe a small change to meetings fixes it?

LudvigHz: Lazyloading won't help here. Since the problem is in the mazemap module using DOM/browser APIs that are not available in Node.js. So the only way to fix the issue is as I stated in the comment and not rendering it on the server at all. 

Nice work with events, although I don't see why you need to dynamically import it? Unless the mazemap module contains code that runs on require? (which it really should not)

If that is the case, you could try lazy loading I guess. As you stated, native lazy loading does not support SSR yet, so we use react-loadable for this purpose. See the documentation and dynamic imports elsewhere in the app for usage.

ollfkaih: I tried using require('mazemap'), but then the element not defined error happens with both events and meetings, so it would seem mazemap runs some code on import.

LudvigHz: Right. But did you also just use es `import` as before? Importing es modules should not run any code, in contrast to cjs `require` statements.

ollfkaih: I used `import * as Mazemap from 'mazemap'` before, which gave the same error.

LudvigHz: Hmm that's strange. Did you remember to run `yarn build` and restart the SSR+server?

ollfkaih: I rebuilt and restarted everything on the frontend.

LudvigHz: Hmmm. Then it might run some code on import then. I would imagine the code you have with dynamic imports works then. So very strange of it doesn't work only on Meetings.

ollfkaih: Exactly, I cannot understand why it fails only with meetings when events use the exact same component.

LudvigHz: Soo, I just tested this with the latest from your branch, and both events and meetings seem to work just fine? :smiling_face_with_tear: 

Note that sometimes just after you start up the server, you might need to hard refresh a couple times before it's completely stable (due to caching and some weird states and all that). But anyways, this looks really good now. And afaikt this is good to go!

ollfkaih:
Oh, if that's the case I guess you can merge it.😁
________________________________

LudvigHz:
> Oh, if that's the case I guess you can merge it.grin

Actually, hold that thought. It seems that for me, joblistings is broken with the `Element is not defined` error coming from Mazemp :thinking:. I don't know why that is, but it seems that the mazemap library is being loaded in some places when it absolutely should not. So I think this needs some more investigating sadly

app/components/MazemapEmbed/index.js
LudvigHz: Riiight! @ollfkaih I found the problem. The issue is not in the mazemapEmbed component. That works great. But the same issue occurs in the `mazemapAutocomplete` component of course. As that loads the same library. Why it works in some places and not others probably has to do with the way `reactDOM`, `webpack` and `node` traverses the dependency tree when rendering the app.

So just do the same thing in the `mazemapAutocomplete` component and then everything should work just dandy!

ollfkaih:
Could it be in or related to `SelectInput.js`?

LudvigHz:
> Could it be in or related to `SelectInput.js`? Edit: had some faulty uncommited code locally, but reproduced the errors with joblistnings.

See my latest reply in the thread here: https://github.com/webkom/lego-webapp/pull/2383#discussion_r802080658

app/components/Search/mazemapAutocomplete.js
```diff
@@ -51,67 +41,64 @@ function mazemapAutocomplete<Props>({
...
+                ),
+              });
+            }
+          })
+          .finally(() => {
+            if (this._isMounted) {
+              this.setState({ searching: false });
+            }
+          });
+      };
```
LudvigHz, line 96: Hmm, why did you move these methods into `render()`? This will make every render heavier as it sets up new functions for every render and not when creating the component as before.

Also, it's probably alot better to initialize the `Mazemap.Search.SearchController` only once in the mount when you do the import, and rather store that in state instead of the entire Mazemap lib, since it seems that that's the only thing used here.

ollfkaih: Good questions all around, I've done some cleanup and moved the methods back out of render(), in what is hopefully the final™ commit on this pr.

LudvigHz (approved):
Nice! :shipit: :rocket: :rocket: :rocket:

eikhr (commented):
LTGM! 💯 But it seems like there are some small flow problems

eikhr (approved):
Great!

PR #2422, opened 2022-01-18
Remove slate-editor cypress workarounds

eikhr:
As of Cypress v5.5 the type() command works fine with the slate editor so our workarounds are no longer needed. This also allows the editor test to be run on all browsers, not just Firefox, so i moved it into the same file as all the other event tests (event_editor_spec.js)

I also removed all editorFocus() commands, as it seems that clicking the editor also focuses it.

Closes webkom/lego#2353


EikHR:
That makes sense. I just assumed the reason it was in its own file was the browser incompatibility.

SmithPeder (approved):

PR #2425, opened 2022-01-18
Remove unused and outdated code docs

erlingfn:
As of now, we still pretend like we actually have code documentation in the frontend. Let us stop pretending. Nobody has written docs for components in the last four years, and the styleguidist server is broken. This leaves us with unnecessary dependencies, an outdated readme and lots of files that serve no purpose. While code docs are great, and it would be useful to have it, it does not warrant leaving useless and unused docs around. If we were to begin documenting our components and styles it would be simpler to set it up from scratch rather than basing it on 4 year old code written by people that have left.


docs/layout.md
```diff
@@ -1,38 +0,0 @@
...
-```
-
-and put stuff inside them.
-
-They support the props `alignItems`, `justifyContent`, `padding`, `flex` and
-`width`, which should be sufficient for most purposes (the default values are
-usually fine).
-
-These components are only responsible for how items are laid out, and have no
-opinions on colors etc. They are just `div` elements with flex styling.
```
LudvigHz, line 38: Some of there are quite nice (and still relevant) actually

LudvigHz (approved):

docs/layout.md
erlingfn: I agree that they are nice to have, but if they are just laying around as an md file, how often would they be used? We could of course leave them around, but imo they have little use until we get a system and habits for frontend docs up and running.

SmithPeder: @LudvigHz Do you really think so? I feel like everyone looks at the current code, and not the docs. Have you ever looked at these before today? I say we just delete them, and can `git-revert` them back if you ever need them

SmithPeder (approved):

PR #2432, opened 2022-01-25
Fix react-helmet bug by introducing react-helmet-async

ivarnakken:
`react-helmet-async` seems like just the thing we need, but it was a bit difficult to test locally, but it seemed to work. Notice that I had to remove a test because it was behaving weirdly and impossible to debug, but was extensively tested manually and it worked as expected.


SmithPeder (approved):
Nice. Let's see if it works in staging

PR #2439, opened 2022-01-30
Add qoute preview

SmithPeder:
This is the most useless feature ever, but people keep creating quotes "wrong". This will show them how their quote will look

image


ivarnakken (approved):
Not useless at all, it has been bugging me a lot as well! 💯

erlingfn (approved):
This is not the most by the book way to create it, but it bugs me a lot as well, so I will accept it.
Ideally one would refactor the old `RandomQuote` component, but one has to be pragmatic.

PR #2503, opened 2022-02-22
Bump stylelint

ivarnakken:


app/components/Button/Button.css
```diff
@@ -30,7 +30,7 @@
   }
 
   &:hover:not(:disabled) {
-    box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
+    box-shadow: 0 1px 2px rgba(0, 0, 0, 10%);
```
LudvigHz, line 33: I hate this rule change, but oh well :shrug:

app/routes/quotes/components/Quotes.css
```diff
@@ -123,15 +123,6 @@
   }
 }
 
-.rightQuote {
```
LudvigHz, line 126: Are these all unused?

LudvigHz (approved):
niiiiice :100:

SmithPeder (approved):
Looks good, but should still some testing 😂

app/components/Button/Button.css
ivarnakken: I kinda regret not changing the rule and sticking with the decimal format 🤔 I might change back, but not tonight ...

app/routes/quotes/components/Quotes.css
ivarnakken: Indeed

PR #2529, opened 2022-03-01
Add anonymous attendance for deleted users

theodorsm:
Before user (webkom) deletion: 

image
image

After user (webkom) deletion:

image
image


theodorsm:
In the screenshots above a deleted user will increase the legacy count since the event has already started. If a user is registered for an event, but the event is in the future, a user deletion would not increase the legacy count.

PeterJFB (approved):
Simple and structured, lgtm^^

PR #2534, opened 2022-03-06
General small style changes

SmithPeder:
There have always been small CSS things that bug me every day. This fixes some of them.

I know the new frontpage will eventually fix this, but I just need something temp

### Festet Oppslag content spacing was off
image
image

### Articles are too big, and people don't write that much "description", and again the content spacing was off
image
image

### It annoyed me that the bottom of the articles and events didn't line up
image
image

### Title [Bedpress og Kurs] was not indented like all the other headers
image
image

### Saving an article was dumb
image
image

### DIFF
--- a/app/components/Reactions/assets/AddReactionEmoji.js
+++ b/app/components/Reactions/assets/AddReactionEmoji.js
@@ -10,7 +10,7 @@ const AddReactionEmoji = ({ color }: Props) => (
   <div className={styles.container}>
     <span>
       <svg
-        ariaHidden="true"
+        aria-hidden="true"
         focusable="false"
         width="0.44em"
         height="1em"
@@ -27,7 +27,7 @@ const AddReactionEmoji = ({ color }: Props) => (
     </span>
     <span className={styles.smiley}>
       <svg
-        ariaHidden="true"
+        aria-hidden="true"
         focusable="false"
         width="1em"
         height="1em"
--- a/app/routes/overview/components/CompactEvents.css
+++ b/app/routes/overview/components/CompactEvents.css
@@ -10,7 +10,7 @@
 
 .compactEvents h3 {
   margin-bottom: 10px;
-  padding: 0;
+  padding: 0 10px;
 }
 
 .compactLeft {
--- a/app/routes/articles/components/ArticleEditor.js
+++ b/app/routes/articles/components/ArticleEditor.js
@@ -70,7 +70,6 @@ const ArticleEditor = ({
       <Form onSubmit={handleSubmit}>
         <Field
           name="cover"
-          label="Cover"
           component={ImageUploadField}
           aspectRatio={20 / 6}
           img={article && article.cover}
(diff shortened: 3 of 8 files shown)


app/components/Reactions/assets/AddReactionEmoji.js
```diff
@@ -10,7 +10,7 @@ const AddReactionEmoji = ({ color }: Props) => (
   <div className={styles.container}>
     <span>
       <svg
-        ariaHidden="true"
+        aria-hidden="true"
```
SmithPeder, line 13: and this Warning annoyed the shit out me

LudvigHz: LOL yeah, this one's a goodamn bi***

app/components/Button/Button.css
```diff
@@ -89,3 +89,8 @@
   border: none;
   cursor: pointer;
 }
+
+.danger {
+  background: var(--color-red-3);
+  color: var(--color-white);
+}
```
LudvigHz, line 96: Don't we already have a "delete" button styling?

app/routes/overview/components/CompactEvents.css
```diff
@@ -10,7 +10,7 @@
 
 .compactEvents h3 {
   margin-bottom: 10px;
-  padding: 0;
+  padding: 0 10px;
```
LudvigHz, line 13: I'm not so sure about this. I feel like the old style is better. The indented title looks kinda "off" to me :thinking:

LudvigHz (approved):
Nice :100: Always nice to fix the small, annoying things.

app/components/Button/Button.css
SmithPeder: We had a danger prop, that did nothing 🤷🏻‍♂️

SmithPeder: Ahh, wait, the "dark" is red... So dark is actually "danger/delete"
That's...

app/routes/overview/components/CompactEvents.css
SmithPeder: I was going back and forth for like 15min, haha. Both look kinda "off", but I feel this looks "less" off somehow. But I'm open for a vote

erlingfn: I agree with @LudvigHz on that it looks a bit off, but perhaps we should be consistent on the  headers?

ivarnakken (approved):
✨✨

app/routes/overview/components/CompactEvents.css
juniwbjerde: I don't really care if they're indented or not, but they should for sure be the same as the rest of the headers. So if we go back to the old style we should change all the other header to fit that styling too

juniwbjerde (approved):

app/routes/overview/components/CompactEvents.css
eikhr: I agree that it looks a little off, but I think it looks better than it being inconsistent

eikhr (approved):

PR #2541, opened 2022-03-08
Disallow eslint warnings

eikhr:


app/routes/events/components/StripeElement.js
```diff
@@ -283,7 +278,7 @@ const PaymentRequestForm = (props: PaymentRequestFormProps) => {
     return () => {
       completePaymentManual('fail');
     };
-  }, []);
+  }, [completePaymentManual]);
```
eikhr, line 281: @LudvigHz Can I do this? I'm assuming it won't change anything as completePaymentManual is a const, but I don't understand all the Stripe code.

LudvigHz: The point of this code is just to cancel the payment in case the component unmounts. So if the payment status is suddenly accepted and we need to cancel a started payment, we can do that safely.

As far as I can see, this shouldn't break anything though. But we should test it to make sure.

ivarnakken (approved):
It'll be nice to get rid of these 💯

SmithPeder (approved):

LudvigHz (approved):

app/routes/meetings/components/MeetingEditor.js
```diff
@@ -60,8 +60,12 @@ function MeetingEditor({
   const [useMazemap, setUseMazemap] = useState<boolean>(false);
 
   useEffect(() => {
-    setUseMazemap(isEditPage && meeting.mazemapPoi?.value > 0);
```
ivarnakken, line 63: @eikhr, I just noticed we removed the `isEditPage` condition, why? 🤔 The whole reason we added the optional chaining yesterday was becuase `isEditPage` was already somehow changed to `meeting.mazemapPoi`, right?

PR #2542, opened 2022-03-09
Don't show "0" character when legacyRegistrationCount is 0

eikhr:
Screenshot 2022-03-09 at 11 31 03


theodorsm (commented):
Woooopsie, good catch!

LGTM

PR #2545, opened 2022-03-09
Hide registrationDeadline when heedPenalties is false

AdrianAndersen:
Hide the registration deadline when heedplenalties is false. This deadline only determines when you should get a penalty for unregistering too late.\nWhen an event does not heed penalties, this deadline is redundant.


ivarnakken (approved):
I like 💯

eikhr (approved):
Makes sense

SmithPeder (approved):

PR #2554, opened 2022-03-12
Improve button class design

AdrianAndersen:
The current button design is quite poor. It needs more difference between the disabled and clickable button. (Especially in dark mode) Therefore, I suggest making the following changes.

Feel free to suggest changes to the colours / shadows etc! 🧑‍🎨 

#### Before / After
imageimage 

#### Before / After
imageimage

### DIFF
--- a/app/components/Button/Button.css
+++ b/app/components/Button/Button.css
@@ -1,17 +1,16 @@
 @import '~app/styles/variables.css';
 
 .button {
+  box-shadow: 1px 1px 3px 1px var(--color-gray-5);
   font-weight: 600;
-  background-color: var(--color-mono-gray-4);
-  background-image: linear-gradient(
+  background: linear-gradient(
     -180deg,
-    var(--color-mono-gray-5) 0%,
-    var(--color-mono-gray-4) 90%
+    var(--color-red-3) 0%,
+    var(--color-red-2) 90%
   );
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):
Nice! I've wanted to change up our buttons for so long!😍

LudvigHz (commented):
Nice! I definitely agree our buttons don't look the best.

Although I strongly suggest **not** using red as a main button color, as we use that for all delete and danger buttons.

AdrianAndersen:
> 

Yeah, it resembles that. However I think it is fine since it matches our theme. We could do something about our critical buttons. For instance having an icon, or a “danger zone” like GitHub. What do you think, @LudvigHz?

LudvigHz:
> > 
> 
> Yeah, it resembles that. However I think it is fine since it matches our theme. We could do something about our critical buttons. For instance having an icon, or a “danger zone” like GitHub. What do you think, @LudvigHz? Or do you have a suggestion for another good color?

I think that red buttons are so unanimously used as a "danger" button across the web that it doesn't matter if it matches our primary color. I think it's a bad idea (tbh my brain hurts when looking at the pool example :upside_down_face:). So for how they should look I'm not too sure, maybe a white/gray button with some border or something? Or we could use another color in our palette (or add one?)

PR #2560, opened 2022-03-15
Fix progressive images showing wrong image

eikhr:
Fixes a bug where progressive images don't update when the src prop is changed. I also rewrote `ImageComponent` to be a functional component.

Also fixes https://github.com/webkom/lego/issues/2558

### DIFF
--- a/app/components/Image/Image.js
+++ b/app/components/Image/Image.js
@@ -20,6 +20,7 @@ const ImageComponent = (props: Props) => {
     props.placeholder || EMPTY_IMAGE
   );
   const [imageLoaded, setImageLoaded] = useState<boolean>(false);
+  const [imageError, setImageError] = useState<boolean>(false);
   const [loadStart, setLoadStart] = useState<number>(0);
   const [loadEnd, setLoadEnd] = useState<number>(0);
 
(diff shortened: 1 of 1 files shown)


LudvigHz:
_Posting an excerpt of the discussion on slack so everyone can see it:_

The reason this component was not written as a functional component, is due to incompatibility with `react-prepare`, specifically with some hooks. For the uninitiated, this means that the use of this component with hooks will break the server-side renderer (SSR). It's a bit hit-and-miss whether some components will work or not, but this did not work at all, no matter how I changed the component and tried different combinations of hooks and logic.

We know this to be an issue, and to be able to fix these issues, we would need to do a lot of changes in `react-prepare`, which as we have discussed earlier, is a huge PITA and just not worth it. Class components still work great, and there's no rush to convert everything to functional components just yet, even though it would be nice.

So I would recommend to just fix the issue and ignore the refactoring to function components and hooks.

eikhr:
Seems to work well with the updated [react-prepare](https://github.com/webkom/react-prepare/pull/7) :) Once we get that merged and published on npm this should be good to go!

ivarnakken (approved):
😍

erlingfn (approved):
LGTM

LudvigHz (approved):
Seems to work nicely
LGTM :shipit:

PR #2567, opened 2022-03-18
Remove excess quote#2587

ulrik2204:
Closes webkom/lego#2587 

Now the random loaded quote is not in the list of quotes when going to the "/quotes" route. 

This is achieved by clearing the list of quotes in the redux state the first time the request for quotes it done, but not when "Last inn flere" is clicked.

### DIFF
--- a/app/actions/QuoteActions.js
+++ b/app/actions/QuoteActions.js
@@ -1,13 +1,13 @@
 // @flow
 
+import callAPI from 'app/actions/callAPI';
+import { addToast } from 'app/actions/ToastActions';
+import type { ID } from 'app/models';
+import { quoteSchema } from 'app/reducers';
+import type { Thunk } from 'app/types';
 import { push } from 'connected-react-router';
 import { startSubmit, stopSubmit } from 'redux-form';
-import { quoteSchema } from 'app/reducers';
-import callAPI from 'app/actions/callAPI';
 import { Quote } from './ActionTypes';
-import { addToast } from 'app/actions/ToastActions';
-import type { Thunk } from 'app/types';
-import type { ID } from 'app/models';
 
 const getEndpoint = (state, loadNextPage, queryString) => {
   const pagination = state.quotes.pagination;
@@ -38,6 +38,9 @@ export const fetchAll =
     const endpoint = getEndpoint(getState(), loadNextPage, queryString);
     if (!endpoint) {
       return Promise.resolve(null);
+    }
+    if (!loadNextPage) {
+      dispatch({ type: Quote.CLEAR });
     }
     return dispatch(
       callAPI({
--- a/app/actions/ActionTypes.js
+++ b/app/actions/ActionTypes.js
@@ -215,6 +215,7 @@ export const Quote = {
   UNAPPROVE: (generateStatuses('Quote.UNAPPROVE'): AAT),
   DELETE: (generateStatuses('Quote.DELETE'): AAT),
   ADD: (generateStatuses('Quote.ADD'): AAT),
+  CLEAR: 'Quote.CLEAR',
 };
 
 /**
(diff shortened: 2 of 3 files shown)


ulrik2204:
This is the current state when loading an old quote. 
image

After the fix
image

ulrik2204:
Before the fix:
Screenshot from 2022-03-19 00-49-50


After the fix:
Screenshot from 2022-03-19 00-52-47

erlingfn (approved):
Looks nice!

eikhr (approved):
LGTM 👍

app/actions/QuoteActions.js
```diff
@@ -39,6 +39,9 @@ export const fetchAll =
     if (!endpoint) {
       return Promise.resolve(null);
     }
+    if (!loadNextPage) {
+      dispatch({ type: Quote.CLEAR });
+    }
```
LudvigHz, line 44: When we navigate from another part of the page, this will clear the quote state regardless of the fact that the requested items have already been loaded. I.e. this will fix the edgecase you want, but at the same time, it will cause the page to load every time we navigate from somewhere to `/quotes`.

You can probably see this if you go frontpage -> /quotes -> <somewhere else, f.ex /events> -> /quotes.
In the current version, the page won't load in the last step.
With this change, it will load everything again.

LudvigHz (changes requested):
The idea is nice, however the drawback in the mentioned comment is in my mind larger than the positive sides of this. So if you want to fix the mentioned issue, you need to find someway to do it without introducing this behavior. Some ideas are:
- You might be able to modify the selector, or add a new one, that selects only quotes fetched from the backend directly based on the sorting, or omit when the id gap is too large (prob not a good idea). 
- Maybe add a flag to the "random quote" and either remove that or omit it from the selector.
- Or some other way.

app/actions/QuoteActions.js
ulrik2204: Your mentioned issue is the behaviour I expect when re-entering a list view. Every time I click on "Overhørt" again after being somewhere else and get routed to the middle of the previously loaded list of quotes, I am surprised and may sometimes reload the page. To me, going away from the list view of quotes and re-entering it, is like the user telling us "please update the list of quotes". Thus, I do not see this as an issue. 

This is, of course, just my opinion. If you want to keep the current behaviour where all the quotes are kept in redux while you navigate other parts of the website, I think there are two alternatives:
1. As you mentioned, to add a flag to each randomly retrieved quote and only clear the quotes with this flag. 
2. To make the random quote not be added to the redux state. This does, however, conflict with the structure where every request goes through redux.

I would like your answer before any more changes are made. Should we keep it as I suggested, add the flag and filter (1), or remove random quotes from redux (2). I think (2) is best, but it conflicts with the rest of the project structure.  

The overarching problem is that most API requests on the website are going through redux and are added to the global redux state. This makes the code overcomplicated and creates problems like this. However, fixing the redux problem will take a lot more time.

AdrianAndersen: We discussed this at our meeting, and think that it could be a viable solution to make a separate state for random Quotes in Redux. This would be in line with our architecture, even though it is a bit excessive. (Since it is only used for one component) What do you think about this approach, @ulrik2204 ?

LudvigHz: You could also use the same top-level `quote` state, and store the random quote under another key. So you just modify the reducer to store it here, and modify the selector to use this as well. In this case you don't need to create a whole new entityReducer and everything.

ulrik2204: Yeah, using another key seems like the best option. I'll see what I can do.

eikhr:
This was fixed in #2669

PR #2568, opened 2022-03-20
Add photo album download

celinaryholt:
Closes webkom/lego#2066 

I made it possible to download an entire photo album instead of having to download each photo one by one.

image


app/actions/GalleryPictureActions.js
```diff
@@ -30,7 +33,7 @@ export function fetch(
         },
         propagateError: true,
       })
-    );
+    ).then(() => SelectGalleryPicturesByGalleryId(getState(), { galleryId }));
```
erlingfn, line 36: I think this should be moved into the `downloadNext` function inside `Gallerydetail`, as it is only useful for that call, and is unnecessary for the other uses of fetch.

erlingfn (changes requested):
Great work! This creates a useful feature, and this code implements it well. Just a little nitpick.

app/actions/GalleryPictureActions.js
celinaryholt: Mathias Oterhals Myklebust
Good point! I was not able to move the selector to GalleryDetail, as this component does not have access to getState(). But I have now changed it to use the existing pictures prop instead, as this seems to be in sync. 
In addition, I have added a gallery clear action. This is because the image URLs expire one minute after retrieval. So if the user has had the page open for longer than this, the download will fail for the already fetched images. Let me know if this is a suitable solution.

celinaryholt: Good point! I was not able to move the selector to GalleryDetail, as this component does not have access to getState(). But I have now changed it to use the existing pictures prop instead, as this seems to be in sync. 
In addition, I have added a gallery clear action. This is because the image URLs expire one minute after retrieval. So if the user has had the page open for longer than this, the download will fail for the already fetched images. Let me know if this is a suitable solution.

LudvigHz (approved):
Nice! :100: 
This looks very good!

Ideally, we would do this on the server, both to offload the client, save on bandwidth to S3, and be able to cache the archive.

I think this a nice addition in any case, but IMO we should move this functionality to the backend in the future if we merge this.

erlingfn (approved):
Looks good! Agree that we should ideally have this in the backend, yet it is better to have it implemented here than to not have it implemented at all.

PR #2615, opened 2022-04-04
Fix fields using deprecated simpleValue prop

erlingfn:
The very old version we were using for react select used an old prop that made us able to simply use a string for the value, instead of an object. This was removed quite a few versions about. This is a fix that simply takes modifies the value sent in `onSubmit`. A more sophisticated alternative would be to implement the old `simpleValue` prop ourselves, but the only way I saw that working was by adding internal state to our react select field.


LudvigHz (approved):
This made it kinda ugly, but I don't see any other way to do it simply.

LGTM :shipit:

PR #2630, opened 2022-04-07
Automatically merge devdeps PR's from dependabot

SmithPeder:
I enabled it for dev-deps `minors`. This might a bit much. If it works we can maybe enable it for `patch` on normal deps :o


LudvigHz (approved):
LGTM :shipit: 
I don't think minor is much at all for dev deps. Ideally, if CI passes for dev-deps updates, regardless of version, it should be good to go :)

PR #2638, opened 2022-04-08
Change (fix) dark theme

LudvigHz:
This is my proposition for a **new** and **better** dark mode :art:! (I.e. just switching the main colors)

This also fixes some annoying stuff, like react-select components and input fields.

IMO this looks 1000 times better. Notably, the blue hint in the foreground colors are gone, and the foreground is lighter than the background. The current, opposite style where the foreground is darker than the background, is just jarring and makes no sense to me :slightly_smiling_face:  You notice this especially on event details and the info pages.

Generally, the foreground is lighter, with a bit less contrast. This means the whole page is lighter on the eyes.

Of course, everything is subjective though, so let me know what you think. (Maybe we could ask some users as well?) Personally, this is a middle ground for me, where it's at the limit of what I feel is acceptable in terms of darkness and contrast. It's _almost_ like I could use it myself :smile:.

## Frontpage

Before:
image

After:
image

## Event list

Before:
image

After:
image

## Event details

Before:
image

After:
image

## Om Abakus

Before:
image

After:
image

### DIFF
--- a/app/components/Form/SelectInput.js
+++ b/app/components/Form/SelectInput.js
@@ -24,6 +24,42 @@ type Props = {
   options?: {}[],
 };
 
+export const selectStyles = {
+  control: (styles: Object) => ({
+    ...styles,
+    cursor: 'pointer',
+  }),
+  option: (
+    styles: Object,
+    { isDisabled, isSelected }: { isDisabled: boolean, isSelected: boolean }
+  ) => ({
+    ...styles,
+    cursor: isDisabled ? 'not-allowed' : 'pointer',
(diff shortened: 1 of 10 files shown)


ivarnakken (approved):
Definitely a huge improvement! 😍 It might never be perfect, but this is a big step in the right direction!! 💯 Looks very good to me!

juniwbjerde (approved):
I totallt agree, this looks way better! Great work :heart_eyes:

SmithPeder (approved):

PR #2651, opened 2022-04-13
Improve styling of the Button component

ivarnakken:
I've introduced _danger_ and _success_ buttons, and changed up the default one as well. Please branch out to _styling_ to experience the buttons live, as I believe the images below is not sufficient.

And before you say so, yes, I was greatly inspired by GitHub's buttons 🙃

I'm planning on making a _rounded_, _ghost_ and _bordered_ button as well, but that can wait.

#### After
Both the green success button and red danger button are disabled, hence the opacity. 
<p float="left">
  
  
</p>

------

#### After
Default button and success button.
<p float="left">
  
  
</p>

------

#### After
Here, the "danger" button has a _dark_ style, as the danger's ghost style looked ugly next to the success button.

------

The danger button also works great with an icon, so I added some.
<p float="left">

</p>


app/components/Button/Button.css
```diff
@@ -1,52 +1,62 @@
...
+  padding: 5px 15px;
+  box-shadow: 0 1px 0 rgba(27, 31, 36, 10%),
+    inset 0 1px 0 rgba(255, 255, 255, 3%);
+  font-weight: 600;
+  background-color: var(--color-light-gray-4);
+  color: var(--color-dark-gray-5);
+  outline: none;
+  border-radius: 6px;
+  border: 1px solid #30404d77;
+  transition: 200ms cubic-bezier(0.3, 0, 0.5, 1) ease-in-out;
```
SmithPeder, line 19: This feels a bit much for me.

app/routes/articles/components/ArticleEditor.js
```diff
@@ -152,10 +152,12 @@ const ArticleEditor = ({
           initialized={initialized}
         />
         <div>
-          <Button submit>{!isNew ? 'LAGRE ENDRINGER' : 'OPPRETT'}</Button>
+          <Button submit success={!isNew}>
+            {!isNew ? 'Lagre endringer' : 'Opprett'}
```
SmithPeder, line 156: I never understood why we had these in Capital letters in the first place 🙃

app/routes/surveys/components/SurveyEditor/SurveyEditor.js
```diff
@@ -239,7 +239,12 @@ class SurveyEditor extends Component<Props, State> {
             rerenderOnEveryChange={true}
           />
           <div className={styles.clear} />
-          <Button className={styles.submit} disabled={submitting} submit>
+          <Button
+            success
+            className={styles.submit}
```
SmithPeder, line 244: This style only adds margin, maybe we can remove this.

app/routes/users/components/GroupChange.js
```diff
@@ -75,7 +75,11 @@ class GroupChange extends Component<Props, State> {
           isClearable={false}
         />
         {this.state.selectedOption && (
-          <Button className={styles.saveButton} onClick={this.handleOnClick}>
+          <Button
+            className={styles.saveButton}
```
SmithPeder, line 79: This also only adds margin-top, and that is the only thing in that entire file :D Maybe we can remove that

app/routes/users/components/RemovePicture.js
```diff
@@ -33,20 +34,23 @@ export default class RemovePicture extends Component<Props, State> {
 
   render() {
     return (
-      <Flex justifyContent="center">
-        <Button className={styles.saveButton} onClick={this.toggleSelected}>
-          {this.state.selected ? 'Angre' : 'Slett bilde'}
-        </Button>
+      <div className={styles.buttons}>
+        {this.state.selected ? (
+          <Button className={styles.saveButton} onClick={this.toggleSelected}>
```
SmithPeder, line 39: Here styles.savebutton is used on "Avbryt" to get 10px margin 👎🏻

app/components/Button/index.js
```diff
@@ -40,22 +42,30 @@ function Button({
...
+              className
+            )
       }
       type={submit ? 'submit' : 'button'}
       {...(rest: Object)}
     >
       <LoadingIndicator small margin={0} loading={pending} />
-      {pending && <span className={styles.loading}>Laster</span>}
-      {!pending && children}
+      {pending ? <span className={styles.loading}>Laster</span> : children}
```
SmithPeder, line 68: 🤣 The prev code 💯

app/components/Button/Button.css
```diff
@@ -75,20 +85,60 @@ html[data-theme='dark'] .button {
...
+
+  &:disabled {
+    color: var(--color-red-2);
+  }
 }
 
-.large {
-  font-size: 1.4rem;
+.success {
+  background: #2da44e;
```
SmithPeder, line 123: Should add these greens to the variables file

app/components/Button/Button.css
```diff
@@ -75,20 +85,60 @@ html[data-theme='dark'] .button {
...
+  color: var(--color-white) !important;
+
+  &:hover:not(:disabled) {
+    background: #2c974b;
+  }
 }
 
-.flatButton {
+html[data-theme='dark'] .success {
+  background: #347d39;
```
SmithPeder, line 133: This green as well

erlingfn:
Can you include some pictures please :pleading_face:

app/routes/meetings/components/MeetingDetail.css
```diff
@@ -21,3 +21,24 @@
...
+
+/* stylelint-disable no-descending-specificity */
+html[data-theme='dark'] .declineBtn {
+  background: var(--color-red-1);
+  color: var(--color-black) !important;
+
+  &:hover:not(:disabled) {
+    background: var(--color-red-2) !important;
+  }
+}
```
SmithPeder, line 44: hmm, that's a lot of extra code for a small thing, but okey.

SmithPeder (commented):
Disabled normal dark button looks a bit off to me
image
i don't know if i love it when its enabled either
image
It's a bit too blue i feel 🟦

SmithPeder:
>  Can you include some pictures please 🥺

image

app/components/Button/Button.css
ivarnakken: Yes, I'm thinking of changing the variables up, as the current green colors are very ugly

app/routes/meetings/components/MeetingDetail.css
ivarnakken: Well, to me, it was worth it 🙃 It looked very off

app/routes/users/components/RemovePicture.js
ivarnakken: It was actually not used for anything, but should be removed anyway lol

app/routes/users/components/GroupChange.js
ivarnakken: A quick `git blame`: _Peder Smith, 4 years ago via PR #1255_

SmithPeder: 😡

app/routes/meetings/components/MeetingDetail.css
ivarnakken: I removed this btw, and made it the "dark" button

juniwbjerde:
This is a huge improvement in my opinion! Looks way better and now the action of the buttons actually match their design like they should:) And the added icons are great too!

A tiny bit of constructive criticism:
1. I think the hover color for the default button is a bit to lite, as it's a bit hard to tell the difference when i hover over it. It's better for the danger and success buttons, but the contrast could be a bit higher there too in my opinion
2. It's a bit inconsistent that the default button has a border, and the danger and success buttons don't. I see why this is: the default button is so lite-colored that would be weird without a border, but it still bothers me a bit that it's not consistent. Maybe add a border too all the buttons? Or maybe that would be ugly, i'm not completely sure about this one...
3. I'm not too big of a fan of the type of danger button where the background color fills in when you hover over it (e.g. the ones that you find on settings for a user, like the "slett profilbilde" or "slett bruker" buttons). Especially in dark mode they look a bit off to me. I think it might look better if the buttons were red by default and then a darker red when you hovered over them, like the "Avslå" button for meetings

Buuuut, this is all very subjective. I think you've done a great job and it's up to you to decide if you want to implement any of the changes i mentioned above. If we merged this as it is now I would still think it's a big improvement:)

app/routes/podcasts/components/PodcastEditor.js
```diff
@@ -75,7 +75,11 @@ class PodcastEditor extends Component<Props, *> {
             filter={['users.user']}
             component={SelectInput.AutocompleteField}
           />
-          <Button disabled={pristine || submitting} submit>
+          <Button
+            success={!this.props.new}
```
erlingfn, line 79: Perhaps extract this with the rest of the props?

app/routes/users/components/ChangePassword.js
```diff
@@ -48,7 +48,7 @@ const ChangePassword = ({
         type="password"
         component={TextInput.Field}
       />
-      <Button disabled={disabledButton} submit>
+      <Button disabled={disabledButton} submit danger>
```
erlingfn, line 51: Not very crucial, but I think we should have a consistent order for the props, ie. `danger` always coming before or after `disabled`

erlingfn (approved):
Definitively an improvement, yet a few nitpicks. As the others have mentioned, there might be something to discuss as to the look of the buttons. Yet as far as I see it, this is matter of taste, so I will leave it up those with a better taste than me.

app/components/Button/Button.css
```diff
@@ -1,52 +1,68 @@
...
+    margin-right: 7px;
   }
 
   > a {
     color: var(--color-dark-gray-3);
   }
+
+  &:hover:not(:disabled) {
+    background: var(--color-light-gray-3);
+    border-color: #394b59aa;
```
LudvigHz, line 52: Is this supposed to be the same for dark and light mode?
And is there no variable that can be used for this?

app/components/Button/Button.css
```diff
@@ -58,37 +74,83 @@ html[data-theme='dark'] .button {
 }
 
 .dark {
-  background: var(--color-red-3);
-  color: var(--color-white);
-  border-color: var(--color-red-1);
+  background: var(--color-red-2);
+  border: none;
+  color: var(--color-white) !important;
```
LudvigHz, line 79: NO :no_good:
If you want to override other styles, use the "cascading" property of CSS.

app/components/Button/Button.css
```diff
@@ -58,37 +74,83 @@ html[data-theme='dark'] .button {
...
+
+  &:hover:not(:disabled) {
+    background: var(--color-red-2);
+    border: 1px solid var(--color-red-2);
+    color: var(--color-dark-gray-5);
+  }
+
+  &:disabled {
+    color: var(--color-red-2);
+  }
```
LudvigHz, line 131: Most of these are the same as the above `.dark`. Just override the relevant properties if you _have_ to do this. (I would prefer the colors to be correct between dark/light. If some of the darkmode colors can be changed, just do that instead.

LudvigHz (changes requested):
Nice :100: 
I really like the danger and success buttons. The gray one looks "off" to me. I'm not quite sure what it is, but it just doesn't look right. I did some quick testing and it looks more fitting without the border IMO. And the blue hint in the color is a bit too much still IMO.

app/components/Button/Button.css
ivarnakken: I would love to use a variable and I tried, but since our color variables are on hex format (and not rgb), it's not possible to apply an alpha value on the variable itself. I'm considering changing all the variables to rgb 🤔

LudvigHz: aaah I see now. Yeah I had the same issue with the dark mode update. So if you want to change some variables to rbg, I would say go ahead.

app/components/Button/Button.css
ivarnakken: The `.danger` and `.dark` class do not share any properties. `.danger` is a bit darker than `.dark` 🙃 Should maybe change the name from `.dark` to `.red` or something? Not sure

app/routes/users/components/ChangePassword.js
ivarnakken: Sure, it kinda bothered me as well. It's tedious work, but I'll do it.

app/components/Button/Button.css
LudvigHz: Whoops! I meant `.danger` :sweat_smile:. The point was that the normal `.danger` and `html[data-theme='dark'] .danger` have duplicated a lot of properties.

app/routes/podcasts/components/PodcastEditor.js
ivarnakken: Sure, it was previously done this way because `new` is reserved

app/components/Button/Button.css
ivarnakken: 🙃

app/components/Button/Button.css
ivarnakken: Changed them now 👍🏼

ivarnakken:
> It's a bit inconsistent that the default button has a border, and the danger and success buttons don't.

@juniwbjerde. Added border colors to all style variants now 😄

app/components/Button/Button.css
```diff
@@ -1,52 +1,68 @@
...
+  width: fit-content;
+  padding: 5px 15px;
+  box-shadow: 0 1px 0 rgba(27, 31, 36, 10%),
+    inset 0 1px 0 rgba(255, 255, 255, 3%);
+  font-weight: 600;
+  background-color: var(--color-light-gray-4);
+  color: var(--color-dark-gray-5);
+  outline: none;
+  border-radius: 6px;
+  border: 1px solid rgba(--color-dark-gray-4, 40%);
```
LudvigHz, line 18: Does this work without the `var()`??

app/components/Button/Button.css
```diff
@@ -57,38 +73,118 @@ html[data-theme='dark'] .button {
...
+
+  &:focus-visible {
+    box-shadow: 0 0 0 3px rgba(--color-red-5, 25%);
+  }
+
+  &:hover:not(:disabled) {
+    background: var(--color-red-2);
+    border: 1px solid rgba(--color-light-gray-5, 20%);
+  }
+}
```
LudvigHz, line 110: Most of these are still the same as the `.button.dark` above. Pls remove all the duplicate stuff, and it would be way better if you could just modify the dark-theme color variables to be proper, or add new ones, instead of having custom dark-theme color adjustments here.

app/components/Button/Button.css
```diff
@@ -57,38 +73,118 @@ html[data-theme='dark'] .button {
...
+
+  &:hover:not(:disabled) {
+    background: var(--color-red-2);
+    border: 1px solid var(--color-red-2);
+    color: var(--color-black);
+  }
+
+  &:disabled {
+    color: var(--color-red-2);
+  }
```
LudvigHz, line 152: Same here. Remove all the duplicate styles

app/routes/surveys/components/surveys.css
```diff
@@ -406,3 +402,13 @@
...
 }
+
+.copied {
+  border-color: var(--color-green-3);
+  color: var(--color-green-3);
+  outline-color: var(--color-green-3) !important;
+
+  &:hover {
+    border-color: var(--color-green-3) !important;
+  }
```
LudvigHz, line 413: Why are these `!important`?

Is it impossible to to fix the order so these are prioritized?

ivarnakken: hmm I forgot about these ... I'll look into it 🤔

app/components/Button/Button.css
ivarnakken: lol not at all, but for some reason it still worked last night 🙃 Nice catch!

ivarnakken: Perhaps it's better to just let the colors be hard coded instead ... 😞

LudvigHz: Hmm, yeah. I think we would need this to be stable in most browsers :/ https://www.w3.org/TR/css-color-5/#relative-colors
Sadly, no browsers have implemented it yet :cry:

app/components/Button/Button.css
ivarnakken: Sure, but if you don't want duplicates on the `&:hover`, you'll have to accept a bit of `!important` 😄 Or else, the hover effect cascaded from `.button` will take effect.

ivarnakken: I managed to remove all duplicates everywhere else though

app/components/Button/Button.css
ivarnakken: Removed all duplicate code 👍🏼

ivarnakken:
@LudvigHz. I turned all button colors into variables, thus removing the need for `html[data-theme='dark']` in `button.css`. Ultimately, I think this will make it easier to further develop the code in the future.

LudvigHz (approved):
LGTM :shipit:. The variables made it a lot cleaner💯

SmithPeder (approved):
Looks good to me! Good work ⭐

PR #2662, opened 2022-04-19
replace react-copy-to-clipboard with custom code

eikhr:
The entire package could be replaced with **1** line of code. We are losing IE11 support, but I think that's fine, especially since it's in an admin page.


ivarnakken (approved):
No IE11 support? 😭 I love it

erlingfn (approved):
The fewer dependencies the better

SmithPeder (approved):
⭐

PR #2669, opened 2022-04-24
Start using paginationNext in quotes

odinuge:
This will as a side effect remove the "random quote" from the list, make
pagination cleaner and easier, as well as reuse code => less code.

Otherwise, this works as before. I'm unable to test the 'approved=false' flow since I'm no longer an admin, but I can test if you give me access to it on staging. I caaaaaaaaaaan start the backed, but my xps still only has 8GiB of memory :rip:

This will also as a sideeffect do as @LudvigHz suggested here; https://github.com/webkom/lego-webapp/pull/2567#pullrequestreview-915770283 to fix the issue https://github.com/webkom/lego/issues/2587 :eyes: 

@LudvigHz:
> - Or some other way.

### DIFF
--- a/app/reducers/quotes.js
+++ b/app/reducers/quotes.js
@@ -53,12 +53,6 @@ export default createEntityReducer({
   mutate,
 });
 
-const compareByDate = (a, b) => {
-  const date1 = new Date(a.createdAt);
-  const date2 = new Date(b.createdAt);
-  return date2.getTime() - date1.getTime();
-};
-
 export const selectQuotes = createSelector(
   (state) => state.quotes.byId,
   (state) => state.quotes.items,
@@ -77,22 +71,6 @@ export const selectQuoteById = createSelector(
   (quotes, quoteId) => {
     if (!quotes || !quoteId) return {};
     return quotes.find((quote) => Number(quote.id) === Number(quoteId));
-  }
-);
-
-export const selectSortedQuotes = createSelector(
-  selectQuotes,
-  (state, props) => ({ filter: props.filter }),
-  (_, props) => props && props.pagination,
-  (quotes, query, pagination) => {
-    return quotes
-      .filter(
-        (quote) =>
-          typeof quote !== 'undefined' &&
-          quote.approved.toString() ===
-            (pagination.query.approved && pagination.query.approved.toString())
-      )
-      .sort(compareByDate);
   }
 );
 
--- a/app/reducers/quotes.js
+++ b/app/reducers/quotes.js
@@ -62,9 +62,12 @@ const compareByDate = (a, b) => {
 export const selectQuotes = createSelector(
   (state) => state.quotes.byId,
   (state) => state.quotes.items,
-  (quotesById, ids) => {
+  (_, props) => props && props.pagination,
+  (quotesById, ids, pagination) => {
     if (!quotesById || !ids) return [];
-    return ids.map((quoteId) => quotesById[quoteId]);
+    return (pagination ? pagination.items : ids).map(
+      (quoteId) => quotesById[quoteId]
+    );
   }
 );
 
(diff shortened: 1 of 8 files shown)


odinuge:
My js skills are a bit rusty tho., so I can have some dodgy code here and there :eyes:

app/reducers/quotes.js
```diff
@@ -80,12 +83,14 @@ export const selectQuoteById = createSelector(
...
-  (quotes, query) => {
+  (_, props) => props && props.pagination,
+  (quotes, query, pagination) => {
     return quotes
       .filter(
         (quote) =>
           typeof quote !== 'undefined' &&
-          quote.approved === (query.filter !== 'unapproved')
+          quote.approved.toString() ===
+            (pagination.query.approved && pagination.query.approved.toString())
```
ivarnakken, line 93: Do you think it would be cleaner to add an optinal chain to this?😄 `pagination.query.approved?.toString()`

app/routes/quotes/components/QuotePage.js
```diff
@@ -18,7 +18,7 @@ type Props = {
   approve: (number) => Promise<*>,
   unapprove: (number) => Promise<*>,
   deleteQuote: (number) => Promise<*>,
-  fetchMore: ({ approved: boolean }) => Promise<*>,
+  fetchAll: ({ query: { approved?: string } }, next?: boolean) => Promise<*>,
```
ivarnakken, line 21: Could you explain why you treat `approved` as a string?🤔 It makes sense in regards to the rest of the code, but is it not a bad practice?

app/reducers/quotes.js
```diff
@@ -80,12 +83,14 @@ export const selectQuoteById = createSelector(
 export const selectSortedQuotes = createSelector(
   selectQuotes,
   (state, props) => ({ filter: props.filter }),
```
LudvigHz, line 85: Maybe remove this since it's not used anymore

app/routes/quotes/QuotesRoute.js
```diff
@@ -12,40 +11,39 @@ import prepare from 'app/utils/prepare';
...
-import { selectPagination } from '../../reducers/selectors';
+import { selectPaginationNext } from 'app/reducers/selectors';
 import { addReaction, deleteReaction } from 'app/actions/ReactionActions';
 import { selectEmojis } from 'app/reducers/emojis';
 import { fetchEmojis } from 'app/actions/EmojiActions';
 import qs from 'qs';
 
+const qsParamsParser = (search) => ({
+  approved: qs.parse(search, { ignoreQueryPrefix: true }).approved || 'true',
+});
```
LudvigHz, line 22: I kind of like having the default `true` in the selector (as it was before) and just pass `undefined` in the query here tbh. It just makes it more consistent, but not really a big deal. As it is now, if you omit `approved` in the `props` to `sortedSelector` you'd just get `[]`. Which could be okay, but I think defaulting to approved makes more sense.

app/reducers/quotes.js
LudvigHz: And ref my other comment, default `true`?

LudvigHz (commented):
Awesome! :100: 
This is for sure a lot cleaner and nicer :heart_eyes:.
Just add some optional chaining here and there to clean it up a bit, and I think this is good!

app/reducers/quotes.js
odinuge: Yeee, will do :ok_hand:

app/routes/quotes/QuotesRoute.js
odinuge: Switched this to a bool to make it more consistent ref. @ivarnakken's comment. Passing an empty "approved" will select both approved and unapproved, ref how pagination works, and creating the default in the selector would be a bit awkward imo. and will not be consistent with the other use of paginationNext..

app/reducers/quotes.js
odinuge: Hmm. Tbh, after thinking, we don't need this filtering, since it is already filtered on the pagination :eyes:  Will remove it

app/routes/quotes/components/QuotePage.js
odinuge: Mostly for simplicity and consistency with other things. But yeah, changed to boolean now, and ensured that we only pass it around as a boolean

app/reducers/quotes.js
odinuge: Hmmmm. The `approve` and `unapprove` logic will then be kinda strange.... :/

app/routes/quotes/components/QuotePage.js
odinuge: Blæh. That makes the "query object" => "querystring" convertion return a querystring dropping the `approved` elem when it is false. String it is....

LudvigHz (approved):
LGTM :shipit:

PR #2681, opened 2022-04-26
Allow adding non-obligatory questions in event-editor

eikhr:
Currently in the event editor, the only way to create non-obligatory questions is to check the box for "Obligatorisk spørsmål", add the question, and then uncheck "Obligatorisk spørsmål" again. Although this hides the text field, the question is still in form-state, so it works fine. But it's not very intuitive.

1. Screenshot 2022-04-26 at 12 45 04
2. Screenshot 2022-04-26 at 12 45 11

Because of this people often set questions as obligatory when they shouldn't be. (of course allergies shouldn't be in this question field anyways, I added a warning for that)

I've changed it to be one checkbox for having a question, and a separate checkbox for marking the question as obligatory.

1. Screenshot 2022-04-26 at 12 43 35
2. Screenshot 2022-04-26 at 12 43 28

### DIFF
--- a/app/routes/events/components/EventEditor/index.js
+++ b/app/routes/events/components/EventEditor/index.js
@@ -616,6 +616,9 @@ const validate = (data) => {
   if (data.useMazemap && !data.mazemapPoi) {
     errors.mazemapPoi = 'Sted eller Mazemap-rom er påkrevd.';
   }
+  if (data.hasFeedbackQuestion && !data.feedbackDescription) {
+    errors.feedbackDescription = 'Spørsmål er tomt.';
+  }
   if (!data.id && !data.cover) {
     errors.cover = 'Cover er påkrevet';
   }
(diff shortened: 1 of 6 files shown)


app/routes/events/components/EventEditor/index.js
```diff
@@ -88,10 +88,13 @@ function EventEditor({
...
+  const [hasFeedbackQuestion, setHasFeedbackQuestion] =
+    useState<boolean>(false);
 
   useEffect(() => {
     if (initialized) {
       setUseMazemap(event?.mazemapPoi?.value > 0);
+      setHasFeedbackQuestion(!!event?.feedbackDescription);
     }
     // Should only run once when initialized to initialize useMazemap state
     // eslint-disable-next-line react-hooks/exhaustive-deps, react-app/react-hooks/exhaustive-deps
```
norbye, line 100: Would be nice if this wasn't necessary, as it's not done this way for other boolean values in events.

Other than that LGTM:)

eikhr: I figured out how to do it without the `useState` 😁

norbye (approved):
Looks great:)

PR #2688, opened 2022-04-27
Remove useMazemap state from meeting and event editors

eikhr:
It turns out we don't need the `useState` for `useMazemap` in the meeting or event editors as this is already saved in form state. 

I also set `useMazemap` to `true` by default in the hope that it will be used a bit more often.

### DIFF
--- a/app/routes/events/EventCreateRoute.js
+++ b/app/routes/events/EventCreateRoute.js
@@ -86,6 +86,7 @@ const mapStateToProps = (state, props) => {
       useConsent: false,
       feedbackDescription: '',
       pools: [],
+      useMazemap: true,
       separateDeadlines: false,
       unregistrationDeadline: time({ hours: 12 }),
       registrationDeadlineHours: 2,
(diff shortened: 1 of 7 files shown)


SmithPeder (commented):
I tested it, and it works. 

Pooling the value from the eventEditorForm is a lot smarter ⭐ 

One note on the `useMazeMap: True` that can make it even better is setting useMazeMap to `true` when we edit an event from `TBA` to `NORMAL`. Now, when I create an event and set it to TBA _(which is what happens 95% of the time)_, the location value will be set to the string "TBA" and useMazeMap will be set to `false`. 

Therefore, when I edit the event a month later and switch from `TBA` to `NORMAL`, the value will be `useMazemap: false` 😢

eikhr:
> I tested it, and it works.
> 
> Pooling the value from the eventEditorForm is a lot smarter ⭐
> 
> One note on the `useMazeMap: True` that can make it even better is setting useMazeMap to `true` when we edit an event from `TBA` to `NORMAL`. Now, when I create an event and set it to TBA _(which is what happens 95% of the time)_, the location value will be set to the string "TBA" and useMazeMap will be set to `false`.
> 
> Therefore, when I edit the event a month later and switch from `TBA` to `NORMAL`, the value will be `useMazemap: false` 😢

Good point, I'll add this

SmithPeder (approved):
LGTM

norbye (approved):
LGTM:)

PR #2695, opened 2022-04-28
Avoid loading zxcvbn unneccesarily

eikhr:
We use `zxcvbn` for checking strength of new passwords in the frontend, and it's a pretty large package (400kb minified). It is currently loaded on all pages, and makes up about 20% of loaded javascript on the front page.

Luckily we have code-splitting and dynamic imports, so with some refactoring it is now only loaded once you actually need to use it 😁 

I've never used dynamic imports before, so please tell me if I'm doing something dumb or not following best practice. But it seems to work well.

### DIFF
--- a/app/routes/users/components/PasswordStrengthMeter.js
+++ b/app/routes/users/components/PasswordStrengthMeter.js
@@ -1,9 +1,7 @@
 // @flow
 
-import { Component, Fragment } from 'react';
-import { pick } from 'lodash';
+import loadable from '@loadable/component';
 import moment from 'moment-timezone';
-import zxcvbn from 'zxcvbn';
 import Bar from '@webkom/react-meter-bar';
 import '@webkom/react-meter-bar/dist/Bar.css';
 import styles from './PasswordStrengthMeter.css';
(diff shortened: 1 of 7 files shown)


app/routes/users/components/PasswordStrengthMeter.js
```diff
@@ -19,60 +17,76 @@ type Props = {
...
+        <span>
+          <strong>Laster inn passordstyrke...</strong>
+        </span>
+      </>
     );
   }
-}
+
+  let tips = zxcvbnValue.feedback.suggestions;
+  tips.push(zxcvbnValue.feedback.warning);
```
SmithPeder, line 43: Is there any change .feedback is undefined because then this will fail. Maybe add optional chaining here?

app/routes/users/components/PasswordStrengthMeter.js
```diff
@@ -19,60 +17,76 @@ type Props = {
...
+  tips = tips.map((tip) => passwordFeedbackMessages[tip]).filter(Boolean);
```
SmithPeder, line 44: `.filter(Boolean)` 😍

app/routes/users/components/PasswordStrengthMeter.js
```diff
@@ -19,60 +17,76 @@ type Props = {
...
+
+  const crackTimeSec = Number(
+    zxcvbnValue.crack_times_seconds.offline_slow_hashing_1e4_per_second
```
SmithPeder, line 47: Here as well, can `crack_times_seconds` be undefined?

app/routes/users/components/PasswordStrengthMeter.js
```diff
@@ -19,60 +17,76 @@ type Props = {
...
-            {tips.map((value, key) => {
-              return <li key={key}>{value}</li>;
-            })}
-          </ul>
-        )}
-      </Fragment>
+      <>
+        <PasswordStrengthBar strengthScore={0} />
+        <span>
+          <strong>Laster inn passordstyrke...</strong>
```
SmithPeder, line 36: Kinda odd formulation, what about "Kalkulerer passordstyrke".

app/routes/users/components/PasswordStrengthMeter.js
SmithPeder: According to the docs it's only set when 
`result.feedback   # verbal feedback to help choose better passwords. set when score <= 2. `
, but i see that `zxcvbnValue.feedback.suggestions;` was present before the changes as well. hmm

app/routes/users/components/PasswordStrengthMeter.js
```diff
@@ -19,60 +17,76 @@ type Props = {
...
-      .humanize();
-    const crackTime =
-      crackTimeSec > 2 ? crackTimeDuration : crackTimeSec + ' sekunder';
+  useEffect(() => {
+    import('zxcvbn').then((zxcvbn) => {
+      setZxcvbnResult(
+        zxcvbn.default(password, [user.username, user.firstName, user.lastName])
+      );
+    });
+  }, [password, user.firstName, user.lastName, user.username]);
```
SmithPeder, line 29: Why do we need to recalculate when these `[user.firstName, user.lastName, user.username]` change? I see that the `zxcvbn` is used by the `.default`, but hmmm

SmithPeder (commented):
Daam, looks very good! I didn't know zxcvbn was this heavy 📦 
The dynamic imports look correct to me ™️

app/routes/users/components/PasswordStrengthMeter.js
eikhr: It's because they are added to `zxcvbn`'s word-list. So if your password contains your usename it will show as a weaker password.

app/routes/users/components/PasswordStrengthMeter.js
eikhr: True. It's not technically spending much time calculating the strenght, though. Most of the time will go to downloading the `zxcvbn` package, but I guess that doesn't matter to the end user.

LudvigHz (commented):
This is very needed. The frequency list bundle size has annoyed me immensely forever.

However, I think you could do this a bit simpler. Instead of using dynamic imports just supported by ES and webpack, and manually handling the loading of the library, you could also just use loadable-components to defer the loading and split the chunk (like we do in all other cases :smile:)

Ref. https://loadable-components.com/docs/library-splitting/

With that, at least the code for `PasswordStrengthMeter` will be alot simpler and cleaner. 

You could also use it for the validator, but I don't know how much cleaner it will be. You _could_ use it to defer the loading of the entire `ChangePassword` component and avoid the `asyncValidate` as well, which _is_ simpler. But I'll let you be the judge.

app/routes/users/components/PasswordStrengthMeter.js
eikhr: I didn't actually change this code, so I assume it is working. But I guess it doesn't hurt to add some optional chaining and a default value in case it is undefined.

LudvigHz (approved):

app/utils/asyncValidator.js
```diff
@@ -0,0 +1,53 @@
...
+  formData: Object
+): Promise<[string, string[]][]> => {
+  return Promise.all(
+    Object.keys(fieldValidators).map((field) =>
+      getFieldErrors(fieldValidators[field], field, formData)
+    )
+  ).then((fieldArray) =>
+    fieldArray.filter(([, fieldErrors]) => fieldErrors.length)
+  );
+};
```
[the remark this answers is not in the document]
erlingfn: I agree that it is generally better with less code, yet functions should do one thing and be rather short. While this is not long, it contains a log of logic and chained calls, so IMO it is more readable with a helper functions.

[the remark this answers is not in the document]
erlingfn: To quote the linux kernel coding style

> However, if you have a complex function, and you suspect that a less-than-gifted first-year high-school student might not even understand what the function is all about, you should adhere to the maximum limits all the more closely. Use helper functions with descriptive names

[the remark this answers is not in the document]
LudvigHz: I generally agree, but when the helper functions are < 10 lines and it's _not_ easy to understand what it does, it could have the opposite effect. I don't necessarily mean the example above is the best. As I said, maybe just merging the two helper functions into one will help.

Generally this is a JS issue though :shrug:, with a mix of chaining, function calls and Promise/awaits it becomes ugly no matter how you do it.

LudvigHz (approved):
LGTM :shipit:

app/utils/asyncValidator.js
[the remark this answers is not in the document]
eikhr: I'm a big fan of splitting functions into simpler helper functions to increase readability. In this case I don't think it is very easily readable, it's a bit tricky because of the Promises and nested mapping, but putting it all in a single function makes it even less readable imo. even though it does use a lot fewer lines of code.

PeterJFB (approved):
Ouf, dynamic imports seems both cursed and nice:) lgtm :100:

PR #2700, opened 2022-05-02
Fix dark mode detection on SSR to avoid blinking 

odinuge:
This ensures that the correct theme is applied _before_ the SSR content is show to the user. That removes the blinking for users using the auto theme and are using dark mode.

https://user-images.githubusercontent.com/1467188/166219820-aab09c44-d834-4803-bc3b-74285f9e808d.mp4

PS: This also adds 'auto' as the selected theme for unauthenticated users.

### DIFF
--- a/server/pageRenderer.js
+++ b/server/pageRenderer.js
@@ -74,17 +74,15 @@ export default function pageRenderer({
 }: PageRendererProps = {}): string {
   const isSSR = app === undefined ? 'false' : 'true';
   const { helmet } = helmetContext;
-  const getDataTheme = () => {
-    if (!isEmpty(state)) {
-      let selectedTheme = selectCurrentUser(state).selectedTheme;
-      return selectedTheme !== 'auto' ? selectedTheme : 'light';
-    }
-  };
+  const selectedTheme: string =
+    (!isEmpty(state) && selectCurrentUser(state).selectedTheme) || 'auto';
   const { body, scripts, styles, links } = readyHtml(app);
 
   return `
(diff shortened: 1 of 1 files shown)


server/pageRenderer.js
```diff
@@ -120,6 +118,21 @@ export default function pageRenderer({
...
       </head>
       <body>
+        <script>
+          ${
+            // If user has selected auto and device is in dark mode; ensure we update
+            // the theme before first render to screen
+            selectedTheme === 'auto'
+              ? `(function () {
+            try {
+              if (window.matchMedia('(prefers-color-scheme: dark)').matches === true) {
```
odinuge, line 128: It's also possible to subscribe to changes here, but that is outside the scope here, and others can feel free to do that :ok_hand:

server/pageRenderer.js
```diff
@@ -120,6 +118,21 @@ export default function pageRenderer({
...
+                document.documentElement.setAttribute('data-theme', 'dark');
+              }
+            } catch (e) {}`
+              : ''
+          }
+          })();
+        </script>
```
LudvigHz, line 135: Might it be an idea to put the whole `script` tag in the ternary, so we don't end up with an empty tag that might confuse people?

LudvigHz (approved):
LGTM :shipit:

server/pageRenderer.js
odinuge: Skal være good nå

LudvigHz (approved):

PR #2706, opened 2022-05-03
When you finally the bastard who have ruined your life

SmithPeder:
Don't git blame plz. Fix the shitty frontend css

image


ivarnakken (approved):
Better, but more can still be done 🙃

PR #2697, opened 2022-04-30
[WIP] New frontpage 2022 release

PeterJFB:
## :sparkles: Loading new frontpage ...

Preview-images are below, all feedback is appreciated :100: 

### :information_source:  PRs
However, commenting on code will be done in individual PRs and merged into this branch, please have a look:
- webkom/lego#2594
- #2696 
- #2707 
- more coming ...

Once above PRs are done we will ready to merge the new frontpage!

## New look (29.04):

image
image

### 50% (29.04)

<details>
  <summary>Click to expand!</summary>

image
image
image
</details>

### Mobile (29.04)
<details>
  <summary>Click to expand!</summary>

image
image
image
</details>

### Dark (29.04)

image
image

### Poll (<29.04)
Peek 2022-01-25 17-14

PR #2707, opened 2022-05-03
Redesign poll component

PeterJFB:
# New poll waddup

https://user-images.githubusercontent.com/42469466/175818348-4593ca08-e491-4157-95e2-ae44118466bc.mp4

Tooltip is a bit hidden, but will be a one-liner with the new forntpage:)

### DIFF
--- a/app/components/Poll/Poll.css
+++ b/app/components/Poll/Poll.css
@@ -1,15 +1,83 @@
-@import '~app/styles/variables.css';
+.poll {
+  background-color: var(--lego-card-color);
+  width: 100%;
+  min-width: 250px;
+  max-width: 500px;
+  margin: 0 auto;
+  overflow-y: hidden;
+  border-radius: 5px;
+}
 
-.optionWrapper {
-  justify-content: center;
-  min-height: 120px;
-  cursor: pointer;
+.topBar {
+  background-color: var(--lego-red);
   position: relative;
+  width: 100%;
+  height: 70px;
+  padding-bottom: 25px;
+  box-shadow: 0 2px 2px rgba(0, 0, 0, 20%);
+}
+
+.pollIcon,
+.arrowIcon {
+  color: var(--lego-card-color);
+}
+
+.headerBar {
+  color: var(--lego-font-color);
+  background-color: var(--lego-card-color);
+  position: absolute;
+  bottom: 0;
+  left: 50%;
+  width: 80%;
+  height: 50px;
+  transform: translate(-50%, 50%);
+  font-weight: 900;
+  text-align: center;
+  border-radius: 10px;
+  box-shadow: 0 3px 3px rgba(0, 0, 0, 20%);
+}
+
+.contentWrapper {
+  overflow-y: hidden;
+  width: 100%;
+  transition: 0.5s height;
+}
+
+.voteOptionsWrapper {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 100%;
+  height: fit-content;
+}
+
+.voteButton,
+.voteButton + .voteButton {
+  color: var(--color-white);
+  background: var(--lego-red);
+  font-weight: 500;
+  width: 90%;
+  height: 2rem;
+  padding: 5px;
+  margin: 2px;
+}
+
+.voteOptions {
+  width: 100%;
+  padding-top: 35px;
+}
+
+.resultsHiddenInfo {
+  font-style: italic;
 }
 
 .pollTable {
   width: 100%;
   font-size: 14px;
+}
+
+.pollTable tr {
+  width: 100%;
 }
 
 .pollTable td {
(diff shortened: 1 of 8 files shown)


app/components/Poll/index.js
```diff
@@ -9,247 +9,225 @@ import { Link } from 'react-router-dom';
...
-    const totalVotes = options.reduce((a, option) => a + option.votes, 0);
-    const ratios = options.map((option) => {
-      return { ...option, ratio: (option.votes / totalVotes) * 100 };
-    });
-    return this.perfectRatios(ratios);
-  };
+const Poll = ({
+  poll,
+  handleVote,
+  backgroundLight,
```
eikhr, line 69: `backgroundLight` seems to be unused. Remove?

app/components/Poll/index.js
```diff
@@ -9,247 +9,225 @@ import { Link } from 'react-router-dom';
...
+      </Flex>
+      <Flex
+        column
+        alignItems="center"
+        className={styles.contentWrapper}
+        style={{
+          height:
+            isExpanded && optionsRef.current
+              ? `${optionsRef.current.clientHeight}px`
+              : '0',
```
eikhr, line 123: I guess this is kind of a design opinion, but i think it looks better if the opening transition doesn't play while loading default-expanded polls. This is the code I used to make it not do that.

```suggestion
          height: isExpanded
            ? optionsRef.current
              ? `${optionsRef.current.clientHeight}px`
              : 'auto'
            : 0,
```

eikhr (commented):
While testing this I noticed that old polls give you the opiton to vote, even though voting is impossible because the polls are no longer valid. This happens with the old component too, so I guess it's not really your responsibility, but it would be super nice if you fixed it. Should be as simple as checking if `poll.validUntil` is before now, and f.ex. showing results if this is the case.

app/components/Poll/index.js
PeterJFB: yeet

app/components/Poll/index.js
PeterJFB: I love small moving things, but i realize it's not ideal for [everyone using our plattform](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion). I'll remove it for now :3

PeterJFB:
> While testing this I noticed that old polls give you the opiton to vote, even though voting is impossible because the polls are no longer valid. This happens with the old component too, so I guess it's not really your responsibility, but it would be super nice if you fixed it. Should be as simple as checking if `poll.validUntil` is before now, and f.ex. showing results if this is the case.

Nicely spotted! :o 
Changed it to instead show results (given they're not hidden) :)

app/components/Poll/Poll.css
```diff
@@ -1,185 +1,167 @@
...
-  font-size: 14px;
-}
-
-.pollTable td {
-  border: 0;
-  padding: 5px;
+  min-width: 250px;
+  max-width: 500px;
+  overflow-y: hidden;
+  border-radius: 15px;
```
ivarnakken, line 8: I'd prefer if the `border-radius` would match the rest of the front page, as it would make it stick out less. Or do you think that looks ugly?

app/components/Poll/Poll.css
```diff
@@ -1,185 +1,167 @@
...
+  transform: scale(150%);
 }
 
-.pollLight {
-  background: var(--lego-card-color);
+.notAnswered {
+  color: var(--color-white);
+  font-size: 30px;
+  font-weight: 900;
+  line-height: 45px;
```
ivarnakken, line 30: Isn't this a bit hacky?

app/components/Poll/Poll.css
```diff
@@ -1,185 +1,167 @@
...
+.arrowUp {
+  margin-bottom: 25px;
+}
 
-  .blurEffect {
-    filter: blur(3px);
-    pointer-events: none;
-  }
+.arrowDown {
+  margin-bottom: 10px;
```
ivarnakken, line 166: I _really_ don't like the way you're changing the margin. The way the icon slightly moves is very ugly, as it makes the component look "non-robust" or sort of "hacky" - difficult to explain.

app/components/Poll/index.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import { Component } from 'react';
+import React, { useRef, useState } from 'react';
```
ivarnakken, line 3: You don't need to import `React`.

app/components/Poll/Poll.css
```diff
@@ -1,185 +1,167 @@
...
-.optionWrapper {
-  justify-content: center;
-  min-height: 120px;
-  cursor: pointer;
-  position: relative;
-}
-
-.pollTable {
+.poll {
+  margin: 0 auto;
```
ivarnakken, line 2: Are you sure you still need this? `.poll` is now a `Flex` component.

ivarnakken (changes requested):
This is definitely a step in the right direction! 😍 However, I do have some comments

ivarnakken:
> Tooltip is a bit hidden, but will be a one-liner with the new forntpage:)

Also, I don't get this. Please make it perfect from the beginning. I thought we agreed that we'd merge these into master, and get one component at a time into prod? At least that's more agile™️

PeterJFB:
> > Tooltip is a bit hidden, but will be a one-liner with the new forntpage:)
> 
> Also, I don't get this. Please make it perfect from the beginning. I thought we agreed that we'd merge these into master, and get one component at a time into prod? At least that's more agiletm

No:)
See https://github.com/webkom/lego-webapp/pull/2697

app/components/Poll/index.js
```diff
@@ -9,247 +9,237 @@ import { Link } from 'react-router-dom';
...
+        style={{ cursor: !alwaysOpen ? 'pointer' : null }}
+        onClick={!alwaysOpen ? toggleTruncate : null}
+      >
+        {!alwaysOpen ? (
+          <Icon
+            className={styles.arrowIcon}
+            size={30}
+            name={isExpanded ? 'arrow-up' : 'arrow-down'}
+          />
+        ) : null}
```
ivarnakken, line 239:
```suggestion
        {!alwaysOpen && (
          <Icon
            className={styles.arrowIcon}
            size={30}
            name={isExpanded ? 'arrow-up' : 'arrow-down'}
          />
        )}
```

app/components/Poll/Poll.css
```diff
@@ -1,185 +1,151 @@
...
 }
 
-.alignItems {
+.fullGraph {
   display: flex;
-  justify-content: center;
+  background-color: var(--color-mono-gray-5);
+  color: #000;
+  word-wrap: break-word;
+  width: 100%;
```
ivarnakken, line 120: In general, please group your CSS better: everything regarding display/position/size etc. at the top, and things related to the content (colors, fonts, shadows) at the bottom. The position (not just the CSS `position` property) of an element is often the most tricky part to get right and usually has the most "impact", and should therefore come first. Makes it easier to read and develop. You usually group well, but it's nice to be consistent.

PeterJFB: thanks for pointing out, it's always nice to read files with consistency ^:heart:

app/components/Poll/index.js
```diff
@@ -9,247 +9,234 @@ import { Link } from 'react-router-dom';
...
+                        </td>
+                      </tr>
+                    );
+                  })}
+                </tbody>
+              </table>
+            </Flex>
+          )}
+          <Flex column alignItems="center">
+            <Flex className={styles.totalVotesInfo}>
```
ivarnakken, line 202: Remove this unused classname

app/components/Poll/Poll.css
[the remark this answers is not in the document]
ivarnakken: I prefer if it matches the other cards. You don't just want the front page to look good, but the entire site. We therefore need consistency.

Also, the quote card looks too large.

PeterJFB:
Looks way cleaner now, thanks for all the input @ivarnakken! :heart_eyes:  Merging this in two days if no additional comments are made, We'll be able to make final design-adjustments in the `nfp-release`-branch o/:sparkles:

ivarnakken:
Closing this due to being stale for 2 years

PR #2710, opened 2022-05-04
BETTER fix bug with multiple image upload

LudvigHz:
Slight miscommunication, so this is kinda duplicate of @erlingfn PR. But this also adds a nice small
preview of the images next to the filename so that it's easier to see what is selected and what you
want to delete before uploading.
image

### DIFF
--- a/app/components/Upload/ImageUpload.js
+++ b/app/components/Upload/ImageUpload.js
@@ -46,46 +46,56 @@ type UploadAreaProps = {
   image: ?string,
 };
 
-const FilePreview = ({ file, onRemove, index }: FilePreviewProps) => (
-  <Flex
-    wrap
-    className={styles.previewRow}
-    alignItems="center"
-    justifyContent="space-between"
-  >
-    <div style={{ width: '90%' }}>
-      <TextInput
-        disabled
-        value={file.name}
-        style={{ width: '100%', height: 50 }}
+const FilePreview = ({ file, onRemove, index }: FilePreviewProps) => {
+  const [previewUrl, setPreviewUrl] = useState();
+  useEffect(() => {
+    !previewUrl && setPreviewUrl(URL.createObjectURL(file));
+    return () => {
+      if (previewUrl) URL.revokeObjectURL(previewUrl);
+    };
+  }, [file, previewUrl]);
+
+  return (
+    <Flex
+      wrap
+      className={styles.previewRow}
+      alignItems="center"
+      justifyContent="space-between"
+    >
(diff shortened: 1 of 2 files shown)


app/components/Upload/ImageUpload.js
```diff
@@ -46,27 +46,40 @@ type UploadAreaProps = {
...
+    <Flex
+      wrap
+      className={styles.previewRow}
+      alignItems="center"
+      justifyContent="space-between"
+    >
+      <div style={{ width: '90%', display: 'flex' }}>
+        <img width="60" alt="preview" src={previewUrl} />
+        <div style={{ flex: '1 1 auto' }}>
+          <TextInput
```
ivarnakken, line 68: I find it a bit weird to use the `<TextInput />` - just print out the file name like a normal kid

app/components/Upload/ImageUpload.js
```diff
@@ -46,27 +46,40 @@ type UploadAreaProps = {
...
+            disabled
+            value={file.name}
+            style={{ width: '100%', height: 50 }}
+          />
+        </div>
+      </div>
+      <Icon
+        name="close"
```
ivarnakken, line 76: I think a trash icon is more fitting, especially since we're using a close icon at the top with a different functionality

erlingfn (approved):
LGTM

app/components/Upload/ImageUpload.js
LudvigHz: I think it's just to get the style of the input field. Having it disabled just means we can reuse the styling without any more work. Kinda weird, but also kinda nice.

app/components/Upload/ImageUpload.js
LudvigHz: This is unchanged in this PR btw.

Thought about this myself. It's also really tiny and hard to click.

app/components/Upload/ImageUpload.js
ivarnakken: Ye i get the styling thing, but it's just weird and bad practice

LudvigHz: I disagree. It's a pattern we use all over the app, and I think it makes sense for items that are "user-defined" but not possible to change directly. (the same is done for emails, f.ex.

But I see your point in it being kinda weird. Out of scope for this PR though.

app/components/Upload/ImageUpload.js
LudvigHz: Done!

PR #2711, opened 2022-05-05
Fix bad index in ImagePreview in ImageUpload

LudvigHz:
I found some more bugs in the ImageUpload component :upside_down_face:

### DIFF
--- a/app/components/Upload/ImageUpload.js
+++ b/app/components/Upload/ImageUpload.js
@@ -36,8 +36,7 @@ type State = {
 
 type FilePreviewProps = {
   file: DropFile,
-  onRemove: (index: number) => void,
-  index: number,
+  onRemove: () => void,
 };
 
 type UploadAreaProps = {
@@ -46,7 +45,7 @@ type UploadAreaProps = {
   image: ?string,
 };
 
-const FilePreview = ({ file, onRemove, index }: FilePreviewProps) => {
+const FilePreview = ({ file, onRemove }: FilePreviewProps) => {
   const [previewUrl, setPreviewUrl] = useState();
   useEffect(() => {
     !previewUrl && setPreviewUrl(URL.createObjectURL(file));
@@ -80,7 +79,7 @@ const FilePreview = ({ file, onRemove, index }: FilePreviewProps) => {
         size={32}
         name="trash"
         prefix="ion-md-"
-        onClick={() => onRemove(index)}
+        onClick={onRemove}
         className={styles.removeIcon}
       />
     </Flex>
@@ -252,7 +251,6 @@ export default class ImageUpload extends Component<Props, State> {
                 />
               </div>
             )}
-            {/* $FlowFixMe */}
             {preview && (
               <Cropper
                 ref={(node) => {
@@ -268,10 +266,9 @@ export default class ImageUpload extends Component<Props, State> {
               <Flex wrap column>
                 {files.map((file, index) => (
                   <FilePreview
-                    onRemove={this.onRemove}
+                    onRemove={() => this.onRemove(index)}
                     file={file}
-                    index={index}
-                    key={index}
+                    key={file.name}
                   />
                 ))}
               </Flex>


app/components/Upload/ImageUpload.js
```diff
@@ -268,10 +266,9 @@ export default class ImageUpload extends Component<Props, State> {
               <Flex wrap column>
                 {files.map((file, index) => (
                   <FilePreview
-                    onRemove={this.onRemove}
+                    onRemove={() => this.onRemove(index)}
                     file={file}
-                    index={index}
-                    key={index}
+                    key={file.name}
```
ivarnakken, line 271: Aren't file names more prone to duplicates than an index?

LudvigHz: Well, most likely, files with the same name are _the same file_. So it shouldn't be a problem :tm:. The only case this could be an issue, is if the user uploads two _different_ files with the same name, which I kinda see as very unlikely. Sure, a hash of the file content is the only _proper_ key, but I think this is good enough :shrug:

Index is **NEVER** a proper key. The position of elements in the array is of course not stable. The immediate moment you change the `files` array, the keys are not unique for the files and you get a bunch of bugs. I.e. delete a file from the list and you are in trouble, since the keys and files are suddenly mismatched.

ivarnakken: > which I kinda see as very unlikely

Saving this to when we get an angry mail from PR 😬

LudvigHz: If you have a better suggestion please let me know :/

ivarnakken (approved):
Nice

erlingfn (approved):
LGTM

app/components/Upload/ImageUpload.js
SmithPeder: {file.name}-{index}

LudvigHz: That doesn't really make sense either. The whole point of the key is to identify the _same_ object within a list of similar components. Having the index as part of the key will make it unstable and change every time the object changes position within the array, which is undesirable.

SmithPeder: What, no. Using composite keys will always increase stability. It's the de facto recommendation in these cases. Of course, it's not 100% stable, but it's more stable than using just the name.

If you have a list of items, where two of them have the same name, and you rearrange or make a change to the list, react will have no idea what the difference between the two are. But with a composite index, you have a chance that one of the same name elements is above the change, and will keep its key/order.

SmithPeder: But, the really cool key here would be `file.name-file.size` ⭕  Two images of the exact same name and exact same size gotta be rare

LudvigHz: I do agree that composite keys will work and is generally a good idea if you can't find a unique value. However using the `index` as part of it will as you say increase uniqueness, but it's not a stable value for each element. If you want to use a composite key, at least use different values which are all stable for each element :upside_down_face:. Using the index will _decrease_ stability and thus performance, since react will re-evaluate the entire component if it just changes index (which does not affect the component or object itself in any way)

I did consider using a combination of `file.name` and `file.lastModified` (or `file.size`), but concluded that noone is going to upload different files with the same name, so didn't bother.

Another option, which is arguably better, is to generate a unique ID, such as a UUID, for each element when we add them to the `files` list in the first place.

For anyone wondering, the ONLY time it's okay to use `index` as a key (composite or otherwise) is if all these are true:
1. the list and items are static – they are not computed and do not change;
1. the items in the list have no ids;
1. the list is never reordered or filtered.

References:
- https://reactjs.org/docs/reconciliation.html#keys
- https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

SmithPeder: Of course, the sources you post are correct for the single `index` as a key. I never stated otherwise.

I might have been bamboozled by one of my old sources. I read once that adding the composite key could give you a small boost, as you could be lucky and find have elements previously un-unique elements gain a 50% of being unique and stable in the next render. But looking into it this is still far from ideal, particularly if the items in the list are reordered.

PR #2723, opened 2022-05-05
Don't block user zooming on mobile

LudvigHz:
Setting `maximum-scale` to a value <= 1 will make prevent users to zoom
on several mobile browsers.


SmithPeder (approved):
I can zoom on all my browsers, so that sounds like a aNdRoiD problem to me 😎

LudvigHz:
>  I can zoom on all my browsers, so that sounds like a aNdRoiD problem to me :sunglasses:

Well yes, safari is one of few that ignore this property. Rightfully so, since it's really stupid.

ivarnakken (approved):
You'd think setting `maximum-scale` would only affect zooming out, but well then🤷🏼‍♂️

PR #2772, opened 2022-05-27
Prevent page overflow on mobile devices and small style changes

norbye:
* Add links to the tags in the bottom of articles to other articles with
the same tag.
* Flex-wrap and word-break in headers and profile badges
* Change color of active tag in /articles/ to the same color as hovered
tag, as it feels unintuitive to have them be different colors when it is
not a multi-select
* Change :hover to :active to prevent it displaying as if it were active
when disabling on mobile.


eikhr (approved):
nice

PR #2780, opened 2022-05-28
Add link to deployment in webkom/lego to README.md

norbye:
Add link to the deployment section in the webkom/lego README.md, and added a top-level table of contents for the README.md of this repo


SmithPeder (approved):
Nice!
Would have been nice with a `taget="_blank"` on this, but I see that is not possible with [GitHub markdown engine](https://stackoverflow.com/questions/41915571/open-link-in-new-tab-with-github-markdown-using-target-blank)

PR #2786, opened 2022-06-01
Block accidental scroll/pan on mazemap embed

ollfkaih:
Visste du at Samsung Internet Browser av Samsung Electronics Co., Ltd. ignorerer `touch-events: none` fordi den har en helt egen zoom-implementasjon?  
Det betyr at brukere av Samsung Internet Browser nå får mulighet til å zoome inn nettsiden på kartet og sette seg fast der💯 Det kan de takke Lee Jae-yong for, fordi han vet best hvordan en nettleser skal oppføre seg🤞
Uansett, preview på desktop:
anim

Er en liten bug hvor zoom av og til slutter å fungere, men ser ut til å funke igjen så snart man klikker på kartet (og dermed tvinger det til å oppdatere seg, da det av en eller annen grunn slutter å oppdatere seg). Mulig man kan tvinge en refresh, men er lite sannsynlig at noen faktisk driver og zoomer masse inn og ut på embedden...
Closes webkom/lego#2531


app/components/MazemapEmbed/index.js
```diff
@@ -32,10 +35,60 @@ export const MazemapEmbed = ({ mazemapPoi, ...props }: Props) => {
...
+
+    embeddedMazemap.on('wheel', () => {
+      if (controlPressed) {
+        embeddedMazemap.scrollZoom.enable();
+      } else {
+        embeddedMazemap.scrollZoom.disable();
+        setBlockScrollZoom(true);
+        setTimeout(() => {
+          setBlockScrollZoom(false);
+        }, 500);
```
eikhr, line 67: When scrolling over the map, the overlay will sometimes flicker a bit. This is because the timeout will set `setBlockScrollZoom(false);` even if we are still scrolling. To fix it you should save the timeout ID of the timeouts created, so you can cancel the previous timeout (if any) on each wheel event and start a new one. Something like this:
```javascript
if (blockScrollZoomTimeout) {
  clearTimeout(blockScrollZoomTimeout);
}
blockScrollZoomTimeout = setTimeout(() => {
  setBlockScrollZoom(false);
}, 500);
```

app/components/MazemapEmbed/index.js
```diff
@@ -32,10 +35,60 @@ export const MazemapEmbed = ({ mazemapPoi, ...props }: Props) => {
...
       doubleClickZoom: false,
+      dragRotate: false,
+      dragPan: false,
       touchZoomRotate: false,
+      touchPitch: false, //this is a horrible feature
+      pitchWithRotate: false,
     });
+    embeddedMazemap.dragPan._mousePan.enable();
+
+    window.addEventListener('keydown', (e) => {
```
eikhr, line 48: The window events should be removed when unmounting the component. As they are attached to the global `window`-object, they will stay there after closing the map and stack up when you open new maps.

If you return a function from the useEffect this will be run on unmount, so something like this should fix it:
```javascript
const onKeyDown = (e) => { ... }
const onKeyUp = (e) => { ... }
window.addEventListener('keydown', onKeyDown);
window.addEventListener('keyup', onKeyUp);
```
and at the end of the useEffect
```javascript
return () => {
  window.removeEventListener('keydown', onKeyDown);
  window.removeEventListener('keyup', onKeyUp);
};
```

eikhr (changes requested):
Love the feature, good job! :)

I have commented on some small things that should be fixed/cleaned up before merging

app/components/MazemapEmbed/index.js
ollfkaih: Clever 👍

app/components/MazemapEmbed/index.js
```diff
@@ -32,10 +34,67 @@ export const MazemapEmbed = ({ mazemapPoi, ...props }: Props) => {
...
+    };
+    const onKeyUp = (e) => {
+      if (e.key === 'Control') {
+        controlPressed = false;
+      }
+    };
+    window.addEventListener('keydown', onKeyDown);
+    window.addEventListener('keyup', onKeyUp);
+
+    let blockScrollZoomTimeout = undefined;
```
eikhr, line 61:
```suggestion
    let blockScrollZoomTimeout;
```

eikhr (approved):
LGTM 💯

app/components/MazemapEmbed/index.js
```diff
@@ -90,7 +153,35 @@ export const MazemapEmbed = ({ mazemapPoi, ...props }: Props) => {
...
+              top: '50%',
+              left: '50%',
+              width: '100%',
+              transform: 'translate(-50%,-50%)',
+              zIndex: 5,
+              pointerEvents: 'none',
+            }}
+          >
+            {blockScrollZoom
+              ? 'Hold ctrl for å zoome'
```
SmithPeder, line 180: Is it ctrl for mac as well?  :O

erlingfn: Why must you people always be like this

app/components/MazemapEmbed/index.js
```diff
@@ -90,7 +153,35 @@ export const MazemapEmbed = ({ mazemapPoi, ...props }: Props) => {
...
+              fontSize: '1.5rem',
+              textAlign: 'center',
+              color: '#111',
+              position: 'absolute',
+              top: '50%',
+              left: '50%',
+              width: '100%',
+              transform: 'translate(-50%,-50%)',
+              zIndex: 5,
+              pointerEvents: 'none',
```
ivarnakken, line 176: Could we move this to a CSS file please? 😄

app/components/MazemapEmbed/index.js
ollfkaih: > Is it ctrl for mac as well? :O

Yes, or you could have needed a special case for macs.

eikhr: command definitely feels more natural than ctrl on mac. I think it's worth adding a special case.

eikhr (approved):

ivarnakken (approved):
Looks good! 💯

PR #2782, opened 2022-05-31
Navigation cleanup

ivarnakken:
## I made some changes to the navigation
- Employ the generic `back` property of `NavigationTab` and do some general clean up of the navigation.
- Improve `NavigationLink` with active and hover effects. The active effect was no longer working due to it being removed from `Link`. I think the active effect will make the navigation tab more "visible", e.g. in the settings page, on which _Notifikasjoner_ and _OAuth2_ is quite hidden and unknown to many users.  
- Move meeting deletion to the edit page, as I find it more fitting.

### Showcase of the new `NavigationTab` style
newNavigationTab

### The _back_ link also got a retouch
newback

PR #2791, opened 2022-06-04
Fix quotes

erlingfn:
This PR fixes a few things that has been bugging me about the quotes page. First and foremost this PR fixes the issue where the page would show all the quotes in the redux state, regardless of approved status. So approved quotes would be shown when viewing unapproved quotes, and vice versa.
This PR also fixes the NavLink component so that it takes query params into account. Why react-router does not do this by default is beyond me, but at this point I have come to believe that the maintainers of that repo makes changes just to mess with people. 
In addition it adds a loading indicator to the quotelist. 

I am not sure if I did the selector stuff correctly, so feel free to correct me if it is wrong.

### DIFF
--- a/app/components/NavigationTab/NavigationLink.js
+++ b/app/components/NavigationTab/NavigationLink.js
@@ -12,9 +12,16 @@ type Props = {
 };
 
 const NavigationLink = (props: Props) => {
+  // Custom isActive function as NavLink does not compare query params by default
+  const isActive = (match, location) => {
+    const regex = /\/+$/i; // Regex to remove trailing / for comparison
+    const comparePath = location.pathname + location.search;
+    return props.to.replace(regex, '') === comparePath.replace(regex, '');
+  };
   return (
     <NavLink
       exact
+      isActive={isActive}
       to={props.to}
       onClick={props.onClick}
       className={styles.link}
--- a/app/reducers/quotes.js
+++ b/app/reducers/quotes.js
@@ -84,7 +84,7 @@ export default createEntityReducer({
 export const selectQuotes = createSelector(
   (state) => state.quotes.byId,
   (state) => state.quotes.items,
-  (_, items, props) => props?.pagination,
+  (_, props) => props?.pagination,
   (quotesById, items, pagination) =>
     (pagination ? pagination.items : items).map(
       (quoteId) => quotesById[quoteId]
--- a/app/routes/quotes/QuotesRoute.js
+++ b/app/routes/quotes/QuotesRoute.js
@@ -29,7 +29,9 @@ const mapStateToProps = (state, props) => {
   })(state);
   const emojis = selectEmojis(state);
   return {
-    quotes: selectQuotes(state, { pagination }),
+    quotes: selectQuotes(state, {
+      pagination,
+    }),
     query: pagination.query,
     actionGrant: state.quotes.actionGrant,
     showFetchMore: pagination.hasMore,
--- a/app/routes/quotes/components/QuotePage.js
+++ b/app/routes/quotes/components/QuotePage.js
@@ -9,6 +9,7 @@ import Button from 'app/components/Button';
 import type { ActionGrant, ID } from 'app/models';
 import type { QuoteEntity } from 'app/reducers/quotes';
 import type { EmojiEntity } from 'app/reducers/emojis';
+import LoadingIndicator from 'app/components/LoadingIndicator';
 
 type Props = {
   reactions: Array<Object>,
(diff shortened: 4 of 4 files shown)


app/components/NavigationTab/NavigationLink.js
```diff
@@ -12,9 +12,16 @@ type Props = {
 };
 
 const NavigationLink = (props: Props) => {
+  // Custom isActive function as NavLink does not compare query params by default
+  const isActive = (match, location) => {
+    const regex = /\/+$/i; // Regex to remove trailing / for comparison
+    const comparePath = location.pathname + location.search;
+    return props.to.replace(regex, '') === comparePath.replace(regex, '');
```
eikhr, line 19: If there were multuple query params, they could be in the wrong order, and this would return a false negative. This is very much an edge case though, and I'm pretty sure it won't happen in lego-webapp, so might be better just to keep it simple and leave it like this.

app/reducers/quotes.js
```diff
@@ -91,6 +91,17 @@ export const selectQuotes = createSelector(
...
 
+export const selectQuotesByApproval = createSelector(
+  (state) => state.quotes.byId,
+  (state) => state.quotes.items,
+  (state, props) => props?.pagination,
+  (state, props) => props?.approvalStatus,
+  (quotesById, items, pagination, approvalStatus) =>
+    (pagination ? pagination.items : items)
+      .map((quoteId) => quotesById[quoteId])
+      .filter((quote) => quote.approved === approvalStatus)
```
eikhr, line 102: I believe this is done correctly, but it can be simplified by using the existing `selectQuotes`-selector :)

```suggestion
  selectQuotes,
  (_, props) => props?.approvalStatus,
  (quotes, approvalStatus) =>
    quotes.filter((quote) => quote.approved === approvalStatus)
```

eikhr (changes requested):
Good job, this is badly needed considering the state this is in in prod.

app/reducers/quotes.js
erlingfn: Agreed

app/components/NavigationTab/NavigationLink.js
```diff
@@ -12,9 +12,16 @@ type Props = {
 };
 
 const NavigationLink = (props: Props) => {
+  // Custom isActive function as NavLink does not compare query params by default
+  const isActive = (match, location) => {
```
ivarnakken, line 16: Very smart! I noticed this was any issue back in #2782, but somehow forgot about it 🙃

app/routes/quotes/QuotesRoute.js
```diff
@@ -22,14 +22,17 @@ const qsParamsParser = (search) => ({
 });
 
 const mapStateToProps = (state, props) => {
+  const { approved } = qsParamsParser(props.location.search);
   const { pagination } = selectPaginationNext({
     endpoint: `/quotes/`,
-    query: qsParamsParser(props.location.search),
+    query: { approved },
```
ivarnakken, line 28: Does this really have to be a variable? 🤔 I don't seem to understand how this changes anything

ivarnakken (approved):
Looks good! 😍 Tried it out, and it seems to work great!

app/routes/quotes/QuotesRoute.js
erlingfn: No it does not have to be a variable, its a leftover from when it was passed to the selector

app/components/NavigationTab/NavigationLink.js
LudvigHz, line 16: Move this function out of the component definition.

LudvigHz (changes requested):
Looks good. Just a small comment

app/components/NavigationTab/NavigationLink.js
erlingfn: That does not work as it requires the to prop in order to make the comparison. I could get most of the function body of the component by making it a curry function and only instantiating it inside the component, but imo that would be overkill for a function used one place.

LudvigHz: Ah, I didn't see the usage of `props.to`, sorry! Then this is fine

app/reducers/quotes.js
```diff
@@ -84,7 +84,7 @@ export default createEntityReducer({
 export const selectQuotes = createSelector(
   (state) => state.quotes.byId,
   (state) => state.quotes.items,
-  (_, items, props) => props?.pagination,
+  (_, props) => props?.pagination,
```
odinuge, line 87: Thanks for fixing by very shitty code 👀 🥇

PR #2792, opened 2022-06-05
Implement styling changes, particularly on the Joblistings page

ivarnakken:
- Replace JoblistingEditor header with `NavigationTab`
- Add a rotation animation to the dropdown caret in `JoblistingRightNav`
- Clean up `JoblistingRightNav`
- Rewrite `JoblistingsRightNav` to a functional component
- Prevent active effect on Button when disabled
- Redesign `JoblistingList`. Notice that the entire item functions as a link, as opposed to the current situation in which the link is only accessible on the title and the image.
- Untrack mazemap.min.* files and ignore them when running eslint

## The new look
showcase

### DIFF
--- a/app/routes/joblistings/components/JoblistingEditor.css
+++ b/app/routes/joblistings/components/JoblistingEditor.css
@@ -1,19 +1,3 @@
-@import '~app/styles/variables.css';
-
-.root {
-  composes: container from '~app/styles/utilities.css';
-  margin-top: 2rem;
-}
-
-.heading {
-  color: var(--color-dark-mono-gray-3);
-  letter-spacing: 2px;
-  text-transform: uppercase;
-  border-bottom: 1px solid var(--color-mono-gray-4);
-  font-weight: 400;
-  font-size: 24px;
-}
-
 .descriptionField,
 .textField {
   border: 1px;
--- a/app/routes/joblistings/components/JoblistingEditor.css
+++ b/app/routes/joblistings/components/JoblistingEditor.css
@@ -14,10 +14,6 @@
   font-size: 24px;
 }
 
-.submit {
-  width: 130px;
-}
-
 .descriptionField,
 .textField {
   border: 1px;
--- a/app/routes/joblistings/components/JoblistingList.css
+++ b/app/routes/joblistings/components/JoblistingList.css
@@ -1,10 +1,5 @@
 @import '~app/styles/variables.css';
 
-.root {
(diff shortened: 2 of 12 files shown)


SmithPeder (approved):
Looks good to me <3

app/routes/joblistings/components/JoblistingList.css
```diff
@@ -63,43 +60,36 @@
...
   min-width: 120px;
   object-fit: contain;
+  margin-right: 20px;
 }
 
 .listItem {
+  display: flex;
   flex: 1;
   justify-content: space-between;
+  font-size: 14px;
```
erlingfn, line 70: I may just be getting old, but I thought this made the text a little to small

app/routes/joblistings/components/JoblistingEditor.css
```diff
@@ -14,10 +14,6 @@
   font-size: 24px;
 }
```
norbye, line 15: If I haven't missed any usages this class could probably be deleted

app/routes/joblistings/components/JoblistingList.css
```diff
@@ -63,43 +60,36 @@
...
 }
 
 .joblistingItemTitle {
-  color: var(--color-black);
-  font-weight: 550;
-  margin: 0;
-  text-transform: uppercase;
+  color: var(--color-dark-gray-1);
+  font-weight: 600;
   font-size: 15px;
```
erlingfn, line 80: Perhaps increase this a bit to compensate for making the text lower case? Again I find a little too small

norbye (changes requested):

erlingfn (approved):
Looks good! I think the text is a bit small in some places, but that is subjective so I will leave it up to you

app/routes/joblistings/components/JoblistingEditor.css
ivarnakken: Nice catch! `.root` was not being used as well 🙃

norbye (approved):
Very nais(:

app/routes/joblistings/components/JoblistingList.css
ivarnakken: I definitely see your point. I think it stands out a bit much since the rest of the text on the page is quite larger. In general we use very large font sizes on the webapp, which I don't really like. Nevertheless, I bumped it up to `15px`, and the title to `16px`. It now matches the rest, so I hope that's not too small he he

PR #2810, opened 2022-06-10
Give tags a retouch

ivarnakken:
## Regarding the tag:
- Remove unused `small` prop
- Introduce a new way to set the color of a tag
- Introduce new tag colors
- Fix alignment of tags on pinned article
- Introduce a `active` prop

Please make notice of the difference between light and dark theme tags 😉. It's also worth noting that the way we handle tag colors can be scaled up to give each tag their own respective color, but this would require a lot of rewriting, so I think a different PR would then be fitting.

### After
Red is the new default color.

------

### After
In these examples, the _genfors_ tag is active and employing the `active` prop.

------

### Before
<p align=center">
  
  
</p>

### After
<p align="center">
 
  
</p>


app/routes/overview/components/Overview.css
```diff
@@ -52,11 +33,19 @@
...
-.item:hover .itemInfo {
-  color: black;
+.type {
+  margin-right: 4px;
+}
+
+.tagline {
+  display: inline-block;
+  white-space: nowrap;
+  transform: translateY(-7.5%);
```
ivarnakken, line 46: I take no criticism on this. It's the only way in which we keep the `vertical-align: middel`, which is needed for this component to work on both the pinned article and the event items below 🙃 FlexBox? Nope, that would ruin the centered alignment of the text on the event items below. Should probably be separated, but that's out of scope of what I'm trying to accomplish here. It still looks good on mobile, I checked.

app/routes/articles/components/Overview.js
```diff
@@ -80,12 +80,17 @@ export default class Overview extends Component<Props> {
               <Tag
                 tag={tag.tag}
                 key={tag.tag}
-                color={isSelected ? 'red' : ''}
+                color={isSelected ? 'purple' : 'blue'}
```
erlingfn, line 83: Hmmmm, I am not sure about this choice of colours, blue and purple does not fit with our theme imo

ivarnakken: Abakus don't really have a color palette besides red and gray, sadly, and I think we should be more "open-minded" with introducing more colors here and there. Nevertheless, I'm okay with changing these specific colors to something else though, as I'm not sure if I like them either. Any suggestions?

LudvigHz:
Very cool! The dark mode improvements are especially nice, and the tags "pop" a lot more now so hopefully they are used more.

Although I have to agree that the tag cloud colors (blue & purple) look very strange. Why not just keep the old colors?

app/components/NavigationTab/NavigationTab.css
```diff
@@ -39,7 +39,7 @@ span.backLabel {
 }
 
 .backIcon {
-  transition: transform 150ms ease-in-out;
+  transition: transform 200ms cubic-bezier(0.175, 0.885, 0.32, 1.275);
```
SmithPeder, line 42: Why is this in a tag pr?

app/components/Tags/Tag.css
```diff
@@ -1,72 +1,156 @@
...
-  color: var(--lego-color-gray);
-}
+  --lightness-threshold: 0.59;
+  --border-threshold: 0.96;
+  --perceived-lightness: calc(
+    (
+        (var(--label-r) * 0.2126) + (var(--label-g) * 0.7152) +
+          (var(--label-b) * 0.0722)
+      ) / 255
+  );
```
SmithPeder, line 30: Is this how prettier wants it to look???

SmithPeder: Jeeeezzeee

app/routes/articles/components/Overview.js
ivarnakken: It is now solely blue, and instead employing a `active` prop to give feedback.

SmithPeder (approved):
Although I'm not a huge fan of some of the complicated CSS calculations, this looks really good and is a huge improvement. LATM :shipit:

PR #2824, opened 2022-06-19
bug fix: don't require meeting "location" if useMazemap is selected

eikhr:
Fix a bug in the meeting editor where new meetings cannot be created without filling in the "location"-field, even if a mazemap location is entered.


LudvigHz (approved):
LGTM :shipit:

ivarnakken (approved):
Nice fix! 💯

PR #2832, opened 2022-06-21
fadderperioden 2022

SmithPeder:
- Add prop destructuring
- Add fadderperioden banners to public and internal frontpage

### DIFF
--- a/app/routes/overview/components/Overview.js
+++ b/app/routes/overview/components/Overview.js
@@ -18,7 +18,7 @@ import type { PollEntity } from 'app/reducers/polls';
 import RandomQuote from 'app/components/RandomQuote';
 import { renderMeta } from './utils';
 
-//import Banner, { COLORS } from 'app/components/Banner';
+import Banner, { COLORS } from 'app/components/Banner';
 
 type Props = {
   frontpage: Array<Object>,
@@ -192,12 +192,13 @@ class Overview extends Component<Props, State> {
     return (
       <Container>
         <Helmet title="Hjem" />
-        {/*<Banner
-          header="itDAGENE 20. & 21. september"
-          subHeader="Kom til U1 i Realfagsbygget, kanskje finner du din drømmebedrift her?"
-          link="https://itdagene.no"
-          color={COLORS.itdageneBlue}
-        />*/}
+        <Banner
+          header="Velkommen til fadderperioden 2022!"
+          subHeader="Trykk her for mer informasjon"
+          link="articles/414"
+          color={COLORS.red}
+          internal
+        />
         <Flex className={styles.desktopContainer}>
           <Flex column className={styles.leftColumn}>
             {compactEvents}
--- a/app/routes/overview/components/PublicFrontpage.js
+++ b/app/routes/overview/components/PublicFrontpage.js
@@ -15,7 +15,7 @@ import truncateString from 'app/utils/truncateString';
 import Time from 'app/components/Time';
 import { readmeIfy } from 'app/components/ReadmeLogo';
 
-//import Banner, { COLORS } from 'app/components/Banner';
+import Banner, { COLORS } from 'app/components/Banner';
 
 type Props = {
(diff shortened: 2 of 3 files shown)


app/routes/overview/components/PublicFrontpage.js
```diff
@@ -81,13 +81,13 @@ class PublicFrontpage extends Component<Props, State> {
...
       <Container>
-        {/*<Banner
-          header="Velkommen alle nye studenter!"
-          subHeader="Trykk her for informasjon om fadderperioden"
-          internal
-          link="/articles/359"
-          color={COLORS.lightBlue}
-        />*/}
+        <Banner
+          header=" Velkommen til fadderperioden 2022!"
```
LudvigHz: Is this space intented?

app/routes/overview/components/PublicFrontpage.js
```diff
@@ -81,13 +81,13 @@ class PublicFrontpage extends Component<Props, State> {
...
-          internal
-          link="/articles/359"
-          color={COLORS.lightBlue}
-        />*/}
+        <Banner
+          header="Velkommen til fadderperioden 2022!"
+          subHeader="Trykk her for mer informasjon"
+          link="articles/414"
+          color={COLORS.red}
+          internal={true}
```
LudvigHz, line 89:
```suggestion
          internal
```

app/routes/overview/components/Overview.js
```diff
@@ -192,12 +192,13 @@ class Overview extends Component<Props, State> {
...
-          subHeader="Kom til U1 i Realfagsbygget, kanskje finner du din drømmebedrift her?"
-          link="https://itdagene.no"
-          color={COLORS.itdageneBlue}
-        />*/}
+        <Banner
+          header="Velkommen til fadderperioden 2022!"
+          subHeader="Trykk her for mer informasjon"
+          link="articles/414"
+          color={COLORS.red}
+          internal={true}
```
LudvigHz, line 200:
```suggestion
          internal
```

LudvigHz (approved):
LGTM :shipit:

app/routes/overview/components/PublicFrontpage.js
SmithPeder: THERE NEVER WAS ANY SPACE! 😬

SmithPeder: image
😬

app/routes/overview/components/PublicFrontpage.js
SmithPeder: No, generally, I'm not too fond of this. As I pointed out when we had the react course, this is possible but highly unnecessary

ivarnakken: I vouch for @LudvigHz suggestion. We should remain consistent across the webapp.

SmithPeder: Okey! I added the dumbest shorthand in the universe! Still hate it though 🙃 

Generally, the reason I do not like it is laid out here (https://stackoverflow.com/a/60817451/10002175). This shorthand saves 0.0002 seconds of typing, and lays a trap for future developers that forget that `true===true` but `false=!=undefined`. As he says

> JSX's behavior of defaulting a prop to true is just an added feature that matches with [HTML5's boolean attributes behavior](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes).

and

> This is where JSX's behavior differ from HTML5's boolean attributes behavior. There is not such thing as defaulting to false in JSX; it is only applicable for passing an explicit true. In contrast to HTML5, if you do not pass anything, you're really passing undefined, and your defined default values will be used instead. This is contrary to the HTML5 specs which would say it's false

This is just begging for something to go wrong. Where one removes the props, hoping the code on the other side is written to handle true, false and undefined correctly. Of course here, and everywhere a type system this will be less of an issue. You save so little to gain so little. 😞

PR #2838, opened 2022-06-26
Sort imports according to eslint guidelines

ivarnakken:
Used [eslint-plugin-simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort) to sort imports according to [eslint guidelines](https://eslint.org/docs/latest/rules/sort-imports).

`git blame --ignore-revs-file .git-blame-ignore-revs` will now output a useful git blame. Sadly, the flag must always be passed, so I recommend setting a config for this; `git config blame.ignoreRevsFile .git-blame-ignore-revs`. Note that this needs to be set per-project, as `git blame` will raise an error if it cannot find the file.


LudvigHz:
The idea is nice, and I do agree that imports can quickly become messy and disorganized. I think we've discussed some "manual" sorting of imports before, as an introduction to the project, but it probably isn't enforced or introduced that well. But, our conventions do seem to match that of the eslint plugin's, so it seems like a good fit. I.e in rough terms:
1. Package imports, with react type imports first
2. Absolute imports
3. relative imports


However, as it is now, the `plugin-simple-import-sort` doesn't sort correctly due to our webpack alias. So you would need to create a config to make it work correctly: https://github.com/lydell/eslint-plugin-simple-import-sort#custom-grouping.


Also, this type of bulk change should be done with @webkom-R2D2. And the commit should be put in a [`.git-blame-ignore-revs` file](https://git-scm.com/docs/git-config#Documentation/git-config.txt-blameignoreRevsFile) (which we haven't done before, but really should)

ivarnakken:
> But, our conventions do seem to match that of the eslint plugin's, so it seems like a good fit.

Yes, it matched perfectly - very pleasant indeed!

> Also, this type of bulk change should be done with @webkom-R2D2.

That was my plan, but I wanted to hear you guy's thoughts before moving on. `.git-blame-ignore-revs` seems both overkill and nice, so I'll check it out!

ivarnakken:
> However, as it is now, the plugin-simple-import-sort doesn't sort correctly due to our webpack alias. So you would need to create a config to make it work correctly: https://github.com/lydell/eslint-plugin-simple-import-sort#custom-grouping.

@LudvigHz. I configured some custom groups, what do you think of it?

eikhr (approved):
Looks great! Kind of interesting that import sorting isn't more common, it really makes it easier to get an overview of the imports.

I don't know about the git-blame-ignore-revs thing though. Seems like it doesn't track the moved import, so git blame shows whoever added the import that was there before the sorting. In this case I think it might be better to just show R2D2 in the blame, and if you really need to know who added the import you could check the blame from before the sorting.

ivarnakken:
> Seems like it doesn't track the moved import, so git blame shows whoever added the import that was there before the sorting.

That's the entire point of the feature though, to ignore mass reformatting commits. I, for instance, religiously use GitLens, and it's really annoying whenever all I see is a prettier/black commit. Besides, when are you ever gonna _want_ to see the sorting commit? It's simply practical to easily look up the commit, instead of having to git blame from before.

eikhr:
> it's really annoying whenever all I see is a prettier/black commit. Besides, when are you ever gonna _want_ to see the sorting commit? It's simply practical to easily look up the commit, instead of having to git blame from before.

I agree it's a little annoying, I've seen the "Update files for prettier 2" commit on a lot of code-lines. You never _want_ to see the sorting commit, I'm just not sure it's an improvement to see the _wrong_ commit. 

It doesn't hurt to have the file there though, as long as it's optional.

ivarnakken:
> It doesn't hurt to have the file there though, as long as it's optional.

Yes, completely optional, but _strongly_ recommended. In my eyes, it's not really a "wrong" commit if the code simply changed its position. It's like when you rename a file, you still want the content to point to the commit which introduced the code, not entirely to the rename commit.

eikhr:
Screenshot 2022-07-08 at 23 42 48

My point is just that in this screenshot, the hightlighted line `import { push} ...` was _not_ added in the commit "Add fetch for semesters..." like it seems. "Add fetch for semesters..." actually added the line `import { semesterToText } ...` near the bottom. In my opinion that blame is wrong.

ivarnakken:
ah yes, I've been testing it more thoroughly now, and you're right. I only checked one file in which it seemed to work as intended, sorry. The idea is nice though, and I hope it will be improved in the future. I suppose it works best with simple code formatting where the code doesn't really change its position, unlike here. On that thought, I might consider adding the other _prettier_ bulk changes to `.git-blame-ignore-revs` as well.

LudvigHz:
I didn't think about the fact that lines changed order and the git blame doesn't follow :/

But on another note, why is this not just using `eslint-plugin-import` which supports import order enforcement?[^1]
This will save us another npm package and eslint-plugin-import is more widely used and maintained as well. It seems to have the same features as `eslint-plugin-simple-import-sort`, if not more, so there is no reason for not using it.

[^1]: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

ivarnakken:
Replaced by https://github.com/webkom/lego-webapp/pull/3251

PR #2864, opened 2022-07-02
Change color of front page banner

ivarnakken:
As requested by the PR committee, the banner on the front page should employ the same background-color as the chosen "buddy week" color for 2022.


eikhr (approved):
lgtm😁

PR #2898, opened 2022-07-19
Update text in contact form

juniwbjerde:
Alva requested that we update the text in the contact form to make it easier for new members to understand what it can be used for, and I totally agree that this was needed. I only added a part stating that this is where we want to receive "avvik", and restructured the text a bit to make it tidier.

### DIFF
--- a/app/routes/contact/components/ContactForm.js
+++ b/app/routes/contact/components/ContactForm.js
@@ -57,14 +57,21 @@ const ContactForm = (props: Props) => {
     <Form onSubmit={props.handleSubmit(submit)}>
       <p>
         Dette skjemaet kan benyttes for å kontakte Abakus sine komiteer eller
-        Hovedstyret. Sender du til en spesifikk komité er det kun lederen av
-        komiteen som vil motta meldingen. Om du sender til Hovedstyret vil hele
-        styret få meldingen.
+        Hovedstyret. Dersom du har opplevd avvik eller ubehagelige hendelser i
+        forbindelse med Abakus, ønsker vi gjerne at dette meldes fra om her.
+        Send oss gjerne også spørsmål, feedback eller ting du ønsker å fortelle
+        oss, for eksempel om du ønsker å avtale tid for innsyn i HSP-avtalen.
       </p>
       <p>
-        Send oss gjerne spørsmål, feedback eller ting du ønsker å fortelle oss.
-        Du kan også benytte dette skjemaet om du ønsker å avtale tid for innsyn
-        i HSP-avtalen.
+        Sender du meldingen til en spesifikk komité er det kun lederen av
+        komiteen som vil motta meldingen. Om du sender til Hovedstyret vil hele
+        styret motta meldingen.
+      </p>
+      <p>
+        Skjemaet kan både benyttes til å sende inn en anonym henvendelse, eller
+        en med avsender for videre oppfølging. Både komitéledere og Hovedstyret
+        som mottar henvendelser har signert taushetserklæring, og de vil
+        kontakte deg og følge opp saken dersom det ønskes.
       </p>
       <Field
         placeholder="Velg mottaker"


app/routes/contact/components/ContactForm.js
```diff
@@ -57,14 +57,22 @@ const ContactForm = (props: Props) => {
     <Form onSubmit={props.handleSubmit(submit)}>
       <p>
         Dette skjemaet kan benyttes for å kontakte Abakus sine komiteer eller
-        Hovedstyret. Sender du til en spesifikk komité er det kun lederen av
+        Hovedstyret. Dersom du har opplevd avvik eller ubehagelige hendelser i
+        forbindelse med Abakus, ønsker vi gjerne at dette meldes fra om her.
+        Send oss gjerne også spørsmål, feedback eller ting du ønsker å fortelle
+        oss. Du kan også benytte dette skjemaet om du ønsker å avtale tid for
+        innsyn i HSP-avtalen.
```
ivarnakken, line 64: > Du kan også benytte dette skjemaet om du ønsker å avtale tid for innsyn i HSP-avtalen.

This felt like sort of out of place, following the more "serious" matters mentioned in the paragraph. Perhaps make it clear that it's just a random example, because I guess the point of it is to show how "anything" is welcome. Does that make sense?

ivarnakken: I now see that the example was there from previous - is it still even relevant? Maybe I'm overthinking.

app/routes/contact/components/ContactForm.js
```diff
@@ -57,14 +57,22 @@ const ContactForm = (props: Props) => {
...
+        styret motta meldingen.
       </p>
       <p>
-        Send oss gjerne spørsmål, feedback eller ting du ønsker å fortelle oss.
-        Du kan også benytte dette skjemaet om du ønsker å avtale tid for innsyn
-        i HSP-avtalen.
+        Skjemaet kan både benyttes til å sende inn en anonym henvendelse, eller
+        en med avsender for videre oppfølging. Både komitéledere og Hovedstyret
+        som mottar henvendelser har signert taushetserklæring, og vil følge opp
+        saken dersom det ønskes.
```
ivarnakken, line 75: > og vil følge opp saken dersom det ønskes.

hmm, should we add what we mean by this? Will they contact anyone else etc. etc. Or, besides, does it not go without saying? Again, might be overthinking, but I'm simply trying to picture myself in the shoes of someone in desperate need of help 😅

ivarnakken (approved):
A lovely change! Some small nitpick comments though, but I'll leave it up to you to decide what's best 🤗

app/routes/contact/components/ContactForm.js
```diff
@@ -57,14 +57,22 @@ const ContactForm = (props: Props) => {
...
+        Sender du meldingen til en spesifikk komité er det kun lederen av
         komiteen som vil motta meldingen. Om du sender til Hovedstyret vil hele
-        styret få meldingen.
+        styret motta meldingen.
       </p>
       <p>
-        Send oss gjerne spørsmål, feedback eller ting du ønsker å fortelle oss.
-        Du kan også benytte dette skjemaet om du ønsker å avtale tid for innsyn
-        i HSP-avtalen.
+        Skjemaet kan både benyttes til å sende inn en anonym henvendelse, eller
```
ivarnakken, line 72: Also, now that I think about it, should we explain how the message is sent anonymously? And how the receiver actually receives it. I would image that would increase trust and make it easier for people to reach out!

app/routes/contact/components/ContactForm.js
juniwbjerde: If I were to guess, I would think the example was originally added because people were confused about where they should contact Abakus in that exact case. (In my experience, when something is stated THAT clearly it's usually a reason for it)

But I kind of agree that it feels a bit out of place now that I added the part about "avvik". I could change the wording a bit so it's just one sentence about the serious stuff, and one about other things to use the contact form for. What do you think about: "Dersom du har opplevd avvik eller ubehagelige hendelser i forbindelse med Abakus, ønsker vi gjerne at dette meldes fra om her. Send oss gjerne også spørsmål, feedback eller ting du ønsker å fortelle oss, for eksempel om du ønsker å avtale tid for innsyn i HSP-avtalen"

app/routes/contact/components/ContactForm.js
juniwbjerde: I feel like it's nice to state explicitly that the case will be followed up on, so there's no doubt about it. It's probably very uncomfortable to be unsure of whether or not your message will lead to any action if you've just reported an uncomfortable situation. But I can change it to "og vil kontakte deg og følge opp saken dersom det ønskes." so they know that when someone sees the message, they will get in contact.

app/routes/contact/components/ContactForm.js
ivarnakken: I like the new suggestion! Next to the other stuff, "HSP-avtalen" seemed like some holy thing within Abakus and felt weirdly special, although I understand it was just an absurd example.

app/routes/contact/components/ContactForm.js
juniwbjerde: I think we should leave any technical explanation out, if that's what you meant. In my opinion it would just make the text longer and more confusing for most users who probably just care about being anonymous, not how we do it.

app/routes/contact/components/ContactForm.js
ivarnakken: Yes, the new suggestion is good! I think it's nice to "know" for certain that you will be contacted, sort of assuring in a way.

app/routes/contact/components/ContactForm.js
ivarnakken: Yes, wouldn't want to make it longer than needed. But, maybe a tooltip could be nice? Or a footnote at the bottom hidden away with a tooltip. The tooltip would only require an extra sentence. Sort of like the tooltip in the corner on pools (by the question mark). 

> most users who probably just care about being anonymous, not how we do it.

Maybe, but, I know that I personally would be extremely skeptical of a system like abakus.no made by a bunch of students, especially if I was a first grader fresh out of buddy week. It was just a thought.

juniwbjerde: A tooltip could actually be nice, good idea! I'll see what I can do:)

juniwbjerde: When i think about it, I have to catch a flight tomorrow requiring me to be up in 3 hours, so i should probably get some sleep. I'll make the other changes, and then add the tooltip when I have some free time on vacation (as adding an explanatory sentence about how we handle anonymity would actually require me to look into how we handle anonymity first hehe)

SmithPeder (approved):

eikhr (approved):

LudvigHz (approved):

PR #2918, opened 2022-08-02
Update path to find eventId of event to copy

norbye:
Resolves https://github.com/webkom/lego/issues/2380

Apparently it seems the location of where props are passed through react-router-dom have changed. I didn't find the changelog for exactly when the change occured, but in my search for answers I found [this StackOverflow answer](https://stackoverflow.com/questions/38085893/react-router-link-pass-in-params) that used `match.params` for react-router-dom v2, and the other sources I visited (e.g. [this guide](https://ui.dev/react-router-v5-pass-props-to-link)) used `location.state` for v5 which we currently use.


ivarnakken (approved):
Nice, now it works like a charm! 💯 I appreciate the thorough explanation!

PR #2935, opened 2022-08-10
Update lego-editor to v2.0.1

LudvigHz:
As you can see from the release notes of lego-editor, this should fix alot of issues that have been plaguing us for a while.
Notably we have:
- Android support
- Fixing (hopefully) all the editor crashes everywhere
- General bugfixes and increased smoothness

These should not be any visual or functional changes with this release.


ivarnakken (approved):
About time 🎉

PR #2946, opened 2022-08-12
Fix event detail crashing

LudvigHz:
This fixes an issue where the event detail page would crash for some events that you follow.

Now quite sure why this was happening, and I couldn't find when `follower` was undefined when
debugging.

Anyways, this seems to fix the issue :shrug:.


ivarnakken (approved):
I don't seem to understand the bug, but this looks good to me anyway so :shrug:

erlingfn (approved):
LGTM

PR #2948, opened 2022-08-12
Update create new email form to autofill the new email

norbye:
Recruitment-season do be closing in, and as it's not unheard of to write the wrong name when creating new emails I decided it could be nice to give a helping hand.

Set æøå to be replaced by ae, oe and aa. Seemed to be kinda standard, but if anyone has any objections I'll do it.  
Set to remove any characters except a-z, 0-9, - and . after the three above are replaced.

image
image
image


LudvigHz (approved):
:shipit:

PR #2682, opened 2022-04-26
Minimal redesign of event details

amalieeh:
It has finally come! The redesign of the sidebar. 

It is rather simple, mostly rearranging of code.
Move some of the information and add some icons 'cause people like icons.
And move the name of the company and put the event type to the coloured bar under the header.
Also, the mazemap thing is now something that opens and close when hitting the link.

Before:
image

After (the faces dont show, but now there is _one_ row of faces rather than two!):
image

And with payment and mazemap:
image

PR #2949, opened 2022-08-12
Relocate event rules and allergy/preferences link

norbye:
Change allergy/preferences link to go directly to the users' settings

Preview when logged in
image

Preview when not logged in
image

### DIFF
--- a/app/routes/events/components/EventDetail/index.js
+++ b/app/routes/events/components/EventDetail/index.js
@@ -449,6 +449,23 @@ export default class EventDetail extends Component<Props, State> {
               )}
               <Line />
               <InfoList items={eventCreator} className={styles.infoList} />
+              <Line />
+              {loggedIn && (
+                <Link
+                  to={`/users/${currentUser.username}/settings/profile`}
+                  className={styles.iconWithText}
+                >
+                  <Icon name="create-outline" className={styles.infoIcon} />
+                  <span>Oppdater allergier og preferanser</span>
+                </Link>
+              )}
+              <Link
+                to="/pages/arrangementer/26-arrangementsregler"
+                className={styles.iconWithText}
+              >
+                <Icon name="document-outline" className={styles.infoIcon} />
+                <span>Arrangementsregler</span>
+              </Link>
               {(actionGrant.includes('edit') ||
                 actionGrant.includes('delete')) && <Line />}
               <Flex column>
@@ -460,21 +477,6 @@ export default class EventDetail extends Component<Props, State> {
               </Flex>
             </ContentSidebar>
           </ContentSection>
-          <Link
-            to="/pages/arrangementer/26-arrangementsregler"
-            style={{ marginTop: 0 }}
-          >
-            <Flex alignItems="center">
-              <Icon name="document" style={{ marginRight: '4px' }} />
-              <span>Regler for Abakus&#39; arrangementer</span>
-            </Flex>
-          </Link>
-          {loggedIn && (
-            <p>
-              Du kan oppdatere dine allergier og preferanser
-              <Link to="/users/me"> her</Link>.
-            </p>
-          )}
           {event.contentTarget && (
             <CommentView
               style={{ marginTop: 20 }}
--- a/app/routes/events/components/Admin.js
+++ b/app/routes/events/components/Admin.js
@@ -84,7 +84,7 @@ const Admin = ({ actionGrant, event, deleteEvent }: Props) => {
     ) < 2;
 
   return (
-    <div style={{ marginTop: '10px' }}>
+    <div>
       {(canEdit || canDelete) && (
         <ul>
           <li>
--- a/cypress/e2e/event_visit_spec.js
+++ b/cypress/e2e/event_visit_spec.js
@@ -98,7 +98,7 @@ describe('View event', () => {
 
     // With out custom methods for interacting with the editor, we need to fire events on some
     // other elements first.
-    cy.contains('allergier').click();
+    cy.contains('Kommentarer').click();
     cy.get(c('CommentForm') + ' [data-slate-editor="true"]')
       .first()
       .click()


ivarnakken:
I would actually prefer to have "the rules" where it already was, as it's preferable if the sidebar isn't too _crowded_. 

And in addition, with this, we're not consistent in our use of links: with the allergy link, it's at the end of the sentence, whereas with the rules, the link is the the entire sentence. Does that make sense? It looks a bit strange next to each other. I would prefer to not use links in sentences on a webapp (e.g. "klikk _her_ for å [...]").

Can we also keep the icon? I like icons 😅

app/routes/events/components/EventDetail/index.js
```diff
@@ -460,21 +476,6 @@ export default class EventDetail extends Component<Props, State> {
               </Flex>
             </ContentSidebar>
           </ContentSection>
-          <Link
-            to="/pages/arrangementer/26-arrangementsregler"
-            style={{ marginTop: 0 }}
-          >
-            <Flex alignItems="center">
-              <Icon name="document" style={{ marginRight: '4px' }} />
-              <span>Regler for Abakus&#39; arrangementer</span>
```
norbye, line 469: > I would actually prefer to have "the rules" where it already was, as it's preferable if the sidebar isn't too crowded (ref. the PR we just merged addressing this very issue).

I agree that it's important to keep it light, but I've talked to people that weren't even aware that we have event rules - so I think the place where it is now is way to hidden, and I don't know any other places where it would fit, so I'd like to keep it in the bottom of the sidebar.

app/routes/events/components/EventDetail/index.js
```diff
@@ -460,21 +476,6 @@ export default class EventDetail extends Component<Props, State> {
...
-            </Flex>
-          </Link>
-          {loggedIn && (
-            <p>
-              Du kan oppdatere dine allergier og preferanser
```
norbye, line 474: > And in addition, with this, we're not consistent in our use of links: with the allergy link, it's at the end of the sentence, whereas with the rules, the link is the the entire sentence. Does that make sense? It looks a bit strange next to each other. I would prefer to not use links in the middle or at the end of sentences on a webapp (e.g. "klikk her for å [...]"), it should in stead be intuitive by means of buttons, icons or other practices.

Very much agree, it looks a bit strange with the inconsistency. I could change them out, but so far I haven't come up with any other clear way to format this than as links - as they either require too much context to fit in a button, and they don't really perform any actions - they are just links..  

I could switch them both out for full-width links though, might be a little bit better. Open for suggestions if you have any other ideas.

app/routes/events/components/EventDetail/index.js
```diff
@@ -460,21 +476,6 @@ export default class EventDetail extends Component<Props, State> {
               </Flex>
             </ContentSidebar>
           </ContentSection>
-          <Link
-            to="/pages/arrangementer/26-arrangementsregler"
-            style={{ marginTop: 0 }}
-          >
-            <Flex alignItems="center">
-              <Icon name="document" style={{ marginRight: '4px' }} />
```
norbye, line 468: > Can we also keep the icon? I like icons :sweat_smile:

I'll try to make it work. So far the formatting of this thing is what I've been struggling the most with, as the inconsistency you've mentioned was a bit too much with the text for allergies combined with the event-rules still having the icon..  
Might work better if some more of the inconsistencies are worked out though

app/routes/events/components/EventDetail/index.js
eikhr: The allergy link might make sense as a button with a text like "Oppdater allergier og preferanser". It's not really an action, but it is a link to a page where you would perform an action. 

Not sure how it would look with one button and one link though. 

Or maybe if you shortened down the text a bit, made the whole thing a link, and then threw in some food icon, so it matches the rules with the document icon.

app/routes/events/components/EventDetail/index.js
ivarnakken: I'll supplement that [we now have](https://github.com/webkom/lego-webapp/pull/2682) a tooltip on the "Frist for prikk" containing a link to the rules page. I'd argue the page is mostly used in regards to the "prikk" logic, and with the (very visible in my opinion) tooltip, this addition becomes less important, don't you think?

app/routes/events/components/EventDetail/index.js
norbye: Good ideas! Updated both to links with icons now, with the same formatting as the time and place. Updated the pics in the description

Didn't find any nice icons with food in ion, so I went with an edit-icon

eikhr: There should be a pizza icon, but the edit-icon also works fine

app/routes/events/components/EventDetail/index.js
norbye: Good point. With the updated design I put it in the bottom of the sidebar, so it's practically in the same place as it currently is on mobile (except for elevated users) - but a bit more visible on desktop, and shortened it down as the point of it including "prikker" is already stated in the prikkefrist tooltip.

app/routes/events/components/EventDetail/index.js
norbye: True - tried it out now but as the link also refers to preferences, and they are on the same page, it ended up being clearer that you can edit both than using some variant of pizza and cog.

app/routes/events/components/EventDetail/index.js
ivarnakken: It's much clearer now! What do you think about using the outlined document icon instead of the default filled one? Wouldn't that match the others more? 😄 Yes, I do realize it was filled from before.

ivarnakken: Also, the icons used on this page (including the follow star) are `ios-` icons, but, in my opinion the `md-` ones are much prettier. Just a though given, and I know it's out of scope of what you're trying to accomplish with this PR!

norbye: Agree that they should match, frankly the thought that document-outline existed never hit me :o  
Updated to both being outlined now

ivarnakken (approved):
This looks fantastic to me now! ❤️

eikhr (approved):
Looks great!

PR #2956, opened 2022-08-16
Fix event-detail title overflow, make participants-list fit content

norbye:
## ~Full-width
Only change here is that the list of participants has no height if there are no participants.
### Before
image

### After
image

## Smaller width
List of participants has no height if there are no participants  
Title and event-type wraps so none overflow and the title gets to use the entire width
### Before
image

### After
image


LudvigHz (approved):
LGTM :shipit:

eikhr (approved):
lgtm

ivarnakken (approved):
Nice 💯 The overflow was really annoying me

PR #2959, opened 2022-08-17
Upgrade Ionicons to version 5

ivarnakken:
## About time ...
The new version is very similar in looks, but is in my opinion much better, less _edgy_. 

Most notably though, icons are now SVGs, instead of being a font. This is clever, as the icon font would require us to bundle several megabytes of font data, even if we didn't use all the icons. Now with this, we only include the icons we actually need to use in our app.

I went through and checked every single use of `<Icon />` in our webapp, and they all work with the new version.

## Example

**Before**

**After**

-----

**Before**

**After**

### DIFF
--- a/app/components/Icon/index.js
+++ b/app/components/Icon/index.js
@@ -1,49 +1,46 @@
 // @flow
 
-import cx from 'classnames';
 import styles from './Icon.css';
 
 type Props = {
   /** Name of the icon can be found on the webpage*/
   name: string,
-  /** Name of the icon can be found on the webpage*/
   scaleOnHover?: boolean,
-  /** Name of the icon can be found on the webpage*/
   className?: string,
-  /** Name of the icon can be found on the webpage*/
   size?: number,
-  /** Name of the icon can be found on the webpage*/
-  prefix?: string,
-  /** Icon prefix. defaults to "ion-ios-" */
   style?: Object,
 };
 
 /**
- * Render an Icon
+ * Render an Icon like this with the name of your icon:
  *
- * <Icon name="add" > </Icon>
+ * <Icon name="add" />
  *
- * Just like this...
+ * Names can be found here:
+ * https://ionic.io/ionicons
  *
- * https://infinitered.github.io/ionicons-version-3-search/
  */
-function Icon({
+const Icon = ({
   name = 'star',
-  prefix = 'ion-ios-',
   scaleOnHover = false,
   className,
   style = {},
   size = 24,
   ...props
-}: Props) {
+}: Props) => {
   return (
-    <i
-      className={cx(`${prefix}${name}`, styles.icon, className)}
-      style={{ fontSize: `${size.toString()}px`, lineHeight: 1, ...style }}
+    <ion-icon
+      name={name}
+      class={className}
+      style={{
+        fontSize: `${size.toString()}px`,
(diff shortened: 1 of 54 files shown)


app/components/Icon/index.js
```diff
@@ -1,49 +1,46 @@
...
   ...props
-}: Props) {
+}: Props) => {
   return (
-    <i
-      className={cx(`${prefix}${name}`, styles.icon, className)}
-      style={{ fontSize: `${size.toString()}px`, lineHeight: 1, ...style }}
+    <ion-icon
+      name={name}
+      class={className}
```
ivarnakken, line 34: I know it's semi bad practice to use `class` instead of `className`, but it is the only way to override the cascaded styling for some dumb reason.

eikhr (commented):
😍

eikhr (approved):
Love the new icons!

Seems like you need to update the cypress tests, though. And some icons have gotten slightly misaligned with the text.
Screenshot 2022-08-18 at 09 48 29
image

ivarnakken:
> And some icons have gotten slightly misaligned with the text.

@eikhr. Those icons were slightly off from before:
image
image

But for some reason it is now more visible. I'll try to fix it up!

mattiastofte:
So sexy 😫

ivarnakken:
@eikhr. The icons are now aligned properly with the text. It still seems a bit off, but that's due to the font itself being ascended, thus there's nothing more to do (other than adding some margin to the text, but I would rather kill myself). Luckily, I'm currently working on replacing our font with `Inter`, which would eradicate this issue.  
image
image

norbye (approved):
Me likey!

If it's not too much hazzle to align the span and Icon in https://github.com/webkom/lego-webapp/blob/master/app/components/Poll/index.js#L122-L129 that would be awesome, but not critical as that will probably be taken care of with the new frontpage whenever that happens

ivarnakken:
@norbye, I aligned the icon you suggested as well 😄

cypress/e2e/create_event_spec.js
```diff
@@ -16,7 +16,9 @@ describe('Create event', () => {
   const uploadHeader = () => {
     // Upload file
     cy.upload_file(
-      c('ImageUploadField__coverImage') + ' ' + c('UploadImage__dropArea'),
+      c('ImageUploadField__coverImage') +
+        ' ' +
+        c('UploadImage__placeholderTitle'),
```
ivarnakken, line 21: Thanks to @eikhr for spotting that Cypress was failing because it was trying to click in the center of the image drop area, where the image icon was located, and that simply making it click on the text below instead solved the issue. Strange behavior by Cypress, but it's because the icon is now a div instead of being text (a font to be precise).

app/components/Upload/ImageUpload.js
```diff
@@ -111,7 +110,7 @@ const UploadArea = ({ multiple, onDrop, image }: UploadAreaProps) => {
       <div {...getRootProps({ className: styles.dropArea })}>
         <div className={styles.placeholderContainer}>
           <Icon size={82} name="image" />
-          <h2 className={styles.placeholdeTitle}>
+          <h2 className={styles.placeholderTitle}>
```
ivarnakken, line 113: This took me 3 hours to debug 👍🏼

PR #2964, opened 2022-08-19
Convert prikkfrist from datetime to time from now

norbye:
The only interesting part about the prikkfrist is when it is going to happen in relation to the current time, and as such I'd like it to show exactly that.

Although I think the feature would be nice, I'm torn about whether it would be confusing or not that it is currently not rerendering as the time from now changes. That could either be done manually with a timeout which would add more rerenders to the page, or with the react-moment library as suggested [here](https://stackoverflow.com/a/61947050) which would add yet another dependency. Or not at all.

Thoughts?

Preview:
image

image

image

### DIFF
--- a/app/routes/events/components/EventDetail/index.js
+++ b/app/routes/events/components/EventDetail/index.js
@@ -185,10 +185,12 @@ export default class EventDetail extends Component<Props, State> {
       ? () => unfollow(currentUserFollowing.id, event.id)
       : () => follow(currentUser.id, event.id);
 
+    const currentMoment = moment();
     const eventRegistrationTime = moment(event.activationTime).subtract(
       penaltyHours(penalties),
       'hours'
     );
+    const unregistrationDeadlineMoment = moment(event.unregistrationDeadline);
 
     const deadlines = [
       event.activationTime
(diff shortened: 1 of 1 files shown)


ivarnakken:
I totally agree that the only interesting part of the "prikkfrist" is when it is actually going to happen in relation to the current time, but whenever the "prikkfrist" is many days in the future, it actually gives me very little. In the example below, 8 days is difficult to "remember" because I can't see an actual date right off the bat (I know there's a tooltip, but it's not obvious that a tooltip exists either).
image

In other words, it makes a lot more sense when it's only hours or minutes till deadline, like in your examples. Could we maybe meet in the middle, and only show the time remaining whenever it is less than, let's say, 12 hours? 😄

ivarnakken:
Now that I think about it, does it even make sense to show the "Frist for prikk" _before_ registration has opened? I can't imagine anyone actually caring about the "prikkfrist" before they are able to register?

----

On that note, there's definitely room for improvement here:

- We don't need to specify when the registration opens two places, literally next to each other.
- We don't need to tell the user they are not registered ("Du er ikke påmeldt") when the registration hasn't even opened yet, _and_ it right above says no one is registered yet.

The more noise we can remove, the better.

LudvigHz:
> Now that I think about it, does it even make sense to show the "Frist for prikk" before registration has opened? I can't imagine anyone actually caring about the "prikkfrist" before they are able to register?

I very much disagree. I myself, and I think many others want to know if they can register and not get a prikk for unregistering without getting a prikk. It's very common that people register to save the spot and figure out later if they want to/can go or not.

ivarnakken:
>  I think many others want to know if they can register and not get a prikk for unregistering

Sure, I would want that myself, but I don't need to know about the specific deadline before I even have the opportunity to register? There's always a deadline, so there's no question of whether "it is safe" _prior_ to the registration opening. Remember, I'm talking about _before_ registration has opened.

norbye:
> Now that I think about it, does it even make sense to show the "Frist for prikk" before registration has opened? I can't imagine anyone actually caring about the "prikkfrist" before they are able to register?

Agree with @LudvigHz here. Even though the specific time you get the prikk is not really relevant before the registration opens, that is the only place you can see that you will eventually get a prikk if you don't unregister.

norbye:
Update to only display prikkfrist as timediff when there are <=12 hrs remaining.

norbye:
> We don't need to specify when the registration opens two places, literally next to each other.

Personally I like having both, as I never remember the current date, so it's nice to have a more relative format as well to let me know how long it is until the registration opens.

> We don't need to tell the user they are not registered ("Du er ikke påmeldt") when the registration hasn't even opened yet, and it right above says no one is registered yet.

Agreed, but we should still be able to display the "Du er påmeldt" if that would be the case, as people could have been registered by an admin before they are able to register themselves.

PR #2965, opened 2022-08-19
Convert registration-start to registration-end after registrations open

norbye:
During the redesign it was concluded to remove the unregistre/register deadline, on the grounds that it would be confusing to have both "avregistreringsfrist" and "avregistrering stenger".

I argue that these premisses have changed, as we no longer have "avregistreringsfrist", but "frist for prikk" - which in my opinion has a different meaning.  
In addition, "påmelding åpner" makes no sense after the registration has already opened.  
We also have some events that close the day before before to be able to order food, and it would be nice to show that as everyone might not realize that before it is too late to unregister.  
Aaand not all events use the prikk-system, but they still have times where the registration closes.

Preview:
image

image

image

image

### DIFF
--- a/app/routes/events/components/EventDetail/index.js
+++ b/app/routes/events/components/EventDetail/index.js
@@ -16,7 +16,11 @@ import { FormatTime, FromToTime } from 'app/components/Time';
 import InfoList from 'app/components/InfoList';
 import { Flex } from 'app/components/Layout';
 import Tooltip from 'app/components/Tooltip';
-import { colorForEvent, penaltyHours } from '../../utils';
+import {
+  colorForEvent,
+  penaltyHours,
+  registrationCloseTime,
+} from '../../utils';
 import Admin from '../Admin';
 import RegistrationMeta from '../RegistrationMeta';
 import DisplayContent from 'app/components/DisplayContent';
@@ -185,13 +189,16 @@ export default class EventDetail extends Component<Props, State> {
       ? () => unfollow(currentUserFollowing.id, event.id)
       : () => follow(currentUser.id, event.id);
 
-    const eventRegistrationTime = moment(event.activationTime).subtract(
+    const currentMoment = moment();
+    const activationTimeMoment = moment(event.activationTime);
+    const eventRegistrationTime = activationTimeMoment.subtract(
       penaltyHours(penalties),
       'hours'
     );
+    const registrationCloseTimeMoment = registrationCloseTime(event);
 
     const deadlines = [
-      event.activationTime
+      event.activationTime && currentMoment.isBefore(activationTimeMoment)
         ? {
             key: 'Påmelding åpner',
             value: (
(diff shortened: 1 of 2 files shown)


app/routes/events/components/EventDetail/index.js
```diff
@@ -230,6 +237,32 @@ export default class EventDetail extends Component<Props, State> {
...
+                  <span>
+                    Etter påmeldingen stenger er det hverken mulig å melde seg
+                    på eller av arrangementet.
+                  </span>
+                }
+              >
+                {currentMoment.isBefore(registrationCloseTimeMoment)
+                  ? 'Påmelding stenger'
+                  : 'Påmelding stengte'}{' '}
+                <Icon name="help-circle-outline" size={16} />
```
ivarnakken, line 255: Could you flex this [like this](https://github.com/webkom/lego-webapp/pull/2959/files#diff-74791df4b2365e0328f80b092e14dd504f8584ca1d6dcfc829e336d6542071c3R222-R225), so it'll still look good  once that is merged 😅

app/routes/events/components/EventDetail/index.js
```diff
@@ -230,6 +237,32 @@ export default class EventDetail extends Component<Props, State> {
...
         : null,
+      activationTimeMoment.isBefore(currentMoment)
+        ? {
+            key: 'Frist for av/påmelding',
+            keyNode: (
+              <Tooltip
+                content={
+                  <span>
+                    Etter påmeldingen stenger er det hverken mulig å melde seg
+                    på eller av arrangementet.
```
ivarnakken, line 248: It might be an idea to add a more thorough explanation of this to the rules page, or? E.g. why it was introduced in the first place and what happens if you still don't show up to the event.

ivarnakken (changes requested):
You're right, the premises have definitely changed, and this conveys it much clearer! 😍

app/routes/events/components/EventDetail/index.js
norbye: Updated now(: although gap={1} was a bit close for my taste, so I upped it to gap={4} - thoughts about doing the same to your PR?

ivarnakken: haha, sure, I'll do it right away!

app/routes/events/components/EventDetail/index.js
norbye: Probably a good idea, but that might be a matter we should probably include inter in. Although I think @juniwbjerde talked about updating those rules last semester, not sure what the status is on that. Might write a suggestion for updated rules some time

ivarnakken: I'll add it to our meeting agenda for next Tuesday!

ivarnakken (approved):
Looking good now!

LudvigHz (approved):

erlingfn (approved):
Looks good!

PR #2966, opened 2022-08-19
Rename allergies to 'matellergier/preferanser'

norbye:
Renamed with the purpose of making it more intuitive exactly what info we are requesting to prevent storing more sensitive info than we use.


ivarnakken (approved):

eikhr (approved):
No more "pollenallergi" 😆

PR #2972, opened 2022-08-22
Show ical and google calendar integration on mobile

eikhr:
We should probably do more to ensure people know this exists, but this a bare minimum. I honestly don't understand why it was hidden on mobile at all.

Before: 
Screenshot 2022-08-22 at 13 48 27
After:
Screenshot 2022-08-22 at 13 46 03


LudvigHz (approved):

norbye (approved):
Sweet

PR #2973, opened 2022-08-22
Make "breakfast talk" an event type

ivarnakken:
Requested by Fagkom.

Related to https://github.com/webkom/lego/pull/2861.

I took the liberty to use the color from the cover image that they usually use as the event color. 

image
image


eikhr (approved):
tet

PR #2974, opened 2022-08-22
update "Nyttige linker" link to this years buddy week

eikhr:
We probably should have updated this _before_ the "buddy week" started. But hey, better late than never!


ivarnakken (approved):
oooopps ....

PR #2975, opened 2022-08-22
use white background behind company and event images

eikhr:
Makes it easier to see dark logos with transparency in dark mode. In light mode it is invisible as all the backgrounds were already white.

I also changed `height` to `max-height` a few places, this was to avoid unneccesary white background around the images. The only layout change i noticed was that job-listings became slightly thinner.

## Events
before:
Screenshot 2022-08-22 at 20 47 14
Screenshot 2022-08-22 at 21 05 13

after:
Screenshot 2022-08-22 at 20 47 04
Screenshot 2022-08-22 at 21 05 05

## Front page
before:
Screenshot 2022-08-22 at 21 10 18
after:
Screenshot 2022-08-22 at 21 10 12

## Job listings
before:
Screenshot 2022-08-22 at 20 52 55
after:
Screenshot 2022-08-22 at 20 53 06

### DIFF
--- a/app/components/Content/Content.css
+++ b/app/components/Content/Content.css
@@ -36,6 +36,7 @@
   img {
     object-fit: cover;
     display: block;
+    background: white;
   }
 }
 
--- a/app/components/EventItem/styles.css
+++ b/app/components/EventItem/styles.css
@@ -27,6 +27,7 @@
 
 .companyLogo img {
   min-width: 120px;
-  height: 80px;
+  max-height: 80px;
   object-fit: contain;
+  background: white;
 }
--- a/app/components/Upload/UploadImage.css
+++ b/app/components/Upload/UploadImage.css
@@ -62,6 +62,7 @@ html[data-theme='dark'] .dropArea {
   height: inherit;
   width: 100%;
   object-fit: cover;
+  background: white;
 }
 
 .cropper {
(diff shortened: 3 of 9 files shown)


norbye (approved):
Nice, I like it!

ivarnakken (approved):
This won't work with white logos though, but that's kinda of rare and this is still a good temporary solution to https://github.com/webkom/lego/issues/2747. For the future, it'd be nice to support transparent images, so if you'd like to tackle the issue further, please do 😅

eikhr:
Yeah, something fancy to make it work with white transparent images would be cool, but I can't really think of any simple solutions off the top of my head. Two separate images is a possibility, but would require more work when creating events.

I force-pushed a small change now though, that makes the ImageUpload preview also have a white background, so at least it should be obvious when creating an event that a white transparent logo will be invisible.

ivarnakken:
>  but I can't really think of any simple solutions off the top of my head.

I've been pondering of a simple solution all summer as well. I'm probably gonna go at it some time this fall, but go ahead if you'd like to.

> I added a small change now though, that makes the ImageUpload preview also have a white background,

Nice!

----

Just realized that it doesn't seem like you've added a background to article covers, would you like to do so as well now while you're at it? 😄 I've previously spotted people employing the transparent Abakus logo sometimes multiple times ...

ivarnakken (approved):

PR #2980, opened 2022-08-23
Add button to toggle theme in header

LudvigHz:
The button is an icon that switches based on the current theme. If the
user is logged in, the theme will also be persisted to the
backend.

### DIFF
--- a/app/components/Header/ToggleTheme.js
+++ b/app/components/Header/ToggleTheme.js
@@ -1,0 +1,64 @@
+// @flow
+import { useCallback, useState, type Node } from 'react';
+import cx from 'classnames';
+import Icon from '../Icon';
+import styles from './toggleTheme.css';
+import { applySelectedTheme, getTheme } from 'app/utils/themeUtils';
+
+const getIcon = () => (getTheme() === 'dark' ? 'sunny' : 'moon');
+
+type Props = {
+  loggedIn: boolean,
+  username?: string,
+  updateUserTheme: (username: string, theme: string) => Promise<*>,
+  className?: string,
+  children?: Node,
+  isButton?: boolean,
+};
+
+const ToggleTheme = ({
+  loggedIn,
+  username,
+  updateUserTheme,
+  className: classN,
+  children,
+  isButton = true,
+}: Props) => {
+  const [icon, setIcon] = useState(getIcon());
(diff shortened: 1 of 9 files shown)


app/components/Header/ToggleTheme.js
```diff
@@ -0,0 +1,43 @@
...
+      const currentTheme = getTheme();
+      const newTheme = currentTheme === 'dark' ? 'light' : 'dark';
+      applySelectedTheme(newTheme);
+      setIcon(getIcon());
+      loggedIn && username && updateUserTheme(username, newTheme);
+    },
+    [loggedIn, username, updateUserTheme]
+  );
+
+  console.log(icon);
```
ivarnakken, line 29: :eyes:

app/components/Header/index.js
```diff
@@ -213,6 +215,12 @@ class Header extends Component<Props, State> {
             </div>
 
             <div className={styles.buttonGroup}>
+              <ToggleTheme
+                loggedIn={loggedIn}
+                updateUserTheme={this.props.updateUserTheme}
+                username={this.props.currentUser?.username}
+              />
```
ivarnakken, line 222: How does this look on mobile? The `.buttonGroup` is already very cramped (_note: we should reduce the gap on mobile_). I think I would prefer _not_ to have the toggle in the header on mobile, which sites like [Vite](https://vitejs.dev/) and [Docker](https://docs.docker.com/) do as well.

ivarnakken (changes requested):
This seems to work nicely! 💯 Next out is making the dark theme actually functional 😆

app/components/Header/index.js
eikhr: I agree, but I think it should be in the user drop-down or something on mobile, so it's not too hard to get to.

LudvigHz: I have added it to the dropdown now (when logged in)

LudvigHz:
[Screencast from 08-30-2022 11:25:57 PM.webm](https://user-images.githubusercontent.com/42850232/187546334-92446d8c-c3d5-425a-b5fe-3fa8e0522211.webm)


image
image

ivarnakken (approved):
Looks awesome now 💯

PR #2982, opened 2022-08-23
Set survey default activation time to noon the day after

ingraso:
Small fix to do what is described here: https://github.com/webkom/lego/issues/1998

> It is currently the time that you make the survey. We should set it to something like 24h in the future, so that people don't accidentally create it and send it out immediately, when it is not ready.


norbye (approved):

ivarnakken (approved):
Awesome 💯

SmithPeder (approved):
:shipit:

PR #2998, opened 2022-08-29
change banner to opptak

Arashfa0301:
changed the banner from "fadderuke" to "opptak". Though I'm not a big fan of the plain red colour that it uses. Maybe some patters would be nice 👀


danielyanghansen (commented):
LGTM, you can push to prod ASAP :)

danielyanghansen (approved):

ivarnakken (approved):

PR #3000, opened 2022-08-30
fix banner typo

Arashfa0301:
Abakus has opptak


norbye (approved):
Might be cleaner to revert the previous PR and update it there, but idk if we usually do this so it's probably fine

eikhr (approved):

ivarnakken (approved):

PR #3002, opened 2022-08-30
Make the PhotoConsent buttons more legible

LudvigHz:
Disable when the option is already selected, and remove the border when
selected.

Also fix some event link issues.


ivarnakken (approved):
Much better 💯

AdrianAndersen:
Nice!

PR #3017, opened 2022-09-05
ABA-19: Add the possibility to scan QR codes instead of searching for users in Abacard

AdrianAndersen:
Made it possible to scan QR codes instead of searching for the user's name. The QR code is available on all users profiles, and is just their username encoded. 

image
image

### DIFF
--- a/app/components/Search/SearchPage.js
+++ b/app/components/Search/SearchPage.js
@@ -14,7 +14,7 @@ type Props = {
   onQueryChanged: (string) => void,
   placeholder?: string,
   results: Array<SearchResult>,
-  handleSelect: (SearchResult) => Promise<*>,
+  handleSelect: (string) => Promise<*>,
 };
 
 const SearchPage = (props: Props) => {
@@ -59,7 +59,7 @@ const SearchPage = (props: Props) => {
   const handleSelect = (result: SearchResult) => {
     setQuery('');
     setSelectedIndex(0);
-    props.handleSelect(result);
+    props.handleSelect(result.username ?? '');
   };
 
   const handleQueryChange = ({
--- a/app/components/Search/SearchPage.js
+++ b/app/components/Search/SearchPage.js
@@ -10,7 +10,7 @@ import qs from 'qs';
 type Props = {
   searching: boolean,
   location: Object,
-  inputRef?: (?HTMLInputElement) => void,
+  inputRef?: {| current: ?HTMLInputElement |},
   onQueryChanged: (string) => void,
   placeholder?: string,
   results: Array<SearchResult>,
@@ -31,7 +31,7 @@ const SearchPage = (props: Props) => {
     );
 
     setSelectedIndex(adjustedSelectedIndex);
-  }, [props, selectedIndex, setSelectedIndex]);
+  }, [props.results, selectedIndex, setSelectedIndex]);
 
   const handleKeyDown = (e: KeyboardEvent) => {
     if (props.results.length === 0) return;
--- a/app/components/Search/SearchPage.js
+++ b/app/components/Search/SearchPage.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { Component } from 'react';
+import { useEffect, useState } from 'react';
 import SearchPageInput from 'app/components/Search/SearchPageInput';
 import SearchPageResults from 'app/components/Search/SearchPageResults';
 import { Keyboard } from 'app/utils/constants';
@@ -17,52 +17,38 @@ type Props = {
   handleSelect: (SearchResult) => Promise<*>,
 };
 
-type State = {
-  query: mixed,
-  selectedIndex: number,
-};
+const SearchPage = (props: Props) => {
+  const [selectedIndex, setSelectedIndex] = useState<number>(0);
+  const [query, setQuery] = useState<mixed>(
+    qs.parse(props.location.search, { ignoreQueryPrefix: true }).q || ''
+  );
 
-class SearchPage extends Component<Props, State> {
-  state = {
-    selectedIndex: 0,
-    query:
-      qs.parse(this.props.location.search, { ignoreQueryPrefix: true }).q || '',
-  };
-
-  // eslint-disable-next-line
-  componentWillReceiveProps(nextProps: Props) {
+  useEffect(() => {
     // Make sure the selectedIndex is within 0 <= index < results.length:
-    const selectedIndex = Math.min(
-      this.state.selectedIndex,
-      Math.max(nextProps.results.length - 1, 0)
+    const adjustedSelectedIndex = Math.min(
+      selectedIndex,
+      Math.max(props.results.length - 1, 0)
     );
 
-    this.setState({ selectedIndex });
-  }
+    setSelectedIndex(adjustedSelectedIndex);
+  }, [props, selectedIndex, setSelectedIndex]);
 
-  handleKeyDown = (e: KeyboardEvent) => {
-    if (this.props.results.length === 0) return;
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (props.results.length === 0) return;
     switch (e.which) {
       case Keyboard.UP:
         e.preventDefault();
-        this.setState({
-          selectedIndex: Math.max(0, this.state.selectedIndex - 1),
-        });
+        setSelectedIndex(Math.max(0, selectedIndex - 1));
         break;
 
       case Keyboard.DOWN:
         e.preventDefault();
-        this.setState({
-          selectedIndex: Math.min(
-            this.props.results.length - 1,
-            this.state.selectedIndex + 1
-          ),
-        });
+        setSelectedIndex(Math.min(props.results.length - 1, selectedIndex + 1));
         break;
 
       case Keyboard.ENTER:
         e.preventDefault();
-        this.handleSelect(this.props.results[this.state.selectedIndex]);
+        handleSelect(props.results[selectedIndex]);
         break;
 
       default:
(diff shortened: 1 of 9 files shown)


app/components/UserValidator/index.js
```diff
@@ -56,39 +59,64 @@ class Validator extends Component<Props, State> {
...
+                content: '',
+                icon: '',
+                label: '',
+                link: '',
+                path: '',
+                picture: '',
+                value: '',
+              });
+            }
           }}
```
danielyanghansen, line 106: Do we need some sort of error handling here (or atleast a console log)? Could be an idea to just add a pop up or something, any kind of error feed back is nice.
```suggestion
          onResult={(res, error) => {
            if (!processing && res) {
              handleSelect({
                username: res.getText(),
                result: '',
                color: '',
                content: '',
                icon: '',
                label: '',
                link: '',
                path: '',
                picture: '',
                value: '',
              });
            }
            if (!!error) {
                console.info(error)
            }
          }}
```

AdrianAndersen: Sure, should not really matter tho, as when it throws an error, the scanner will prob not be displayed. I added it

app/routes/users/components/UserProfile.js
```diff
@@ -177,7 +181,14 @@ const ListEvents = ({ events, noEventsMessage, loggedIn }: EventsProps) => (
   </div>
 );
 
-export default class UserProfile extends Component<Props, EventsProps> {
+type UserProfileState = { showAbaId: boolean };
+
+export default class UserProfile extends Component<Props, UserProfileState> {
+  constructor(props: Props) {
```
ivarnakken, line 187: Bold of you to not rewrite this to a functional component, but okay

package.json
```diff
@@ -82,6 +82,8 @@
     "react-notification": "^6.8.5",
     "react-overlays": "5.2.0",
     "react-phone-number-input": "^3.2.11",
+    "react-qr-reader": "^3.0.0-beta-1",
```
ivarnakken, line 85: Nice, I love beta™️

app/components/UserValidator/index.js
```diff
@@ -56,39 +60,73 @@ class Validator extends Component<Props, State> {
...
-    );
-  }
-}
+      </Modal>
+      <Button
+        className={styles.scannerButton}
+        onClick={() => setShowScanner(true)}
+      >
+        <Icon name="qr-code" size={18} />
+        Vis Scanner
```
ivarnakken, line 118: I prefer to not capitalize every word.
```suggestion
        Vis scanner
```

app/components/UserValidator/index.js
```diff
@@ -56,39 +60,73 @@ class Validator extends Component<Props, State> {
...
-          <i className="fa fa-check" />
-        </div>
-        <SearchPage
-          {...this.props}
-          placeholder="Skriv inn brukernavn eller navn"
-          handleSelect={this.handleSelect}
-          inputRef={(input) => {
-            this.input = input;
+  return (
+    <div>
```
ivarnakken, line 70:
```suggestion
    <>
```

app/components/UserValidator/Validator.css
```diff
@@ -1,6 +1,7 @@
 @import '~app/styles/variables.css';
 
 .overlay {
+  z-index: 100;
```
ivarnakken, line 4: :shipit:

package.json
ivarnakken: I also despise the fact that we need two whole deps for this

ivarnakken (changes requested):
This definitely has potential 💯 Some minor comments

package.json
AdrianAndersen: Tbf, this is the official released latest version (dunno why they branded it as beta), and this package is the most popular qr-scanner for React out there. Also, these two packages combined only amounts to less than 100kb increase of the bundle size. So this should not be a concern.

app/components/UserValidator/index.js
AdrianAndersen: ✅

app/components/UserValidator/index.js
AdrianAndersen: ✅

app/components/UserValidator/Validator.css
AdrianAndersen: This is required to display the overlay with feedback after scanning in front of the scanner modal. I agree that you could lower the number, but really does it make a difference? If you have an alternative suggestion to this approach, please let me know. ✌🏻

app/routes/users/components/UserProfile.js
AdrianAndersen: Yeah, that was kinda lazy. Rewrote it now. ✌🏻

app/components/UserValidator/index.js
```diff
@@ -21,28 +22,31 @@ type Props = {
...
-class Validator extends Component<Props, State> {
-  input: ?HTMLInputElement;
-  state = {
-    showCompleted: false,
-  };
+const Validator = (props: Props) => {
+  const [input, setInput] = useState<?HTMLInputElement>();
+  const [completed, setCompleted] = useState(false);
+  const [showScanner, setShowScanner] = useState(false);
+  const [processing, setProcessing] = useState(false);
```
danielyanghansen, line 29: Small thing, but is it necessary to use separate states for processing and completion?
Might be better to use `useState<'completed' | 'not_started' | 'processing'>('not_started')` or something along those lines

package.json
eikhr: 100Kb is not really a _small_ increase in bundle size. It will add about 5% to the total bundle size of `lego-webapp`, but as it should only be loaded on the user-validator page it's not really a problem.

app/components/UserValidator/index.js
```diff
@@ -56,39 +60,73 @@ class Validator extends Component<Props, State> {
...
+                result: '',
+                color: '',
+                content: '',
+                icon: '',
+                label: '',
+                link: '',
+                path: '',
+                picture: '',
+                value: '',
+              });
```
eikhr, line 104: Instead of doing this, could you instead refactor `handleSelect` to only take the username? Only the username is used anyways, and with it nothing else would even work.

eikhr (changes requested):

package.json
AdrianAndersen: Yeah, that is a good point. But as you said, since this increase only is in the validator, it is acceptable. QR code generator on the user profile page adds about 4kb for reference.

app/components/UserValidator/index.js
AdrianAndersen: Good point. Refactored it to only take the username. Also refactored SearchPage and Abacard to functional along the way.

eikhr:
You will need to rebase this

app/components/Search/SearchPage.js
```diff
@@ -59,7 +59,7 @@ const SearchPage = (props: Props) => {
   const handleSelect = (result: SearchResult) => {
     setQuery('');
     setSelectedIndex(0);
-    props.handleSelect(result);
+    props.handleSelect(result.username ?? '');
```
eikhr, line 62: Do you know if `username` in `SearchResult` can ever be undefined? It seems to pass flow fine without being optional, and I can't really think of a situation where it would be undefined.

AdrianAndersen: Yeah, it happens when you search for other things, such as events, meetings. However, in this search it will filter for only users. The nullish check is only needed because of the SearchResult flow type.

Branch is up to date now btw

AdrianAndersen: Without the null check, flow fails for me at least ✌🏻

eikhr: Ohh, right. It's the same type for all search results. That makes sense.

eikhr (approved):
Good work, can't wait to try it!

Cypress is being a little weird today though, so it has failed, but that is not caused by anything in this pr.

app/components/UserValidator/index.js
```diff
@@ -9,40 +9,44 @@ import styles from './Validator.css';
...
   searching: boolean,
 };
 
-class Validator extends Component<Props, State> {
-  input: ?HTMLInputElement;
-  state = {
-    showCompleted: false,
-  };
+const Validator = (props: Props) => {
+  const [input, setInput] = useState<?HTMLInputElement>();
```
LudvigHz, line 26: This should be a `useRef`

LudvigHz (approved):
Just fix up some small things and we're good to go!
The diff on some of the class -> func was pretty impossible to read so I'm assuming it's correct :/

app/components/UserValidator/index.js
```diff
@@ -56,39 +60,62 @@ class Validator extends Component<Props, State> {
...
+      >
+        <h1>Scan ABA-ID</h1>
+        <QrReader
+          onResult={(res, error) => {
+            if (!processing && res) {
+              handleSelect(res.getText());
+            }
+            if (error) {
+              console.info(error);
+            }
```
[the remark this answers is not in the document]
AdrianAndersen: Yeah, it is only for if the scanner does not open, you can check console to see why it broke. (very edge case tho). I think it is unnecessary to send this to SendGrid, so I just kept it for convenience.

app/components/Search/SearchPage.js
```diff
@@ -14,97 +14,85 @@ type Props = {
...
-      Math.max(nextProps.results.length - 1, 0)
+    const adjustedSelectedIndex = Math.min(
+      selectedIndex,
+      Math.max(props.results.length - 1, 0)
     );
 
-    this.setState({ selectedIndex });
-  }
+    setSelectedIndex(adjustedSelectedIndex);
+  }, [props, selectedIndex, setSelectedIndex]);
```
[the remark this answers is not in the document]
AdrianAndersen: ✅

app/components/UserValidator/index.js
AdrianAndersen: ✅

LudvigHz (approved):
LGTM :shipit:

PR #3021, opened 2022-09-06
fix bug where opening a modal scrolls to bottom of page

eikhr:
😡😠😠 This took waaaaay too long to debug

The styling needs to be on the modal-element, not on the div inside it...

fixes https://github.com/webkom/lego/issues/2841


ivarnakken (approved):
😆

LudvigHz (approved):

PR #3022, opened 2022-09-06
Add auto-labeler action

LudvigHz:
Fixes ABA-27


ivarnakken (approved):
I would still like @webkom-R2D2 to do the labeling but oh well 🤷🏼

PR #3040, opened 2022-09-13
fix bug where comment form would not clear after submit

eikhr:
LegoEditor is uncontrolled, so we cannot set the editor text from our code. What we can do though, is make sure that each separate comment uses a separate editor-instance by setting a unique key. As each new editor is intially empty, this will clear the comment form whenever the form changes to a new comment.

For the key i just used Math.random(), as we don't have a UUID library. I assume this will be good enough.

Fixes https://github.com/webkom/lego/issues/1464

### DIFF
--- a/app/components/CommentForm/index.js
+++ b/app/components/CommentForm/index.js
@@ -57,6 +57,7 @@ const CommentForm = ({
     <LegoFinalForm
       initialValues={{
         text: EDITOR_EMPTY,
+        commentKey: Math.random(),
       }}
       validate={validate}
       onSubmit={({ text }) => {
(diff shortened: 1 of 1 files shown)


app/components/CommentForm/index.js
```diff
@@ -69,7 +70,14 @@ const CommentForm = ({
         );
       }}
     >
-      {({ handleSubmit, pristine, submitting, form }) => {
+      {({
+        handleSubmit,
+        pristine,
+        submitting,
+        form,
+        // $FlowFixMe
```
ivarnakken, line 78: Was this too difficult to type? Our webapp has already enough technical debt 😬

ivarnakken (approved):
Nice! 💯 The random key is definitely very hacky, but it's better than nothing!

app/components/CommentForm/index.js
eikhr: yes

LudvigHz (approved):
Not the best key no, but if it works it works :shrug:

PR #2381, opened 2021-12-23
[WIP] Redesign Header

ivarnakken:
# ⚠️ Under construction 👷🏼‍♂️ 

My proposition for a ✨redesigned✨ Header

**TODO**
- [x] New dropdown functionality
- [x] Improve visuals
- [ ] Register dropdown
- [ ] Mobile support
- [ ] Dark-theme check

---

Closes ABA-143

PR #3043, opened 2022-09-13
Give indicator styling a more modern look

mattiastofte:
Not a very invasive change although I believe it contributes a lot to both the look and feel of the webpage. The animations feel smoother (expands only on the x-axis and not upwards) and the border is not rounded (why would that make sense?). Screenshots are provided under. I've also changed the colour of the navigator which is more controversial 😬.

image
image

### DIFF
--- a/app/components/Header/Header.css
+++ b/app/components/Header/Header.css
@@ -157,12 +157,10 @@
   &::after {
     content: '';
     height: 10px;
-    left: 0;
-    right: 0;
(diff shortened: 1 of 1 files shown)


eikhr (approved):
I like it, great to improve this now even though it will eventually be changed in #2381

ivarnakken (approved):
Awesome 💯

LudvigHz (approved):

PR #3045, opened 2022-09-14
Linkify and truncate links in announcements

ollfkaih:
Before
image
After
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,8 @@
     "jsdom": "^20.0.0",
     "jszip": "^3.10.1",
     "jwt-decode": "3.1.2",
+    "linkify-react": "^3.0.4",
+    "linkifyjs": "^3.0.5",
     "lodash-es": "^4.17.21",
     "mazemap": "file:mazemap",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -60,6 +60,8 @@
     "jsdom": "^20.0.0",
     "jszip": "^3.10.1",
     "jwt-decode": "3.1.2",
+    "linkify-react": "^3.0.4",
+    "linkifyjs": "^3.0.5",
```
ivarnakken, line 64: Is `linkifyjs` also needed?

ivarnakken (approved):
Looks very nice! 💯

package.json
LudvigHz: It's a peer-dependency, so unfortunately yes :'(

LudvigHz (approved):
Nice :100: 

It would be awesome if you could add this to polls (description) as well :grimacing:

ivarnakken:
You'll have to lint your new changes. Please just amend it 😄

PR #3056, opened 2022-09-19
fix: add a z-index to Tooltip

olestole:
Resolves webkom/lego#2941

image


eikhr (approved):
Thanks for fixing this!

LudvigHz (approved):

PR #3062, opened 2022-09-22
Make Past Events on Profile Page Extra Compact

danielyanghansen:
Just making the components for previous events more compact.
Planning on using pagination in a later PR 

ExtraCompactOldEvents

### DIFF
--- a/app/components/EventItem/index.js
+++ b/app/components/EventItem/index.js
@@ -18,6 +18,8 @@ type AttendanceProps = {
   event: Event,
 };
 
+export type EventStyle = 'default' | 'extra-compact';
+
 const Attendance = ({ event }: AttendanceProps) => {
   const attendance = eventAttendance(event);
   return (
@@ -57,6 +59,7 @@ type EventItemProps = {
   field?: EventTimeType,
   showTags?: boolean,
   loggedIn: boolean,
+  eventStyle?: EventStyle,
 };
 
 const EventItem = ({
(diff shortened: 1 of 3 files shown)


eikhr (changes requested):
This creates a lot of duplicate code, could you try to add a prop `compact` to `EventItem` that toggles between the normal and compact version. This way we only need one component and avoid duplicating all the parts that are the same.

app/components/EventItem/index.js
```diff
@@ -18,6 +18,8 @@ type AttendanceProps = {
   event: Event,
 };
 
+export type EventStyle = 'default' | 'extra-compact';
```
ivarnakken, line 21: I think I'd prefer a `compact` boolean prop

app/components/EventItem/index.js
```diff
@@ -57,45 +59,72 @@ type EventItemProps = {
...
+            {showTags && (
+              <Flex wrap>
+                {event.tags.map((tag, index) => (
+                  <Tag key={index} tag={tag} />
+                ))}
+              </Flex>
+            )}
+          </div>
+
+          <div className={styles.companyLogo}>
```
ivarnakken, line 120: Turn this `div` into a `Flex`, and remove the unnecessary CSS. Same goes for the compact version above.

ivarnakken (changes requested):

app/components/EventItem/index.js
danielyanghansen: The reason why I'm not using a boolean here is because I'm working on another style: `compact`,
so I'm aiming to have atleast 3 styles with ` 'default' | 'compact'| 'extra-compact'`;

Should I just "pit stop" with a `compact` boolean here then?

ollfkaih (approved):

eikhr (approved):

PR #3075, opened 2022-09-28
Fix poll info icon & remove tooltip with icon names

LudvigHz:
The name attribute would make it so the default tooltip whenever you
hovered over an ion-icon element.

Before:
image

After:
image

### DIFF
--- a/app/components/Poll/index.js
+++ b/app/components/Poll/index.js
@@ -117,7 +117,7 @@ class Poll extends Component<Props, State> {
 
     return (
       <div className={cx(styles.poll, backgroundLight ? styles.pollLight : '')}>
-        <Flex wrap justifyContent="space-between" alignItems="center">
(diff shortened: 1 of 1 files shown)


app/styles/globals.css
```diff
@@ -10,6 +10,10 @@
   -webkit-font-smoothing: antialiased;
 }
 
+ion-icon {
+  pointer-events: none;
```
eikhr, line 14: I think we need to find a different solution than this. This removes the hover-effect from icons, and makes cypress refuse to click any icons.

eikhr (changes requested):

app/styles/globals.css
LudvigHz: I just removed this then :/ I couldn't really find another good way to get rid of it, and it's not _that_ important.

ivarnakken (approved):

PR #3090, opened 2022-10-04
Remove noreferrer from job listing links

trthingnes:
Removes `rel="noreferrer"` from job listing links to allow businesses to see where the traffic is coming from.
This will incentivize businesses to post more job listings because they can see how the listing performs.

### DIFF
--- a/app/routes/company/components/CompanyDetail.js
+++ b/app/routes/company/components/CompanyDetail.js
@@ -160,6 +160,7 @@ const CompanyDetail = (props: Props) => {
             <a
               href={joblisting.applicationUrl}
               target="_blank"
+              rel="noopener"
             >
               <strong>SØK HER</strong>
             </a>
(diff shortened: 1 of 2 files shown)


LudvigHz:
Thanks for the contribution! Although we have to discuss a bit before we can accept this.

Also, you would need to change this to use `noopener` instead of just removing `noreferrer` so we can keep the `noopener` attribute for security reasons (see the [failed CI](https://ci.webkom.dev/webkom/lego-webapp/5274) for more details).

trthingnes:
It seems like `noreferrer` should not be removed without also removing `_blank`.

ivarnakken:
@trthingnes, yes, we're not too happy about removing `_blank`, but is it still needed now that you've added `noopener`? It seems, from what I've checked that you only really need `noopener` to prevent security vulnerabilities, by preventing the new page from being able to access the `window.opener` property, which ensures it runs in a separate process. Or, were you thinking of something else?

trthingnes:
I tested out adding no-opener. It is still a vulnerability in older browsers. Because of this it still fails during lint.

ivarnakken:
Right, that makes sense, I didn't check the linter hehe. Then I suppose we'll leave this be. Thanks a lot for the input though, it made me more aware of the existing vulnerabilities!

PR #3094, opened 2022-10-05
Fix typos in tooltip and validation in the event editor

ingraso:
The first typo, "Manuell betaling kan også av i etterkant" does not really make sense, and the second one is using the wrong term "avmelding" instead of "avregistrering".

image


ivarnakken (approved):

ivarnakken:
@ingraso, could you rebase this to `master`? We just merged a fix for the CI, so it should pass now.

LudvigHz (approved):

PR #3103, opened 2022-10-07
Fix cypress tests by finding the correct date

ivarnakken:
When `tomorrowDate` would for example be "8", cypress would falsely try to click on "28" (from the previous month) because this comes before "8" in the DOM. The dates from the previous month would, for obvious reasons, be disabled and thus cause the tests to fail.

This was solved by simply utilizing a regex.


eikhr (approved):
lgtm

PR #3116, opened 2022-10-12
Improve layout and design of sidebar

mattiastofte:
First change of the planned improvements of the /pages/ route layout. 

This PR includes:
- Improved alignment of list items
- New icon for dropdown buttons
- Flip animation/transition for icon
- Cleaner design with small tint of background instead of right border/divider
- Links are gray while the "abakule" circle is kept red
- Mobile look is vastly improved (although not perfect)

Light mode:
image

Dark mode:
image

### DIFF
--- a/app/routes/pages/components/Sidebar.js
+++ b/app/routes/pages/components/Sidebar.js
@@ -51,9 +51,10 @@ class Sidebar extends Component<Props, State> {
                 <Icon name="close" size={50} />
               </button>
               <h3 className={styles.sidebarHeader}>Om Abakus</h3>
+              {/*
               <h4 className={styles.sidebarSubtitle}>
                 {categorySelected ? categorySelected : 'Generelt'}
-              </h4>
+              </h4>*/}
               {/* <div className={styles.sidebarPicture}>
                 <h4 className={styles.pictureHeader}> {"Abakus' Fortid"}</h4>
                 <a href="https://abakus.no/">
--- a/app/routes/pages/components/PageDetail.css
+++ b/app/routes/pages/components/PageDetail.css
@@ -14,7 +14,8 @@
 }
 
 .page {
-  margin: -20px 0px 0px 0px;
+  margin: 0;
+  margin-top: -20px;
   width: 100%;
 }
 
--- a/app/routes/pages/components/PageDetail.css
+++ b/app/routes/pages/components/PageDetail.css
@@ -14,7 +14,7 @@
 }
 
 .page {
-  margin: 0;
+  margin: -20px 0px 0px 0px;
   width: 100%;
 }
 
--- a/app/routes/pages/components/PageHierarchy.css
+++ b/app/routes/pages/components/PageHierarchy.css
@@ -11,25 +11,28 @@
 
 .pageList {
   text-align: left;
-  line-height: 18px;
-  width: 220px;
+  line-height: 28px;
+  width: 100%;
 }
 
 .pageList a {
-  font-size: 18px;
+  font-size: 17px;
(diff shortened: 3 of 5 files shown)


Arashfa0301:
Could you include an illustration of the mobile view in this PR ?

app/routes/pages/components/Sidebar.js
```diff
@@ -51,9 +51,10 @@ class Sidebar extends Component<Props, State> {
                 <Icon name="close" size={50} />
               </button>
               <h3 className={styles.sidebarHeader}>Om Abakus</h3>
+              {/*
               <h4 className={styles.sidebarSubtitle}>
                 {categorySelected ? categorySelected : 'Generelt'}
-              </h4>
+              </h4>*/}
```
Arashfa0301, line 57: Do you need this comment here ?

ivarnakken: I'd say you can just remove it entirely.

LudvigHz:
Nice! :100: 
Some of my thoughts:
- This new design looks a lot _cleaner_ (especially on dark mode).
- However, IMO it might be a bit _too clean_. In that I miss some contrast and indication that the different menu items are clickable. Could f.ex. having some background for either theall the different items, or just the selected one work?
Here is one example with a border between each item, and some different color on hover:
image

- It still looks kinda off to me that the red bullet point is inset in the list item :thinking:. I wonder how it would look if it was just before the text without "shifting" it to the right. I.e. like this:
  image

LudvigHz:
Another Idea I had is to also change the bullet point with a border, if you go for the "block" click target as above (i.e. each element is more a button than just a link).
Something like this:
image

mattiastofte:
Great feedback! I agree that the design could be more readable so I'll keep experimenting with different approaches to achieve this. Although I think it's kinda cool to use a red circle for indication, since it emphasises the abakule:)

eikhr:
I really like the newest design, however I think the border between items would look better if they were lighter than the background in dark mode.

<table>
<tr>
	<td>Like this</td>
	<td>Instead of this</td>
<tr>
	<td>

Screenshot 2022-10-24 at 14 54 22
</td>
	<td>

Screenshot 2022-10-24 at 14 54 31
</td>
</table>

ivarnakken:
@mattiastofte Is this turning stale? I feel like we had something very nice. If you're bored of working on it, we can fix up the small stuff, get it merged in, and then iterate on it later. I think anything is better than the current design.

mattiastofte:
It's kinda stale since the branch is quite far behind the main branch so I'd have to rebase and I'm not super confident with that. I also would like to improve the design futher as I've gotten some ideas and I've done something similar to the sidenav og kateter.no. I'll look at it tonight!

mattiastofte:
> @mattiastofte Is this turning stale? I feel like we had something very nice. If you're bored of working on it, we can fix up the small stuff, get it merged in, and then iterate on it later. I think anything is better than the current design.
> 
> 
> 
> BTW: I think it would look good to use the `var(--additive-background)` as a hover background, and no border.

It's kinda stale since the branch is quite far behind the main branch so I'd have to rebase and I'm not super confident with that. I also would like to improve the design futher as I've gotten some ideas and I've done something similar to the sidenav og kateter.no. I'll look at it tonight!

ivarnakken:
>I also would like to improve the design futher as I've gotten some new ideas and I've done something similar to the sidenav on kateter.no. I'll look at it tonight!

Awesome! 💯 Excited to see the result 😄 Let me know if you need any help rebasing.

ivarnakken:
Closed in favor of #3553

PR #3117, opened 2022-10-12
Fix attendees stats bug on admin page

ivarnakken:
Introduced in #3107.

People that had paid "manual" were not taken into account, and instead of comparing the stats to capacity it should be compared to registration count.


LudvigHz (approved):
:eyes: 

:shipit:

PR #3121, opened 2022-10-12
Add analytics snippet

LudvigHz:
This adds an analytics snippet to start tracking some analytics again!

This points to our self-hosted plausible instance, so all data is stored in the same method as our
other data.

Additionally, this analytics snippet is GDPR-compliant and does _not_ store any cookies, saving us
from the dreaded cookie-popup hell.

Fixes ABA-93

### DIFF
--- a/server/pageRenderer.js
+++ b/server/pageRenderer.js
@@ -68,6 +68,14 @@ const readyHtml = (app) => {
   }
 };
 
+const analytics = __DEV__
+  ? ''
+  : // The .replace() remove the protocol (https://) part of the url, leaving just the domain
+    `<script defer data-domain=${config.webUrl.replace(
+      /(^\w+:|^)\/\//,
+      ''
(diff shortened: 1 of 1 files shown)


server/pageRenderer.js
```diff
@@ -68,6 +68,10 @@ const readyHtml = (app) => {
   }
 };
 
+const analytics = __DEV__
+  ? ''
+  : `<script defer data-domain=${config.webUrl} src="https://analytics.webkom.dev/js/plausible.js"></script>`;
```
ivarnakken, line 73: `config.webUrl` includes `https://`, but does that matter?

LudvigHz: Fixed now. I added a quick regex replace for all URL protocols

ivarnakken (approved):
Looks good now 💯 Excited ...

eikhr (approved):

PR #3122, opened 2022-10-12
improved styling of the notification dropdown

haukkagu:
* Moved the icons to the top of each notification
* Added a separator between each notification
* Decreased line height from 1.8 to 1.5
* Increased the height of the dropdown by 100 pixels

Screenshot:
notification4

### DIFF
--- a/app/components/HeaderNotifications/HeaderNotifications.css
+++ b/app/components/HeaderNotifications/HeaderNotifications.css
@@ -22,10 +22,11 @@ html[data-theme='dark'] .notifications {
 
 .notification {
   border-bottom: 1px solid lightgray;
-  padding: 15px;
+  padding: 15px 15px 0;
   display: flex;
   flex-direction: row;
   color: var(--lego-font-color);
+  line-height: 1.5;
 
   &:last-child {
     border-bottom: none;
@@ -41,12 +42,14 @@ html[data-theme='dark'] .notifications {
 }
 
 .innerNotification {
+  width: 100%;
   display: flex;
   flex-direction: row;
-  align-items: center;
   padding-right: 15px;
   word-break: break-word;
   white-space: pre-line;
+  border-bottom: 1px solid var(--color-mono-gray-3);
+  padding-bottom: 15px;
 }
 
 .unRead {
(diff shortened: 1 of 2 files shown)


app/components/HeaderNotifications/HeaderNotifications.css
```diff
@@ -22,10 +22,13 @@ html[data-theme='dark'] .notifications {
 
 .notification {
   border-bottom: 1px solid lightgray;
-  padding: 15px;
+  padding-top: 15px;
+  padding-right: 15px;
+  padding-left: 15px;
```
ivarnakken, line 27: This is the same as `padding: 15px 15px 0`

LudvigHz, line 27: You can actually rewrite this to
```suggestion
  padding: 15px 15px 0 15px;
```

See https://developer.mozilla.org/en-US/docs/Web/CSS/padding#syntax

app/components/HeaderNotifications/HeaderNotifications.css
```diff
@@ -41,9 +44,10 @@ html[data-theme='dark'] .notifications {
 }
 
 .innerNotification {
+  border-bottom: 1px solid var(--color-mono-gray-3);
+  padding-bottom: 15px;
```
ivarnakken, line 48: Veeery picky, but nice to get in early: please move color and "styling" css to the bottom, and let position/display/size lines come first.

ivarnakken (changes requested):
This looks much better! 💯 

But, it's kinda annoying that the border line is not as long as the text - any way you can change that?

app/components/HeaderNotifications/HeaderNotifications.css
haukkagu: fixed

app/components/HeaderNotifications/HeaderNotifications.css
haukkagu: fixed

app/components/HeaderNotifications/HeaderNotifications.css
haukkagu: fixed

LudvigHz (approved):

PR #3149, opened 2022-10-19
Improve Styling On Interestgroup Leaders

jonasdeluna:
Closes https://github.com/webkom/lego/issues/2880 

image

*Bildet er bare broken på min lokale versjon

### DIFF
--- a/app/routes/interestgroups/components/InterestGroupMemberList.css
+++ b/app/routes/interestgroups/components/InterestGroupMemberList.css
@@ -1,3 +1,17 @@
+:root {
+  --color-gold-blur: rgba(255, 172, 47, 0%);
+  --color-silver-blur: rgba(192, 192, 192, 0%);
+  --color-leader-gold: rgb(233, 150, 6);
+  --color-co-leader-silver: #c0c0c0;
+}
+
+[data-theme='dark'] {
+  --color-gold-blur: rgba(248, 203, 22, 80%);
+  --color-silver-blur: rgba(192, 192, 192, 80%);
+  --color-leader-gold: #f8cb16;
+  --color-co-leader-silver: #c0c0c0;
+}
+
 .list {
   width: 100%;
   min-height: 100px;
@@ -31,4 +45,32 @@
 
 .row i {
   margin-right: 5px;
+}
+
+.leader {
+  font-weight: 900;
+  text-shadow: 0.0625em 0.125em 0.875em var(--color-gold-blur);
+  color: var(--color-leader-gold);
+}
+
+.coleader {
+  font-weight: 750;
+  text-shadow: 0.0625em 0.125em 0.875em var(--color-silver-blur);
+  color: var(--color-co-leader-silver);
+}
+
+.leadericon {
+  margin: 0.2em 0.2em 0;
+  filter: drop-shadow(0.0625em 0.125em 0.625em var(--color-gold-blur));
+  color: var(--color-leader-gold);
+}
+
+.coleadericon {
+  margin: 0.2em 0.2em 0;
+  filter: drop-shadow(0.0625em 0.125em 0.625em var(--color-silver-blur));
+  color: var(--color-co-leader-silver);
+}
+
+.suffix {
+  font-style: italic;
 }
--- a/app/routes/interestgroups/components/InterestGroupMemberList.tsx
+++ b/app/routes/interestgroups/components/InterestGroupMemberList.tsx
@@ -1,4 +1,4 @@
-import { Component } from 'react';
+import { Component, ReactNode } from 'react';
 import Modal from 'app/components/Modal';
 import styles from './InterestGroupMemberList.css';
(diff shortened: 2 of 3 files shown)


jonasdeluna:
det ser ok ut på light mode men hvis man først bruker light mode kan man skylde seg selv...

image

LudvigHz (commented):
Looks okay :100: 

Please also attach a before image, as well as an after image in light mode :)

app/routes/interestgroups/components/InterestGroupMemberList.js
```diff
@@ -40,7 +73,10 @@ const ListedUser = ({ user, role }: { user: User, role: string }) => (
     <Flex className={styles.row}>
       <ProfilePicture size={30} user={user} />
       <RoleIcon role={role} />
-      <Link to={`/users/${user.username}`}>{user.fullName}</Link>
+      <Link to={`/users/${user.username}`}>
+        <span className={roleStyle(role)}>{user.fullName} </span>
+        <span className={styles.suffix}>{suffix(role)}</span>
```
ivarnakken, line 78: This leaves a bunch of empty spans in the DOM, since most members don't have a role. You should instead check for a role before creating the element 😄

app/routes/interestgroups/components/InterestGroupMemberList.css
```diff
@@ -32,3 +40,31 @@
...
+
+.coleader {
+  font-weight: 750;
+  text-shadow: 0.0625em 0.125em 0.875em var(--silverblur);
+  color: var(--coleadersilver);
+}
+
+.leadericon {
+  margin: 0.2em 0.2em 0;
+  filter: drop-shadow(0.0625em 0.125em 0.625em var(--goldblur));
```
ivarnakken, line 58: Could we maybe reduce the shadow a tiny bit? I think it's cool on dark mode, but on light mode it feels kinda over the top. I'm all for adding shadow, but I think I'd prefer a smaller one.

app/routes/interestgroups/components/InterestGroupMemberList.css
```diff
@@ -32,3 +40,31 @@
...
+.leader {
+  font-weight: 900;
+  text-shadow: 0.0625em 0.125em 0.875em var(--goldblur);
+  color: var(--leadergold);
+}
+
+.coleader {
+  font-weight: 750;
+  text-shadow: 0.0625em 0.125em 0.875em var(--silverblur);
+  color: var(--coleadersilver);
```
ivarnakken, line 53: Please add some hyphens on these variables - coleadersilver was kinda difficult to read 😅

ivarnakken (changes requested):
Looks a lot better! 🤗 I've made some tiny comments though

Also, please add a "before" photo of the change to the PR, just for comparison reasons.

jonasdeluna:
Before:

image

After:

image

image

app/styles/variables.css
```diff
@@ -126,6 +126,11 @@
   --button-success-border: #a7b6c233;
   --button-success-border-hover: #bfccd644;
   --button-success-box-shadow-focus: #2c974b44;
+
+  --color-gold-blur: rgba(255, 172, 47, 80%);
+  --color-silver-blur: rgba(192, 192, 192, 80%);
+  --color-leader-gold: rgb(233, 150, 6);
+  --color-co-leader-silver: #c0c0c0;
```
ivarnakken, line 133: Now that I think about it, it's better to move these variables to the `InterestGroupMemberList.css` file, since this is the only place where they are used. Variables in `variables.css` should be more "global" variables used throughout the app, to prevent it from becoming too crowded. The button variables above should likewise be moved, but you don't have to do that.

app/routes/interestgroups/components/InterestGroupMemberList.js
```diff
@@ -11,24 +11,65 @@ import Icon from 'app/components/Icon';
...
+    const roleStyle = _roleStyle;
+
+    return (
+      <span>
+        <span className={roleStyle}>{user.fullName} </span>
+        <span className={styles.suffix}>{title}</span>
+      </span>
+    );
+  }
+  return <span>{user.fullName}</span>;
```
ivarnakken, line 44: It's not so good practice to "encapsulate" much good inside if-statements. Better to check for the opposite and return something earlier. See https://www.geeksforgeeks.org/writing-clean-else-statements/ for a better explanation. This isn't so bad though, but it's nice to build good habits early 🙌🏼

ivarnakken (changes requested):
Still looking good! I have some comments on your code changes though

jonasdeluna:
image

image

ivarnakken (approved):
Looks good to me now! 😍 Fantastic job 💯 

By the way, I see that you edit your commit message when you amend, which isn't really optimal. Preferably, either make a new commit or don't edit the commit your're amending to, since editing  sort of makes the git history more confusing. Nevertheless, awesome job!

PR #3154, opened 2022-10-19
Check required photo consents from events

LudvigHz:
LEGO now passes the required photo consent objects for the user in the
event object.

### DIFF
--- a/app/models.js
+++ b/app/models.js
@@ -227,6 +227,7 @@ export type Event = EventBase & {
   registrationCloseTime?: ?Dateish,
   unregistrationCloseTime?: ?Dateish,
   mazemapPoi: number,
+  photoConsents?: Array<PhotoConsent>,
 };
 
 type EventTransformPool = EventPoolBase & {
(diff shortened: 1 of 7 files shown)


eikhr (approved):
Nice!
I feel like maybe `hasRegisteredConsent` should be renamed now that it only checks if the given list of consents are all registered, something like `areConsentsRegistered` or somthing to make it clearer what it actually does.

eikhr:
> > Nice! I feel like maybe `hasRegisteredConsent` should be renamed now that it only checks if the given list of consents are all registered, something like `areConsentsRegistered` or somthing to make it clearer what it actually does.
> 
> hmm, I thought of it like "user has registered their consent", and I can't really find a better name? (it's better than `areConsentsRegistered` tbh)

When i think of a function that checks if a user has registered their consent I expect it to do something related to the user. This function will check any consents given as arguments, and return false if one of them is not answered. It has nothing to do with the user, we just happen to only call it with a single user's relevant consents. I agree `areConsentsRegistered` is not perfect, but what the function is doing is literally checking if all the given consent-objects are registered/answered and I think the name should reflect that.

LudvigHz:
@eikhr I changed it to `allConsentsAnswered` now.

erlingfn (approved):
This looks good!

PR #2670, opened 2022-04-24
Remove podcasts from main menu

odinuge:
It's broken, and has been for ages... :/

PR #3158, opened 2022-10-20
Remove broken podcast routes

eikhr:
The Podcast pages have been broken for ages, and there are no plans to fix them. The link was removed from the menu in #2670 . As i see it there is no reason to keep the broken routes in the code.


LudvigHz:
:cry: yeah I agree with the change mostly. But I do think this was a nice feature. Has anyone checked whether the limitations from SoundCloud still are there, or if there is a way to  fix this without too much hassle?

ivarnakken (approved):
💀

eikhr:
> 😢 yeah I agree with the change mostly. But I do think this was a nice feature. Has anyone checked whether the limitations from SoundCloud still are there, or if there is a way to fix this without too much hassle?

I have not checked, and I am not going to spend time checking. It is not a priority. We can add it as an issue, and look into it later (and just revert this to reuse the old code).

LudvigHz:
> > cry yeah I agree with the change mostly. But I do think this was a nice feature. Has anyone checked whether the limitations from SoundCloud still are there, or if there is a way to fix this without too much hassle?
> 
> I have not checked, and I am not going to spend time checking. It is not a priority. We can add it as an issue, and look into it later (and just revert this to reuse the old code).

Fair enough, although I kind of doubt anyone will bother fixing it if we remove all the code.

For reference, the fix is to register an application with soundcloud, provide the details to the backend, and create and return the token (as per soundcloud API flow) with the podcasts API in LEGO.

PR #3165, opened 2022-10-22
Use react-prepare "usePreparedEffect" instead of "prepared"

eikhr:
Replace all the "prepare" function calls with "withPreparedDispatch". These are both wrappers for react-prepare functions, providing dispatch as a prop, so that we can run redux-actions in the prepare-functions. While "prepare" was a wrapper for the "prepared" function from react-prepare, "withPreparedDispatch" is wrapping the new "withPreparedEffect". 

This changes one main thing. Instead of the prepare-functions beeing run on both the server and the client, they will now only run on client if they were not run on the server, or if the dependency-array changes. As most of our requests get cached, this will not make a significant difference (some exceptions exist, like /events, which will no longer reload after SSR).

In addition it allows for better type-checking of the dependency-array (used for determining whether or not to update), as we now pass the actual prop instead of a string that later gets translated to the prop.


app/routes/admin/email/EmailUsersRoute.js
```diff
@@ -52,15 +52,14 @@ const mapStateToProps = (state) => {
...
 
 export default compose(
-  prepare(
-    (_, dispatch) =>
-      Promise.all([
-        dispatch(fetchAllWithType(GroupTypeCommittee)),
-        dispatch(fetchAllWithType(GroupTypeGrade)),
-      ]),
-    [],
-    { awaitOnSsr: false }
```
LudvigHz, line 62: Is the new function equal in behavior here with the empty dep array and `awaitOnSsr: false`?

app/routes/app/AppRoute.js
```diff
@@ -201,32 +201,27 @@ const mapDispatchToProps = {
...
 export default compose(
-  prepare((_, dispatch) => dispatch(loginAutomaticallyIfPossible()), [], {
-    componentDidMount: false,
-    componentWillReceiveProps: false,
-  }),
-  prepare((_, dispatch) => dispatch(fetchMeta()), [], {
-    componentDidMount: false,
-    componentWillReceiveProps: false,
-    awaitOnSsr: false,
-  }),
```
LudvigHz, line 212: How has this changed now? if at all?
I would assume that this never refetches? Even on re-mount, but it might be hard to test considering the AppRoute never unmounts?

app/routes/app/AppRoute.js
```diff
@@ -201,32 +201,27 @@ const mapDispatchToProps = {
...
-    [],
-    { awaitOnSsr: false }
+  withPreparedDispatch(
+    'login',
+    (_, dispatch) => dispatch(loginAutomaticallyIfPossible()),
+    () => [],
+    { runSync: true }
+  ),
+  withPreparedDispatch('fetchMeta', (_, dispatch) => dispatch(fetchMeta())),
+  withPreparedDispatch('fetchDivData', (props, dispatch) =>
```
LudvigHz, line 211: These can be merged into one as well if the behaviour is as it should without the options on `fetchMeta`

app/utils/withPreparedDispatch.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+export default function withPreparedDispatch(
+  key: string,
+  prepareDispatchFn: (props: Object, dispatch: Dispatch<*>) => ?Promise<*>,
+  depsFn?: (props: Object) => Array<*>,
+  options?: PreparedEffectOptions
+): any {
+  return compose(
+    injectDispatch,
```
LudvigHz, line 32: Why is this here? All the routes should be connected already, so this will just add a bunch of overhead?

LudvigHz (commented):
Looks good overall, but I'll have to run this on my own machine and look around if anything is off. 

Also planning to do some profiling to see if there are any performance/stability improvements or changes with the new prepare :smile:

app/routes/admin/email/EmailUsersRoute.js
eikhr: It should have equal behaiviour, `runSync: false` (which is roughly equivelant to `awaitOnSsr: false`) is now default. Empty dep array is also default in `withPreparedEffect`.

This does mean that most prepared effects have changed behaiviour from `awaitOnSsr: true` (which was previously default) to `false`, but it shouldn't matter as there are no other prepared effects deeper in the component tree.

app/routes/app/AppRoute.js
eikhr: Hmm, ok this one is interesting. I previously assumed it would not change as it was never run on the client before, and shouldn't now, but I didn't think of the case where we do not server-side render.

These prepare-functions previously never ran on client at all. However the same actions are dispatched when initializing the store on the client when _not_ server-side rendered. 

Now it will also dispatch them from the `withPreparedEffect` whenever we do not server-side render, which might not be great. I believe the behaiviour we want is basically the default for the new hooks, only run on client if not ran on the server. I will look into this further.

eikhr: It will also re-run on re-mount, but App should never unmount so that shoudn't matter.

erlingfn (approved):
Looks good! This seems to be merge-ready when you look into the issues Ludvig found

app/routes/app/AppRoute.js
eikhr: It should now work exactly as it did before:)

app/routes/app/AppRoute.js
eikhr: Yup

app/utils/withPreparedDispatch.js
eikhr: No, they are connected after `withPreparedDispatch`. This makes sense because, in theory‚ there could be another prepared effect deeper in the component tree, that uses a prop from `connect` that is fetched in the `withPreparedDispatch`. The `connect` needs to be below `withPreparedDispatch`, so that the data will be awaited before connecting.

However we do need access to `dispatch` in order to execute the fetch action, so we need another `connect` before our `witchPreparedEffect`, only providing us with `dispatch`.

app/routes/interestgroups/InterestGroupDetailRoute.ts
```diff
@@ -1,6 +1,7 @@
+// @flow
```
ivarnakken, line 1: :eyes:

app/routes/pages/PageEditRoute.ts
```diff
@@ -1,6 +1,7 @@
+// @flow
```
ivarnakken, line 1: :eyes:

server/ssr.tsx
```diff
@@ -52,12 +52,14 @@ const createServerSideRenderer = (
       | React.ReactElement<React.ComponentProps<any>, any>
       | null
       | undefined = undefined,
-    state: State | Record<string, never> = Object.freeze({})
+    state: State | Record<string, never> = Object.freeze({}),
+    prepareStateCode?: string
```
ivarnakken, line 56: `prepareStateCode` -> `preparedStateCode`

ivarnakken (approved):
Looks good. Must be tested extensively, I guess in prod.

PR #3178, opened 2022-10-26
Fix cypress bug where it selects wrong date at end of month

eikhr:
At the end of the month, the cypress tests selected the date from the previous month instead of the date tomorrow, if it was visible. I added ":not(:disabled)" to the selector, which works because the previous month's dates are all disabled.
Screenshot 2022-10-26 at 17 46 29

I also moved the date picking code into a function as it was duplicated.


danielyanghansen (approved):
Looks good :shipit:

PR #3180, opened 2022-10-26
Fix mobile page overflow with long words in articles

eikhr:
Before:
Screenshot 2022-10-26 at 19 28 09
After:
Screenshot 2022-10-26 at 19 23 50

It doesn't look great to break words without a hyphen, but that is surprisingly hard with only css.

closes ABA-103

### DIFF
--- a/app/routes/overview/components/ArticleItem.css
+++ b/app/routes/overview/components/ArticleItem.css
@@ -27,7 +27,7 @@
   font-size: 1.2rem;
   letter-spacing: 2px;
   text-transform: uppercase;
-  word-break: normal;
+  word-break: break-word;
 }
 
 .articleMeta {


LudvigHz (approved):
Nice :100: 

Note that you _can_ insert hyphens quite easily with only CSS:
- https://css-tricks.com/almanac/properties/w/word-break/#aa-values
- https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

ivarnakken (approved):

eikhr:
> Nice 💯
> 
> Note that you _can_ insert hyphens quite easily with only CSS:
> 
> * [css-tricks.com/almanac/properties/w/word-break/#aa-values](https://css-tricks.com/almanac/properties/w/word-break/#aa-values)
> * [developer.mozilla.org/en-US/docs/Web/CSS/hyphens](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens)

I did try this, but "hyphens: auto" seems to have no effect on chrome or firefox.

PR #3181, opened 2022-10-26
ABA-110: QR attendance fixes

AdrianAndersen:
* Styling fixes
* Only display the ABA-ID button on your own profile
* Move ABA-ID button down if you have a FRAME
* Eliminate double scans after a successful scan

### DIFF
--- a/app/components/UserValidator/index.js
+++ b/app/components/UserValidator/index.js
@@ -103,7 +103,7 @@ const Validator = (props: Props) => {
         className={styles.scannerButton}
         onClick={() => setShowScanner(true)}
       >
-        <Icon name="qr-code" size={18} />
+        <Icon className={styles.qrIcon} name="qr-code" size={18} />
         Vis scanner
       </Button>
       <SearchPage
--- a/app/components/UserValidator/index.js
+++ b/app/components/UserValidator/index.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { useRef, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { get } from 'lodash';
 import cx from 'classnames';
 import SearchPage from 'app/components/Search/SearchPage';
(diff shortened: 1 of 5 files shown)


app/components/UserValidator/index.js
```diff
@@ -26,15 +26,18 @@ const Validator = (props: Props) => {
...
   const [completed, setCompleted] = useState(false);
   const [showScanner, setShowScanner] = useState(false);
-  const [processing, setProcessing] = useState(false);
+  let processing = false;
 
   const showCompleted = () => {
     setCompleted(true);
-    setTimeout(() => setCompleted(false), 2000);
+    setTimeout(() => {
+      processing = false;
```
ivarnakken, line 34: Why not keep the `useState` and add run the `setProccessing` here? This seems very weird

AdrianAndersen: Because the state does not update quickly enough. When having a state variable, the handleSelect function will be called multiple times, since if (!processing) does not refer to the updated `processing` variable when using state variables.

AdrianAndersen: This causes double and triple scans, where the user receives messages that the user is already present, even though the just scanned them 100ms ago.

AdrianAndersen: This is because useState is asynchronous. ✌🏻

eikhr: I don't see how this could work, won't `processing` always be false in the actual component rendering?

eikhr: You might want to use a [useRef](https://reactjs.org/docs/hooks-reference.html#useref) if you need it to update before rerendering.

AdrianAndersen: This solution actually worked, @eikhr, probably because the component did not re render when processing was set? (Dunno)

However, I solved this now with hooks using a useEffect to detect when the scanned result changes, and then run the handleSelect. ✅

eikhr (approved):
Looks good, we should get this merged before deploying to prod.

app/components/UserValidator/index.js
```diff
@@ -26,45 +26,51 @@ const Validator = (props: Props) => {
...
-      });
-  };
+        )
+        .then(() => {
+          if (input?.current) {
+            input.current.focus();
+          }
+        });
+    },
+    [props]
```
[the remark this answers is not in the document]
AdrianAndersen: It is needed because it is needed as a dependency for the useEffect. This way, it is only created when it is needed in the useEffect.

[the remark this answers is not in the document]
AdrianAndersen: And ESLint wanted med to do it, soooo.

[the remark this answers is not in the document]
AdrianAndersen: For reference:
image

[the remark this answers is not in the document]
LudvigHz: You can have individual object properties in the dependency array, so something like this works:
`[props.clearSearch, props.handleSelect]`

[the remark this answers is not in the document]
AdrianAndersen: I did some refactoring in my latest commit however, such that only the needed functions are passed to the dep array. 😉

LudvigHz (approved):
LGTM

PR #3188, opened 2022-10-27
Fix icon alignment in buttons

eikhr:
Before: 
Screenshot 2022-10-27 at 19 29 55

After:
Screenshot 2022-10-29 at 13 55 11


ivarnakken (approved):
Just fix the linting issue and we're good💯 Thanks!

PR #3189, opened 2022-10-27
Fix Turnstile theme discrepany

ollfkaih:
Couple of minor issues: 
* The light version seems to be loaded before being replaced when running with csr (not an issue with ssr).
* When switching themes, the Turnstile widget remains the old theme (I don't know how to cascade this🤷).

Closes https://github.com/webkom/lego/issues/3036

### DIFF
--- a/app/components/Form/Captcha.js
+++ b/app/components/Form/Captcha.js
@@ -7,6 +7,7 @@ import { createField } from './Field';
 import Turnstile from 'react-turnstile';
 import styles from './Captcha.css';
 import cx from 'classnames';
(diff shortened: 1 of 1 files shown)


LudvigHz (approved):
LGTM :100: 

Yeah, the theme changing is not perfect here, just due to the way we handle the theme stuff. To change the theme dynamically we would ideally store it in redux or a context or something and use that in the component to force a re-render. Since the component won't re-render when we change the theme here.

I am planning on improving the situation, with also storing the theme in localstorage, so unauthed users don't have their theme reset every time they reload the page :upside_down_face:

PR #3195, opened 2022-10-28
fix(UserProfile): FRAME styling

AdrianAndersen:
The FRAME has some whack styling that caused this visual mishap. Did not spot this in the original PR, since profile pictures are not loaded in the staging environment. 

In my LEGO PR [#3041](https://github.com/webkom/lego/pull/3041) I made it so that the default (placeholder) profile picture is always used in staging. This makes it easier to spot these kind of issues.

## Before and After
imageimage

### DIFF
--- a/app/routes/users/components/UserProfile.js
+++ b/app/routes/users/components/UserProfile.js
@@ -392,9 +392,11 @@ const UserProfile = (props: Props) => {
       </Modal>
 
       <Flex wrap className={styles.header}>
-        <div className={cx(styles.sidebar, styles.picture)}>
-          {hasFrame && <Image className={styles.frame} src={frame} />}
-          <ProfilePicture user={user} size={150} />
+        <Flex column alignItems="center" className={styles.sidebar}>
+          <Flex alignItems="center" justifyContent="center">
+            {hasFrame && <Image className={styles.frame} src={frame} />}
+            <ProfilePicture user={user} size={150} />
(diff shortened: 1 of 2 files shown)


app/routes/users/components/UserProfile.js
```diff
@@ -393,8 +393,10 @@ const UserProfile = (props: Props) => {
 
       <Flex wrap className={styles.header}>
         <div className={cx(styles.sidebar, styles.picture)}>
-          {hasFrame && <Image className={styles.frame} src={frame} />}
-          <ProfilePicture user={user} size={150} />
+          <div className={styles.picture}>
```
ivarnakken, line 396: I assume you use the `picture` class for its flex-box styling, but I think this should be switched to a `<Flex />` component instead. The class name does not convey what it does either.

AdrianAndersen: Yeah, sure. I switched to Flex now ✅

eikhr (approved):
Looks good!

app/routes/users/components/UserProfile.js
ivarnakken: I was thinking you could do it on the other div employing the same styling as well, and remove the `picture` class entirely.

AdrianAndersen: Sure. Removed the class now 🚀

PR #3196, opened 2022-10-30
Set end time explicitly in cypress meeting test

eikhr:
The test fails on the two days a year we set the time because of daylight savings, because the default values of the form will be an hour off. Setting the time explicitly in the test fixes it, and I don't really see a need to test the default values.

ABA-115


ivarnakken (approved):
🙃

PR #3208, opened 2022-11-01
Fix clicking links on /search page

eikhr:
[This](https://github.com/webkom/lego-webapp/pull/3017/commits/0053f16b4360d5cc00b6463ba5af5e34a6dc42da) commit from https://github.com/webkom/lego-webapp/pull/3017, accidentally broke all `handleSelect`-calls when searching for other things than users, making links unclickable in the search-page.

If only we had working type-checking...


ivarnakken (approved):
:shipit:

erlingfn (approved):

PR #3210, opened 2022-11-01
ABA-127: Attendee Statistics in Event Administrate

AdrianAndersen:
In this PR, I have implemented visualisations of attendees for each event in. /administrate/. The following statistics are displayed:
- Gender distribution (male, female, other)
- Datatek class distribution
- Komtek class distribution
- Datatek vs Komtek (+ PhD and graduated people) distribution
- Group affiliation (Abakom, Revy or Abakus)
- Registration/Unregistrations per day [This might be useful for seeing how effective "blesting" for a given event is]

Hopefully this functionality can help event organisers be more data driven when planning and executing events. 

## Screenshot of the graphs from from this years Charity Party (Veldedighetsfest)
imageimage

### DIFF
--- a/app/components/Chart/utils.js
+++ b/app/components/Chart/utils.js
@@ -1,0 +1,60 @@
+// @flow
+
+export interface DistributionDataPoint {
+  name: string;
+  count: number;
+}
+
+export const CHART_COLORS = [
+  'var(--lego-red)',
+  'var(--color-blue-4)',
+  'var(--color-orange-3)',
+  'var(--color-green-5)',
+  'var(--lego-chart-purple)',
+  'var(--lego-chart-yellow)',
+  'var(--lego-chart-green)',
+  '#ff87eb',
+  'var(--color-black)',
+];
+
+type GraphProps = {
+  cx: number,
+  cy: number,
+  midAngle: number,
+  innerRadius: number,
+  outerRadius: number,
+  percent: number,
+  index: number,
+};
+
+const RADIAN = Math.PI / 180;
+
+export const renderCustomizedLabel = ({
+  cx,
+  cy,
+  midAngle,
+  innerRadius,
+  outerRadius,
+  percent,
+}: GraphProps) => {
+  const radius = innerRadius + (outerRadius - innerRadius) * 0.5;
+  const x = cx + radius * Math.cos(-midAngle * RADIAN);
+  const y = cy + radius * Math.sin(-midAngle * RADIAN);
+
+  // Hide the percent if it is too small to fit the chart
+  if (percent < 0.04) {
+    return;
+  }
+
+  return (
+    <text
+      x={x}
+      y={y}
+      fill="white"
+      textAnchor="middle"
+      dominantBaseline="central"
+    >
+      {`${(percent * 100).toFixed(0)}%`}
+    </text>
+  );
+};
--- a/app/routes/events/EventStatisticsRoute.ts
+++ b/app/routes/events/EventStatisticsRoute.ts
@@ -1,0 +1,43 @@
+import { compose } from 'redux';
+import { connect } from 'react-redux';
+import { getRegistrationGroups } from 'app/reducers/events';
+import Statistics from 'app/routes/events/components/EventAdministrate/Statistics';
+import { fetchAllWithType } from 'app/actions/GroupActions';
+import prepare from 'app/utils/prepare';
+import { selectGroupsWithType } from 'app/reducers/groups';
+import { GroupTypeCommittee, GroupTypeRevue } from 'app/models';
+
+const groups: {
+  committee: typeof GroupTypeCommittee;
+  revue: typeof GroupTypeRevue;
+} = {
+  committee: 'komite',
+  revue: 'revy',
+};
+
+const mapStateToProps = (state, props) => {
+  const { registered, unregistered } = getRegistrationGroups(state, {
+    eventId: props.eventId,
+  });
+
+  return {
+    committees: selectGroupsWithType(state, {
+      groupType: groups.committee,
+    }),
+    revueGroups: selectGroupsWithType(state, {
+      groupType: groups.revue,
+    }),
+    registered,
+    unregistered,
+  };
+};
+
+export default compose(
+  prepare((_, dispatch) =>
+    Promise.all([
+      dispatch(fetchAllWithType(groups.committee)),
+      dispatch(fetchAllWithType(groups.revue)),
+    ])
+  ),
+  connect(mapStateToProps, {})
+)(Statistics);
--- a/app/routes/events/EventStatisticsRoute.ts
+++ b/app/routes/events/EventStatisticsRoute.ts
@@ -3,9 +3,9 @@ import { connect } from 'react-redux';
 import { getRegistrationGroups } from 'app/reducers/events';
 import Statistics from 'app/routes/events/components/EventAdministrate/Statistics';
 import { fetchAllWithType } from 'app/actions/GroupActions';
-import prepare from 'app/utils/prepare';
 import { selectGroupsWithType } from 'app/reducers/groups';
 import { GroupTypeCommittee, GroupTypeRevue } from 'app/models';
+import withPreparedDispatch from 'app/utils/withPreparedDispatch';
 
 const groups: {
   committee: typeof GroupTypeCommittee;
@@ -33,7 +33,7 @@ const mapStateToProps = (state, props) => {
 };
 
 export default compose(
-  prepare((_, dispatch) =>
+  withPreparedDispatch('fetchStatisticsGroups', (_, dispatch) =>
     Promise.all([
       dispatch(fetchAllWithType(groups.committee)),
       dispatch(fetchAllWithType(groups.revue)),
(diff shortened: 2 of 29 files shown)


eikhr:
I think it's a great idea! I can imagine that f.ex. the distribution of what year the attendees are going could be very useful for somone arranging a course. I think we have a lot of very useful data that is currently too diffucult to access, so I'm all for making this more easily available.

It could also be interesting to get the share of attendees who are comitee (and maybe revue) members, although this might be a bit more difficult to fetch data for.

ollfkaih:
Good idea, looks good overall. Couple of nitpicks:  
Localize the labels for gender.
Also, maybe we should hide numbers below a certain percentage or show the outside the pie, if that is possible?

app/routes/events/components/EventAttendeeStatistics.js
```diff
@@ -0,0 +1,123 @@
...
+                &nbsp;
+              </span>
+              <span> {dataPoint.name}</span>
+            </li>
+          ))}
+        </ul>
+      </div>
+    </Flex>
+  );
+};
```
eikhr, line 56: I love that the pie chart is a separate component and abstract enough to be used for other things as well! 

I feel like it should be moved into a separate file in /components, and possibly also be reused in the survey result page, as that chart seems to be very similar.

eikhr: Actually, while you're at it you should probably move the bar chart from survey results into a separate component as well. No reason to have all this chart-creation code inside the survey route folder if we also use charts outside of surveys.

AdrianAndersen: Yeah, I agree! ✅

AdrianAndersen:
Thank you so much for the initial discussion and feedback for the usability of this feature! 😄 

I have now implemented your suggestions and refactored out some of the code from the surveys to make it more modular. Hope you have time to take a look at the changes. (Even though it turned out the be quite a few lines of code 👨🏻‍🎨 )

app/routes/events/components/EventAttendeeStatistics.js
```diff
@@ -0,0 +1,283 @@
...
+import moment from 'moment';
+
+interface RegistrationDateDataPoint {
+  name: string;
+  registrations: number;
+  unregistrations: number;
+}
+
+const ABAKOM_GROUP_IDS = [4, 5, 6, 7, 8, 9, 10, 11, 98, 231];
+const REVY_GROUP_IDS = [59, 119, 120, 121, 123, 124, 153, 154, 214, 222, 225];
```
AdrianAndersen, line 26: It is kinda ugly to hard code the group IDs here, since it can get outdated if we for instance add another Abakom/Revy group. if someone has any suggestions for how to improve this, please let me know. 👀

eikhr: Yeah, we do have `GroupTypeCommitee` and `GroupTypeRevue`, but group type isn't fetched for users in registrations so you'd have to make some backend changes, or do some extra API-calls.

LudvigHz: Yeah, I think the most sensible here if you want it, is to fetch `/api/v1/groups/?type=<komite/revy>` and use the data from that to find what group type people are in.

LudvigHz:
As far as I understood, this is currently on the `/administrate/attendees` route?

If so, I think we need to move this to another route, f.ex. `/administrate/statistics`. This is so we don't make the page _enourmous_ and can split out the data fetching.

AdrianAndersen:
> As far as I understood, this is currently on the /administrate/attendees route?

Yup, currently it is. It makes sense to move it to a separate page, I fix.

app/routes/events/components/EventAttendeeStatistics.js
```diff
@@ -0,0 +1,283 @@
...
+    totalDistribution: [],
+    registrationTimeDistribution: [],
+  };
+
+  const dataTekTotal: DistributionDataPoint = {
+    name: 'Datateknologi',
+    count: 0,
+  };
+  const komTekTotal: DistributionDataPoint = {
+    name: 'Komunikasjonsteknologi',
```
ingraso, line 138: As the name of the study has changed I suggest also updating it here, in addition to adding an "m".
```suggestion
    name: 'Kommunikasjonsteknologi og digital sikkerhet',
```

app/routes/events/components/EventAttendeeStatistics.js
erlingfn: I agree with Ludvig. If we add more committees I bet we will not remember to add them here

app/routes/events/components/EventAttendeeStatistics.js
```diff
@@ -0,0 +1,283 @@
...
+  );
+};
+
+const toLocalizedGender = (gender: string) => {
+  if (gender === 'male') return 'Mann';
+  if (gender === 'female') return 'Kvinne';
+  return 'Annet';
+};
+
+const createGenericDataPoint = (
```
erlingfn, line 55: Perhaps a better name for this might be either `addGenericDataPoint` or `addOrIncrementDataPoint`? I was a bit confused by the naming :grimacing:

app/routes/events/components/EventAttendeeStatistics.js
```diff
@@ -0,0 +1,283 @@
...
+    (entry) => entry.name === selectedDataPoint
+  );
+  if (dataPointEntry) {
+    dataPointEntry.count++;
+  } else {
+    selectedDistribution.push({ name: selectedDataPoint, count: 1 });
+  }
+};
+
+const createGroupDataPoint = (
```
erlingfn, line 69: Same as above applies to this.

app/components/Chart/utils.js
```diff
@@ -0,0 +1,60 @@
+// @flow
+
+export interface DistributionDataPoint {
+  name: string;
+  count: number;
+}
+
+export const CHART_COLORS = [
```
erlingfn, line 8: Would it be possible to include more chart colours? I have sometimes experienced charts being difficult to read because it ran out of colours.

erlingfn (approved):
Great work! A few nitpicks, but they may just be me.

app/routes/events/components/EventAttendeeStatistics.js
AdrianAndersen: Yeah, that sounds like a good solution. Moved it now ✅

app/routes/events/components/EventAttendeeStatistics.js
AdrianAndersen: ✅

app/routes/events/components/EventAttendeeStatistics.js
AdrianAndersen: ✅

app/routes/events/components/EventAttendeeStatistics.js
AdrianAndersen: ✅

app/components/Chart/utils.js
AdrianAndersen: Yup, I added some more colors now, so that we now have 12 colors. That should suffice for all charts. ✅

AdrianAndersen:
Also, kinda weird that `tsc` and some seemingly unrelated Cypress tests fails in CI. 👀

app/routes/events/EventStatisticsRoute.ts
```diff
@@ -0,0 +1,43 @@
...
+import { selectGroupsWithType } from 'app/reducers/groups';
+import { GroupTypeCommittee, GroupTypeRevue } from 'app/models';
+
+const groups: {
+  committee: typeof GroupTypeCommittee;
+  revue: typeof GroupTypeRevue;
+} = {
+  committee: 'komite',
+  revue: 'revy',
+};
```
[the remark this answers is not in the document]
erlingfn: Are they in the redux state at all? I do not know if they are in the serializer.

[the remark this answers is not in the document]
LudvigHz: Hmm, I tried using them directly and it worked just fine :thinking: 

Ideally though, if you would be a champ @AdrianAndersen, you can just yeet all the `GroupType****` stuff from `app/models` and replace it with an enum, which can be used both as a type and value everywhere :)

So something like this in app/models:
```ts
export enum GroupType {
  committee = 'komite',
  Board = 'styre',
  Revue = 'revy',
  Interest = 'interesse',
  Grade = 'klasse',
  SubGroup = 'under',
  Other = 'annen',
}
```
And use that in the route.

It's not a HUGE deal though, I can add a task for converting every string type in app/models to enums, but it's nice to just do it while you're working with it :)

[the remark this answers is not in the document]
AdrianAndersen: Huh, interesting. It actually worked to access the variables directly today. I am guessing it was just a hot reloading issue I experienced earlier. 🤔 

I refactored all GroupType** usages to an enum now. ✅

eikhr:
I just noticed that looking at stats for old events might be a bit misleading. It will show the status of each attendee now, not when the event took place. This means that an event that happened two years ago will show that no 1st or 2nd graders attended. It also makes it look like more non-committee/revue members attended than you would expect, as all the people who are now done studying are no loger counted as committee/revue-members.

I can't think of any easy way to fix this, but perhaps a warning on the statistics page for old events would be a good idea.

AdrianAndersen:
> I can't think of any easy way to fix this, but perhaps a warning on the statistics page for old events would be a good idea.

@eikhr, I guess one could fetch a users previous memberships, but I think that is too cumbersome. I therefore added a warning on events from a previous semester as you suggested:

image

LudvigHz (approved):
LGTM :shipit:

eikhr (approved):

PR #3211, opened 2022-11-01
Use proxied url for plausible script.

LudvigHz:
URLs with `/plausible.js` are blocked by most adblockers, and the
`analytics` subdomain is also susceptible to blocking. Changing to a URL
that avoids this should make sure we get more accurate analytics.

Fixes ABA-126

### DIFF
--- a/server/pageRenderer.js
+++ b/server/pageRenderer.js
@@ -70,11 +70,11 @@ const readyHtml = (app) => {
 
 const analytics = __DEV__
   ? ''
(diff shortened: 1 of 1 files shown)


ingraso (approved):
LGTM. I like the change from `plausible.js` to `script.js`. As discussed during today's meeting, if we add some kind of documentation/explanation about what https://ls.webkom.dev is and can be used for, I think the name is okay. However, maybe someone has any other suggestions instead of `ls`?

ivarnakken (approved):

PR #3213, opened 2022-11-02
Change group leader name to editor for readme

ingraso:
As the leader of readme is called editor ("Redaktør") and not leader, this is now updated on the flatpages.

image

Fixes ABA-132


ivarnakken (approved):
Nice! I think this is applicable for a lot of our groups (both committee and revue).

PR #3214, opened 2022-11-02
Improve Checkbox component

ivarnakken:
The `<Checkbox />` component is in great need of an improvement. It now also looks much better in dark mode.

It now works to pass a `label` into the component as a prop, making it align to the checkbox. It is now also clickable, but only if an `id` is given.  

The CSS is completely rewritten, so no need to compare the diffs.

## Result
Keep in mind that the .gif is laggy, causing the transitions to seem slow and less smooth. 

checkbox

Closes ABA-133

### DIFF
--- a/app/components/Form/CheckBox.css
+++ b/app/components/Form/CheckBox.css
@@ -1,108 +1,106 @@
 @import '~app/styles/variables.css';
 
+.checkbox {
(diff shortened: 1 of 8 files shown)


ingraso:
Great work! I really like the visual changes, and the animations :rocket:

PR #3215, opened 2022-11-02
:red_circle:  Typescript (V2.0) :red_circle: 

LudvigHz:
We're trying again!! :partying_face: 

For this time, the plan is to just merge this in with shit types, and incrementally fix the types. The reasoning is that we are more likely to fix bad TS types than bad Flow types. TS is also more strict which will help here. As of now, eslint has about ~1300 warnings, most of which are the `unexpected any`. TSC however is a bit worse and complains about a _bunch_ of stuff, some of which is due to that `flow-to-ts` didn't know how to handle it. These are quite easy to fix incrementally.

So everything works as intended, and there are no errors from `eslint`, but a thousand warnings and tsc will take some time to make perfect.

_It's a start :slightly_smiling_face:_


erlingfn (approved):
This does not make LEGO any worse

.babelrc
```diff
@@ -1,5 +1,5 @@
 {
-  "presets": ["@babel/preset-flow", ["@babel/preset-env", {
+  "presets": ["@babel/preset-typescript", "@babel/preset-flow", ["@babel/preset-env", {
```
norbye, line 2: Is there a reason that this still includes `@babel/preset-flow`?

config/webpack.server.js
```diff
@@ -70,7 +70,7 @@ module.exports = (env, argv) => {
     module: {
       rules: [
         {
-          test: /\.js$/,
+          test: /\.[tj]sx?$/,
```
norbye, line 73: Is there actually an extension we use that matches this regex? Wouldn't they be ts or tsx here as well?

norbye (approved):
Awesome! tried to find the few files that were not just renamed and updated and review those, and found a few minor things I didn't quite get but other than that it looks nice

tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+  "compilerOptions": {
+    "target": "esnext",
+    "jsx": "react-jsx",
+    "module": "commonjs",
+    "moduleResolution": "node",
+     "baseUrl": "./",
+     "paths": {
+       "app/*": ["app/*"],
+     },
+    "include": ["app/**/*"],
```
eikhr, line 11: Shouldn't we include `server` as well here?

config/webpack.server.js
eikhr: This regex matches .ts, .tsx, .js, and .jsx. But we don't have any js or jsx files, so I suppose it could be simplified

eikhr: This regex matches .ts, .tsx, .js and .jsx, so it should work fine. All the tests are .ts or .tsx though afaik, so I suppose it could be simplified

Arashfa0301 (commented):
A little difficult to review 🥸

.babelrc
LudvigHz: ah, I had it there while testing flow + ts, but it was decided against, I'll remove all the flow stuff :)

eikhr (approved):
:shipit: There are an insane amount of typescript-errors, especially in Prop-types, but the config looks good and the errors will have to be fixed over time.

PR #3225, opened 2022-11-04
Support interest group links on events

ivarnakken:
Now that interest groups are more regularly creating events, I think it's fitting to link to their page.

## Result

**Before**
image

**After**
image

---

To help in review, this is what `resolveGroupLink` does:
```tsx
export const resolveGroupLink = (group: { type: string; id: ID }) => {
  switch (group.type) {
    case GroupTypeInterest:
      return `/interest-groups/${group.id}`;

    case GroupTypeCommittee:
      return `/pages/komiteer/${group.id}`;

    default:
      return null;
  }
};
```

---

Closes ABA-136

### DIFF
--- a/app/routes/events/components/EventDetail/index.tsx
+++ b/app/routes/events/components/EventDetail/index.tsx
@@ -45,6 +45,7 @@ import type {
 import type { CommentEntity } from 'app/reducers/comments';
 import type { UserEntity } from 'app/reducers/users';
 import { MazemapEmbed } from 'app/components/MazemapEmbed';
+import { resolveGroupLink } from 'app/reducers/groups';
 type InterestedButtonProps = {
   isInterested: boolean;
 };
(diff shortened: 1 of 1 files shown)


itsisak (approved):
Looks good! I agree it makes sense to link to the interest groups aswell

ollfkaih (approved):
LGTM💯

LudvigHz (approved):
:100:

haukkagu (approved):
LGTM

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -297,15 +298,15 @@ export default class EventDetail extends Component<Props, State> {
           }
         : null,
     ];
+
+    const groupLink = resolveGroupLink(event.responsibleGroup);
```
eikhr, line 302:
```suggestion
    const groupLink = event.responsibleGroup && resolveGroupLink(event.responsibleGroup);
```
Seems like `event.responsibleGroup` is sometimes undefined while loading. This causes a crash, and is why the cypress tests fail.

eikhr (changes requested):

app/routes/events/components/EventDetail/index.tsx
ivarnakken: Ye, gotta love our SSR 🙃

eikhr: It's not really ssr-related, I think it's just caused by the way we load data with redux. I could reproduce the error without ssr enabled by just opening some different events.

ivarnakken: Ye I kinda realized after commenting, but I like to blame our SSR for all of my problems :)

PR #3228, opened 2022-11-05
Fix registration selector when user is missing

odinuge:
When a user does not have perms to register on an event, websocket registration messages in the `Event.SOCKET_REGISTRATION.SUCCESS` action will use the `RegistrationAnonymizedReadSerializer` containing the fields `("id", "pool", "status")`. This is done to make those user able to watch registrations real time, without seeing who the users are.

For those events, we don't hit this code path, since it will only use counters. However, if the user has normal access to an event and can see the registrations, the frontend will currently crash if one such websocket event (for another event without access) has occurred.

We could in theory handle the case here as well, but it will currently not be hit unless other code is refactored; and in that case even more code has to support it.

Fixes https://github.com/webkom/lego/pull/1929

### DIFF
--- a/app/reducers/events.ts
+++ b/app/reducers/events.ts
@@ -309,8 +309,9 @@ export const selectAllRegistrationsForEvent = createSelector(
   (state, props) => props.eventId,
   (registrationsById, registrationItems, usersById, eventId) =>
     registrationItems
-      .map((regId, i) => {
-        const registration = registrationsById[regId];
+      .map((regId) => registrationsById[regId])
+      .filter((registration) => registration.event === Number(eventId))
+      .map((registration) => {
         const user = registration.user.id
           ? registration.user
           : usersById[registration.user];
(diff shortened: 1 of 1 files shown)


app/reducers/events.ts
```diff
@@ -309,8 +309,9 @@ export const selectAllRegistrationsForEvent = createSelector(
   (state, props) => props.eventId,
   (registrationsById, registrationItems, usersById, eventId) =>
     registrationItems
-      .map((regId, i) => {
-        const registration = registrationsById[regId];
+      .map((regId) => registrationsById[regId])
```
ivarnakken, line 312: I might be dumb, but is this map really needed now that you map on registration after the filter? Is it not the equivalent to filter directly on `registrationItems`?

odinuge: Yeah, you are right, not fully necessary. The reason I added it was to avoid doing `registrationsById[regId]` twice, since that is more verbose and maybe harder to understand. Happy to switch to using `registrationsById[regId]` if you prefer that

PR #3152, opened 2022-10-19
Add the Abakus Revue to pages

erlingfn:
To better showcase the Abakus Revue it will now be displayed in its own tab, in the same way as the committees.

Depends on: https://github.com/webkom/lego/pull/3010

PR #3229, opened 2022-11-06
Fix Link For Revue Badges

haukkagu:
Thanks to #3152, every group in the revue has its own page on abakus.no. That means that we now can link the revue badges on your profile to their respective page, like what we've done for committees and interest groups. This PR adds that feature.

For example, I have a Teknikk-badge on my profile. Previously, nothing would happen if you clicked on it, but now it will route you to the Teknikk-page under Revy.


LudvigHz (approved):
LGTM :100:

PS: Remember to sign your commits :)

ivarnakken (approved):
Nice 💯

PR #3237, opened 2022-11-07
Fix Alarm Icon

haukkagu:
Currently, the alarm icon under Påmeldinger is a custom svg in the assets folder and not an [ion-icon](https://ionic.io/ionicons) like every other icon on the webapp. Not only is the icon very ugly, it also doesn't change color by theme, which makes it difficult to spot on dark mode. This PR changes the alarm icon from an svg to an ion-icon (in addition to some small styling changes).

Before:
image
image
After:
image
image


haukkagu:
Any feedback is appreciated of course. Do you guys think that the icon is too bright on dark mode?

app/routes/overview/components/NextEvent.css
```diff
@@ -12,6 +12,13 @@
   margin: 0;
 }
 
+.info {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  margin: -3px 10px;
```
LudvigHz, line 19: Try to avoid negative margins. I'm assuming this is for the while event text thing? Of so try to rather decrease the spacing thats created by some other css :)

app/routes/overview/components/NextEvent.css
```diff
@@ -12,6 +12,13 @@
...
+}
+
 .info span {
   margin: 0 10px;
 }
```
LudvigHz, line 24: Remove this as afaics it's not used anymore

LudvigHz (approved):
Nice :100:
I _kinda_ think the old icon looks slightly better but I agree with this change overall :+1:

haukkagu:
What about this?
image
image

app/routes/overview/components/NextEvent.css
haukkagu: It's used by the text to the right of the icon ("Påmelding åpnet...")

app/routes/overview/components/NextEvent.css
haukkagu: Not sure how to go about this. The negative margin makes the title of the event and the icons n stuff closer together, which looks nicer imo. The margin of the title and the padding of the eventItem is not set, so I can't see how I could drecrease the spacing elsewhere.

app/routes/overview/components/NextEvent.css
```diff
@@ -12,6 +12,13 @@
   margin: 0;
 }
 
+.info {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
```
ivarnakken, line 18: Preferable use the `<Flex />` component for this, especially if you end up getting rid of the margin.

app/routes/overview/components/NextEvent.tsx
```diff
@@ -64,9 +63,7 @@ class EventItem extends Component<Props, State> {
           </Link>
 
           <div className={styles.info}>
-            <span>
-              <img alt="alarm-icon" className={styles.alarm} src={alarm} />
-            </span>
+            <Icon name="alarm-outline" className={styles.alarmIcon} size={20} />
```
ivarnakken, line 66: The `alarmIcon` class seems to be unused

ivarnakken (changes requested):
I like that the icon shares the same color as the text 😄 Good job!

Just some tiny cleanup left, and we're good

app/routes/overview/components/NextEvent.css
haukkagu: Fixed

ivarnakken (approved):
Much cleaner!

PR #3251, opened 2022-11-09
Enable ESLint import order rule

LudvigHz:
This replaces #2838

Instead of using `eslint-simple-import` this uses `eslint-plugin-import` which is much more maintained and easier to use.
Fixes ABA-158


app/actions/CompanyActions.ts
```diff
@@ -8,11 +8,12 @@ import {
   joblistingsSchema,
 } from 'app/reducers';
 import createQueryString from 'app/utils/createQueryString';
-import { startSubmit, stopSubmit } from 'redux-form';
-import { push } from 'connected-react-router';
 import type { Thunk } from 'app/types';
 import { addToast } from 'app/actions/ToastActions';
```
ivarnakken, line 12: This does not sort as well though. You'd preferably want `import { addToast } from 'app/actions/ToastActions';` to come next to `import callAPI from 'app/actions/callAPI';`

LudvigHz: Hmm, true. I can try enabeling alphabetical sorting

LudvigHz: @ivarnakken What do you think now. Now all imports in each group are sorted alphabetically

app/actions/CompanyInterestActions.ts
```diff
@@ -1,9 +1,10 @@
-import { CompanyInterestForm } from './ActionTypes';
+import type { Thunk } from 'app/types';
+import type { CompanyInterestEntity } from 'app/reducers/companyInterest';
+import { companyInterestSchema } from 'app/reducers';
 import callAPI from 'app/actions/callAPI';
 import { addToast } from 'app/actions/ToastActions';
-import { companyInterestSchema } from 'app/reducers';
-import type { CompanyInterestEntity } from 'app/reducers/companyInterest';
-import type { Thunk } from 'app/types';
+import { CompanyInterestForm } from './ActionTypes';
```
ivarnakken, line 6: Do we not want type imports to come last?

LudvigHz: So, right now, type imports are between external (npm) modules and internal imports (app/...).
I thought that would be nice since then the type imports are sort of inbetween internal and external things, considering we import types from both groups.

ivarnakken: hmm, okay, I see your point. I'm just used to putting them at the bottom. I'll let you decide then 👍🏼 Perhaps somebody else got an input on this?

LudvigHz: People at the office agreed with types at the bottom, so I'll do that :+1:

PR #3252, opened 2022-11-09
Fix plenty of TS errors & icon badge z-index

LudvigHz:
The notifications badge was behind the notification bell icon, most
likely as a result of the recent ion-icon upgrade. This fixes that so
that it's in front. Also reduces the padding a bit in the badge.

### DIFF
--- a/app/components/Header/index.tsx
+++ b/app/components/Header/index.tsx
@@ -1,5 +1,5 @@
 import cx from 'classnames';
-import { Component } from 'react';
+import { Component, MouseEventHandler } from 'react';
 import { Modal } from 'react-overlays';
 import { Link, NavLink, withRouter } from 'react-router-dom';
 import logoLightMode from 'app/assets/logo-dark.png';
@@ -40,11 +40,19 @@ type Props = {
   loading: boolean;
   updateUserTheme: (username: string, theme: string) => Promise<void>;
 };
+
+enum LoginMode {
+  LOGIN = 'login',
+  REGISTER = 'register',
+  FORGOT_PASSWORD = 'forgotPassword',
+}
+
 type State = {
   accountOpen: boolean;
   shake: boolean;
-  mode: 'login' | 'register' | 'forgotPassword';
+  mode: LoginMode;
 };
+
 type AccountDropdownItemsProps = {
   logout: () => void;
   onClose: () => void;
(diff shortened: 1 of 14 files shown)


app/components/Header/index.tsx
```diff
@@ -363,11 +375,9 @@ class Header extends Component<Props, State> {
           >
             <Search
               loggedIn={loggedIn}
-              isOpen={this.props.searchOpen}
               onCloseSearch={this.props.toggleSearch}
               updateUserTheme={this.props.updateUserTheme}
               username={this.props.currentUser?.username}
-              className={utilStyles.hiddenOnMobile}
```
ivarnakken, line 370: Are you sure about this?

app/components/Icon/index.tsx
```diff
@@ -33,19 +34,21 @@ const Icon = ({
         fontSize: `${size.toString()}px`,
         ...style,
       }}
-      {...(props as Record<string, any>)}
+      {...props}
     >
+      {/* eslint-disable-next-line*/}
+      {/* @ts-ignore*/}
```
ivarnakken, line 40: You'd think it'd be enough to just ts-ignore it, but oh well 🤷🏼

app/globals.d.ts
```diff
@@ -3,5 +3,10 @@ declare module '*.css' {
   export default classes;
 }
 
+declare module '*.png' {
```
ivarnakken, line 6: Could you not add the other image file formats that we use as well?

ivarnakken (approved):
💯

app/components/Header/index.tsx
ivarnakken: But you tested it, right?

LudvigHz: I just checked with prod, and there's no difference. I have no idea why this was here either, since you would want to be able to search on mobile :shrug:

app/globals.d.ts
LudvigHz: Done :handshake:

app/globals.d.ts
```diff
@@ -3,5 +3,20 @@ declare module '*.css' {
...
+
+declare module '*.svg' {
+  const value: string;
+  export default value;
+}
+
+declare module '*.jpg' {
+  const value: string;
+  export default value;
+}
```
ivarnakken, line 19: Can you not do a regex? :)

LudvigHz: I tried. It didn't seem to work...

ivarnakken: :(

haukkagu (approved):

PR #3262, opened 2022-11-12
Fix a bunch of TS types

LudvigHz:
- Fix TS types in server/
- Fix types in comments, images, user components

### DIFF
--- a/app/components/Image/ProfilePicture.tsx
+++ b/app/components/Image/ProfilePicture.tsx
@@ -1,16 +1,18 @@
 import type { UserEntity } from 'app/reducers/users';
 import CircularPicture from './CircularPicture';
+import type { ComponentProps } from 'react';
 
-type Props = {
-  user: UserEntity;
-  alt: string;
-  size: number;
-  style?: Record<string, string>;
-};
+type Props = Omit<
+  {
+    user: UserEntity;
+    size: number;
+  } & ComponentProps<typeof CircularPicture>,
+  'placeholder' | 'src' | 'alt'
+>;
 
-const ProfilePicture = ({ alt, user, size = 100, style, ...props }: Props) => (
+const ProfilePicture = ({ user, size = 100, style, ...props }: Props) => (
   <CircularPicture
-    alt={alt}
+    alt={`${user.username} profile picture`}
     src={user.profilePicture}
     placeholder={user.profilePicturePlaceholder}
     size={size}
--- a/app/components/Layout/Flex.tsx
+++ b/app/components/Layout/Flex.tsx
@@ -1,6 +1,6 @@
 import cx from 'classnames';
 import styles from './Flex.css';
-import type { ReactNode, ElementType } from 'react';
+import type { ReactNode, ElementType, HTMLAttributes } from 'react';
 
 type Props = {
   children?: ReactNode;
@@ -29,8 +29,7 @@ type Props = {
   margin?: number | string;
   width?: number | string;
   gap?: number | string;
-  style?: Record<string, string>;
-};
+} & HTMLAttributes<HTMLDivElement>;
 
 /**
  * Simple FlexBox component
--- a/app/models.ts
+++ b/app/models.ts
@@ -96,7 +96,7 @@ export type UserMembership = {
 };
 export type Group = {
   id: ID;
-  actionGrant: Array<string>;
+  actionGrant: ActionGrant;
   type: GroupType;
   name: string;
   numberOfUsers?: number;
@@ -143,8 +143,18 @@ type EventBase = {
   useContactTracing: boolean;
   legacyRegistrationCount: number;
 };
+
 export type Company = Record<string, any>;
-export type Comment = Record<string, any>;
+
+export type Comment = {
+  id: ID;
+  text: string;
+  author: User;
+  createdAt: Dateish;
+  updatedAt: Dateish;
+  parent?: ID;
+};
+
 export type Permission = string;
 export type EventRegistrationPresence = 'PRESENT' | 'NOT_PRESENT' | 'UNKNOWN';
 export type LEGACY_EventRegistrationPhotoConsent =
@@ -194,7 +204,7 @@ export type EventPool = EventPoolBase & {
   permissionGroups: Array<Record<string, any>>;
 };
 export type Event = EventBase & {
-  actionGrant: Array<string>;
+  actionGrant: ActionGrant;
   activationTime: Dateish | null | undefined;
   isAdmitted: boolean | null | undefined;
   activeCapacity: number;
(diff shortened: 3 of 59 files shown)


app/routes/events/utils.ts
```diff
@@ -272,21 +274,24 @@ export const eventStatusTypes = [
...
   return (
     find(eventStatusTypes, {
       value: eventStatusType,
     }) || eventStatusTypes[0]
   );
 };
+
 export const getEventSemesterFromStartTime = (
-  startTime: moment
+  startTime: Moment
```
eikhr, line 287:
```suggestion
  startTime: Dateish
```
It is used with `event.startTime` which is `Dateish`

app/routes/events/utils.ts
```diff
@@ -10,10 +10,6 @@ import type {
   EventSemester,
 } from 'app/models';
 
-export const PHOTO_CONSENT_DOMAINS = {
-  WEBSITE: 'WEBSITE',
-  SOCIAL_MEDIA: 'SOCIAL_MEDIA',
-};
```
eikhr, line 16: This is still imported and (attempting to be) used in `RegistrationMeta.tsx` and `RegistrationTables.tsx`, update them to the new enum.

eikhr (changes requested):
Great work! 

I've just commented some nitpicking, and also you forgot to update to your new enum a couple of places causing a crash.

app/reducers/users.ts
```diff
@@ -18,7 +18,7 @@ export type UserEntity = {
   profilePicture: string;
   profilePicturePlaceholder?: string;
   emailListsEnabled?: boolean;
-  selectedTheme: string;
+  selectedTheme?: string;
   photoConsents?: Array<PhotoConsent>;
```
ivarnakken, line 22: What's the point of keeping the `UserEntity` type when we have a `User` model in _app/models.ts_? Are they not the same? It seems like the only difference is `selectedTheme` and `emailListEnabled`.

app/reducers/comments.ts
```diff
@@ -9,10 +9,11 @@ export type CommentEntity = {
   id: ID;
   parent?: ID;
   createdAt: string;
-  children: Array<Record<string, any>>;
+  children?: Array<Record<string, any>>;
```
ivarnakken, line 12: Could you type this better while you're at it :))

app/components/Layout/Flex.tsx
```diff
@@ -1,6 +1,6 @@
 import cx from 'classnames';
 import styles from './Flex.css';
-import type { ReactNode, ElementType } from 'react';
+import type { ReactNode, ElementType, HTMLAttributes } from 'react';
```
ivarnakken, line 3: I now realized that I'd prefer if our import sort put `react` imports at the top and style imports at the bottom. Just a thought.

app/components/Image/ProfilePicture.tsx
```diff
@@ -1,16 +1,18 @@
 import type { UserEntity } from 'app/reducers/users';
 import CircularPicture from './CircularPicture';
+import type { ComponentProps } from 'react';
```
ivarnakken, line 3: Is this not sorted wrongly? Why is it not caught by our linter?

ivarnakken (approved):
This is extremely satisfying to review! 😍

app/components/Layout/Flex.tsx
ivarnakken: I now also realized this is a type import, so forget what I said about react, but the style import remains.

app/components/Image/ProfilePicture.tsx
LudvigHz: heh, our sorting rules are a bit _off_. Everything is sorted automatically, so this is how it wants it. 
I think the `app/` group overrides type imports :/ Should look at it, but not in this PR

ivarnakken: Yes, definitely out of scope here.

LudvigHz:
Heads up! I originally wanted to merge this, and do incremental type updates. But I touched on a lot of the same code in my other branch, so I decided to include those commits here as well. It will be a bit larger, but I think it's better considering some of the issues mentioned here might be fixed by those commits :)

app/reducers/users.ts
LudvigHz: Yeah, the whole double typing thing is ridiculous. But I'm trying to move more and more types into the `app/models`. The `***Entity` types seem to be more for creating new entities and typing the actions, so we'll keep them for now.

In some of the changes here, you can see I replace a bunch of the `Entity` types with better types from `app/models` and I think that's the way to go here.

app/reducers/comments.ts
LudvigHz: I don't really want to :grimacing:.

Trying to focus on the TS errors, and typing the `any`s that are pretty easy. This one I don't really know what should be so I don't want to spend a lot of time on it.

app/reducers/meetings.ts
```diff
@@ -8,19 +8,22 @@ export type MeetingEntity = {
   id: number;
   title: string;
   location: string;
-  startTime: moment$Moment;
-  endTime: moment$Moment;
+  startTime: Moment;
+  endTime: Moment;
```
ivarnakken, line 12: Are these not `Dateish`?

app/models.ts
```diff
@@ -233,16 +300,35 @@ export type CompanySemesterContactedStatus =
...
   | 'not_contacted';
+
+export type Meeting = {
+  id: ID;
+  createdBy: ID;
+  title: string;
+  location: string;
+  startTime: Moment;
+  endTime: Moment;
+  reportAuthor: number | null;
```
ivarnakken, line 311: Shouldn't this rather be of type `ID`?

app/models.ts
```diff
@@ -233,16 +300,35 @@ export type CompanySemesterContactedStatus =
...
   | 'contacted'
   | 'not_contacted';
+
+export type Meeting = {
+  id: ID;
+  createdBy: ID;
+  title: string;
+  location: string;
+  startTime: Moment;
+  endTime: Moment;
```
ivarnakken, line 310: Same as comment above.

app/routes/events/components/RegistrationMeta.tsx
```diff
@@ -58,7 +58,7 @@ const ConsentStatus = ({
   if (!useConsent) return null;
 
   if (photoConsents && allConsentsAnswered(photoConsents)) {
-    const { WEBSITE, SOCIAL_MEDIA } = PHOTO_CONSENT_DOMAINS;
+    const { WEBSITE, SOCIAL_MEDIA } = PhotoConsentDomain;
```
ivarnakken, line 61: Is it not bad practice to destruct an enum?

LudvigHz: Nah, it works just like an object, just with proper types. So it doesn't really matter

ivarnakken: I just felt like it kinda defeated its purpose, even thought it kinda doesn't ..

eikhr (approved):

PR #3263, opened 2022-11-12
Upgrade to Yarn 3 (without PnP)

AdrianAndersen:
With this PR, we can start using Yarn 3.

This should be a non breaking switch, since we are not adopting the PnP features yet.
The benefits of yarn 3 are many, but mostly performance (install speed) and reliability in our case.
If you are interested in reading more about why we should upgrade:
[Why should you upgrade to Yarn Modern?](https://yarnpkg.com/getting-started/qa)


norbye (approved):
LGTM:)
Appreciate the linked article

LudvigHz:
Shouldn't the `.yarn/cache` folder be gitignored if you're not using PnP? Ref. https://yarnpkg.com/features/zero-installs#how-do-you-reach-this-zero-install-state-youre-advocating-for

LudvigHz (commented):
I'm not _entirely_ sure what I think about this (although I'm somewhat positive IF we use PnP).
The main reason to switch to modern yarn versions is for large projects that are monorepos or have a ton of packages. We don't, so the benefits will be minimal.

However, I have read good things about `yarn >= 2`, so it might be a good idea to switch. The only thoughts I have are:
- How will this affect our resolution? Webpack probably works fine, but if we want to switch to vitejs or turbopack, are we still covered?
- zero-installs: There are plenty of good reasons to use zero-installs (commiting the `.yarn/cache` folder to git), but I just kinda don't like it. Just because of the fact that having a bunch of binary files in the repo seems _off_ to me. And again, we don't have that many dependencies that it will save a huge amount of time. Running `yarn` on branch switches already takes `< 1 min`.
- The main benefits seem to come from using PnP and zero-installs, which this migration does _not_ do. So with that, I really don't see a reason to migrate. I think either we should either use PnP (zero-installs or not), or stay on yarn classic.

AdrianAndersen:
> Shouldn't the .yarn/cache folder be gitignored if you're not using PnP? Ref. https://yarnpkg.com/features/zero-installs#how-do-you-reach-this-zero-install-state-youre-advocating-for

Since this PR does not use PnP, yes you could omit the cache folder. But, it makes the install faster nevertheless since it does not need to fetch the package versions from upstream each install. (It only needs to move the package versions into node_modules)

> I'm not _entirely_ sure what I think about this (although I'm somewhat positive IF we use PnP). The main reason to switch to modern yarn versions is for large projects that are monorepos or have a ton of packages. We don't, so the benefits will be minimal.

Yeah, I agree that the benefits are not huge, but I think it is good to be on a version that is actually in active development, rather than on a version that is actually dead. 

> However, I have read good things about `yarn >= 2`, so it might be a good idea to switch. The only thoughts I have are:
> 
> * How will this affect our resolution? Webpack probably works fine, but if we want to switch to vitejs or turbopack, are we still covered?

Webpack, ViteJS and Turbopack all support Yarn 3 with and without PnP


> * zero-installs: There are plenty of good reasons to use zero-installs (commiting the `.yarn/cache` folder to git), but I just kinda don't like it. Just because of the fact that having a bunch of binary files in the repo seems _off_ to me. And again, we don't have that many dependencies that it will save a huge amount of time. Running `yarn` on branch switches already takes `< 1 min`.
> * The main benefits seem to come from using PnP and zero-installs, which this migration does _not_ do. So with that, I really don't see a reason to migrate. I think either we should either use PnP (zero-installs or not), or stay on yarn classic.

Zero Installs are the new ✨shiny✨ way to do it, and it is kinda useless to have yarn3 without it. I think the benefits are smaller with PnP, but it still has some minor improvements. 

I tried to enable PnP, but there were some issues that I need to look into. Would be nice to have, but I do not think it is critical for this PR at least.

AdrianAndersen:
So, I dug a bit deeper, and found out that there are a few issues stopping us from adopting PnP right away:

1. In many cases, we import peerDependencies directly in our code, which is not possible in PnP. For instance, we some places import stuff from "react-router", even though that package is not explicitly installed. (It is peer dependency by connected-react-router). In these cases, we need to install the dependencies needed explicitly or change the imports. 
2. Some places we import stuff directly from node_modules, such as css. This is not supported with PnP (since there is no node_modules) and we need to find a way around it. 
3. The Webpack config has references to node_module stuff, so it needs to be updated. 

These issues can be uncovered by running 
'yarn dlx @yarnpkg/doctor', but that is only possible if we have adopted Yarn 3. The above issues are arguably something we should take care of anyways, so therefore I suggest the following:

To adopt Yarn 3 using node_module resolution for now, either with or without the cache folder. (I suggest we keep it, since it provides faster installs and no real drawbacks other than a slightly larger repo size) 

Then, once we have adopted Yarn 3, we can gradually fix our bad practices in terms of dependencies, and eventually adopt PnP. 

What do you think about this approach, @LudvigHz?

LudvigHz:
> So, I dug a bit deeper, and found out that there are a few issues stopping us from adopting PnP right away:

Yeah, PnP is not at all important for us. As previously mentioned the benefits are zero to none.

> To adopt Yarn 3 using node_module resolution for now, either with or without the cache folder. (I suggest we keep it, since it provides faster installs and no real drawbacks other than a slightly larger repo size)

I'm fine with that. But I'm a _bit_ confused about the cache folder. Yarn themselves recommend **not** committing the cache folder unless you're using zero-installs[^1]. So I'm not sure exactly what's best here, although I still don't like, or see the reason for, storing the cache in the repo.

> Then, once we have adopted Yarn 3, we can gradually fix our bad practices in terms of dependencies, and eventually adopt PnP.

Again, I don't think we need to rush to use PnP at all. There are a hundred other tooling improvements we rather should spend time on than saving some seconds during dependency resolution.


[^1]: https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

AdrianAndersen:
Imma kill dis for now

PR #3264, opened 2022-11-13
Implement a registration skeleton on event page

ivarnakken:
To prevent the register button from moving once people start registering (which we've recieved a ton of feedback on). The skeleton will, however, only be visible once there's less than 5 minutes until activation.

### When someone has registered:

---

Closes https://github.com/webkom/lego/issues/3004 and closes ABA-122

### DIFF
--- a/app/components/UserAttendance/AttendanceStatus.css
+++ b/app/components/UserAttendance/AttendanceStatus.css
@@ -13,13 +13,13 @@
 .poolBox {
   display: flex;
   flex: 1;
-  background-color: rgba(200, 200, 200, 30%);
   max-width: 100%;
   min-width: 25%;
   flex-direction: column;
   align-items: center;
+  background-color: rgba(200, 200, 200, 30%);
 }
 
 html[data-theme='dark'] .poolBox {
-  background-color: rgba(50, 50, 50, 30%);
+  background-color: rgba(80, 80, 80, 30%);
 }
--- a/app/components/UserGrid/index.tsx
+++ b/app/components/UserGrid/index.tsx
@@ -1,7 +1,9 @@
+import cx from 'classnames';
 import { Link } from 'react-router-dom';
 import { ProfilePicture } from 'app/components/Image';
 import Tooltip from 'app/components/Tooltip';
 import type { User } from 'app/models';
+import styles from './UserGrid.css';
 
 const UserGrid = ({
   users,
@@ -10,7 +12,7 @@ const UserGrid = ({
   minRows = 0,
   padding = 3,
 }: {
-  users: Array<User>;
+  users: User[];
 
   /* profile picture size */
   size?: number;
@@ -26,17 +28,23 @@ const UserGrid = ({
       }px, 1fr))`,
       gridTemplateRows: users.length === 0 ? 0 : size + padding,
       overflow: 'hidden',
-      minHeight: minRows * size + (minRows - 1) * padding,
-      maxHeight: maxRows * size + (maxRows - 1) * padding,
+      minHeight: minRows * size + (minRows - 1) * padding + 3.3,
+      maxHeight: maxRows * size + (maxRows - 1) * padding + 3.3,
     }}
   >
     {users.map((user) => (
-      <RegisteredCell key={user.id} user={user} />
+      <RegisteredUserCell key={user.id} user={user} />
     ))}
+
+    {/* Only 15 users are passed into the users prop */}
+    {minRows > 0 &&
+      Array.from({ length: 15 - users.length }, (_, index) => (
+        <SkeletonUserCell key={index} />
+      ))}
   </div>
 );
 
-export const RegisteredCell = ({ user }: { user: User }) => (
+const RegisteredUserCell = ({ user }: { user: User }) => (
   <Tooltip
     style={{
       marginTop: '-7px',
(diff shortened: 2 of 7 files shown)


app/components/UserAttendance/AttendanceStatus.css
```diff
@@ -13,13 +13,13 @@
...
   max-width: 100%;
   min-width: 25%;
   flex-direction: column;
   align-items: center;
+  background-color: rgba(200, 200, 200, 30%);
 }
 
 html[data-theme='dark'] .poolBox {
-  background-color: rgba(50, 50, 50, 30%);
+  background-color: rgba(80, 80, 80, 30%);
```
ivarnakken, line 24: This is kinda unrelated to the PR, but it has bugged me a long time and I refuse to open a separate PR for a color change.

app/components/UserGrid/index.tsx
```diff
@@ -26,17 +28,23 @@ const UserGrid = ({
       }px, 1fr))`,
       gridTemplateRows: users.length === 0 ? 0 : size + padding,
       overflow: 'hidden',
-      minHeight: minRows * size + (minRows - 1) * padding,
-      maxHeight: maxRows * size + (maxRows - 1) * padding,
+      minHeight: minRows * size + (minRows - 1) * padding + 3.3,
+      maxHeight: maxRows * size + (maxRows - 1) * padding + 3.3,
```
ivarnakken, line 32: `3.3` pixels are added here and in `UserGrid.css` so that, once people start registering, the skeleton and profile pictures will be aligned. This is because the images contain margins. It's not very clean, but it's better than nothing, and I've spent way too many hours trying to find a better solution ..

app/components/UserAttendance/AttendanceStatus.css
```diff
@@ -13,13 +13,13 @@
 .poolBox {
   display: flex;
   flex: 1;
-  background-color: rgba(200, 200, 200, 30%);
   max-width: 100%;
   min-width: 25%;
   flex-direction: column;
   align-items: center;
+  background-color: rgba(200, 200, 200, 30%);
```
norbye, line 20: Seems unnecessary, but sure

app/components/UserGrid/index.tsx
```diff
@@ -26,17 +28,23 @@ const UserGrid = ({
...
     }}
   >
     {users.map((user) => (
-      <RegisteredCell key={user.id} user={user} />
+      <RegisteredUserCell key={user.id} user={user} />
     ))}
+
+    {/* Only 15 users are passed into the users prop */}
+    {minRows > 0 &&
+      Array.from({ length: 15 - users.length }, (_, index) => (
```
norbye, line 41: Why is this not something like `maxRows*5 (images pr row) - users-length`? Wouldn't 15 - users.length produce three rows of 5 images rather than two, or am I missing something here

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -205,16 +205,29 @@ export default class EventDetail extends Component<Props, State> {
...
       'hours'
     );
+
     const registrationCloseTimeMoment = registrationCloseTime(event);
+
+    // The UserGrid is expanded when there's less than 5 minutes till activation
+    const minUserGridRows = currentMoment.isAfter(
+      activationTimeMoment.subtract(5, 'minutes')
+    )
+      ? 2
```
norbye, line 228: Could this, and the value of the maxRows property passed in line 434 be combined into a const?

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -417,7 +431,7 @@ export default class EventDetail extends Component<Props, State> {
                   {registrations ? (
                     <Fragment>
                       <UserGrid
-                        minRows={0}
+                        minRows={minUserGridRows}
                         maxRows={2}
```
norbye, line 435: This line 434

norbye (commented):
Awesome! Got a few clarifying questions and minor stuff, but other than that it seems good 💯

app/components/UserAttendance/AttendanceStatus.css
ivarnakken: I prefer to have visual stylings that don't affect size/position/etc. at the bottom (color, background, font, border).

app/components/UserGrid/index.tsx
ivarnakken: The amount of images per row is not always 5. Try to resize your viewport (to just before it converges to mobile view) and you'll see.

app/routes/events/components/EventDetail/index.tsx
ivarnakken: Sure thing

norbye (approved):

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -51,6 +51,9 @@ type InterestedButtonProps = {
   isInterested: boolean;
 };
 
+const MIN_USER_GRID_ROWS = 2;
+const MAX_USER_GRID_ROW = 2;
```
LudvigHz, line 55: It's just too annoying that you don't have an `S` here that I have to comment on it :grimacing:

LudvigHz (approved):
LGTM

app/routes/events/components/EventDetail/index.tsx
ivarnakken: oops, ye that annoys me as well

PR #3278, opened 2022-11-15
Fix Tag Links On "Festet Oppslag"

haukkagu:
The tags under "Festet Oppslag" weren't proper links, so you couldn't click on them and be redirected to the tag page. Now this is fixed.

### DIFF
--- a/app/routes/overview/components/utils.tsx
+++ b/app/routes/overview/components/utils.tsx
@@ -40,7 +40,7 @@ export const renderMeta = (
       {item.tags?.length > 0 && (
         <Tags className={styles.tagline}>
           {item.tags.slice(0, 3).map((tag) => (
-            <Tag tag={tag} key={tag} />
+            <Tag tag={tag} key={tag} link={`/articles/?tag=${tag}`} />
           ))}
         </Tags>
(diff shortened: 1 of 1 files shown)


app/routes/overview/components/utils.tsx
```diff
@@ -40,7 +40,7 @@ export const renderMeta = (
       {item.tags?.length > 0 && (
         <Tags className={styles.tagline}>
           {item.tags.slice(0, 3).map((tag) => (
-            <Tag tag={tag} key={tag} />
+            <Tag tag={tag} key={tag} link={'/articles/?tag=' + tag} />
```
ivarnakken: Okay, so, I now realized that events _can_ hold tags, despite there being no possible way to actually tag an event in the GUI ... so `item.tags.length` will always be zero. Not sure if we want to check for this here though ... could use `isEvent`

LudvigHz (approved):
I guess this is fine, although I feel like it _could_ be annoying if you expect to get to the article but click on the tag instead. Especially on mobile.

Arashfa0301 (approved):
LGTM:)

PR #3279, opened 2022-11-15
Remove redux cache

eikhr:
As far as I can tell, the cache was only used to avoid duplicate requests after SSR. This is now handled by `react-prepare`, so it is no longer needed.


LudvigHz:
I disagree quite strongly with this. The point of the cache is _not_ just to avoid duplicate requests from SSR as you mention. It is there as a general cache.

The caching here is quite simple and we get it "for free". It also should be enabled by default for all GET requests. I do however agree that it is not used that much (it might be that the caching has been a bit broken for a while, since I couldn't get it to properly return cached requests in prod :upside_down_face: ), but that is rather an argument for using it _more_, not less.

eikhr:
Right now there is code specifically disabling all caching client side, see the [fetchHistory reducer](https://github.com/webkom/lego-webapp/blob/e9980de64063cca2118365b0036ff56f91ac0509/app/reducers/fetchHistory.ts#L23). 

The way I see it an extra request once in a while is a small price to pay for a more consistently updated state. The entities already in the redux state also work as a type of "optimistic" state until the new fetch has completed. In my opinion we should refetch, but just avoid hiding the entire page behind a loading animation every time it refetches.

LudvigHz:
> Right now there is code specifically disabling all caching of requests sent client side, see the [fetchHistory reducer](https://github.com/webkom/lego-webapp/blob/e9980de64063cca2118365b0036ff56f91ac0509/app/reducers/fetchHistory.ts#L23).

Hmm, right I forgot about that one :upside_down_face:. (it should still cache 3 times no?) (although as you say, the new prepare behavior should cover that use-case)
 
> The way I see it an extra request once in a while is a small price to pay for a more consistently updated state. The entities already in the redux state also work as a type of "optimistic" state until the new fetch has completed. In my opinion we should refetch, but just avoid hiding the entire page behind a loading animation every time it refetches.

I see what you mean, but I still kinda disagree with this. For a lot of things, there is absolutely no need to refetch. F.ex. pages, groups, joblistings and similar that almost never update and are very unlikely to update in the same session or a certain cache time. Caching will avoid unnecessary requests, which is useful not only to reduce load on our servers, but also quite nice for slow connections.

I do think "strongly disagree" was a bit too strong words though. I can understand that you want to remove "unused" code. But I do still think that we should cache more, and I think removing this code will remove code that could be modified very little to functional. The reason caching was disabled was partly due to weird behavior with returning incorrect request for pagination and requests with query string etc. Alot of that was fixed recently and the caching behavior should:tm: be correct for those actions using the improved pagination handler.


But yeah, I realize I'm saying quite a lot of "everything was better before" here, so don't listen too much to me if people want to change things :slightly_smiling_face:

erlingfn:
I do not really know how the caching we currently have works well enough to know what changes this will bring, so I cannot vouch for these changes before reading up a bit. However, I personally do not like having custom caching. Premature optimization is the root of a lot of evil, and there is already a lot of weird behavior stemming from the redux state, such as which events are displayed on the user page. If this is a big problem for users we could spend time on it, but otherwise it just makes the frontend more complex.

LudvigHz (approved):

ivarnakken (approved):
💯

ivarnakken:
@eikhr, what's the status on this? Any reason to not get it in?

PR #3283, opened 2022-11-16
Implement new event components for upcoming events on profile page.

danielyanghansen:
### (Revamped design after feedback)
Future Event Components now combines waiting-list events and user-admitted events . Includes tooltip for event attendance status.

## New Design
new_dark_mode
new_light_mode

### Vs the Old Design
old_dark_mode
old_light_mode

### DIFF
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -1,4 +1,3 @@
-import moment from 'moment-timezone';
 import { Link } from 'react-router-dom';
 import Icon from 'app/components/Icon';
 import { Image } from 'app/components/Image';
@@ -6,57 +5,59 @@ import { Flex } from 'app/components/Layout';
 import Pill from 'app/components/Pill';
 import Tag from 'app/components/Tags/Tag';
 import Time from 'app/components/Time';
-<<<<<<< HEAD
+import Tooltip from 'app/components/Tooltip';
 import type { Event, EventTime } from 'app/models';
 import { colorForEvent } from 'app/routes/events/utils';
-import { eventStatus, eventAttendance } from 'app/utils/eventStatus';
-=======
-import Tooltip from 'app/components/Tooltip';
-import type { Event, EventTimeType } from 'app/models';
-import { colorForEvent } from 'app/routes/events/utils';
-import { EVENTFIELDS } from 'app/utils/constants';
-import {
-  eventStatus,
-  eventAttendance,
-  eventAttendanceAbsolute,
-} from 'app/utils/eventStatus';
->>>>>>> 164a29ac (Implement new event components for upcoming events on profile page.)
+import { eventAttendanceAbsolute } from 'app/utils/eventStatus';
 import styles from './styles.css';
 import type { ReactNode } from 'react';
-
-type AttendanceProps = {
-  event: Event;
-};
 
 export type EventStyle = 'default' | 'extra-compact' | 'compact';
 
 type statusIconProps = {
+  status: string;
   icon: string;
   color: string;
   tooltip: string;
 };
 
-const statusIconDict = {
-  'Du er påmeldt': {
-    icon: 'checkmark-circle-outline',
-    color: '#0eb30e',
-    tooltip: 'Du er påmeldt',
-  },
-  'Ingen påmeldingsrett': {
-    icon: 'timer-outline',
-    color: '#fc9003',
-    tooltip: 'Du er på ventelisten',
-  },
+const eventStatusObject = (event: Event): statusIconProps => {
+  const { isAdmitted, eventStatusType } = event;
+
+  switch (eventStatusType) {
+    case 'NORMAL':
+    case 'INFINITE':
+      if (isAdmitted) {
+        return {
+          status: 'Admitted',
+          icon: 'checkmark-circle-outline',
+          color: '#1ec91e',
+          tooltip: 'Du er påmeldt',
+        } as statusIconProps;
+      }
+      return {
+        status: 'Waitlist',
+        icon: 'timer-outline',
+        color: '#fc9003',
+        tooltip: 'Du er på ventelisten',
+      } as statusIconProps;
+    default:
+      return {
+        status: 'Error',
+        icon: 'help-outline',
+        color: 'blue',
+        tooltip: 'Det har oppstått en feil',
+      } as statusIconProps;
+  }
 };
 
-const Attendance = ({ event }: AttendanceProps) => {
+const Attendance = ({ event }) => {
   const attendance = eventAttendanceAbsolute(event);
   return (
-    attendance && (
+    !!attendance && (
       <Pill
         style={{
           marginLeft: '5px',
-          marginBottom: '5px',
           color: 'black',
           whiteSpace: 'nowrap',
         }}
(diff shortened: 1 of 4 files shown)


ivarnakken:
Here's everything I notice regarding the visual changes:

- Icons (calendar and clock) are not aligned to the text. Simply add a `align-items: center`.
- Cover image should be way larger.
- More spacing is needed between cover image and text. Consider putting the text on the right side, instead of below.
- Tooltip content should not be capitalized.
- I like that you replaced the text of whether you're registered or on waiting list with an icon, but I'm not a fan of it being inside a gray wrapper. I feel like the placement of elements here should be reconsidered.
- The width of the gray boxes feel too large, in relation to its small content.

app/components/EventItem/index.tsx
```diff
@@ -41,22 +68,58 @@ type TimeStampProps = {
 };
 
 const TimeStamp = ({ event, field, loggedIn }: TimeStampProps) => {
-  const registration = eventStatus(event, loggedIn, true);
   const hasStarted = moment().isAfter(event.startTime);
   const startedStatus = hasStarted ? 'Startet' : 'Starter';
```
eikhr, line 72: unused
```suggestion
```

app/components/EventItem/index.tsx
```diff
@@ -41,22 +68,58 @@ type TimeStampProps = {
 };
 
 const TimeStamp = ({ event, field, loggedIn }: TimeStampProps) => {
```
eikhr, line 70: unused
```suggestion
const TimeStamp = ({ event }: TimeStampProps) => {
```

app/components/EventItem/index.tsx
```diff
@@ -41,22 +68,58 @@ type TimeStampProps = {
...
+        />
+        <Time time={event.startTime} format="HH:mm" />
+      </Flex>
     </div>
   );
 };
 
+//TODO: Create icon with tooltip as registration / eventStatus
+const RegistrationIcon = ({ event, field, loggedIn }: TimeStampProps) => {
+  const attendance = eventAttendanceAbsolute(event);
```
eikhr, line 97: unused
```suggestion
```

app/components/EventItem/index.tsx
```diff
@@ -41,22 +68,58 @@ type TimeStampProps = {
...
+          size={20}
+          style={{ cursor: 'pointer', marginRight: '10px' }}
+        />
+        <Time time={event.startTime} format="HH:mm" />
+      </Flex>
     </div>
   );
 };
 
+//TODO: Create icon with tooltip as registration / eventStatus
```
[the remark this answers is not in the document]
danielyanghansen: Yeah that's not supposed to be there

ivarnakken:
I now also notice that the icons and text are not vertically aligned.

app/components/EventItem/index.tsx
```diff
@@ -1,30 +1,67 @@
...
   return (
-    attendance && (
+    !!attendance && (
       <Pill
         style={{
           marginLeft: '5px',
+          marginBottom: '5px',
           color: 'black',
           whiteSpace: 'nowrap',
+          alignItems: 'center',
```
ivarnakken, line 64: Is the content not aligned already?

app/components/EventItem/index.tsx
```diff
@@ -38,23 +75,44 @@ type TimeStampProps = {
...
   return (
-    <div className={styles.eventTime}>
-      {registration && (
-        <span>
-          {registration}
-          <br />
-        </span>
-      )}
-      {startedStatus} <Time time={event.startTime} format="ll - HH:mm" />
+    <div className={styles.eventTime} style={{ alignItems: 'center' }}>
```
ivarnakken, line 80: Please no inline styling.

app/components/EventItem/index.tsx
```diff
@@ -95,7 +154,58 @@ const EventItem = ({
...
+          style={{ borderColor: colorForEvent(event.eventType) }}
+          className={styles.eventItemCompact}
+        >
+          <div className={styles.eventItemRow}>
+            <Link to={`/events/${event.id}`}>
+              <h3 className={styles.eventItemTitle}>{event.title}</h3>
+            </Link>
+          </div>
+          <div className={styles.eventItemRow}>
+            <div className={styles.eventItemColumn} style={{ width: '72%' }}>
```
ivarnakken, line 169: 🚫 inline styling

ivarnakken: There's more below here as well.

app/components/EventItem/index.tsx
```diff
@@ -95,7 +154,58 @@ const EventItem = ({
...
+          </div>
+          <div className={styles.eventItemRow}>
+            {showTags && (
+              <Flex wrap>
+                {event.tags.map((tag, index) => (
+                  <Tag key={index} tag={tag} />
+                ))}
+              </Flex>
+            )}
+          </div>
```
ivarnakken, line 205: What's the point of an empty div if `showTags` is `false`?

app/components/EventItem/styles.css
```diff
@@ -6,6 +6,26 @@
...
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row;
+}
+
+.eventItemColumn {
+  width: 100%;
+  display: flex;
+  justify-content: flex-start;
+  flex-direction: column;
```
ivarnakken, line 26: Please just use the `<Flex />` component if the css only adds stuff like this.

app/components/EventItem/index.tsx
ivarnakken: And is it even needed?

ivarnakken (changes requested):
Looking much better! Please just fix up on the tiny comments I've made and we can get this in soon 💯

app/components/EventItem/index.tsx
danielyanghansen: I couldn't find a way of removing the inline styling here, as it would force the class to be overwritten and therefore lose a lot of the attributes.
Is there a way around this issue?

app/components/EventItem/index.tsx
danielyanghansen: You're right, it wasn't needed

app/components/EventItem/index.tsx
```diff
@@ -95,7 +153,60 @@ const EventItem = ({
...
+            </Flex>
+          </Flex>
+          {showTags && (
+            <Flex width="100%" justifyContent="flex-start">
+              <Flex wrap>
+                {event.tags.map((tag, index) => (
+                  <Tag key={index} tag={tag} />
+                ))}
+              </Flex>
+            </Flex>
```
ivarnakken, line 205:
```suggestion
            <Flex wrap width="100%" justifyContent="flex-start">
              {event.tags.map((tag, index) => (
                <Tag key={index} tag={tag} />
              ))}
            </Flex>
```

Is this not the same thing?

app/components/EventItem/index.tsx
```diff
@@ -1,28 +1,64 @@
...
+  const attendance = eventAttendanceAbsolute(event);
   return (
-    attendance && (
+    !!attendance && (
       <Pill
         style={{
           marginLeft: '5px',
+          marginBottom: '5px',
           color: 'black',
           whiteSpace: 'nowrap',
```
ivarnakken, line 63: Is this styling really needed though? I feel like the added margin is kind of hacky if it's to get it aligned.

You're saying a classname does not override the styling, but that's just because the `<Pill />` component does not employ the `className` prop from `HTMLAttributes<HTMLSpanElement>` onto its span.

ivarnakken (changes requested):

app/components/EventItem/index.tsx
danielyanghansen: Yeah this is not needed. I forgot to remove it after fixing the aligning to be "not hacky".

ivarnakken (approved):
Awesome job 💯

PR #3284, opened 2022-11-16
Fix Unregistration With Unanswered Survey

itsisak:
Previously if you had an unanswered survey you could not unregister from an event.

Added so that if you are registered for the event it will show the unregister button and not the warning that you have unanswered surveys.


danielyanghansen (approved):

eikhr (approved):
nice!

haukkagu (approved):
LGTM

ivarnakken (approved):

PR #3309, opened 2022-11-23
Bump node to v16

ivarnakken:
Noticing a lot of [deps](https://github.com/webkom/lego-webapp/pull/3297) dropping support for node versions less than latest 14, so it's good to stay on top of things.

Also remove Dockerfile.minimal


.drone.yml
```diff
@@ -233,7 +233,7 @@ steps:
       - yarn ssr
 
   - name: cypress
```
ivarnakken, line 235: `legocypresshelper` is already running on version 16

.drone.yml
```diff
@@ -233,7 +233,7 @@ steps:
...
-    image: cypress/browsers:node14.16.0-chrome89-ff86
+    image: cypress/browsers:node16.18.0-chrome90-ff88
```
LudvigHz, line 236: Shouldn't we update the browser versions here at the same time? These are quite old now

LudvigHz (changes requested):
Nice :100: 
Just remember to update the drone signature and LGTM :+1:

.drone.yml
ivarnakken: This is the latest browser version for node v16.18. See all [here](https://hub.docker.com/r/cypress/browsers/tags). I suppose we could opt to go for a lower node 16 version and a higher chrome version, but I think that's working in the opposite direction.

LudvigHz: Well no. There is a version with node 16.18 and chrome 107 and ff 106: https://hub.docker.com/layers/cypress/browsers/node16.18.0-chrome107-ff106-edge/images/sha256-6205955b1afe6545e9151b3d4ce9e20ae1d05d40052bb3a97669e05cb813a2f6?context=explore

ivarnakken: oops, turns out I'm legally blind. I just read "edge" and immediately skipped it. Updated now.

LudvigHz (approved):

PR #3317, opened 2022-11-26
Fix types in actions/*

LudvigHz:
Just some quick and easy type fixes in app/actions/\* (except for `callAPI.ts`).

### DIFF
--- a/app/actions/EventActions.ts
+++ b/app/actions/EventActions.ts
@@ -249,7 +249,7 @@ export function unregister({
 export function adminRegister(
   eventId: number,
   userId: number,
-  poolId?: number,
+  poolId: number | undefined,
   feedback: string,
   adminRegistrationReason: string
 ): Thunk<any> {
--- a/app/types.ts
+++ b/app/types.ts
@@ -1,5 +1,6 @@
 import type { RootState } from 'app/store/createRootReducer';
 import type { ThunkAction } from '@reduxjs/toolkit';
+import type { JwtPayload } from 'jwt-decode';
 
 export type AsyncActionType = {
   BEGIN: string;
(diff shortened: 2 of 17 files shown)


app/actions/EventActions.ts
```diff
@@ -249,7 +249,7 @@ export function unregister({
 export function adminRegister(
   eventId: number,
   userId: number,
-  poolId?: number,
+  poolId: number | undefined,
```
ivarnakken, line 252: I'm more fan of the `?` as an optional indicator. I assume you changed this because ts doesn't allow required parameters to follow an optional parameter, but then I'd prefer to just move `poolId?: number,` down the list.

app/types.ts
```diff
@@ -35,7 +36,7 @@ export type GalleryEntity = {
   event?: EntityID;
 };
 export type EncodedToken = string;
-export type DecodedToken = {
+export type DecodedToken = JwtPayload & {
```
ivarnakken, line 39: I'm not sure if I see the point of extending this off of `JwtPayload`, but okay. `exp` is a field there, so you could remove it here to avoid dups, but no biggy.

ivarnakken (approved):
💯 

Two minor comments, which I'll let you decide what to do with.

app/types.ts
LudvigHz: Well yes but all properties in `JwtPayload` are optional. Extending this just means the type is more correct.

app/actions/EventActions.ts
LudvigHz: Yeah I do agree, but it's used in this order elsewhere and I didn't want to change that here. Sure it's a little uglier but it's not that big of a deal IMO

eikhr (approved):
Looks good

But regarding the `SearchActions.spec.js` test, why not just delete it? It doesn't test anything.

LudvigHz:
> Looks good
> 
> But regarding the `SearchActions.spec.js` test, why not just delete it? It doesn't test anything.

Oh, haha. I didn't actually look at it. I just renamed it blindly 😅.

I'll delete it then.

PR #3321, opened 2022-11-28
Added role identifier at group flatpages

Arashfa0301:
Implemented role identifier for 4 more pre-defined roles. These are: "pr-ansvarlig", "sosialansvarlig", "opptakansvarlig" and "økonomiansvarlig". I didn't find it necessary to add more since most of the groups I checked (didn't check all) used only these four. Feel free to suggest improvements :)
Screenshot 2022-11-28 at 20 38 36

Closes issue #3050


ivarnakken:
Please provide some illustrations 😄

Arashfa0301:
I have, but can provide more.

ivarnakken:
> I have, but can provide more.

More. It's difficult to get a sense of your change from what you have given. Please just change the names instead of covering them up.

Arashfa0301:
I understand. Working on it 👍.

LudvigHz:
Just so you understand why CI fails: I've set a hard limit for how many ESLint warnings we allow. So even though you have no errors, this PR introduces more warnings than it removes and as such CI fails. To fix it, fix some (preferably most) of the warnings in the files you have changed.

LudvigHz:
I also expect you to fix any TS errors in the same files :smile: (those that were there from before)

ivarnakken:
@Arashfa0301 Do not update this branch by merging master into it. Instead, rebase from `master` and force push to this branch. We don't want any unnecessary commits.

Arashfa0301:
@ivarnakken Yes thank you for pointing it out.

PeterJFB (approved):
LGTM :fire:. Don't really like the interchangeable use of `media_relations` and `pr_responsibles`, but i struggle to come up with a plural of _relations_ (relationees?) (shouldn't be a legal variable name) myself:P

PR #3371, opened 2022-12-14
Fix some TS types

LudvigHz:
- Comment tests
- Conent
- EventItem
- EventList


app/components/Content/Content.tsx
```diff
@@ -49,7 +49,7 @@ function Content({
         <div>
           <LoadingIndicator
             loading={isLoading}
-            className={isLoading ? {} : styles.hidden}
+            className={isLoading ? null : styles.hidden}
```
ivarnakken, line 52: I feel like `isLoading && styles.hidden` is cleaner

app/components/EventItem/index.tsx
```diff
@@ -36,11 +34,11 @@ const Attendance = ({ event }: AttendanceProps) => {
 
 type TimeStampProps = {
   event: Event;
-  field: EventTimeType;
```
ivarnakken, line 39: Why not remove this?

ivarnakken (approved):
💯

norbye (approved):

app/models.ts
```diff
@@ -4,7 +4,10 @@ export type ID = number;
 export type Dateish = Moment | Date | string;
 export type ActionGrant = Array<string>;
 export type IcalToken = string;
-export type EventTimeType = 'activationTime' | 'startTime';
+export enum EventTime {
+  activate = 'activationTime',
+  start = 'startTime',
```
eikhr, line 9: We're not super consistent about it, but we use PascalCase or SCREAMING_SNAKE_CASE for enum values elsewhere. I suggest we descide on one of these and use it everywhere.
TypeScript docs use PascalCase, but I have no strong preference.

eikhr (approved):
Looks good, but it seems like the new eslint type import rule fails. Running eslint with fix flag should fix it automatically.

app/models.ts
LudvigHz: Yeah. I do agree, but I'm kinda lazy and didn't want to rename all these so I kept the original name. It's just to minimize the amount of changes

PR #3381, opened 2022-12-21
Enforce type imports when import not used as value

eikhr:
Enable eslint and tsc rule for enforcing that imports used only as types are imported with the `type` keyword. This allows the transpilers to safely ignore these imports as they will not be used at runtime, but more importantly it stops IntelliJ from automatically adding non-`type` imports from `.d.ts`-files (which does not work).


app/components/Chart/BarChart.tsx
```diff
@@ -1,7 +1,8 @@
 import { BarChart, Bar, CartesianGrid, Cell, XAxis, YAxis } from 'recharts';
+import type {
+  DistributionDataPoint} from 'app/components/Chart/utils';
 import {
-  CHART_COLORS,
-  DistributionDataPoint,
+  CHART_COLORS
 } from 'app/components/Chart/utils';
```
ivarnakken, line 6: Does it really format like that 👀

ivarnakken (approved):

app/components/Chart/BarChart.tsx
eikhr: Oh god, that's horrible. I might have not run prettier 😬

norbye (approved):

PR #3385, opened 2022-12-23
Fix typescript errors with data on window object

eikhr:
Fix typescript errors when accessing f.ex. `window.__IS_SSR__`


LudvigHz (approved):
LGTM :shipit:

PR #3386, opened 2022-12-23
Fix some TypeScript types

eikhr:
- Wrong import of Node instead of ReactNode
- Some other ReactNode / ReactElement related errors
- Pages route

Also I removed `prop-types` and disabled the eslint prop-types rule. It was causing some errors and as far as I can tell it is redundant when we use typescript.

### DIFF
--- a/app/components/Form/RadioButtonGroup.tsx
+++ b/app/components/Form/RadioButtonGroup.tsx
@@ -1,11 +1,11 @@
 import { Children, cloneElement } from 'react';
 import styles from './RadioButtonGroup.css';
-import type { Node } from 'react';
+import type { ReactElement } from 'react';
 
 type Props = {
   name: string;
   label: string;
-  children: Node;
+  children: ReactElement | ReactElement[];
 };
 
 const RadioButtonGroup = ({ name, label, children }: Props) => {
--- a/app/reducers/pages.ts
+++ b/app/reducers/pages.ts
@@ -1,5 +1,6 @@
 import { uniqBy, sortBy, groupBy } from 'lodash';
 import { createSelector } from 'reselect';
+import type { User } from 'app/models';
 import { selectMembershipsForGroup } from 'app/reducers/memberships';
 import createEntityReducer from 'app/utils/createEntityReducer';
 import { Page } from '../actions/ActionTypes';
@@ -13,6 +14,12 @@ export type PageEntity = {
   content: string;
   comments: Array<number>;
   picture: string;
+  picturePlaceholder: string;
+  logo?: string;
+  logoPlaceholder?: string;
+  membershipsByRole: Record<string, { user: User }[]>;
+  text?: string;
+  name?: string;
 };
 export default createEntityReducer({
   key: 'pages',
@@ -29,8 +36,7 @@ export const selectPageBySlug = createSelector(
 );
 export const selectPages = createSelector(
   (state) => state.pages.byId,
-  (pagesBySlug, pageSlug) =>
-    Object.keys(pagesBySlug).map((slug) => pagesBySlug[slug])
+  (pagesBySlug) => Object.keys(pagesBySlug).map((slug) => pagesBySlug[slug])
 );
 export const selectPagesForHierarchy = (category: string) =>
   createSelector(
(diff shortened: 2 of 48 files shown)


app/routes/admin/OverviewRoute.tsx
```diff
@@ -6,19 +6,15 @@ import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
...
 import groups from './groups';
 
-const OverviewRoute = ({
-  children,
-  currentUser,
-  loggedIn,
-  match,
-}: {
-  children: any;
+interface Props {
```
ivarnakken, line 9: I feel like we should be consist of our use of `interface`/`type`.

app/reducers/pages.ts
```diff
@@ -13,6 +13,12 @@ export type PageEntity = {
   content: string;
   comments: Array<number>;
   picture: string;
+  picturePlaceholder: string;
+  logo?: string;
+  logoPlaceholder?: string;
+  membershipsByRole: any;
+  text: any;
+  name: any;
```
ivarnakken, line 21: Preferably type this correctly now that you're at it.

app/routes/bdb/utils.tsx
```diff
@@ -153,7 +153,7 @@ export const getContactedStatuses = (
   });
   return contacted;
 };
-export const ListNavigation = ({ title }: { title: Node }) => (
```
ivarnakken, line 156: Is `title` ever anything else besides a string? It's wrapped by a `h1` tag.

app/routes/pages/components/PageDetail.tsx
```diff
@@ -31,11 +31,18 @@ type Props = {
   selectedPage: any;
   currentUrl: string;
   selectedPageInfo: PageInfo;
-  PageRenderer: (arg0: { page: any }) => Node;
+  PageRenderer: PageRenderer;
   pageHierarchy: Array<HierarchySectionEntity>;
   loggedIn: boolean;
 };
 
+interface PageRendererProps {
```
ivarnakken, line 39: See comment above.

ivarnakken (changes requested):
💯 Good stuff

app/routes/bdb/utils.tsx
eikhr: Yup, in the add company page the title is a textfield.

app/routes/admin/OverviewRoute.tsx
eikhr: Do you have a specific preference? I usually use interface whenever I can, but I rarely use any interface-specific features.

ivarnakken: I feel like we usually use `type` on props here, so I guess I prefer that in terms of consistency.

app/routes/bdb/utils.tsx
ivarnakken: well that's dumb but okay 😬

app/reducers/pages.ts
```diff
@@ -13,6 +13,12 @@ export type PageEntity = {
   content: string;
   comments: Array<number>;
   picture: string;
+  picturePlaceholder: string;
+  logo?: string;
+  logoPlaceholder?: string;
+  membershipsByRole: Record<string, { user: any }[]>;
```
ivarnakken, line 19:
```suggestion
  membershipsByRole: Record<string, { user: User }[]>;
```

ivarnakken (changes requested):

app/components/Form/RadioButtonGroup.tsx
```diff
@@ -1,11 +1,11 @@
...
 import styles from './RadioButtonGroup.css';
-import type { Node } from 'react';
+import type { ReactElement } from 'react';
 
 type Props = {
   name: string;
   label: string;
-  children: Node;
+  children: ReactElement | ReactElement[];
 };
```
LudvigHz, line 9: Just an FYI, you can also use `PropsWithChildren` utility type from React for this.

LudvigHz (approved):
LGTM :shipit:

app/components/Form/RadioButtonGroup.tsx
eikhr: Not really, this component can only have `ReactElement`s as children. `ReactNode` or `PropsWithChildren` would allow f.ex. `string`s.

PR #3407, opened 2023-01-04
Improve the table component

ivarnakken:
I believe it was in great need of a retouch.

---

[Improve color palette](https://github.com/webkom/lego-webapp/commit/922c8ead20dfeda10b02ba8a934e830c04f3b82d)

I think we should support far more shades of each color. I'm going to change the other colors as well and go more thoroughly through our webapp, but that should be a seperate PR. Here I've only changed colors used by the event table (gray, red, green, blue and orange).

---

[Give the table component a retouch](https://github.com/webkom/lego-webapp/commit/07ec000d6d9354373aac3f49a1e131d6f0fc08d9) 

This is mostly in regard to the table used on the event details page.

- Fix issue with `event.isPriced` being undefined and causing the "paid column" to vanish. 
   See https://github.com/webkom/lego/pull/3155
- Align cells in the body and header. However, not all cells should be centered imo.,
  e.g. cells with text such as the usernames or feedbacks/allergies column as I
  believe those need all the space they can get.
- Remove the "Klassetrinn" and "Status" column from the "Avmeldte" table.
  "Klassetrinn" was not showing and "status" is not needed.
- Replace "Meld av" button with an icon, and utilize a confirm modal.
  Saves space.
- Status pills in the "Status" column instead of icons. Does not save
  space, but looks good.
- Remove the ability to sort on the users' names as I believe that's not
  needed and saves us precious space.
- Remove the ability to sort on pools when there's only 1 pool.
- Align checkboxes to their labels in the filter dropdown.
- Do not show presence status when there's more than a day until event start or
  if someone has been set to present.
- General styling here and there. I've tried to cut down on space where I can.

I didn't really touch the "feedbacks/allergies" column because I'm planning on tackling that whole thing in a separate PR.

### Before
Screenshot 2023-01-04 at 21 22 21

### After
Notice that the dropdown to sort on pools is removed, since there's only one pool.

Screenshot 2023-01-06 at 14 17 00

Screenshot 2023-01-06 at 14 15 40

### Before
Screenshot 2023-01-07 at 23 31 36

### After
Notice that the user column has been explicitly set to not be centered. It looks bad because the name + username combo here is shorter than it usually is. Might not look too bad centered here though, so I might change that later. I'm going to tackle this specific table soon (gonna add an edit button), so I'll do it then.
Screenshot 2023-01-07 at 23 34 07

### Before
Screenshot 2023-01-07 at 23 26 18

### After
Notice here that the table does not use the `<Table />` component, but is still somewhat affected by the global cascading table conf. Will do a seperate PR and clean up stuff like this.
Screenshot 2023-01-07 at 23 23 31

---

Resolves ABA-197

### DIFF
--- a/app/components/Button/Button.css
+++ b/app/components/Button/Button.css
@@ -1,5 +1,52 @@
 @import url('~app/styles/variables.css');
 
+:root {
+  --button-color: var(--color-white);
+  --button-box-shadow: 0 1px 0 rgba(27, 31, 36, 10%),
+    inset 0 1px 0 rgba(255, 255, 255, 3%);
+  --button-background: var(--color-light-gray-4);
+  --button-background-hover: var(--color-light-gray-3);
+  --button-border: #30404d77;
+  --button-border-hover: #394b59aa;
+
+  --button-dark-background: var(--color-red-2);
+  --button-dark-background-hover: var(--color-red-1);
+  --button-dark-border: #bfccd633;
+  --button-dark-border-hover: #ced9e044;
+  --button-dark-box-shadow-focus: #db373744;
+
+  --button-danger-box-shadow-focus: #db373744;
+
+  --button-success-background: var(--color-green-7);
+  --button-success-background-hover: var(--color-green-6);
+  --button-success-border: #a7b6c233;
+  --button-success-border-hover: #bfccd644;
+  --button-success-box-shadow-focus: #2c974b44;
+}
+
+[data-theme='dark'] {
+  --button-color: var(--color-black);
+  --button-box-shadow: 0 0 transparent, inset 0 0 transparent;
+  --button-background: var(--color-mono-gray-2);
+  --button-background-hover: var(--color-mono-gray-4);
+  --button-border: #a7b6c277;
+  --button-border-hover: #bfccd6aa;
+
+  --button-dark-background: var(--color-red-1);
+  --button-dark-background-hover: var(--color-red-2);
+  --button-dark-border: #a7b6c233;
+  --button-dark-border-hover: #bfccd644;
+  --button-dark-box-shadow-focus: #ff737344;
+
+  --button-danger-box-shadow-focus: #ff737344;
+
+  --button-success-background: var(--color-green-4);
+  --button-success-background-hover: var(--color-green-3);
+  --button-success-border: #bfccd633;
+  --button-success-border-hover: #ced9e044;
+  --button-success-box-shadow-focus: #2da44e44;
+}
+
 .button {
   display: inline-flex;
   align-items: center;
--- a/app/components/Poll/index.tsx
+++ b/app/components/Poll/index.tsx
@@ -183,6 +183,7 @@ class Poll extends Component<Props, State> {
                             {ratio < 18 && (
                               <span
                                 style={{
+                                  padding: '5px',
                                   marginLeft: '2px',
                                 }}
                               >
--- a/app/routes/events/components/EventAdministrate/Administrate.css
+++ b/app/routes/events/components/EventAdministrate/Administrate.css
@@ -24,33 +24,71 @@
   grid-template-columns: repeat(5, 20%);
 }
 
-.webkomIcon {
-  font-size: 1.1em;
-  background: #e20d13;
-  color: var(--color-white);
+.pill {
+  margin: 0 auto;
   display: inline-block;
-  border-radius: 60px;
-  padding: 0.1em 0.2em;
+  text-transform: uppercase;
+  padding: 0.25rem 0.375rem;
+  font-size: 0.7rem;
+  font-weight: bold;
+  border-radius: 1rem;
+  letter-spacing: 0.6px;
+  line-height: 1;
+  box-shadow: rgb(0, 0, 0 / 5%) 1px 2px 5px 0;
+  align-items: center;
+  align-self: center;
+}
+
+.webkomPill {
+  background: linear-gradient(
+    90deg,
+    rgba(255, 0, 0, 100%) 0%,
+    rgba(255, 154, 0, 100%) 10%,
+    rgba(208, 222, 33, 100%) 20%,
+    rgba(79, 220, 74, 100%) 30%,
+    rgba(63, 218, 216, 100%) 40%,
+    rgba(47, 201, 226, 100%) 50%,
+    rgba(28, 127, 238, 100%) 60%,
+    rgba(95, 21, 242, 100%) 70%,
+    rgba(186, 12, 248, 100%) 80%,
+    rgba(251, 7, 217, 100%) 90%,
+    rgba(255, 0, 0, 100%) 100%
+  );
+  color: white;
+}
+
+.greenPill {
+  background-color: var(--color-green-2);
+  color: var(--color-green-7);
 }
 
 .greenIcon {
-  font-size: 1.5em;
-  color: green;
+  font-size: 1.4em;
+  color: var(--color-green-6);
 }
 
-.orangeIcon {
-  font-size: 1.5em;
-  color: var(--color-orange-3);
+.orangePill {
+  background-color: var(--color-orange-2);
+  color: var(--color-orange-7);
 }
 
 .questionIcon {
-  font-size: 1.5em;
-  color: var(--color-blue-1);
+  font-size: 1.4em;
+  color: var(--color-blue-6);
+}
+
+.bluePill {
+  background-color: var(--color-blue-2);
+  color: var(--color-blue-6);
 }
 
 .redIcon {
-  font-size: 1.5em;
-  color: '#C24538';
+  font-size: 1.4em;
+  color: var(--color-red-3);
+}
+
+.consentIcon {
+  font-size: 1.15em;
 }
 
 .center {
@@ -77,18 +115,13 @@ a.transparent:hover {
   border-bottom: 1px solid var(--color-mono-gray-3);
 }
 
-.presenceIcons {
-  display: flex;
-  flex-direction: row;
-  align-content: space-around;
-  margin: 5px;
-}
-
-.cell {
-  margin: 2px;
-}
-
 .consents {
   display: flex;
   justify-content: center;
+}
+
+.unregisterIcon {
+  justify-content: center;
+  cursor: pointer;
+  color: var(--color-red-3);
 }
--- a/app/routes/events/components/EventAdministrate/RegistrationTables.tsx
+++ b/app/routes/events/components/EventAdministrate/RegistrationTables.tsx
@@ -1,6 +1,7 @@
 import cx from 'classnames';
 import { Component } from 'react';
 import { Link } from 'react-router-dom';
+import Flex from 'app/components/Layout/Flex';
 import Table from 'app/components/Table';
 import Time from 'app/components/Time';
 import Tooltip from 'app/components/Tooltip';
@@ -40,8 +41,8 @@ type Props = {
     registrationId: ID,
     paymentStatus: EventRegistrationPaymentStatus
   ) => Promise<any>;
-  handleUnregister: (registrationId: ID) => void;
-  clickedUnregister: ID;
+  handleUnregister: (registrationId: ID) => Promise<void>;
+  showPresence: boolean;
   showUnregister: boolean;
   event: Event;
   pools: Array<EventPool>;
@@ -52,8 +53,8 @@ const GradeRenderer = (group: { name: string }) =>
     <Tooltip content={group.name}>
       <span>
         {group.name
-          .replace('Kommunikasjonsteknologi', 'Komtek')
-          .replace('Datateknologi', 'Data')}
+          .replace('klasse Kommunikasjonsteknologi', 'komtek')
+          .replace('klasse Datateknologi', 'data')}
       </span>
     </Tooltip>
   );
@@ -65,31 +66,60 @@ const getPoolName = (pools, poolId) => {
   return pool && pool.name;
 };
 
+type RegistrationPillProps = {
+  status: string;
+  reason: string;
+  className: string;
+};
+
+const RegistrationPill = ({
+  status,
+  reason,
+  className,
+}: RegistrationPillProps) => {
+  if (reason.length === 0)
+    return (
+      <div className={cx(styles.pill, className)}>
+        <span>{status}</span>
+      </div>
+    );
+
+  return (
+    <Tooltip content={reason}>
+      <div className={cx(styles.pill, className)}>
+        <span>{status}</span>
+      </div>
+    </Tooltip>
+  );
+};
+
 const getRegistrationInfo = (pool, registration) => {
   const registrationInfo = {
-    reason: 'Venteliste',
-    icon: cx('fa fa-clock-o fa-2x', styles.orangeIcon),
+    status: 'Venteliste',
+    reason: '',
+    className: styles.orangePill,
   };
 
   if (registration.adminRegistrationReason !== '') {
-    registrationInfo.icon = cx('fa fa-user-secret', styles.greenIcon);
+    registrationInfo.className = styles.bluePill;
 
     if (registration.createdBy !== null) {
       if (
         hasWebkomGroup(registration.user) &&
         hasWebkomGroup(registration.createdBy)
       ) {
-        registrationInfo.icon = cx('fa fa-power-off', styles.webkomIcon);
+        registrationInfo.className = styles.webkomPill;
         registrationInfo.reason = `Webkompåmeldt av ${registration.createdBy.username}: ${registration.adminRegistrationReason}`;
       } else {
+        registrationInfo.status = 'Påmeldt';
         registrationInfo.reason = `Adminpåmeldt av ${registration.createdBy.username}: ${registration.adminRegistrationReason}`;
       }
     } else {
       registrationInfo.reason = `Adminpåmeldt: ${registration.adminRegistrationReason}`;
     }
   } else if (pool) {
-    registrationInfo.reason = 'Påmeldt';
-    registrationInfo.icon = cx('fa fa-check-circle', styles.greenIcon);
+    registrationInfo.status = 'Påmeldt';
+    registrationInfo.className = styles.greenPill;
   }
 
   return registrationInfo;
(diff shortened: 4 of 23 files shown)


app/components/Button/Button.css
```diff
@@ -1,5 +1,52 @@
 @import url('~app/styles/variables.css');
 
+:root {
```
ivarnakken, line 3: Figured it was best to move this from `variables.css` to here since this is the only file where it is used.

app/styles/globals.css
```diff
@@ -76,29 +76,32 @@ i {
 }
 
 table {
+  line-height: 1.3;
```
ivarnakken, line 79: We should transition over to `line-height: 1.3` across the entire webapp (since it looks much much better), but that's a lot of work, so I'm doing it in chunks.

app/styles/globals.css
```diff
@@ -76,29 +76,32 @@ i {
...
   border-collapse: collapse;
 }
 
-th {
-  padding: 5px 0;
-  border-top: 1px solid var(--color-mono-gray-1);
-  border-bottom: 1px solid var(--color-mono-gray-1) !important;
+thead {
```
ivarnakken, line 83: This should ideally be in `Table.css`, but since there are tables in our webapp that don't the  `<Table /> ` component, this would have to do for now. I'm gonna fix later.

app/routes/events/components/EventAdministrate/Administrate.css
```diff
@@ -24,33 +24,72 @@
...
+  font-weight: bold;
+  border-radius: 1rem;
+  letter-spacing: 0.6px;
+  line-height: 1;
+  box-shadow: rgb(0 0 0 / 5%) 1px 2px 5px 0px;
+  align-items: center;
+  align-self: center;
+}
+
+.webkomPill {
```
ivarnakken, line 43: :eyes:

amalieeh:
I like it! Looks more modern. 
Just wondering about two things:
1) The red and green "påmeldt". Is it if the person has showed up or not? Is it yellow if it's undecided? In this example it says "0/4 har møtt opp", so then I don't understand...
2) "Dato". What is that date? Could it be called "påmeldingsdato" or "oppmøtedato" or something more explaining?

ivarnakken:
Thanks, @amalieeh! ❤️ 

1. I think you are mistakenly comparing the "status" and "til stede" columns. The red "påmeldt" pill indicates that a user has been admin registered (which you can compare to the before image). I know it's not obvious, but there's a tooltip to explain it further (and I'd argue the previous icon didn't indicate it any better). 
Since the event starts in more than a day, the "til stede" column is not showing in order to save space. The design of the "til stede" column is pretty much the same as before though, except a slight change in color, alignment and size (😅).

3. I agree that "Dato" is not very clear, but it's the date a person registered. I thought about changing it, but couldn't come up with a word that conveyed it nicely while also saving space. Any suggestions? "påmeldingsdato" feels a bit long.

amalieeh:
1 Okay okay, now I get it! It's just that when it's red, I feel like something's wrong... I know you said you'd fix colours later, but what do you think of swapping the waitlist colour and the admin register? :) 
2 Yeah, you're right. And we can have faith that our best and brightest abakuler can understand the date by logical thinking.

LudvigHz:
Awesome work! This really makes the whole page look a lot better and more up to date.

Some quick thoughts:

- Contrast. The contrast as it is now is too low. It's straining for the eyes to look at, especially for those of us with astigmatism and other visual impairments. This goes both for the background, text color, pill text and background and table color stripes.
- The pills. The pills are really cool, but in this case it feels like it's not the right tool for the job. A pill is nice to mark attributes on an object, but here the column represents an enum of sorts. The previous one was more legible and immediately clear (green for registered, yellow for waiting, red for unregistered and further made clear and distinguished with the icons. Yes you can have a tooltip, but if you need a tooltip now to explain what it is that is justification enough that this is worse imo. Usability over looks here.
- user name sort. In the screenshot there is space and it's a feature I use quite a lot at least. I don't care that much but I thought it was a nice to have.


Other than that it looks really cool so great job! With a few touch-ups I think this will be hard to make any better! 🥇

ivarnakken:
@LudvigHz 

1. The attached screenshot is not showing off its true colors. However, I changed them up a bit, and they are now all within [a good contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast). The pills are "enhanced" on dark theme, but "minimum" on light. Everything else is now enhanced. Also, it looks much better with a different font, since the one we use is very thin at bold weight. So I guess that's another reason to want to change our font.

2. There was never any "red" icon for unregistered people. I don't see how an icon is any better. For example, the icon used for admin registered users does not convey it in any better way. I never added the tooltip, it was there to begin with. In my opinion, the usability is now better, as the contrast of red and green is better than two similar green icons.

3. I can't image when you would need a name sorter. There's not really much space to work with, as there will usually be a "Til stede", "Samtykke" and "Betaling" columns. Nevertheless, I'd say the problem is not just space, but the fact that too many features make the header crowded and unclear.

ivarnakken:
@amalieeh 

1. I'm not a fan of switching the colors. As you say, red may imply danger, and I don't want to convey that being on a waiting list is any danger. So, it's kind of bad either way. Also, the waiting list icon has always been orange, so perhaps it's nice to keep it that way. I tried a green pill with the "Adminpåmeldt" text, but that took up too much space.

2. 🤞🏼

erlingfn:
Would it be possible to have the pills say "Adminpåmeldt" instead of just "Påmeldt" with red. I think that would make it much clearer, especially for people who seldom uses this route.

app/routes/events/components/EventAdministrate/Administrate.css
LudvigHz: Maybe move the pill into another component? Also, now we have the pills used for tags and pills used here? What's the difference?

LudvigHz:
>     1. The attached screenshot is not showing off its true colors. However, I changed them up a bit, and they are now all within [a good contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast). The pills are "enhanced" on dark theme, but "minimum" on light. Everything else is now enhanced. Also, it looks much better with a different font, since the one we use is very thin at bold weight. So I guess that's another reason to want to change our font.

Cool, if you want to share an updated screen please do!


>     2. There was never any "red" icon for unregistered people. I don't see how an icon is any better. For example, the icon used for admin registered users does not convey it in any better way. I never added the tooltip, it was there to begin with. In my opinion, the usability is now better, as the contrast of red and green is better than two similar green icons.

Ah right, I agree that it should be something. Having it blank is kinda weird :smile:.
But on the subject of the admin registered thing. I still do think the icons are way more discoverable and clear. It might be that the admin registered thing is throwing me off though. I definitely think it should not be red. Either use another color, or other text, or text + icon. Just something else, cause right now it just looks really strange.

>  I'd argue this conveys things better, as the red color sticks out more.

I disagree. It shouldn't stick out this much. It should be clear that they are also "registered" but it should just be a smaller difference that is easy to see which indicates that they are "admin" registered. So I would argue that the "main" distinguishing feature of both registered types should be identical, and a secondary feature should indicate the "admin registered part" (the opposite of the current, as the color is more eye-catching than the text)

app/routes/events/components/EventAdministrate/Administrate.css
ivarnakken: Well, they are sort of different in the way you interact with them. I thought about "merging" them together though, but that will require more work, so I'll probably create a new PR.

ivarnakken: The tags will be affected by the color changes I'm going to make, so I'll do it then. I imagine next week.

ivarnakken:
> I disagree.

@LudvigHz 
Yes, the red is very eye-catching, but should it not? The only people that _should_ see this table is the people in charge of the event, and I'd argue it's important for them to see things like this. I image it's just something you'll get used to. Like you say, you want both registered type to be identical in terms of their main feature, but they share the same text? The only difference is the secondary feature, like you say, the color, and I can't think of a different color to use than red. Red feels the most fitting. Green, orange and red.

eikhr:
I agree that the red "Påmeldt" pill for admin-registered is confusing. I feel like the obvious mental mapping of the colors is that green is registered, red is not registered, and yellow is waiting list as it is in between. I don't really see a reason to remove the icons, but if you really want to I would propose using blue instead of red for admin-registered. This would make it stand out and avoid the confusion of beeing in the green-red scale.

ollfkaih:
I think the number and unregistering columns have too much margin on either side considering their actual width is only a few pixels. Looks good overall, though.  
Also, maybe the text in the user and feedback columns should be centered as to not look so misaligned.

ivarnakken:
> I think the number and unregistering columns have too much margin on either side considering their actual width is only a few pixels.

Yea, I know. I tried, but it's the table itself that chooses the colspans, and I wanted to keep things as little hacky as possible. I could e.g. give each column a colspan prop or something, but I wanted the component to be easy to use (less config) and as scalable as possible to other use cases. Most other tables don't have that number column.  Also, the unregistering column will be much smaller once the "Tilbakemelding" column fills up with text and the "Samtykke" and "Betaling" columns show up.

> Also, maybe the text in the user and feedback columns should be centered as to not look so misaligned.

It looks worse in the images because there's no feedback or allergies, but on a realistic event, these will be filled, and they need all the space they can get. Like I mentioned, I didn't want to do anything with that particular column since I'm planning on tackling it in a different PR (moving the allergies away from the table etc.).

mattiastofte:
I'm a huge fan of these changes and I think the pills look great! I think it's important as other have mentioned to explain the coloring system, although it's seems that you've taken care of this with the tooltip. 

I was wondering if maybe the header could only have rounded corners in the upper left and right corners? It just feels a bit disconnected/in contrast with the rows under as they are completely square. Either way very nice work!

ivarnakken:
> I was wondering if maybe the header could only have rounded corners in just the upper left and right corners? It just feels a bit disconnected/in contrast with the rows under as they are completely square.

@mattiastofte I tried and I think it looks better with the border radius all around. I know it feels a bit disconnected, but that's kinda the look I was going for as well. We can iterate on it later as well, but right now I would love to get this in so I can continue working. Better to be agile here than wait a long time for it to be perfect the first time.

erlingfn:
> @erlingfn, like I told Amalie, that would take up a lot of space and cause the user's name column to be smaller, which I find to be of more importance. I believe that, if you really need to know, you'll figure it out some way or another, e.g. through the tooltip. It's just a matter of getting used to it. It was the same case for the detective icon that was there before, it conveys in no way that the person as been admin registered. I'd argue this conveys things better, as the red color sticks out more.

I disagree on several points. First and foremost, for this page I believe clarity is far more important than aesthetics, therefore it should immediately be clear if someone is admin registered or not, regardless of whether the user has used the registrations page before. Secondly, while the detective icon previously used was not very clear, it made it very simple to understand that it was not a standard registration and should be looked at. I do not believe a pill with a different colour is distinct enough to show this.

ivarnakken:
> regardless of whether the user has used the registrations page before

So what you're saying is that someone who's never visited the page before would immediately  understand what the detective icon means?

> it made it very simple to understand that it was not a standard registration and should be looked at.

A blue pill stands out far more that a green icon with a subtle change, so by your logic it indicates better that it is not a standard registration and should be looked at? Try to scroll through a list of 200 attendes and see what stands out the most. I've tried.

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
```diff
@@ -169,14 +200,29 @@ export class RegisteredTable extends Component<Props> {
...
       pools,
     } = this.props;
+
+    const gradeColumn = {
+      title: 'Klassetrinn',
+      dataIndex: 'user.grade',
+      render: GradeRenderer,
+      sorter: (a, b) => {
+        if (a.user.grade && b.user.grade) {
+          if (a.user.grade.name > b.user.grade.name) return 1;
```
erlingfn, line 215: I don't know that much about js sorting internals, but it might be beneficial to return 0 on equal grades so it always keeps its original. It already does in my browser, so this is really nitpicky but got to use algdat for something right :shrug:

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
```diff
@@ -230,16 +274,15 @@ export class RegisteredTable extends Component<Props> {
         title: 'Dato',
         dataIndex: 'registrationDate',
         render: (date) => (
-          <Tooltip content={<Time time={date} format="DD.MM.YYYY HH:mm:ss" />}>
-            <Time time={date} format="DD.MM.YYYY" />
+          <Tooltip content={<Time time={date} format="DD/MM/YYYY HH:mm:ss" />}>
+            <Time time={date} format="DD/MM/YYYY" />
```
erlingfn, line 278: Since this page is in norwegian we should use . for dates https://www.sprakradet.no/sprakhjelp/Skriveregler/Dato/

erlingfn (approved):
Good work, it looks really nice! We can, as always, argue about design changes in eternity, but this is a step in the right direction so do what you think is best.

PS: Please do not change from . to / for dates as that is incorrect in norwegian and would make me very sad.

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: Didn't know this was a thing, but okay I'll change it. Slashes are much better in my opinion though 😬

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: I didn't want to touch any of this logic, but sure I can add that.

app/components/Poll/index.tsx
```diff
@@ -183,6 +183,7 @@ class Poll extends Component<Props, State> {
                             {ratio < 18 && (
                               <span
                                 style={{
+                                  padding: '5px',
                                   marginLeft: '2px',
```
LudvigHz, line 187: Isn't this superfluous with the above change?

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
```diff
@@ -65,31 +66,60 @@ const getPoolName = (pools, poolId) => {
   return pool && pool.name;
 };
 
+type RegistrationPillProps = {
+  state: string;
+  reason: string;
```
LudvigHz, line 71: I would prefer if these were a bit more self-explanatory. Especially state is confusing in the react world

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
```diff
@@ -190,32 +236,30 @@ export class RegisteredTable extends Component<Props> {
         title: 'Bruker',
         dataIndex: 'user',
         search: true,
+        notCentered: true,
```
LudvigHz, line 239: I know this isn't you but this is just stupid 🙃

LudvigHz (approved):
Looks good!

Although I am very much for icons >> pills.
And I see you've changed stuff elsewhere as well. Would be nice og you could add some illustrations of those :)

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: Kinda stupid, but also very useful. Not all columns should be centered.

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: Do you have an example? I find `reason` to be good, but I can agree `state` might not be the best. At the same time, the pills do have an x amount of states, so it sort of makes sense. But ye, in the react world, might not be the best.

app/components/Poll/index.tsx
ivarnakken: It's not the same thing though? This was updated to make the alignment better, cause I thought it was affected by the change in `line-height`, but turns out that it was bad to begin with. Doesn't hurt to improve though.

ivarnakken:
> And I see you've changed stuff elsewhere as well. Would be nice og you could add some illustrations of those :)

I added some more now.

ivarnakken:
I'm merging this in tomorrow morning unless I hear anything. Like I've said, it's better to be agile here than wait a long time for it to be perfect the first time. I need stuff here so I can continue iterating.

SmithPeder (approved):

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
LudvigHz: I.e. `registrationState` is a better name here.

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
LudvigHz: The point was just that `notCentered=false === centered=true`

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: @LudvigHz What about `status`? I feel like `registrationInfo.registrationState` is a bit much.

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: Ye, but the default should be centered, so this just makes for less config.

LudvigHz: Ah i thought this was here from before. But I see now that you changed it to the "not" version. Then I can't let it pass. Having any boolean prop/arg beginning with "not" is a huge nono. Please change to centered and adjust the default accordingly.

LudvigHz (changes requested):
Change the `notCentered` prop. And imo make the reg pill prop names more self-documenting

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
LudvigHz: Status is better :+1:

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken: What? You want the default to be not centered? Passing `centered: false` is dumb. But whatever, I'm tired of this.

LudvigHz (approved):

PR #3416, opened 2023-01-08
Improve all modals

ivarnakken:
Hah, you thought I was done!?

---

[Improve the general modal component](https://github.com/webkom/lego-webapp/commit/7211258db5a9036d9ac4968f609554c61b704c40) 

Only in terms of styling. It's now smaller, yet prettier.

---

[Utilize the back prop on gallery pages](https://github.com/webkom/lego-webapp/commit/28a0ed70444fc89f5cc6b60c9bf6ff2ff8d547bb) 

Also move the routing hack to this navigation link. The hack ensures
that the user ends up where it left off in the y-axis.
  
[Improve the image upload modal](https://github.com/webkom/lego-webapp/commit/39c73b4b0b5e1a9be1a90bd8c080a33d042de248) 

Mostly just styling changes.

[Make upload area react to change](https://github.com/webkom/lego-webapp/commit/69c5affc3a26d165b383b0b94133d17bf51b3793) 

It will give feedback when it's focused, or accepts/rejects a file.

Also update some colors.

### Before
Unnecessary big and all the way at the top.

<p align="center">
	
</p>

### After
Images are not correctly scaled here, so these are in reality much smaller than before and centered in the middle of the screen.

<p align="center">
	
	
</p>

Notice the color feedback on the border. It works the same for all image uploaders, e.g. the one for changing your profile picture. The flickering in the middle of the video is me opening finder, relax.

<video src="https://user-images.githubusercontent.com/69514187/211218355-c1796967-89e1-40c8-a95a-28650d8a8775.mov"></video>

---
  
[Improve the attendance modal](https://github.com/webkom/lego-webapp/commit/a8b69dc49e059485651a513e6e4f8882e9615958) 

- Add a search field
- General styling improvements. It now further resembles the webapp's table.

### Before
This modal is huuuge, and ugly.

<p align="center">
	
</p>

### After
<p align="center">
	
	
</p>
---

[Use shared styling for interest group member list](https://github.com/webkom/lego-webapp/commit/cf2215fb87ddb8011fae539211a5df64bb433e16) 

The list now resembles the attendance modal. It uses the same shared
styling and has been given a search field as well.

---
  
[Improve the confirm modal](https://github.com/webkom/lego-webapp/commit/464a84202512b89afcb528f37ecda1e8a5889a59) 

Not only does it look better, but it now better
conveys that an action is destructive (when of type `danger`).

### Before
<p align="center">
	
</p>

### After
Difficult to notice here, but the modal is centered in the middle of the screen, instead of all the way at the top. Looks much better.

<p align="center">
	
	
</p>

---
  
[Clean up the gallery picture modal](https://github.com/webkom/lego-webapp/commit/b1dbc85fb653949be918126e34eb01c85803960c) 

Changes had to be made to "support" the new global modal changes.

---
  
 Resolves ABA-198

### DIFF
--- a/app/components/Modal/ConfirmModal.css
+++ b/app/components/Modal/ConfirmModal.css
@@ -1,39 +1,11 @@
-.modal {
-  position: fixed;
-  top: 0;
-  bottom: 0;
-  left: 0;
-  z-index: 2;
-  right: 0;
+.warningIcon {
+  color: var(--color-orange-6);
 }
 
-.backdrop {
-  position: fixed;
-  background: rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 80%);
-  top: 0;
-  bottom: 0;
-  left: 0;
-  right: 0;
+.dangerTitle {
+  color: var(--color-red-2);
 }
 
-.content {
-  position: fixed;
-  max-width: 680px;
-  top: 100px;
-  left: 0;
-  right: 0;
-  margin: 0 auto;
-  background: var(--color-white);
-  border-radius: 5px;
-  padding: 40px;
-  outline: none;
-  max-height: 80vh;
-  overflow-y: auto;
-}
-
-.closeButton {
-  position: absolute;
-  top: 10px;
-  right: 10px;
-  font-size: 20px;
+.errorMessage {
+  color: var(--color-red-2);
 }
--- a/app/components/Modal/Modal.css
+++ b/app/components/Modal/Modal.css
@@ -1,12 +1,3 @@
-.modal {
-  position: fixed;
-  top: 0;
-  bottom: 0;
-  left: 0;
-  z-index: 2;
-  right: 0;
-}
-
 .backdrop {
   position: fixed;
   background: rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 80%);
@@ -22,28 +13,37 @@ html[data-theme='dark'] .backdrop {
 }
 
 .content {
+  line-height: 1.3;
   position: fixed;
-  max-width: 680px;
-  top: 20px;
-  left: 0;
-  right: 0;
-  margin: 0 auto;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  max-width: 425px;
+  width: 100%;
+  max-height: calc(100vh - 300px);
+  padding: 35px;
   background: var(--color-white);
-  border-radius: 5px;
-  padding: 40px;
   outline: none;
-  max-height: calc(100vh - 40px);
   overflow-y: auto;
   z-index: 4;
+  /* stylelint-disable indentation */
+  box-shadow: 0 5px 10px
+      rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 12%),
+    0 6px 6px rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 15%);
+  border: 1px solid rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 9%);
+  border-radius: 1rem;
 }
 
 .closeButton {
   position: absolute;
-  top: 7px;
-  right: 10px;
-  font-size: 42px;
+  top: 1rem;
+  right: 1rem;
   padding: 0;
-  color: var(--color-black);
+  color: var(--color-dark-mono-gray-3);
+
+  &:hover {
+    color: var(--color-dark-mono-gray-4);
+  }
 }
 
 .confirmContainer {
--- a/app/components/Upload/UploadImage.css
+++ b/app/components/Upload/UploadImage.css
@@ -1,22 +1,10 @@
 @import url('~app/styles/variables.css');
 
-.container {
-  border-radius: inherit;
-  height: inherit;
-  width: inherit;
-}
-
 .modal {
-  box-shadow: 0 4px 4px rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 5%);
-  border: 1px solid rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 9%);
-  border-radius: 5px;
-}
-
-.uploadContainer {
-  position: absolute;
-  border-radius: inherit;
-  height: inherit;
-  width: inherit;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 20px;
 }
 
 .dropArea {
@@ -25,70 +13,67 @@
   width: inherit;
   z-index: 0;
   position: absolute;
-  background-color: gainsboro;
   cursor: pointer;
 }
 
-html[data-theme='dark'] .dropArea {
-  background-color: var(--color-mono-gray-1);
-}
-
 .inModalUpload {
-  height: 320px;
-  width: 600px;
+  width: 325px;
+  height: 200.861px;
+  border-radius: 1rem;
 }
 
 .placeholderContainer {
   overflow: hidden;
   border-radius: inherit;
   height: inherit;
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
   width: inherit;
   position: absolute;
 }
 
 .placeholderTitle {
-  background: rgba(var(--rgb-max), var(--rgb-max), var(--rgb-max), 80%);
-  border-radius: 5px;
-  padding: 0 20px;
+  background: rgba(var(--rgb-max), var(--rgb-max), var(--rgb-max), 70%);
+  padding: 10px 40px;
   text-align: center;
 }
 
+.inModalUpload .placeholderTitle {
+  background: none;
+}
+
 .image {
-  border-radius: inherit;
-  height: inherit;
   width: 100%;
+  height: inherit;
   object-fit: cover;
   background: white;
+  border-radius: inherit;
 }
 
 .cropper {
-  height: 400px;
   width: 100%;
 }
 
-.footer {
-  margin-top: 20px;
-}
-
 .previewRow {
+  width: 325px;
   height: 52px;
-  margin: 5px 0;
+  color: var(--color-light-gray-5);
 }
 
-.previewRowImage {
-  height: 50px;
+.previewImage {
   width: 80px;
+  height: 50px;
   object-fit: cover;
+}
+
+.fileName {
+  width: 180px;
+  font-size: 14px;
+  font-weight: 500;
 }
 
 .removeIcon {
   cursor: pointer;
 
   &:hover {
-    color: var(--lego-red-color);
+    color: var(--color-red-3);
   }
 }
--- a/app/components/Upload/UploadImage.css
+++ b/app/components/Upload/UploadImage.css
@@ -14,6 +14,10 @@
   z-index: 0;
   position: absolute;
   cursor: pointer;
+
+  &:focus {
+    outline: none;
+  }
 }
 
 .inModalUpload {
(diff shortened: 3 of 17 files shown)


app/components/UserAttendance/AttendanceModal.css
```diff
@@ -1,63 +1,135 @@
...
+
+.list {
+  position: relative;
   width: 100%;
-  justify-content: center;
+  min-height: 100px;
+  height: 80%;
+  overflow-y: auto;
+  /* stylelint-disable indentation */
+  background:
```
ivarnakken, line 50: Kind of proud of this one. It adds a shadow to the top and bottom as you scroll. So, no shadow at the top when at the top and no shadow at the bottom when at the bottom.

app/routes/interestgroups/components/InterestGroupMemberList.tsx
```diff
@@ -1,43 +1,36 @@
...
       break;
 
     default:
       break;
   }
-
-  return (
-    <span>
-      <span className={roleStyle}>{user.fullName} </span>
-      <span className={styles.suffix}>{title}</span>
```
ivarnakken, line 38: I found this suffix to be redundant. First of all, it's in the tooltip, but the leaders have a gold shadow around their name and and gold star next to it. Also, their name is in the list of "leaders" next to the modal trigger.

app/routes/photos/components/GalleryDetailsRow.tsx
```diff
@@ -26,7 +26,7 @@ const GalleryDetailsRow = ({
 
       {gallery.takenAt && (
         <span className={styles.detail}>
-          {moment(gallery.takenAt).format('ll')}
+          <Time time={gallery.takenAt} format="DD.MM.YYYY" />
```
ivarnakken, line 29: Happy, @erlingfn? 😉

app/styles/variables.css
```diff
@@ -110,12 +110,15 @@
   --color-green-8: #108944;
   --color-green-9: #06371b;
 
-  --color-red-0: #fcc5d8;
   --color-red-1: #a82a2a;
   --color-red-2: #c23030;
   --color-red-3: #f31260;
   --color-red-4: #f4256d;
-  --color-red-5: #fcc5d8;
+  --color-red-5: #f881ab;
```
ivarnakken, line 117: Goes without saying, this does not affect anything else. As I've said, I'm gonna go through this more thoroughly, but I really needed some more shades of red.

app/components/Modal/ConfirmModal.css
```diff
@@ -1,39 +1,11 @@
...
-  outline: none;
-  max-height: 80vh;
-  overflow-y: auto;
-}
-
-.closeButton {
-  position: absolute;
-  top: 10px;
-  right: 10px;
-  font-size: 20px;
```
LudvigHz, line 38: less css = better css :heart_eyes:. Love to see it!

app/components/Modal/Modal.css
```diff
@@ -22,28 +13,37 @@ html[data-theme='dark'] .backdrop {
...
   outline: none;
-  max-height: calc(100vh - 40px);
   overflow-y: auto;
   z-index: 4;
+  /* stylelint-disable indentation */
+  box-shadow: 0 5px 10px
+      rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 12%),
+    0 6px 6px rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 15%);
+  border: 1px solid rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 9%);
+  border-radius: 1rem;
```
LudvigHz, line 34: I guess you could put the `var(--rgb-min), var(...), var(...)` in a var to make this a bit more compact and readable (sacrificing verbosity) but not that big of a deal.

app/components/Modal/Modal.css
```diff
@@ -22,28 +13,37 @@ html[data-theme='dark'] .backdrop {
...
   position: fixed;
-  max-width: 680px;
-  top: 20px;
-  left: 0;
-  right: 0;
-  margin: 0 auto;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  max-width: 425px;
```
LudvigHz, line 21: is this max-width an effective width? If not will content that does not fit in this clip the modal box?

app/components/Upload/UploadImage.css
```diff
@@ -25,70 +13,92 @@
...
 }
 
+.inModalUpload .placeholderContainer {
+  background-color: var(--color-gray-6);
+  border: 3px dashed var(--color-blue-6);
+}
+
+.focused .placeholderContainer {
+  border: 3px dashed var(--color-orange-6);
+}
```
LudvigHz, line 44: IMO having it default blue threw me off quite a bit as a blue border signifies focus (in FF and in chrome). I think having it just gray default and blue on focus is a better option.

LudvigHz (approved):
Looks really good! fantastic job :1st_place_medal: 

Design wise my only comments are:
1. The blue border around the image dropzone as mentioned.
2. The pool switcher. I feel like this should be a general component as well, but the issue to me is that at first sight I thought it was three buttons mashed together with no spacing. So it looks broken (until you maybe realize the idea). I suggest making it clear that it a complete, single component with an item selected. One idea could be to make the "active" selection a bit shorter, so there's a gap on the top and bottom between the red and gray. This is what most other similar components around the web do to avoid these problems and I think it looks good.

Other than that this is a really nice and subtle (but huge) improvement!

app/components/Modal/Modal.css
ivarnakken: Yes, but you can override it. The photo gallery view modal does this. It's just that it needs a specified width, and might as well set a max as good practice or something.

app/components/Upload/UploadImage.css
ivarnakken: Well, yes but no. It's not like blue is the default of the web at everyone must abide by it. I think a very big problem with our current webapp is our use of gray and shades of it. Having everything be gray makes it look ugly, and we should be better at incorporating more colors where we can. I'm very much against having a gray border, as that will be a gray border around a slightly different gray box, around a slightly different gray bigger box on top of a gray background .. you see what I mean.

app/components/Modal/Modal.css
ivarnakken: hmm I see your point, but then I wouldn't get to specify the alpha value. I don't think I can do `rgba(var(--x), y%)`. At least I believe I've tried before, but I can try again tomorrow.

app/components/Upload/UploadImage.css
LudvigHz: A blue border is literally how FF and chrome shows focus unless overridden. So I would argue that most should not change it unless really required (i.e. pretty much never).

I do understand your point, but I also think that we should be careful with adding _too_ much color everywhere. I agree that a lot of color is nice, but with color shading in gray and white it can also look really nice. I might be more of a minimalist though so it is subjective. But, I'll use GitHub as an example here. They have a tint to their backgrounds and buttons, but still it is gray boxes with gray backgrounds and gray separator lines, gray pills etc. and it works really well. It's pleasant and not distracting. Color is nice, but don't overdo it.

app/components/Modal/Modal.css
LudvigHz: It doesn't make more sense to have it be a `min-width` then? If it should have a "default" but allow to grow according to the content?

app/components/Modal/Modal.css
LudvigHz: You can do
```css
--rgb-mins: var(--rgb-min), var(--rgb-min), var(--rgb-min);

border: 1px solid rgba(var(--rbg-mins), 9%);
```
I think.
but yeah it might be worse :shrug:

app/components/Upload/UploadImage.css
ivarnakken: I see your point, but I don't feel like this is overdoing it. Also, I feel like this isn't just any normal border, but rather something to signify a drop area and something to interacted with, if that makes any sense, so I feel like a color is more greeting and therefore quite fitting.

I think it might look "too much" here because it's a screenshot which keeps the surrounding elements out. In reality, this modal is very small, and the color much less eye-catching together with a background with a bunch of stuff going on. Try it and see what you think. I think that might help.

app/components/Modal/Modal.css
ivarnakken: ah yes, that'll work. I'm sort of  on the fence on this one. Both aren't really ideal, but I guess your option is more compact, so I think I'll change it to that.

app/components/Modal/Modal.css
ivarnakken: Not really, since the contents of the modals often use `width: 100%` and that'll look for the maximum width, not the minimum.

SmithPeder:
Looks really good!

Except this pink! This pink does not spark joy 🥲 
image

ivarnakken:
> Except this pink! This pink does not spark joy 🥲

I don't agree with you, but it will most likely be changed very soon anyway. Besides, this is a modal that basically no one ever sees ..

app/components/Upload/UploadImage.css
LudvigHz: I tested this out, and it's worse than in the screenshot IMO. The effect that my mind immediately draws the conclusion that the box is selected is more apparent. I tried removing the border completely and it actually looks quite nice, maybe better I think. A slight box-shadow also works, but the cleaner  and simpler one is more appealing to me.

ivarnakken: I still prefer a dotted border. I think the dots signify that it's a "zone" better. What you show is just a very plain box which from a distance don't seem to have any purpose whatsoever. I don't see the whole argument of it looking "focused". Firstly, I don't think so. Secondly, how does that even matter?

ivarnakken: It's not like Google never adds a blue border to anything because that's reserved? Just as a quick example, see the document I sent earlier.

app/styles/variables.css
eikhr: These new "reds" are very pink. I suppose light red colors look a bit pink, but it might make more sense to just call them pinks

eikhr (approved):
Huge improvement!
I don't mind the blue border in the image upload, it adds a bit of color and draws attention to the most important element in the modal.
I love the new registration list, a searchbar is a great addition. I don't really like the pool switcher though, I think it would look better if the middle button didn't have rounded corners. Like this:
Screenshot 2023-01-12 at 14 32 58
Also not a fan of the pink danger header, pink doesn't really indicate danger.

app/components/UserAttendance/AttendanceModal.tsx
```diff
@@ -47,11 +50,13 @@ const Tab = ({ name, index, activePoolIndex, togglePool }: TabProps) => (
 
 type State = {
   pools: Pool[];
+  filter: string;
 };
 
 class AttendanceModal extends Component<Props, State> {
   state = {
     pools: [],
+    filter: '',
```
erlingfn, line 59: I would change the name of this variable to `search` or something similar since that is what it actually does. `filter` makes it seem like it is filtering out users in certain grades and such, not displaying users with a matching name.

app/routes/photos/components/GalleryPictureModal.tsx
```diff
@@ -324,14 +323,16 @@ export default class GalleryPictureModal extends Component<Props, State> {
               </Dropdown>
             </Flex>
           </Content>
+
           <Flex className={styles.pictureContainer}>
             <ProgressiveImage
               key={picture.id}
               src={picture.file}
-              alt="some alt"
```
erlingfn, line 331: How was this approved

erlingfn (approved):
Looks really good!

app/components/UserAttendance/AttendanceModal.tsx
ivarnakken: It's literally filtering out user from the list though, and I chose it since `filter` is also used in the table column search.

app/routes/photos/components/GalleryPictureModal.tsx
ivarnakken: After years of looking at our codebase, I can confirm, we used to have quite low standards.

ivarnakken:
The pinkish red color is now gone everybody :/

PR #3449, opened 2023-01-18
Rewrite overview to functional component & refactor

LudvigHz:
Rewrite overview component to be functional, refactor a lot of stuff and fix a bunch of types.

Also :star_struck: :star_struck::

[Screencast from 2023-01-18 13-10-12.webm](https://user-images.githubusercontent.com/42850232/213168103-0b5b0373-9c87-4f11-9bbb-9f8b35c2e9eb.webm)

### DIFF
--- a/app/components/RandomQuote/RandomQuote.css
+++ b/app/components/RandomQuote/RandomQuote.css
@@ -20,3 +20,8 @@
 .fetchNew {
   margin-bottom: 5px;
 }
+
+.rotateIcon {
+  transform: rotate(360deg);
+  transition: 1s;
+}
--- a/app/reducers/events.ts
+++ b/app/reducers/events.ts
@@ -10,7 +10,7 @@ import createEntityReducer from 'app/utils/createEntityReducer';
 import joinReducers from 'app/utils/joinReducers';
 import mergeObjects from 'app/utils/mergeObjects';
 import { Event } from '../actions/ActionTypes';
-import type { PhotoConsent } from '../models';
+import type { Event as EventType, PhotoConsent } from '../models';
 
 export type EventEntity = {
   id: number;
@@ -160,7 +160,7 @@ export default createEntityReducer({
   mutate,
 });
 
-function transformEvent(event) {
+function transformEvent(event: EventType) {
   return {
     ...event,
     startTime: moment(event.startTime),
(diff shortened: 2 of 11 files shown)


app/routes/overview/components/Overview.tsx
```diff
@@ -16,217 +16,227 @@ import styles from './Overview.css';
...
+    [articles, articlesToShow, pinned]
+  );
+
+  const pinnedComponent = pinned && (
+    <div>
+      <Pinned item={pinned} url={itemUrl(pinned)} meta={renderMeta(pinned)} />
+    </div>
+  );
+
+  const readMe = (
```
ivarnakken, line 84: 🤮

app/reducers/events.ts
```diff
@@ -192,7 +195,7 @@ export const selectUpcomingEvents = createSelector(selectEvents, (events) =>
   events.filter((event) => event.isUsersUpcoming)
 );
 export const selectSortedEvents = createSelector(selectEvents, (events) =>
-  events.sort((a, b) => a.startTime - b.startTime)
+  [...events].sort((a, b) => a.startTime.unix() - b.startTime.unix())
```
ivarnakken, line 198: What's the point of `[...events]`? Is `events` not already an array of `Event`?

app/components/RandomQuote/RandomQuote.css
```diff
@@ -20,3 +20,8 @@
 .fetchNew {
   margin-bottom: 5px;
 }
+
+.rotateIcon {
+  transform: rotate(360deg);
+  transition: 1s;
```
ivarnakken, line 26: I feel like a `cubic-bezier` would be very fitting, in which it speeds up at the middle and slows down at the end. 

On that note, we should standardize a `cubic-bezier` for the webapp as a variable. I feel like animations across the webapp is very "random", with different animations, speeds and lengths. I'll write it up as a todo, so you can keep it like this if you want.

ivarnakken (approved):
Looks good! ⭐ A "better" overview component is going to be very handy!

I like the tiny, yet impactful, animation touch! 😍

app/reducers/events.ts
LudvigHz: JS `.sort` is in-place. Both being annoying as mutating arrays when doing things like this is a bad idea IMO, and the fact that `events` already has a type, and the result of the `.sort()` is something else, and TS can't magically change the type of a variable. So creating a copy ensures that TS infers the correct type.

app/components/RandomQuote/RandomQuote.css
LudvigHz: I thought so too, but the animation was just sooo satisfying by default so I saw no reason to create more code for little to no benefit.

app/reducers/events.ts
ivarnakken: ah yes that makes perfect sense, I forgot that it was in-place.

app/components/RandomQuote/RandomQuote.css
ivarnakken: Yea, I'll agree that it looks smooth

PR #3454, opened 2023-01-19
Make table CSS more specific

ivarnakken:
# Description

This was overlooked since it is not an issue in the local build ..

image

# Testing

I have not been able to test it, since it is not a bug locally 🤔, but this will most likely fix it. I have, however, tested that it does not break the notifications table, so it can't get any worse ..


LudvigHz (approved):
This is why I hate `!important` :/

ivarnakken:
> This is why I hate !important :/

It was actually no longer needed, so I removed it 😅

PR #3467, opened 2023-01-24
Standardize animations

mattiastofte:
Closes #ABA-209 by adding a set of premade global variables for easing and linear animations. Both have three variants, in fast (100ms), medium (200ms) and slow (400ms) variants. Linear has the slightest ease to make it just a tad smoother, but this is super easy to change if people don't like that idea.

There are really no reason to attach screenshots as I've just replaced the animations with the most similar animation variable so everything should basicly look the same.

### DIFF
--- a/app/components/Image/Image.css
+++ b/app/components/Image/Image.css
@@ -20,7 +20,5 @@
 }
 
 .finalImage {
-  transition-property: filter, opacity;
-  transition-duration: 500ms;
-  transition-timing-function: cubic-bezier(0, 0, 0.2, 1);
+  transition: filter, opacity var(--easing-slow);
 }
--- a/app/components/Image/Image.css
+++ b/app/components/Image/Image.css
@@ -20,5 +20,5 @@
 }
 
 .finalImage {
-  transition: filter, opacity var(--easing-slow);
+  transition: filter var(--easing-slow), opacity var(--easing-slow);
 }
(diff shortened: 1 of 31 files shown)


AdrianAndersen (approved):
Looks good to me. Makes it a lot more tidy with the animations 🚀

app/components/UserValidator/Validator.css
```diff
@@ -16,7 +16,7 @@
   border-radius: 5%;
   opacity: 0;
   visibility: hidden;
-  transition: opacity 0.6s, visibility 0.6s;
+  transition: opacity, visibility var(--linear-slow);
```
ivarnakken, line 19: I'm pretty sure you need to add the variable to `opacity` as well.

app/components/Image/Image.css
```diff
@@ -20,7 +20,5 @@
 }
 
 .finalImage {
-  transition-property: filter, opacity;
-  transition-duration: 500ms;
-  transition-timing-function: cubic-bezier(0, 0, 0.2, 1);
+  transition: filter, opacity var(--easing-slow);
```
ivarnakken, line 23: See previous comment.

ivarnakken (changes requested):
Love this - exactly what I imagined! 😍 This improves the technical debt by a lot in my opinion! Just fix up the comment I've made. I _miiiiiiiiiiiiiiiiiiight_ be wrong, but then at least check it out.

ivarnakken (approved):

PR #3468, opened 2023-01-24
Fix Event Statistics Overflow Issues

AdrianAndersen:
# Description

Event Statistics on mobile had major overflow issues. This PR makes the stats look prettier on mobile.

# Result

### Before
image

### After
image

# Testing

Tested with responsive mode in the web browser, down to a size of iPhone 8 Plus, which is a quite small phone nowadays. 

---

Resolves ABA-216

### DIFF
--- a/app/routes/events/components/Event.css
+++ b/app/routes/events/components/Event.css
@@ -143,3 +143,11 @@
   margin: 0 auto;
   border-bottom: 1px solid #6772e585;
 }
+
+.noRegistrationsText {
+  text-align: center;
+}
+
+.chartContainer {
+  overflow: hidden;
+}
--- a/app/routes/events/components/EventAttendeeStatistics.tsx
+++ b/app/routes/events/components/EventAttendeeStatistics.tsx
@@ -42,7 +42,7 @@ const PieChartWithLabel = ({
   return (
     <>
       <h4>{label}</h4>
-      <Flex alignItems="center">
+      <Flex alignItems="center" style={{ marginBottom: '3rem' }} wrap={true}>
         <DistributionPieChart
           dataKey="count"
           distributionData={distributionData}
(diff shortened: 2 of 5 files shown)


app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -42,7 +42,10 @@ const PieChartWithLabel = ({
   return (
     <>
       <h4>{label}</h4>
-      <Flex alignItems="center">
+      <Flex
+        alignItems="center"
+        style={{ flexWrap: 'wrap', marginBottom: '3rem' }}
+      >
```
ivarnakken, line 48: The `<Flex />` has a `wrap` prop.

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -265,7 +262,7 @@ const EventAttendeeStatistics = ({
       {registrations.length === 0 ? (
         <p style={{ textAlign: 'center' }}>Ingen er påmeldt enda.</p>
       ) : (
-        <>
+        <div style={{ overflow: 'hidden' }}>
```
ivarnakken, line 265: Not a fan of all this inline styling.

ivarnakken (changes requested):
Good stuff! 💯 Just fix up the minor comments, and let's get this in quick 🚄

app/routes/events/components/EventAttendeeStatistics.tsx
AdrianAndersen: ✅

app/routes/events/components/EventAttendeeStatistics.tsx
AdrianAndersen: Sure I don't mind. Moved it to a css file. 👍🏻

app/routes/events/components/Event.css
```diff
@@ -143,3 +143,11 @@
   margin: 0 auto;
   border-bottom: 1px solid #6772e585;
 }
+
+.noRegistrationsText {
+  text-align: center;
+}
+
+.chartContainer {
+  overflow: hidden;
```
erlingfn, line 152: Personally I find that this works better on mobile with `flex` and `flex-basis column`, so that the labels are centered, but this is very minor

LudvigHz (approved):

app/routes/events/components/Event.css
AdrianAndersen: I do not quite understand what you mean here. This overflow hidden here, is just to remove the ugly scroll bars that occur because of recharts has really a lot of unnecessary padding. Putting flex on this container would probably not do too much about the labels themselves, since they are nested in child components. Can you elaborate?

ivarnakken (approved):

PR #3471, opened 2023-01-24
Change meeting icon from calendar to people

eikhr:
# Description

Changed the meeing icons from a calendar to two people. I think the calendar icon is a bit confusing as it does not lead to the clandar page. The two people is not an ideal icon, but I think it is an improvement and can't really think of any icon that screams meeting.

I did consider something like this:
Screenshot 2023-01-24 at 23 02 54
but I don't think it looks quite right.

# Result

Before: 
Screenshot 2023-01-24 at 23 17 26
After:
Screenshot 2023-01-24 at 23 17 13

# Testing

### DIFF
--- a/app/components/Header/index.tsx
+++ b/app/components/Header/index.tsx
@@ -91,7 +91,7 @@ function AccountDropdownItems({
       <Dropdown.ListItem>
         <Link to="/meetings/" onClick={onClose}>
           Møteinnkallinger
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):

ingraso (approved):
Nice work! :rocket: I didn't find any other suitable icons either, so I think this one is the best option right now.

PR #3486, opened 2023-01-28
Fix bug where questionType selection does not work

LudvigHz:
# Description

Fix a bug where the questionType selector does not work due to incorrect reference to the value in
the redux-form state.

# Result

It works now

# Testing


ivarnakken (approved):
💯

PR #3487, opened 2023-01-28
Introduce the Inter font

ivarnakken:
# Description

Inter is a _very_ popular font-family. It gives the webapp a more modern
look by replacing Raleway and Roboto.

Changes were made to make it fit more, but I believe this is something
we'll have to iterate on multiple times to get right. No more uppercase
headers, and the webapp is less gray.

I'm considering introducing a "cooler" secondary font to use for headers,
but that can come later.

# Result

**Before**
image

**After**
The webapp is a lot less gray, which I find to be a great improvement.

image

# Testing

Should be no "breaking changes".


SmithPeder:
Very controversial change 🤝 After looking at "Inter" all day, I'm sick of it, but I guess Abakus isn't 🙈

Quick nitpick, I like the lower case headers, but I feel like they became a bit small? They kinda drown into the page?

LudvigHz (approved):
LGTM

haukkagu:
I love the new font! However, as you are saying, a different font for the headers would look better.

eikhr (approved):
Looks good

ivarnakken:
> They kinda drown into the page?

I kinda feel the opposite. They were previously very thin with a light gray color, which blended together with the background. But I kinda also see your point. We can iterate on it.

ollfkaih:
Looks much better, but overhørt should maybe have less font-weight.

PR #3488, opened 2023-01-29
Enhance the Card component

ivarnakken:
# Description

[Enhance the Card component](https://github.com/webkom/lego-webapp/pull/3488/commits/7bc069db346a3a7b62dec86c9a00b56dc8381b4c)

Also make all "cards" actually employ the component. The cards now have
a much more modern look and they all share the same styling.

---

[Iterate on the introduction of Inter](https://github.com/webkom/lego-webapp/pull/3488/commits/56dec53c7d4a6f5a0423424e408395d16891eaba)

Small styling changes here and there. Among other things, the background is now less "blue".

# Result

**Before**
image

**After**
image

**Before**
image

**After**
The cards looks much better here once the content background is gone.

image

# Testing

I've tested all use cases of the component.

---

Resolves ABA-225

### DIFF
--- a/app/components/Card/Card.css
+++ b/app/components/Card/Card.css
@@ -1,13 +1,7 @@
 .card {
-  background: var(--color-white);
-  border: 1px solid rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 9%);
+  background: var(--lego-card-color);
   padding: 20px;
-  border-radius: 3px;
-  flex: 1;
-}
-
-html[data-theme='dark'] .card {
-  background: inherit;
+  border-radius: 1rem;
 }
 
 .dark {
@@ -20,5 +14,6 @@ html[data-theme='dark'] .card {
 }
 
 .shadow {
-  box-shadow: 0 1px 4px rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 4%);
+  box-shadow: 0 2px 8px 2px rgba(104, 112, 118, 7%),
+    0 2px 4px -1px rgba(104, 112, 118, 4%);
 }
--- a/app/components/Footer/index.tsx
+++ b/app/components/Footer/index.tsx
@@ -130,7 +130,7 @@ const Footer = ({ loggedIn }: Props) => (
           Informasjonskapsler (cookies)
         </Link>
         <Link to="/pages/personvern/124-personvernserklring">
-          Personvernserklæring
+          Personvernerklæring
         </Link>
         <span>© {moment().year()} Abakus</span>
       </div>
--- a/app/components/Header/toggleTheme.css
+++ b/app/components/Header/toggleTheme.css
@@ -16,7 +16,7 @@
 }
 
 .moon:hover {
-  color: var(--color-blue-1);
+  color: var(--color-blue-8);
 }
 
 .sun:hover {
--- a/app/components/RandomQuote/RandomQuote.css
+++ b/app/components/RandomQuote/RandomQuote.css
@@ -13,9 +13,8 @@
 }
 
 .quoteText {
-  font-weight: 600;
+  font-weight: 500;
   font-size: 15px;
-  color: var(--color-dark-gray-3);
 }
 
 .quoteReactions {
@@ -32,5 +31,5 @@
 
 .rotateIcon {
   transform: rotate(360deg);
-  transition: 1s;
+  transition: var(--easing-slow);
 }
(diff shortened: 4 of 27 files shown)


app/components/RandomQuote/RandomQuote.css
```diff
@@ -13,9 +13,8 @@
 }
 
 .quoteText {
-  font-weight: 600;
```
ivarnakken, line 16: @ollfkaih I reduced the `font-weight` like you suggested in #3487

app/components/Card/Card.css
```diff
@@ -1,13 +1,8 @@
...
-  border: 1px solid rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 9%);
+  background: var(--lego-card-color);
   padding: 20px;
-  border-radius: 3px;
-  flex: 1;
-}
-
-html[data-theme='dark'] .card {
-  background: inherit;
+  border-radius: 1rem;
```
ivarnakken, line 4: This matches the `border-radius` of the newly improved modals.

app/components/Card/Card.css
```diff
@@ -20,5 +15,6 @@ html[data-theme='dark'] .card {
 }
 
 .shadow {
-  box-shadow: 0 1px 4px rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 4%);
+  box-shadow: 0 2px 8px 2px rgba(104, 112, 118, 7%),
+    0 2px 4px -1px rgba(104, 112, 118, 4%);
 }
```
LudvigHz, line 20: how is this on dark-mode now?

app/routes/overview/components/utils.tsx
```diff
@@ -9,16 +9,18 @@ import styles from './Overview.css';
...
 export const renderMeta = (item: Event | Article) => {
   const isEvent = item.eventType ? true : false;
+
+  let format =
+    moment().year() === moment(item.startTime).year()
+      ? 'DD. MMM'
+      : 'DD. MMM YYYY';
+  if (isEvent) {
+    format += ' HH:mm';
+  }
```
ivarnakken, line 19: The exact time the article was posted isn't that important. However, it's different for events because it's important to know when it starts.

LudvigHz (approved):
Awesome. This is a much much needed improvement :rocket:

LudvigHz:
IMO the borders were increased a _tad_ too much. I'd probably enjoy something in-between this and the current more.

app/components/Card/Card.css
ivarnakken: image

Not very visible, but I kinda prefer that on dark mode. A shadow isn't really needed. That's usually how it's done in the "industry".

ivarnakken:
> IMO the borders were increased a tad too much. I'd probably enjoy something in-between this and the current more.

I don't really agree. A `0.2rem` diff isn't going to do that much. I think this looks much more modern.

> Also, the contrast between the card and the background is a bit too small for my liking (on the frontpage as well). I think ether some more pronounced border or more difference in the card/bg color would make this a bit better

Well, yes, but no. You see, that's kind of the point. I tried a higher contrast, but that just looked off. It made the cards stand out way too much and they became too eye-catching, which is not its purpose. You want it to blend in nicely with the surroundings, while still separating its content from the rest.

app/components/Card/Card.css
ivarnakken: I'm planning on standardizing the box-shadows across the webapp soon.

app/components/Footer/index.tsx
```diff
@@ -130,7 +130,7 @@ const Footer = ({ loggedIn }: Props) => (
           Informasjonskapsler (cookies)
         </Link>
         <Link to="/pages/personvern/124-personvernserklring">
-          Personvernserklæring
+          Personvernerklæring
```
ivarnakken, line 133: Had to be done. Sorry. I'm not making a seperate PR for this.

app/components/RandomQuote/RandomQuote.css
ollfkaih: This is definitely better. I think the poll and quote components still look like they're not designed by the same people, but this improves the situation.

app/components/Header/toggleTheme.css
```diff
@@ -16,7 +16,7 @@
 }
 
 .moon:hover {
-  color: var(--color-blue-1);
+  color: var(--color-blue-8);
 }
```
ollfkaih, line 20: Great change, old was almost invisible on my screen

ollfkaih (approved):
Great work!

ivarnakken:
The contrast between the card and the background has been increased.

norbye (approved):

PR #3496, opened 2023-01-30
Fix HMR due to replaceUnlessLoggedIn using class

LudvigHz:
# Description

Don't know exactly what's wrong here, but changing the return component
in replaceUnlessLoggedIn to a function component seems to resolve the
issue.

# Testing

Works.


ivarnakken (approved):
💯 Holy crap this is much needed ...........

PR #3497, opened 2023-01-30
Set default Pill text color to black

haukkagu:
# Description
The Pill component looks ugly on dark mode since its white on light gray text. This PR hard codes the text color to be black on both dark and light mode.

# Result

Before:
image

After:
image

# Testing

I have tested it for all occurrences of the Pill component.

### DIFF
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -58,7 +58,6 @@ const Attendance = ({ event }) => {
       <Pill
         style={{
           marginLeft: '5px',
-          color: 'black',
           whiteSpace: 'nowrap',
         }}
       >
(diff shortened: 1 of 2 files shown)


ivarnakken:
I kind of prefer the lighter look of the pills 😬 I think they would benefit of more color, like this pills on the event administration table / tags.

haukkagu:
@ivarnakken I prefer it this way, but if the general consensus is that the lighter looks better, we can close this. Some pills, like for the event items, are already dark. Do you think it would look better to lighten them as well? It'd at least be more consistent.
image

haukkagu:
On closer inspection, I think that the problem is that the background for the Pill component is way too light on dark mode. The buttons have a darker background with white text, and they look fine on dark mode.

danielyanghansen:
Might I suggest to have the pill color palette vary based on light/dark mode? :eyes:

ivarnakken:
Fixed in #3518

PR #3504, opened 2023-02-01
Fix Helmet for Server Side Rendering

danielyanghansen:
# Description

The helmet component has been set after the `readyHtml(app)` now. This is just because helmetContext is based on the existing scripts and styles that 'readyHtml' produces.

# Result

Helmet should work now, with properly updated / overwritten `og` tags.

# Testing

- [X] I have (somewhat) thoroughly tested my changes.
- I have tested that different routes all successfully have inserted Helmet tags into Head

How testing was done:
Simply:
 - Build the components using `yarn build`
 - Run SSR frontend with 'yarn ssr' (local backend)
 - Inspect that different pages have their correct `<meta>` tags in `<Head>`

---

Resolves [ABA-181](https://linear.app/abakus-webkom/issue/ABA-181/fix-helmet)

### DIFF
--- a/server/pageRenderer.ts
+++ b/server/pageRenderer.ts
@@ -86,10 +86,10 @@ export default function pageRenderer({
   preparedStateCode = '',
 }: PageRendererProps = {}): string {
   const isSSR = app === undefined ? 'false' : 'true';
-  const { helmet } = helmetContext as FilledContext;
   const selectedTheme: string =
     (!isEmpty(state) && selectCurrentUser(state).selectedTheme) || 'auto';
   const { body, scripts, styles, links } = readyHtml(app);
+  const { helmet } = helmetContext as FilledContext;
   return `
     <!DOCTYPE html>
     <html data-theme=${serialize(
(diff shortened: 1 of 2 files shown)


server/ssr.tsx
```diff
@@ -1,6 +1,6 @@
 import * as Sentry from '@sentry/node';
 import { prepare } from '@webkom/react-prepare';
-import { HelmetProvider } from 'react-helmet-async';
```
ivarnakken, line 3: Unused import.

server/pageRenderer.ts
```diff
@@ -86,14 +86,14 @@ export default function pageRenderer({
...
   const isSSR = app === undefined ? 'false' : 'true';
-  const { helmet } = helmetContext as FilledContext;
   const selectedTheme: string =
     (!isEmpty(state) && selectCurrentUser(state).selectedTheme) || 'auto';
   const { body, scripts, styles, links } = readyHtml(app);
+  const { helmet } = helmetContext as FilledContext;
   return `
     <!DOCTYPE html>
     <html data-theme=${serialize(
-      selectedTheme === 'auto' ? 'light' : selectedTheme
```
ivarnakken, line 96: Please add some spacing.

ivarnakken (changes requested):
Awesome! 💯 💯 

Just some linting issues as well

PR #3510, opened 2023-02-02
Add models for data types fetched from LEGO

eikhr:
Adds typescript types for all entity types from LEGO with different types for the different serializers. They are in `app/store/models`. I have also added an Unknown<entityName> type for each entity which is simply a union of the different serializers, which is used f.ex. in the redux store where we can't know which of the types it is. 

I have also gone through the announcement, article and company (not bdb yet) pages and replaced the types with these new ones.

# Result

There should be no functional difference.

# Testing

---

### DIFF
--- a/app/routes/announcements/components/AnnouncementsCreate.tsx
+++ b/app/routes/announcements/components/AnnouncementsCreate.tsx
@@ -5,14 +5,17 @@ import Button from 'app/components/Button';
 import { ContentMain } from 'app/components/Content';
 import { Form, SelectInput, TextArea } from 'app/components/Form';
 import Flex from 'app/components/Layout/Flex';
-import type { ActionGrant, CreateAnnouncement } from 'app/models';
+import type { ActionGrant } from 'app/models';
 import { selectAutocomplete } from 'app/reducers/search';
+import type { DetailedAnnouncement } from 'app/store/models/Announcement';
 import styles from './AnnouncementsList.css';
 
 type Props = {
-  createAnnouncement: (arg0: CreateAnnouncement) => Promise<any>;
+  createAnnouncement: (
+    announcement: DetailedAnnouncement & { send: boolean }
+  ) => Promise<unknown>;
   actionGrant: ActionGrant;
-  handleSubmit: (arg0: (...args: Array<any>) => any) => void;
+  handleSubmit: (arg0: (...args: Array<any>) => any) => any;
   invalid: boolean;
   pristine: boolean;
   submitting: boolean;
--- a/app/store/models/Poll.d.ts
+++ b/app/store/models/Poll.d.ts
@@ -1,0 +1,25 @@
+import type { Dateish } from 'app/models';
+import type { ID } from 'app/store/models/index';
+import type { ContentTarget } from 'app/store/utils/contentTarget';
+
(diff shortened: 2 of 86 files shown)


app/store/models/entities.ts
```diff
@@ -0,0 +1,117 @@
...
+import type { UnknownUser } from 'app/store/models/User';
+import type { ID } from 'app/store/models/index';
+import type OAuth2Grant from './OAuth2Grant';
+
+export enum EntityType {
+  Announcements = 'announcements',
+  Articles = 'articles',
+  Comments = 'comments',
+  Companies = 'companies',
+  CompanyInterests = 'companyInterest', // Why the fuck is this not plural?
```
ivarnakken, line 41: 😆

app/store/models/SurveySubmission.d.ts
```diff
@@ -0,0 +1,9 @@
+import type { SurveyAnswer } from 'app/store/models/SurveyAnswer';
+import type { ID } from 'app/store/models/index';
+
+interface SurveySubmission {
```
ivarnakken, line 4: Is this used?

app/store/models/Poll.d.ts
```diff
@@ -0,0 +1,25 @@
...
+export default interface Poll {
+  id: ID;
+  createdAt: Dateish;
+  validUntil: Dateish;
+  title: string;
+  description: string;
+  options: PollOption[];
+  resultsHidden: boolean;
+  totalVotes: number;
+  comments: unknown; // TODO: is it even possible to comment on polls?
```
ivarnakken, line 20: Interesting.

ivarnakken (changes requested):
Awesome job!!!! 💯 💯 This is very satisfying and removes a ton of technical debt!

Besides my mostly picky comments, I'm not a big fan of having duplicate types, one in our `app/models.ts` and one in `app/store/models/*.d.ts`. I don't see why we need both. Anyways, fantastic work! And just so it's said, you don't need to resolve all of my comments.

eikhr:
I'm intending on removing all the types from `app/models.ts`, but they are still in use in most of the app so it will have to wait.

app/routes/announcements/components/AnnouncementsCreate.tsx
```diff
@@ -5,14 +5,17 @@ import Button from 'app/components/Button';
...
 import styles from './AnnouncementsList.css';
 
 type Props = {
-  createAnnouncement: (arg0: CreateAnnouncement) => Promise<any>;
+  createAnnouncement: (
+    announcement: DetailedAnnouncement & { send: boolean }
+  ) => Promise<unknown>;
   actionGrant: ActionGrant;
-  handleSubmit: (arg0: (...args: Array<any>) => any) => void;
+  handleSubmit: (arg0: (...args: Array<any>) => any) => any;
```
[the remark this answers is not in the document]
eikhr: one step forward, two back or something...

app/store/models/SurveySubmission.d.ts
eikhr: It will be

ivarnakken (approved):
We can iterate on the nitpicks later 💯

PR #3518, opened 2023-02-05
Improve our use of gray

ivarnakken:
# Description

Our gray color palette has been a mess for a long time. There's far too many shades, and our use of them feels more or less random. leading to pages looking different across the webapp. This standarizes our use of it, while also improving the color. The difference looks minimal, but the reduction in technical debt is major.

Not sure how to show this. As I've said, the visual changes are very minimal, but branch out and check it out for yourself. If the chosen color are not satisfactory, they can now easily be changed later due to everything being variables 🤩 

# Testing

Went through ALL use of "gray" ..

### DIFF
--- a/app/components/Banner/Banner.css
+++ b/app/components/Banner/Banner.css
@@ -19,7 +19,7 @@
 }
 
 .gray {
-  background-color: var(--color-dark-mono-gray-5);
+  background-color: var(--color-gray-4);
   color: var(--color-white);
 }
 
--- a/app/components/Button/Button.css
+++ b/app/components/Button/Button.css
@@ -4,10 +4,10 @@
   --button-color: var(--color-white);
   --button-box-shadow: 0 1px 0 rgba(27, 31, 36, 10%),
     inset 0 1px 0 rgba(255, 255, 255, 3%);
-  --button-background: var(--color-light-gray-4);
-  --button-background-hover: var(--color-light-gray-3);
-  --button-border: #30404d77;
-  --button-border-hover: #394b59aa;
+  --button-background: var(--color-gray-1);
+  --button-background-hover: var(--color-gray-2);
+  --button-border: #8c8c8c77;
+  --button-border-hover: #b6b6b6aa;
 
   --button-dark-background: var(--color-red-2);
   --button-dark-background-hover: var(--color-red-1);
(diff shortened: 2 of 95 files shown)


ivarnakken:
image

That's what I call, sexy.

haukkagu (approved):
Looks very good! The pills finally look good on dark mode. Personally though, I'd prefer if there was a higher contrast between the pills and the background. The pills doesn't stick out as much as before.

LudvigHz:
The only thing I could find here was the button color. I liked the blue tint we had there. Made them "pop" a bit more and be a bit more vibrant and interesting.

Also, in the administrate tables, the gray here is also more "brown" as the previous was a bit lighter and blueish. Minor preference for the current one.

| before      | after                   |
|---------------|-------------------------|
| image | image |
| image|  image |


Otherwise really nice :ok_hand:.

ivarnakken:
> The only thing I could find here was the button color. I liked the blue tint we had there. Made them "pop" a bit more and be a bit more vibrant and interesting.

I feel like the new buttons look better. Also, they align with our use of gray, which I feel is very important. By getting rid of the "mono" shades, we now have no "blueish" grays, which I thought you preferred?

> Also, in the administrate tables, the gray here is also more "brown" as the previous was a bit lighter and blueish. Minor preference for the current one.

I agree. Making it lighter now.

LudvigHz:
> I feel like the new buttons look better. Also, they align with our use of gray, which I feel is very important. By getting rid of the "mono" shades, we now have no "blueish" grays, which I thought you preferred?

I did think the old ones might have had just a bit too much blue, but the new ones are just a bit too brown as well. Some lighter color wouldn't hurt IMO. This is nitpicking though

mattiastofte:
Looks great! As you've mentioned this cleans up the color palette and standardizes the webapp, making it easier to stay consistent with the design. I also actually prefer the warmer grays. LGTM!

ivarnakken:
> On a sidenote: I think we should look at modernizing the buttons as a whole. Getting som JavaFX vibes.

Yes, they totally need to be iterated on. I added some more border-radius now, and reduced the border color, which kind of helped.

ivarnakken:
@LudvigHz
I made the administrate table and buttons a little less "brown".

ivarnakken:
Like I've said, if we're not happy with the chosen gray colors, they can very easily be changed later. This PR is mostly for standardizing everything and reducing technical debt.

LudvigHz (approved):
LGTM

PR #3523, opened 2023-02-06
Add Collapse Animation To LatestReadme

AdrianAndersen:
# Description
- refactor(LatestReadme): class component => functional
- feat(LatestReadme): animated collapse

Use the Collapse element from "react-select" to animate the LatestReadme component. This makes it much smoother to look at, as it does not jump from collapsed to expanded.

# Result

#### Before 
https://user-images.githubusercontent.com/26925695/216996880-0bac0748-3082-43c4-9b45-2950496f6725.mov

#### After
https://user-images.githubusercontent.com/26925695/216996393-b108e511-d472-44c2-8f95-140d57df5d77.mov

# Testing

---

Resolves ABA-251

### DIFF
--- a/app/routes/overview/components/LatestReadme.tsx
+++ b/app/routes/overview/components/LatestReadme.tsx
@@ -1,69 +1,50 @@
-import { Component } from 'react';
+import { useEffect, useState } from 'react';
 import Card from 'app/components/Card';
 import Icon from 'app/components/Icon';
 import { Image } from 'app/components/Image';
 import { Flex } from 'app/components/Layout';
 import { readmeIfy } from 'app/components/ReadmeLogo';
+import type { Readme } from 'app/models';
 import styles from './LatestReadme.css';
 
 type Props = {
-  expanded: boolean;
-  readmes: Array<Record<string, any>>;
-};
-type State = {
-  expanded: boolean;
+  expandedInitially: boolean;
+  readmes: Array<Readme>;
 };
 
-class LatestReadme extends Component<Props, State> {
-  state = {
-    expanded: this.props.expanded,
-  };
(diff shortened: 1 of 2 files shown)


ivarnakken (approved):
I'm not a fan of using third-party libraries for this as the animation feels off (especially considering [we _just_ standardized them](https://github.com/webkom/lego-webapp/pull/3467)). It does not conform with out animations and it is slow.

However, this is definitely better than nothing 🤩 . But we should write up a todo for removing `react-collapse` (it's only used one other place).

app/routes/overview/components/LatestReadme.tsx
```diff
@@ -1,69 +1,51 @@
...
-      </Card>
-    );
-  }
-}
+const LatestReadme = ({ readmes, expandedInitially }: Props) => {
+  const [expanded, setExpanded] = useState(expandedInitially);
+
+  useEffect(() => {
+    setExpanded(expandedInitially);
+  }, [expandedInitially]);
```
LudvigHz, line 21: Do you need this `useEffect`? In my mind it has no effect :thinking:

LudvigHz (approved):

app/routes/overview/components/LatestReadme.tsx
AdrianAndersen: In prod, it will not have any effect, since all the fetching is done server side before it reaches the client. However, when running this locally, the value changes as the fetching is happening. And for some reason, the component does not re render when the initial value changes as the fetches happen. Therefore, the initial value ends up wrong, if you do not include this.

ollfkaih (approved):
LGTM

PR #3531, opened 2023-02-07
Increase size of modal on smaller screens

norbye:
# Description

The new modal, although very pretty, was really small on smaller devices. This PR increases the height for screens that are shorter than 900px high, so that there is less whitespace on top and more visible contents.  
It is also in general made a little bit wider.

# Result

|  | Before | After |
:------------------:|:-----------------------------:|:-----------------------------:
iPhone 12 Pro (390x844) | image | image
Surface Duo (540x720)  | image | image
Screen longer than 900px | image | image

# Testing

Went through all the usages of the modal component I could find and checked that looked fine

---

Resolves ABA-258

### DIFF
--- a/app/routes/photos/components/GalleryPictureModal.css
+++ b/app/routes/photos/components/GalleryPictureModal.css
@@ -2,7 +2,7 @@
 
 .content {
   padding: 0;
-  max-width: 1100px;
+  width: var(--lego-max-width);
   max-height: calc(100vh - 100px);
 
   @media (max-width: 800px) {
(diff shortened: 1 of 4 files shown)


app/routes/photos/components/GalleryPictureModal.css
```diff
@@ -2,7 +2,7 @@
 
 .content {
   padding: 0;
-  max-width: 1100px;
+  width: 1100px;
```
ivarnakken, line 5: Should ideally use `var(--lego-max-width)` (which is equal to `1100px`), but oh well 🤷🏼

ivarnakken (approved):
Nice! I'm all for responsive web design 😛

eikhr (approved):

LudvigHz (approved):

PR #3532, opened 2023-02-07
Add admin info about how many unregistered has paid

jonasdeluna:
Add admin info about how many unregistered has paid if any, as well as add it to the list of unregistered people.
Hide payment count if 0.

Before:
image

After:
image

image

image


ivarnakken (approved):
Nice 💯 Good of you to add a check of whether anyone has actually paid before showing the counter!

PR #3537, opened 2023-02-08
Fix sidebar background color

haukkagu:
# Description
The text in the sidebar on mobile wasn't visible on dark mode due to white text on white background. I also changed the color of the sidebar button to red.

# Result
**Before:**
image
**After**
image

Sidebar button:
**Before**
image
image

**After**
image
image

# Testing

Not much to test here.

### DIFF
--- a/app/routes/pages/components/Sidebar.css
+++ b/app/routes/pages/components/Sidebar.css
@@ -100,7 +100,7 @@
   .sidebar {
     width: 100%;
     max-width: 100%;
-    background-color: #f6f6f6;
+    background-color: var(--lego-card-color);
     height: 100%;
   }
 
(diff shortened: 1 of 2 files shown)


app/routes/pages/components/Sidebar.css
```diff
@@ -100,7 +100,7 @@
   .sidebar {
     width: 100%;
     max-width: 100%;
-    background-color: #f6f6f6;
+    background-color: var(--color-white);
```
ivarnakken, line 103: Could we maybe do `var(--lego-card-color)` so that it matches our other components better? 😄 

It's the same on light mode, but slightly different on dark mode.

ivarnakken (approved):
Ooops wasn't aware of this! 😬 Thanks!

ingraso (approved):
Nice work! :rocket:

app/routes/pages/components/Sidebar.css
haukkagu: Fixed!

PR #3540, opened 2023-02-09
Rewrite CompanyDetail to make it look more like the webapp

ingraso:
# Description

The detailed view of a company really needed a retouch. Now it uses the same layouts as events and meetings, with the sidebar and compact information there. Also, the components for joblisting and events (on the user profile) are used, to ensure consistency.

# Result
The detailed view with sidebar and main sections separated. The extra compact version of the events, that is also used on the user profile, is used.
**Before**
Screenshot from 2023-02-09 17-09-16

**After**
Screenshot from 2023-02-09 17-10-14

The previous events now also use the same layout.
**Before**
Screenshot from 2023-02-09 17-10-44

**After**
Screenshot from 2023-02-09 17-10-54

An example of a company without any data.
**Before**
Screenshot from 2023-02-09 17-13-45

**After**
Screenshot from 2023-02-09 17-13-40

An example of a company with a long description.
**Before**
Screenshot from 2023-02-09 17-13-13

**After**
Screenshot from 2023-02-09 17-12-20

It still does not look particularly good on phones, but in my opinion it looks slightly better than the current view.
**Before**
Screenshot from 2023-02-09 17-32-50

**After**
Screenshot from 2023-02-09 17-32-10

# Testing

Tested by clicking around, both in light and dark mode, as well as looked at the design on both smaller and larger screen sizes.

---

Resolves ABA-263


app/routes/company/components/CompanyDetail.tsx
```diff
@@ -23,224 +27,138 @@ type Props = {
...
-        <p>{company.description}</p>
-      </div>
+      <ContentSection>
+        <ContentMain>
+          <i>
+            <DisplayContent content={company.description} />
+          </i>
+          <h3 className={styles.sectionHeader}>Kommende arrangementer</h3>
+          <ListEvents
+            events={upcomingEvents}
```
ingraso, line 102: This complains about the type of events. Haven't looked to much into it yet, so if anyone has any suggestions for a fix I'll gladly accept :innocent: 
 ```Type 'ListEvent[]' is not assignable to type 'Event[]'.ts(2322)```

app/routes/company/components/CompanyDetail.tsx
```diff
@@ -23,224 +27,138 @@ type Props = {
...
+              {viewOldEvents
+                ? 'Skjul tidligere arrangementer'
+                : 'Vis tidligere arrangementer'}
+            </Button>
+          )}
+          {viewOldEvents && (
+            <>
+              <h3 className={styles.sectionHeader}>Tidligere arrangementer</h3>
+              <ListEvents
+                events={oldEvents}
```
ingraso, line 119: Same as above, with the wrong type

app/routes/company/components/CompanyDetail.tsx
```diff
@@ -23,224 +28,138 @@ type Props = {
...
-              </a>
-            )
-          }
-        </td>
-      </tr>
-    ));
   return (
-    <Content>
+    <Content
+      banner={company.logo || company?.logo}
```
ivarnakken, line 84: What's going on here? Is this not the same as
```suggestion
      banner={company?.logo}
```

app/routes/company/components/CompanyDetail.tsx
```diff
@@ -23,224 +28,138 @@ type Props = {
...
+            ))
+          ) : (
+            <i>Ingen tilgjengelige jobbannonser</i>
+          )}
+        </ContentMain>
+        <ContentSidebar>
+          {companyInfo.map(
+            (info) =>
+              info.text && (
+                <span key={info.text} className={styles.iconWithText}>
```
ivarnakken, line 151: Could turn this into a `<Flex />` component since `styles.iconWithText` only adds flex styling.

ivarnakken (approved):
This is a huuuuuuge improvement! 😍😍 Awesome!!

app/routes/company/components/CompanyDetail.tsx
ingraso: Yeah, that's the same! Didn't look too thorough through the old code

haukkagu:
Looks a lot cleaner now!

LudvigHz:
Wow, this is awesome!! :star_struck: Sorely needed!!

Only thing I can comment on the design is maybe having a "show more" thing for the company text. I assume for companies like Netcompany with a lot of text you'd need to scroll through a lot of text to see events and joblistings. So the classic "text fade out" with a "Show more" button is probably a good idea :) Could be added later though.

app/routes/company/components/CompanyDetail.tsx
LudvigHz: The `ListEvent` type used here (in `Props`) (imported from `app/store/models/event` is incorrect. Change the type to `Event` from `app/models` and it should be correct

app/routes/company/components/CompanyDetail.tsx
```diff
@@ -1,19 +1,24 @@
...
-import { Image } from 'app/components/Image';
-import InfoBubble from 'app/components/InfoBubble';
+import { Flex } from 'app/components/Layout';
 import LoadingIndicator from 'app/components/LoadingIndicator';
 import NavigationTab from 'app/components/NavigationTab';
-import { jobType, Year } from 'app/routes/joblistings/components/Items';
-import EventItem from 'app/routes/overview/components/EventItem';
-import { renderMeta } from 'app/routes/overview/components/utils';
+import { JoblistingItem } from 'app/routes/joblistings/components/JoblistingList';
+import { ListEvents } from 'app/routes/users/components/UserProfile';
```
LudvigHz, line 17: Hmm, I'd prefer if you don't mix components from other routes. 

This component is also just a `map` that returns `EventItem` components from `app/components/EventItem`, so I'd use that directly. Or move the EventList  into app/components and import that here and in the userprofile.

app/routes/company/components/CompanyDetail.tsx
```diff
@@ -23,224 +28,138 @@ type Props = {
...
+                <Flex key={info.text}>
+                  <Icon name={info.icon} className={styles.infoIcon} />
+                  {info.link ? (
+                    <a href={info.text}>{company.name}</a>
+                  ) : (
+                    info.text
+                  )}
+                </Flex>
+              )
+          )}
```
LudvigHz, line 160: Could we also make the component used in events more generic and use that?

No biggie though.

LudvigHz (approved):

app/routes/company/components/CompanyDetail.tsx
ingraso: Great, thank you!

app/routes/company/components/CompanyDetail.tsx
ingraso: Sounds like a good idea, but I think it can be in a later PR :raised_hands:

ingraso:
> Only thing I can comment on the design is maybe having a "show more" thing for the company text. I assume for companies like Netcompany with a lot of text you'd need to scroll through a lot of text to see events and joblistings. So the classic "text fade out" with a "Show more" button is probably a good idea :) Could be added later though.

Agreed! I think that would be great in a later iteration :rocket:

ivarnakken:
@ingraso, is this still a WIP?

ivarnakken:
Btw, it would be nice if the "placeholder text" here had the same styling, but that can be fixed later :shipit: 
image

ingraso:
> @ingraso, is this still a WIP?

@ivarnakken yeah, gonna move the JoblistingList to a separate component or something else to not mix components from other routes as Ludvig commented.

Also, I want to do another iteration on the design etc., so regarding the font style I'll look at that in the next iteration.

ivarnakken:
> Also, I want to do another iteration on the design etc., so regarding the font style I'll look at that in the next iteration.

That's what I like to hear! 🚀

LudvigHz (approved):
LGTM

PR #3541, opened 2023-02-09
Fix event following bug by fetching follow-status through event object

eikhr:
# Description

Depends on https://github.com/webkom/lego/pull/3214

There is a bug where the webapp does not correctly load the following status of the current user if an event has a large number of followers. It is caused by pagination on the event follower endpoint, but I figured a better solution than loading all pages in the webapp was to just include the single relevant follow in the event object.

# Result

The bug is no longer there.

# Testing

I have tried following and unfollowing events multiple times. I have also tested that event detail page still works fine in logged out state.

---

Resolves https://github.com/webkom/lego/issues/2521

### DIFF
--- a/app/actions/EventActions.ts
+++ b/app/actions/EventActions.ts
@@ -389,6 +389,7 @@ export function unfollow(
         method: 'DELETE',
         meta: {
           id: followId,
+          eventId,
           errorMessage: 'Avregistering fra interesse feilet',
         },
       })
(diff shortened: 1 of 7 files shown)


LudvigHz (commented):
I mean, it's kind of a hack but if it works it works. 

Is `following` a parameter on the event object from the event detail API?

eikhr:
Yup, I added "following" as a field on `EventReadUserDetailedSerializer` in the backend.

I don't see it as a very hacky solution. Instead of loading every user who is following the event we just include the relevant following information directly in the event.

LudvigHz (approved):

ivarnakken (approved):
Nice! This looks solid!

PR #3548, opened 2023-02-10
Add reactions to meetings. Rewrote meeting from class component -> hook.

jonasdeluna:
image

### DIFF
--- a/app/routes/meetings/MeetingDetailRoute.tsx
+++ b/app/routes/meetings/MeetingDetailRoute.tsx
@@ -2,14 +2,18 @@ import qs from 'qs';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 import { deleteComment } from 'app/actions/CommentActions';
+import { fetchEmojis } from 'app/actions/EmojiActions';
 import {
   fetchMeeting,
   setInvitationStatus,
   answerMeetingInvitation,
   resetMeetingsToken,
 } from 'app/actions/MeetingActions';
+import { addReaction, deleteReaction } from 'app/actions/ReactionActions';
 import type { User } from 'app/models';
+import { selectEmojis } from 'app/reducers/emojis';
 import { selectMeetingById } from 'app/reducers/meetings';
+import type { ReactionEntity } from 'app/reducers/reactions';
 import withPreparedDispatch from 'app/utils/withPreparedDispatch';
 import MeetingDetailLoginRoute from './MeetingDetailLoginRoute';
 import MeetingAnswer from './components/MeetingAnswer';
@@ -22,7 +26,10 @@ const loadMeeting = (
     },
   },
   dispatch
-) => (loggedIn ? dispatch(fetchMeeting(meetingId)) : Promise.resolve());
+) =>
+  loggedIn
+    ? Promise.all([dispatch(fetchMeeting(meetingId), dispatch(fetchEmojis()))])
+    : Promise.resolve();
 
 const loadData = (props, dispatch): any => {
   const search = qs.parse(props.location.search, {
@@ -63,12 +70,15 @@ const mapStateToProps = (state, props) => {
   const showAnswer = Boolean(
     meetingsToken.response === 'SUCCESS' && action && token
   );
+  const emojis = selectEmojis(state);
   return {
     meetingsToken,
     user: props.currentUser,
     showAnswer,
     meeting,
     currentUser,
+    emojis,
+    fetchingEmojis: state.emojis.fetching,
   };
 };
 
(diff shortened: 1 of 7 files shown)


ivarnakken:
Cool! 💯 We can use this to register who wants pizza for our meetings haha! 

Thought; could we perhaps move it down to the comment section? I feel like the sidebar is only for general info + admin stuff.

jonasdeluna:
> 
Can try this :+1:

jonasdeluna:
image
Like this?

app/routes/meetings/MeetingDetailRoute.tsx
```diff
@@ -63,12 +70,16 @@ const mapStateToProps = (state, props) => {
...
+  const emojis = selectEmojis(state);
   return {
     meetingsToken,
     user: props.currentUser,
     showAnswer,
     meeting,
     currentUser,
+    emojis,
+    fetchingEmojis: state.emojis.fetching,
+    fetching: state.meetings.fetching,
```
ivarnakken, line 82: This seems to be unused.

ivarnakken (approved):
Awesome! 🍕

jonasdeluna:
^removed & placed emojis at the bottom
```
fetchingEmojis: state.emojis.fetching,
fetching: state.meetings.fetching,
```

app/routes/meetings/MeetingDetailRoute.tsx
```diff
@@ -22,7 +26,10 @@ const loadMeeting = (
     },
   },
   dispatch
-) => (loggedIn ? dispatch(fetchMeeting(meetingId)) : Promise.resolve());
+) =>
+  loggedIn
+    ? dispatch(fetchMeeting(meetingId), fetchEmojis())
```
LudvigHz, line 31: Are you sure you can dispatch like this? According to the docs, this is not an option.
You could rather do a `Promise.all([dispatch(fetchmeeting(...)), dispatch(fetchemojis())]`

LudvigHz: Also, why are you fetching emojis here and passing `fetchEmojis` into the reaction picker?

app/routes/meetings/MeetingDetailRoute.tsx
ivarnakken: @jonasdeluna You removed `fetchingEmojis`, but that's actually used. I was only referring to `fetching`. Sorry for not being more specific 😅

app/routes/meetings/MeetingDetailRoute.tsx
jonasdeluna: Hmm I'm not entirely sure actually. It was written that way in the ArticleDetailRoute.tsx file, but I'll try the promise.all instead.

jonasdeluna:
image

Changed placement back to the sidebar (?)

ivarnakken:
> I also liked having the reactions someplace more visible than under the content, as I feel they "disappear" if you have just a little text in the meeting

@LudvigHz 
The same goes for comments though. I feel like "interactions" like comments and reactions should be grouped together? Because that's pretty common. Besides, I feel like comments are more important than reactions, and we've put that at the bottom.

LudvigHz:
> The same goes for comments though. I feel like "interactions" like comments and reactions should be grouped together? Because that's pretty common. Besides, I feel like comments are more important than reactions, and we've put that at the bottom.

Yeah, I agree. And it is a bit annoying sometimes that comments are all the way at the bottom, but on the other hand, comments don't really fit anywhere else, whereas reactions can fit in a small space almost anywhere, so it's more flexible where we can put it

jonasdeluna:
Could not reproduce the cypress fail, but may be due to lack of the correct backend version?

cypress/support/component.js
```diff
@@ -0,0 +1,27 @@
+// ***********************************************************
```
LudvigHz, line 1: Remove these files and we're good to go :upside_down_face:

LudvigHz (changes requested):

cypress/support/component.js
jonasdeluna: Hopefully I removed the correct files.

app/routes/pages/components/subcomponents/Statistic/Statistic.css
```diff
@@ -9,7 +9,7 @@
 }
 
 .statistic {
-  font-size: 2.2rem;
+  font-size: 1.2rem;
```
LudvigHz, line 12: What is this?

app/routes/pages/components/LandingPage.css
```diff
@@ -56,7 +56,7 @@ div > .locationContainerItem {
 
 .contactTitle {
   color: var(--lego-font-color);
-  font-size: 1.7rem;
+  font-size: 1rem;
```
LudvigHz, line 59: What does this do and why is it part of this PR?

LudvigHz (approved):

PR #3550, opened 2023-02-10
Only show loading indicator when necesarry

ivarnakken:
# Description

Currently on /events, you'll get a loading indicator even though you have events stored. This is for obvious reasons unnecessary, and is fixed here.

# Result

**Before**

https://user-images.githubusercontent.com/69514187/218099688-faaacea5-44e9-4445-800a-431ae77b2d11.mov

**After**

Whenever the theme changes, that's me doing a hard refresh. Notice that I do a hard refresh on two different routes.

https://user-images.githubusercontent.com/69514187/218099269-ea8c28de-14c3-446c-87df-843a27094ec6.mov

# Testing

Navigated between routes and did hard refreshes. This should not affect anything else.

---

Resolves ABA-267

### DIFF
--- a/app/routes/events/EventListRoute.ts
+++ b/app/routes/events/EventListRoute.ts
@@ -31,7 +31,7 @@ const mapStateToProps = (state, ownProps) => {
       actionGrant,
     })(state, ownProps),
     icalToken,
-    notLoading: !state.events.fetching,
+    fetching: state.events.fetching,
   };
 };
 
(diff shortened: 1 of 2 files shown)


LudvigHz (approved):

LudvigHz:
Thanks, this has annoyed me for a long time :+1:

eikhr (approved):
lgtm

PR #3551, opened 2023-02-11
Give the text input fields a retouch

ivarnakken:
# Description

They look more modern, and conform better with our styling elsewhere.

A `prefix` prop has been added, which takes in the name of an icon and turns it into a prefix.
A `removeBorder` props has also been added, and is used in the registration table (see image below).
This PR also fixes a bug where the disabled text fields on dark mode did not look disabled.

A big problem with the input fields is that the width of them are `100%` of the content. Changing this requires a lot of work, and is therefore separated into a different PR, but it'll come. Also, a different PR will tackle `react-select`, and make the styling of that similar to this.

# Result

**Before**
image

**After**
image

---

**Before**
image

**After**
image

# Testing

All types of text fields were checked for.


mattiastofte:
Looks good! It's nice to have a prop for removing the border, as it looks really clean in the right places. Could we get some screenshots showcasing exactly what the prefix prop does?

LudvigHz (approved):

norbye (approved):

app/components/Form/TextInput.tsx
```diff
@@ -1,40 +1,70 @@
...
+}: Props) => {
+  /* New ref is made because text inputs that are not Fields are not given a ref */
+  const newInputRef = useRef<HTMLInputElement>(inputRef);
+
   return (
-    <span className={cx(suffix && styles.inputGroup)}>
+    <Flex
+      alignItems="center"
+      gap={10}
+      style={{ border: removeBorder && 'none' }}
```
Arashfa0301, line 40: Isn't inline css a bad idea ??

ivarnakken: Not in this case, as it's dynamic and directly related to the prop.

app/components/Form/TextInput.css
```diff
@@ -1,54 +1,64 @@
 .input {
-  padding: 10px;
-  font-size: 16px;
-  border: 1px solid var(--border-gray);
-  border-radius: 3px;
-  width: 100%;
+  height: 38px;
```
eikhr, line 2: This is way too small for textareas! Breaks the company interest form😠

PR #3552, opened 2023-02-11
Improve /companies page

ivarnakken:
# Description

The cards now employ the `<Card />` component, making them resemble the rest of the webapp more. Also some other minimal styling changes. Finally, the whole card is a link, and not just the image, which has been an annoying bug for a really long time.

`CompaniesPage` is rewritten to a functional component.

`isHoverable` prop is added for cards that work as links.

# Result

**Before**

https://user-images.githubusercontent.com/69514187/218262039-63f5871d-414a-4c4e-8f1e-3428489a9dc8.mov

**After**

https://user-images.githubusercontent.com/69514187/218261983-40312029-efa0-4d6e-a58e-bc2eb7ec54a6.mov

# Testing

Tested on light and dark mode, and on mobile (some minimal changes here to the expanded text functionality).

---

Resolves ABA-270

### DIFF
--- a/app/components/Card/index.tsx
+++ b/app/components/Card/index.tsx
@@ -4,15 +4,10 @@ import type { HTMLAttributes } from 'react';
 
 type Props = {
   className?: string;
-
-  /** Small */
   tight?: boolean;
-
-  /** Shadow */
   shadow?: boolean;
-
-  /** Hidden overflow */
-  overflow?: boolean;
+  hideOverflow?: boolean;
+  isHoverable?: boolean;
 } & HTMLAttributes<HTMLDivElement>;
 
 function Card({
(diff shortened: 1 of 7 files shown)


app/components/Card/index.tsx
```diff
@@ -4,23 +4,19 @@ import type { HTMLAttributes } from 'react';
...
   className?: string;
-
-  /** Small */
   tight?: boolean;
-
-  /** Shadow */
   shadow?: boolean;
-
-  /** Hidden overflow */
-  overflow?: boolean;
```
ivarnakken, line 15: If you need to comment on the prop name, you've chosen a bad name.

ivarnakken:
I still think we need the option to filter and sort on these companies, but that requires more work and I have an event to attend.

ingraso (approved):
Great changes! Much more readable code and will be a lot easier to make further changes :rocket: Also, using the Card component makes it look a lot better already.

PR #3553, opened 2023-02-11
Improve page sidebar layout and design

mattiastofte:
# Description

Note: Since the old branch was so behind I just redid everything from scratch with typescript.

So right now the page sidenav looks hideous on the live webapp. It's one of those components that have long been overlooked and been needing a retouch for a while. In this PR I've improved both the user experience and visual aesthetics for the sidenav. I'm not quite finished yet, I still need to do mobile view improvements, but I would still like to get early feedback so this is the reason I've made this PR.

# Result

**before**
image

image

**after**

image

image

Note: The lighter color is the new highlighting color


app/routes/pages/components/PageHierarchy.css
```diff
@@ -3,6 +3,7 @@
 .sidebar {
   display: flex;
   flex-direction: column;
+  margin: 0 -0.2em;
```
ivarnakken, line 6: Why the negative margin? Kind of hacky, but I'll allow it.

app/routes/pages/components/Sidebar.css
```diff
@@ -24,9 +23,13 @@
...
 
 .sidebarHeader {
-  font-size: 2em;
-  margin-bottom: -10px;
+  font-size: 1.8em;
+  margin: 20px 0 0px 0;
   font-weight: 600;
+  width: 95%;
+  text-align: center;
+  border-bottom: 2px solid var(--additive-background);
```
ivarnakken, line 31: I feel like the border color should be `var(--border-gray)` so that it matches the rest of the webapp, but could you maybe instead tweak it to be lighter and better looking? I never changed it up.

app/routes/pages/components/Sidebar.tsx
```diff
@@ -45,9 +45,6 @@ class Sidebar extends Component<Props, State> {
                 <Icon name="close" size={50} />
               </button>
               <h3 className={styles.sidebarHeader}>Om Abakus</h3>
-              <h4 className={styles.sidebarSubtitle}>
-                {categorySelected ? categorySelected : 'Generelt'}
-              </h4>
```
ivarnakken, line 50: Good change. This has annoyed me for a long time.

But, remember to remove the css class from `Sidebar.css` as well.

app/styles/variables.css
```diff
@@ -85,6 +85,7 @@
   --color-green-8: #108944;
   --color-green-9: #06371b;
 
+  --color-red-hightlight: #ffeaea;
```
ivarnakken, line 88: Did you try out the `var( --color-red-hover-alpha)` variable? I feel like its purpose is the same thing, so it would be nice if it matched the other stuff, e.g. when you hover joblistings on _/karriere_). You can choose what looks best though.

ivarnakken (approved):
This looks stunning! 😍 😍

I only have some minor comments on the colors 😄

app/routes/pages/components/PageHierarchy.css
```diff
@@ -36,26 +45,36 @@
...
 }
 
 .selected {
-  font-weight: bold;
-  margin-left: -16px;
-}
-
-.selected::before {
-  content: '\25CF ';
+  padding: 8px 8px 8px 8px;
```
LudvigHz, line 65: `padding: 8px`?

app/routes/pages/components/Sidebar.css
```diff
@@ -24,9 +23,13 @@
 }
 
 .sidebarHeader {
-  font-size: 2em;
-  margin-bottom: -10px;
+  font-size: 1.8em;
+  margin: 20px 0 0px 0;
```
LudvigHz, line 27: `margin-left` (or right?, don't remeber, hence why it's a good idea to be explicit) ?

LudvigHz (approved):
Awesome :heart_eyes:

app/routes/pages/components/Sidebar.css
ivarnakken: I believe this is `margin-top`, so `margin-top: 20px` will do. Remember that `stylelint` will complain when setting `0px` instead of just `0`.

mattiastofte:
I've resolved the issues mentioned and I've also improved the mobile view. The mobile view is still a bit ugly, although it's much better than before and it can be improved upon further later.

image

app/routes/pages/components/PageHierarchy.css
mattiastofte: Good catch, I've removed the negative margin and made the overall margin less!

app/styles/variables.css
mattiastofte: I replaced it with the --color-red-hover-alpha, although from a universal design perspective I think the hover effect should be less transparent so it's more visible. We can look at it later though:)

app/routes/pages/components/Sidebar.tsx
mattiastofte: Nice one, removed now!

PR #3554, opened 2023-02-12
Created Allergies Tab

itsisak:
As discussed previously, I have now created a seperate tab for allergies in EventAdministrate. Currently anyone can still see all the allergies, but we can add an anonymized list later on.  

Before:

Skjermbilde 2023-02-12 kl  00 56 11

After:
Skjermbilde 2023-02-12 kl  00 56 35
Skjermbilde 2023-02-12 kl  00 57 49

Newst version:
Skjermbilde 2023-02-21 kl  22 04 20

Most of the code in Allergies.tsx and EventAllergiesRoute.ts was copied and slightly altered from Attendees.tsx and EventAttendees.ts. Unsure if everything is necessary or if there is anything I should have done differently, please let me know if so. 

Resolves ABA-175

### DIFF
--- a/app/routes/events/components/EventAdministrate/Allergies.tsx
+++ b/app/routes/events/components/EventAdministrate/Allergies.tsx
@@ -1,0 +1,109 @@
+import moment from 'moment-timezone';
+import { Link } from 'react-router-dom';
+import { Flex } from 'app/components/Layout';
+import LoadingIndicator from 'app/components/LoadingIndicator';
+import Table from 'app/components/Table';
+import type {
+  Event,
+  EventPool,
+  ActionGrant,
+  User,
+  ID,
+  EventRegistration,
+  EventRegistrationPaymentStatus,
+  EventRegistrationPresence,
+} from 'app/models';
+import type Comment from 'app/store/models/Comment';
+import type { CurrentUser } from 'app/store/models/User';
+import { RegistrationPill, getRegistrationInfo } from './RegistrationTables';
+
+export type Props = {
+  eventId: number;
+  event: Event;
+  comments: Comment[];
+  pools: Array<EventPool>;
+  loggedIn: boolean;
+  currentUser: CurrentUser;
+  error: Record<string, any>;
+  loading: boolean;
+  registered: Array<EventRegistration>;
+  unregistered: Array<EventRegistration>;
+  usersResult: Array<User>;
+  actionGrant: ActionGrant;
+  onQueryChanged: (value: string) => any;
+  searching: boolean;
+};
+
+const Allergies = ({ eventId, event, error, loading, registered }: Props) => {
+  if (loading) {
+    return <LoadingIndicator loading />;
+  }
+
+  if (error) {
+    return <div>{error.message}</div>;
+  }
+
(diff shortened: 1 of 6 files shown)


haukkagu (approved):
LGTM

itsisak:
Also, I added "Status" and "Til stede" to the Allergies tab, however Im unsure if this is usefull. I figured it might come in handy, put let me know and I can remove it as it is already in the "påmeldinger" tab.

PeterJFB (approved):
LGTM :100:

ivarnakken:
> Also, I added "Status" and "Til stede" to the Allergies tab, however Im unsure if this is usefull. I figured it might come in handy, put let me know and I can remove it as it is already in the "påmeldinger" tab.

@itsisak 
I don't really think we need status.

app/routes/events/components/EventAdministrate/Allergies.tsx
```diff
@@ -0,0 +1,155 @@
...
+  event: Event;
+  comments: Comment[];
+  pools: Array<EventPool>;
+  loggedIn: boolean;
+  currentUser: CurrentUser;
+  error: Record<string, any>;
+  loading: boolean;
+  registered: Array<EventRegistration>;
+  unregistered: Array<EventRegistration>;
+  unregister: (arg0: {
```
erlingfn, line 32: Is it possible to get a better name than arg0?

app/routes/events/components/EventAdministrate/Allergies.tsx
```diff
@@ -0,0 +1,155 @@
...
+    eventId: ID;
+    registrationId: ID;
+    admin: boolean;
+  }) => Promise<void>;
+  updatePresence: (arg0: number, arg1: number, arg2: string) => Promise<any>;
```
erlingfn, line 37: Same here. I know you just copied this from another place, but I think this is an excellent chance to improve it <3

erlingfn:
>  I don't really think we need status.

It agree. Does not hurt, but cannot really see how it is useful

erlingfn (changes requested):
Looks good! Would be great if you improve the types when you still remember how this works, and I think you can remove the presence, but otherwise this is great.

app/routes/events/components/EventAdministrate/Allergies.tsx
itsisak: I just now pushed a new change where I removed both the unregister and updatePresence (comment below) all together as they are not used in the allergies tab.

itsisak:
I removed the "Til stede" column, but I kept the "Status" and added sorting functionality to it, it seems nice to know if the allergies listed are for people that are actually registered or just on the waiting list.

app/routes/events/components/EventAdministrate/Allergies.tsx
itsisak: I have now fixed these issues in the original Attendes file as well as some deprecated syntax

haukkagu (approved):
LGTM💯

itsisak:
Fixed yarn lint

PR #3555, opened 2023-02-12
Include waiting list to summary pool tab

ivarnakken:
# Description

[Include waiting list to summary pool tab](https://github.com/webkom/lego-webapp/commit/445fa3ab8b67b4698733d815fb788dac148a5238) 

It makes more sense to include everyone in this tab, and it's annoying to
have to check multiple tabs when looking for someone.

People on waiting list are given an opacity to their row as a visual label.
`isMeeting` prop had to be added, because meeting registrations don't
have pools the same way events does.

---
  
[Rewrite AttendanceModal to a functional component](https://github.com/webkom/lego-webapp/commit/2c5428b4b0d04332db728ba41557faa30b2158fc)

# Result

Please check out the changes locally, as I'm not going to show the names of registered users.

# Testing

The attendance modal is used on events and meetings, and I've tested for both. Also tested for when there is no waiting list.

### DIFF
--- a/app/components/UserAttendance/AttendanceModal.tsx
+++ b/app/components/UserAttendance/AttendanceModal.tsx
@@ -1,6 +1,6 @@
 import cx from 'classnames';
 import { flatMap } from 'lodash-es';
-import { Component } from 'react';
+import { useState, useEffect } from 'react';
 import { Link } from 'react-router-dom';
 import Button from 'app/components/Button';
 import { TextInput } from 'app/components/Form';
@@ -21,7 +21,7 @@ export type Pool = {
 };
 
 type Props = {
-  pools: Array<Pool>;
+  pools: Pool[];
   title: string;
   togglePool: (arg0: number) => void;
   selectedPool: number;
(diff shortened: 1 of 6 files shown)


app/components/UserAttendance/AttendanceModal.tsx
```diff
@@ -48,98 +48,99 @@ const Tab = ({ name, index, activePoolIndex, togglePool }: TabProps) => (
...
-  generateAmendedPools = (
-    pools: Array<Pool>,
-    allRegistrations?: Registration[]
-  ) => {
-    if (pools.length === 1)
-      return this.setState({
-        pools,
-      });
-    const registrations =
-      allRegistrations || flatMap(pools, (pool) => pool.registrations);
```
ivarnakken, line 78: This is what fixed the problem. `allRegistrations` did not include waiting list registrations.

eikhr (approved):
Cool, This makes more sense!

app/components/UserAttendance/AttendanceModal.tsx
```diff
@@ -1,6 +1,6 @@
 import cx from 'classnames';
-import { flatMap } from 'lodash';
-import { Component } from 'react';
+import { flatMap } from 'lodash-es';
```
LudvigHz, line 2: We have some webpack config to convert all lodash imports to lodash-es, so you can just to `lodash`.

app/components/UserAttendance/AttendanceModal.tsx
```diff
@@ -48,98 +48,99 @@ const Tab = ({ name, index, activePoolIndex, togglePool }: TabProps) => (
...
-  state = {
-    pools: [],
-    filter: '',
-  };
-  static defaultProps = {
-    title: 'Status',
-  };
+  useEffect(() => {
+    generateAmendedPools(pools);
+  }, [pools]);
```
LudvigHz, line 64: You could've cleaned this up and removed the unnecessary hooks at the same time but :shrug:

LudvigHz (approved):

app/components/UserAttendance/AttendanceModal.tsx
ivarnakken: Ye I know, I just wanted to be precise for some reason 🤷🏼

app/components/UserAttendance/AttendanceModal.tsx
ivarnakken: Which hook is unnecessary?

LudvigHz: The `amendedPools`, `registrations` state hooks and the useEffects. So almost all of them :grimacing:

ivarnakken: what, but they are needed? Sure it can be calculated higher up, but I don't see how that's cleaner

PR #3556, opened 2023-02-12
Standardize our shadows

ivarnakken:
# Description

4 shadows of proportionally different sizes have been introduced, which all use the same color. I've not added shadows to any new places.

# Result

Not really any major visual changes. The shadow on the cards remain the same, so you can check them out.

# Testing

Went through all uses of `box-shadow`. Since I've not added any new shadows anywhere, this should not affect anything else.

### DIFF
--- a/app/components/Button/Button.css
+++ b/app/components/Button/Button.css
@@ -2,8 +2,7 @@
 
 :root {
   --button-color: var(--color-white);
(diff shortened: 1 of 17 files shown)


eikhr (approved):

LudvigHz (approved):
LGTM :shipit: 
Taking your word that there are no visual changes here

PR #3566, opened 2023-02-13
Allow multiple and other authors on articles

jonasdeluna:
# Description

Option to add multiple/other users as author on articles.

# Result

Before:

image

After:

image

Before:

image

After:

image

After:

image

# Testing

- [ X] I have thoroughly tested my changes.

---

Depends on: https://github.com/webkom/lego/pull/3220
Resolves: https://github.com/webkom/lego/issues/3195

### DIFF
--- a/app/components/Form/ObjectPermissions.tsx
+++ b/app/components/Form/ObjectPermissions.tsx
@@ -1,5 +1,6 @@
 import { SelectInput, CheckBox } from 'app/components/Form';
 import Tooltip from 'app/components/Tooltip';
+import type { PublicGroup } from 'app/store/models/Group';
 
 /*
  * Usage inside redux-form:
@@ -26,10 +27,10 @@ const ObjectPermissions = ({
   requireAuth,
   ...props
 }: {
-  canEditUsers?: any;
-  canEditGroups?: any;
-  canViewGroups?: any;
-  requireAuth?: any;
+  canEditUsers?: PublicGroup[];
+  canEditGroups?: PublicGroup[];
+  canViewGroups?: PublicGroup[];
+  requireAuth?: boolean;
 }) => {
   return [
     requireAuth && (
@@ -78,28 +79,13 @@ export const normalizeObjectPermissions = ({
   canViewGroups: initialCanViewGroups,
   canEditGroups: initialCanEditGroups,
   canEditUsers: initialCanEditUsers,
+  currentUser: currentUser,
 }: Record<string, any>) => {
-  const canEditUsers = initialCanEditUsers && initialCanEditUsers.map(toIds);
-  const canViewGroups = initialCanViewGroups && initialCanViewGroups.map(toIds);
-  const canEditGroups = initialCanEditGroups && initialCanEditGroups.map(toIds);
   return {
     requireAuth: !!requireAuth,
-    ...(canEditUsers
-      ? {
-          canEditUsers,
-        }
-      : {}),
-    //$FlowFixMe
-    ...(canEditGroups
-      ? {
-          canEditGroups,
-        }
-      : {}),
-    ...(canViewGroups
-      ? {
-          canViewGroups,
-        }
-      : {}),
+    canEditUsers: initialCanEditUsers?.map(toIds) ?? {},
+    canViewGroups: initialCanViewGroups?.map(toIds) ?? {},
+    canEditGroups: initialCanEditGroups?.map(toIds) ?? {},
   };
 };
 export const objectPermissionsToInitialValues = ({
@@ -107,38 +93,19 @@ export const objectPermissionsToInitialValues = ({
   canEditGroups: initialCanEditGroups,
   canEditUsers: initialCanEditUsers,
 }: Record<string, any>) => {
-  const canEditGroups =
-    initialCanEditGroups &&
-    initialCanEditGroups
-      .filter(Boolean)
-      .map((group) => ({ ...group, label: group.name, value: group.id }));
-  const canViewGroups =
-    initialCanViewGroups &&
-    initialCanViewGroups
-      .filter(Boolean)
-      .map((group) => ({ ...group, label: group.name, value: group.id }));
-  const canEditUsers =
-    initialCanEditUsers &&
-    initialCanEditUsers
-      .filter(Boolean)
-      .map((user) => ({ ...user, label: user.fullName, value: user.id }));
+  const canEditGroups = initialCanEditGroups
+    ?.filter(Boolean)
+    .map((group) => ({ ...group, label: group.name, value: group.id }));
+  const canViewGroups = initialCanViewGroups
+    ?.filter(Boolean)
+    .map((group) => ({ ...group, label: group.name, value: group.id }));
+  const canEditUsers = initialCanEditUsers
+    ?.filter(Boolean)
+    .map((user) => ({ ...user, label: user.fullName, value: user.id }));
   return {
-    ...(canEditUsers
-      ? {
-          canEditUsers,
-        }
-      : {}),
-    //$FlowFixMe
-    ...(canEditGroups
-      ? {
-          canEditGroups,
-        }
-      : {}),
-    ...(canViewGroups
-      ? {
-          canViewGroups,
-        }
-      : {}),
+    canEditUsers: canEditUsers ?? {},
+    canEditGroups: canEditGroups ?? {},
+    canViewGroups: canViewGroups ?? {},
   };
 };
 export const objectPermissionsInitialValues = {
--- a/app/routes/articles/ArticleDetailRoute.ts
+++ b/app/routes/articles/ArticleDetailRoute.ts
@@ -25,8 +25,10 @@ const mapStateToProps = (state, props) => {
   const comments = selectCommentsForArticle(state, {
     articleId,
   });
-  const author = selectUserById(state, {
-    userId: article.author,
+  const authors = article.authors?.map((e) => {
+    return selectUserById(state, {
+      userId: e,
+    });
   });
   const emojis = selectEmojis(state);
   return {
@@ -35,7 +37,7 @@ const mapStateToProps = (state, props) => {
     comments,
     article,
     articleId,
-    author,
+    authors,
     emojis,
   };
 };
@@ -56,11 +58,17 @@ export default compose(
   ),
   connect(mapStateToProps, mapDispatchToProps),
   loadingIndicator(['article.content']),
-  helmet((props: { article: PublicArticle; author: PublicUser }, config) => {
+  helmet((props: { article: PublicArticle; authors: PublicUser[] }, config) => {
     const tags = props.article.tags.map((content) => ({
       content,
       property: 'article:tag',
     }));
+
+    const authors = props.authors.map((author) => ({
+      property: 'article:authors',
+      content: `${config.webUrl}/users/${author.username}`,
+    }));
+
     return [
       {
         property: 'og:title',
@@ -103,10 +111,7 @@ export default compose(
         property: 'og:description',
         content: props.article.description,
       },
-      {
-        property: 'article:author',
-        content: `${config.webUrl}/users/${props.author.username}`,
-      },
+      ...authors,
       ...tags,
     ];
   })
--- a/app/routes/articles/ArticleEditRoute.ts
+++ b/app/routes/articles/ArticleEditRoute.ts
@@ -9,6 +9,8 @@ import {
 import { objectPermissionsToInitialValues } from 'app/components/Form/ObjectPermissions';
 import { LoginPage } from 'app/components/LoginForm';
 import { selectArticleById } from 'app/reducers/articles';
+import { selectCurrentUser } from 'app/reducers/auth';
+import { selectUserById } from 'app/reducers/users';
 import loadingIndicator from 'app/utils/loadingIndicator';
 import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
 import withPreparedDispatch from 'app/utils/withPreparedDispatch';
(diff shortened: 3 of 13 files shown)


app/components/Form/ObjectPermissions.tsx
```diff
@@ -24,12 +24,14 @@ const ObjectPermissions = ({
...
   canViewGroups,
   requireAuth,
+  author,
   ...props
 }: {
   canEditUsers?: any;
   canEditGroups?: any;
   canViewGroups?: any;
   requireAuth?: any;
+  author?: any;
```
LudvigHz, line 34: Would be nice if you could set the correct types here :)

app/components/Form/ObjectPermissions.tsx
```diff
@@ -100,12 +113,19 @@ export const normalizeObjectPermissions = ({
           canViewGroups,
         }
       : {}),
+
+    ...(author
+      ? {
+          author,
+        }
+      : {}),
```
LudvigHz, line 121: It would be awesome if you could rewrite this god-aweful code as well!

app/components/Form/ObjectPermissions.tsx
```diff
@@ -122,6 +142,11 @@ export const objectPermissionsToInitialValues = ({
     initialCanEditUsers
       .filter(Boolean)
       .map((user) => ({ ...user, label: user.fullName, value: user.id }));
+  const author =
+    initialAuthors &&
+    initialAuthors
```
LudvigHz, line 147: `initialAuthors?.filter(...`

app/routes/articles/ArticleEditRoute.ts
```diff
@@ -19,13 +20,24 @@ const mapStateToProps = (state, props) => {
   const article = selectArticleById(state, {
     articleId,
   });
+  const author = article?.author?.map((e) => {
+    return selectUserById(state, {
+      userId: e,
+    });
+  });
+  console.log(article);
```
LudvigHz, line 28: Remove this

app/routes/articles/ArticleEditRoute.ts
```diff
@@ -19,13 +20,24 @@ const mapStateToProps = (state, props) => {
...
     articleId,
     isNew: false,
     initialValues: {
       ...article,
-      ...objectPermissionsToInitialValues(article),
+      ...objectPermissionsToInitialValues({
+        canViewGroups: article.canViewGroups,
+        canEditGroups: article.canEditGroups,
+        canEditUsers: article.canEditUsers,
+        author: author,
```
LudvigHz, line 39: Why is this here? :thinking:

LudvigHz: Ah, if you need to add `author` here, change the function name to make sense with this change

app/routes/articles/components/ArticleDetail.tsx
```diff
@@ -70,16 +70,31 @@ const ArticleDetail = ({
...
-        <span className={styles.detail}>
-          Skrevet av
-          <Link to={`/users/${author.username}`}> {author.fullName}</Link>
-        </span>
-        <span className={styles.detail}>
-          {moment(article.createdAt).format('lll')}
-        </span>
-      </div>
-
+      {
```
LudvigHz, line 73: This is not needed here

app/routes/articles/components/ArticleDetail.tsx
```diff
@@ -70,16 +70,31 @@ const ArticleDetail = ({
...
+        <div className={styles.articleDetails}>
+          <span className={styles.detail}>
+            Skrevet av{' '}
+            {author &&
```
LudvigHz, line 77: This is superfluous with the check below.

app/routes/articles/components/ArticleDetail.tsx
```diff
@@ -70,16 +70,31 @@ const ArticleDetail = ({
...
+                return (
+                  <span key={e.username}>
+                    <Link
+                      to={`/users/${e.username}`}
+                      className={styles.overviewAuthor}
+                    >
+                      {' '}
+                      {e.fullName}
+                    </Link>
+                    {e === author[author.length - 1] ? '' : ','}
```
LudvigHz, line 88: I think using the index directly is easier here

app/routes/articles/components/ArticleEditor.tsx
```diff
@@ -192,12 +193,12 @@ const onSubmit = (
     ...normalizeObjectPermissions(data),
     youtubeUrl: data.youtubeUrl,
     title: data.title,
-    author: currentUser.id,
     description: data.description,
     content: data.content,
     tags: (data.tags || []).map((tag) => tag.value.toLowerCase()),
     pinned: data.pinned,
   };
+  body.author = body.author?.length !== 0 ? body.author : [currentUser.id];
```
LudvigHz, line 201: Inline this in the initialization instead of modifying it outside here.

app/routes/articles/components/Overview.tsx
```diff
@@ -24,17 +24,21 @@ const OverviewItem = ({ article }: { article: ArticleWithAuthorDetails }) => (
...
-            to={`/users/${article.author.username}`}
-            className={styles.overviewAuthor}
-          >
-            {' '}
-            {article.author.fullName}
-          </Link>{' '}
-        </span>
-      )}
+      {article?.author &&
+        article?.author?.map((e) => {
```
LudvigHz, line 28: Here as well, when you use the `?.` operator, you don't need the `article?.author &&` check.

LudvigHz (changes requested):
Cool!

You should rename `author` to `authors` to reflect this change. And fix up the small other things and this is good to go :)

Regarding removing the created_by, I think it might be a nice thing to keep?

jonasdeluna:
On default the author field will be the current user, and if the field is empty it will also be set to the current user upon validation as a security measure.

app/routes/articles/ArticleDetailRoute.ts
```diff
@@ -56,7 +58,7 @@ export default compose(
   ),
   connect(mapStateToProps, mapDispatchToProps),
   loadingIndicator(['article.content']),
-  helmet((props: { article: PublicArticle; author: PublicUser }, config) => {
+  helmet((props: { article: PublicArticle; authors: PublicUser }, config) => {
```
LudvigHz, line 61: authors is an array now :) so you need to change this to `PublicUser[]`

app/routes/articles/ArticleDetailRoute.ts
```diff
@@ -104,8 +106,8 @@ export default compose(
         content: props.article.description,
       },
       {
-        property: 'article:author',
-        content: `${config.webUrl}/users/${props.author.username}`,
+        property: 'article:authors',
+        content: `${config.webUrl}/users/${props.authors.username}`,
```
LudvigHz, line 110: Ref. that authors is now an array, this will fail.

See https://ogp.me/#type_article for the correct way to handle multiple authors in the meta tags here

app/routes/articles/ArticleListRoute.ts
```diff
@@ -34,8 +34,10 @@ const mapStateToProps = (state, props) => {
     }
   ).map((article) => ({
     ...article,
-    author: selectUserById(state, {
-      userId: article.author,
+    authors: article?.authors?.map((e) => {
```
LudvigHz, line 37: `article` should not be nullish here?

app/routes/articles/components/ArticleDetail.tsx
LudvigHz: This is still here :grimacing:

app/routes/articles/components/Overview.tsx
```diff
@@ -24,17 +24,17 @@ const OverviewItem = ({ article }: { article: ArticleWithAuthorDetails }) => (
...
-          <Link
-            to={`/users/${article.author.username}`}
-            className={styles.overviewAuthor}
-          >
-            {' '}
-            {article.author.fullName}
-          </Link>{' '}
-        </span>
-      )}
+      {article?.authors?.map((e) => {
```
LudvigHz, line 27: Here as well, `article` should not suddenly be nullish?

LudvigHz: Realisticly, `authors` is also a required field and should always be defined as well no?

app/components/Form/ObjectPermissions.tsx
```diff
@@ -24,12 +25,14 @@ const ObjectPermissions = ({
...
 }: {
-  canEditUsers?: any;
-  canEditGroups?: any;
-  canViewGroups?: any;
-  requireAuth?: any;
+  canEditUsers?: PublicGroup[];
+  canEditGroups?: PublicGroup[];
+  canViewGroups?: PublicGroup[];
+  requireAuth?: boolean;
+  authors?: number[];
```
LudvigHz, line 35: Hmm, I realize now that this is all in ObjectPermission related stuff.

I don't know how fan I am of this, since authors don't really have anything to do with ObjectPermissions and is specific to just articles.

app/routes/articles/components/ArticleEditor.tsx
```diff
@@ -189,15 +190,15 @@ const onSubmit = (
           cover: data.cover,
         }
       : {}),
-    ...normalizeObjectPermissions(data),
+    ...normalizeObjectPermissions({ ...data, currentUser: currentUser }),
```
LudvigHz, line 193: You shouldn't do this. It is more than enough that you set the default to the current user. If they remove themselves from the authors field, but it sets them as the author anyways that would be very strange.

LudvigHz (changes requested):
Sorry for suddenly finding more stuff here, I should have done a more thorough review the first time around :disappointed:.

A couple things (as are mentioned in the comments):
- If authors is a required field in the backend, it should always be defined here, and both the types, and the usage of the `authors` field should reflect that. So no null checks or anything here (array length checks are fine). Let the type system catch it.
- Right now, the `authors` field is set together with the `ObjectPermissions`. This is not really related, and ObjectPermissions are generic fields that can be used on any field that use objectPermissions. I do like the idea of having default fields that can be reused super easy, but I doubt that we would have authors anywhere else than articles so with that in mind, both that it's not generic and that it does not have anything to do with ObjectPermissions, I think you should move it out and put the code for authors in the articleeditor directly.

app/routes/articles/components/ArticleEditor.tsx
ivarnakken: Is this not just to ensure there's always an author? If they remove themselves, and no author has been specified, it should default to `currentUser`, no? However, if they remove themselves and add someone else, it should just be that other user. I feel like that's the "better" MMI.

jonasdeluna: Made the field required with front-end validation. An empty array seems to be allowed in the back-end but unless someone explicitly sends a request this should not happen.

` authors: data.authors.length > 0 ? data.authors.map((e) => e.value) : null,`

app/routes/articles/ArticleEditRoute.ts
```diff
@@ -19,13 +21,26 @@ const mapStateToProps = (state, props) => {
...
       ...article,
-      ...objectPermissionsToInitialValues(article),
+      ...objectToInitialValues({
+        canViewGroups: article.canViewGroups,
+        canEditGroups: article.canEditGroups,
+        canEditUsers: article.canEditUsers,
+      }),
+      authors: authors
+        .filter(Boolean)
+        .map((user) => ({ ...user, label: user.fullName, value: user.id })),
```
LudvigHz, line 43: Is there a reason you spread `...user` here?

app/routes/articles/ArticleListRoute.ts
```diff
@@ -34,8 +34,10 @@ const mapStateToProps = (state, props) => {
     }
   ).map((article) => ({
     ...article,
-    author: selectUserById(state, {
-      userId: article.author,
+    authors: article.authors.map((e) => {
+      return selectUserById(state, {
+        userId: e,
+      });
```
LudvigHz, line 40: You should _ideally_ use a selector for multiple users here. Not critical though

app/routes/articles/components/ArticleEditor.tsx
```diff
@@ -196,24 +213,43 @@ const onSubmit = (
           cover: data.cover,
         }
       : {}),
-    ...normalizeObjectPermissions(data),
+    ...normalizeObjectPermissions({ ...data }),
```
LudvigHz, line 216: Why the spread here?

app/routes/articles/components/ArticleEditor.tsx
```diff
@@ -196,24 +213,43 @@ const onSubmit = (
...
     tags: (data.tags || []).map((tag) => tag.value.toLowerCase()),
     pinned: data.pinned,
   };
+
   return submitArticle(body);
 };
 
+type ValidationError<T> = Partial<{
+  [key in keyof T]: string | Record<string, string>[];
+}>;
```
LudvigHz, line 231: Loving the types😍

LudvigHz (approved):
Nice 👍 looks good now. Up to you if you want to fix up the last issues

app/routes/articles/components/ArticleEditor.tsx
```diff
@@ -196,24 +213,43 @@ const onSubmit = (
...
+    data.youtubeUrl
+  );
+
+  if (!isValidYoutubeUrl) {
+    errors.youtubeUrl = errorMessage;
+  }
+
+  if (!data.authors || data.authors.length === 0) {
+    errors.title = 'Forfatter er påkrevd';
+  }
```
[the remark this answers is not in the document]
ivarnakken: This is entirely fine imo. though.

ivarnakken:
Any reason to not merge this in now? @jonasdeluna You've done great work!

ivarnakken (approved):

jonasdeluna:
> Any reason to not merge this in now? @jonasdeluna You've done great work!

Thanks, not really unless the validator needs changing ill have a final look at the test^

jonasdeluna:
Everything should be good, but messed up the rebase :| also the backend tests are failing...

PR #3567, opened 2023-02-13
Expand sidebar element based on pathname

ivarnakken:
# Description

You can now visit any page and the correct category will be expanded.

This is done by using the url pathname to determine which category we're currently in, instead of it being passed down as prop.

# Result

https://user-images.githubusercontent.com/69514187/218572725-f30825a0-b921-4363-bcb2-3c000e11201c.mov

# Testing

Tested for "info-om-abakus" (which is an edge case), and for when you're not logged in.


eikhr (approved):
Cool

erlingfn (approved):
LGTM

PR #3573, opened 2023-02-14
Validate user allergy input

ivarnakken:
# Description

It's surprisingly common for people to write "ingen/nei".

# Result

image

# Testing

Tested with and without valid inputs.

### DIFF
--- a/app/routes/users/components/UserSettings.tsx
+++ b/app/routes/users/components/UserSettings.tsx
@@ -12,7 +12,12 @@ import type { UserEntity } from 'app/reducers/users';
 import DeleteUser from 'app/routes/users/components/DeleteUser';
 import RemovePicture from 'app/routes/users/components/RemovePicture';
 import { spySubmittable } from 'app/utils/formSpyUtils';
-import { createValidator, required, isEmail } from 'app/utils/validation';
+import {
+  createValidator,
+  required,
+  isEmail,
+  isValidAllergy,
+} from 'app/utils/validation';
 import ChangePassword from './ChangePassword';
 import UserImage from './UserImage';
 import styles from './UserSettings.css';
@@ -52,6 +57,7 @@ const validate = createValidator({
   lastName: [required()],
   gender: [required()],
   email: [required(), isEmail()],
+  allergies: [isValidAllergy()],
 });
 
 const UserSettings = (props: Props) => {
@@ -67,7 +73,9 @@ const UserSettings = (props: Props) => {
     updateUser,
   } = props;
   const showAbakusMembership = user.isStudent;
+
   const onSubmit = (values: FormValues) => updateUser(values);
+
   return (
     <div>
       <div className={styles.pictureSection}>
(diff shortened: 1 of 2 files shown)


app/utils/validation.ts
```diff
@@ -53,6 +53,16 @@ export const timeIsAfter = (otherField, message) => (value, context) => {
...
 };
 
+export const isValidAllergy =
+  (
+    message = 'La feltet stå tomt hvis du ikke har noen allergier/preferanser'
+  ) =>
+  (value: string) => {
+    const notValidAnswers = ['ingen', 'ingenting', 'nei', 'nope', 'nada'];
+
+    return [!notValidAnswers.includes(value?.toLowerCase()), message];
```
LudvigHz, line 63: if `value` is always a `string`, why do you need the `?`?

LudvigHz (approved):

app/routes/users/components/UserSettings.tsx
```diff
@@ -130,8 +138,10 @@ const UserSettings = (props: Props) => {
                 component={RadioButton.Field}
               />
             </RadioButtonGroup>
+
             <Field
               label="Matallergier/preferanser"
+              parse={(value) => value}
```
erlingfn, line 144: Could you explain why this is necessary? I don't doubt that it is, but does seem a bit weird.

erlingfn (approved):
Looks good!

app/utils/validation.ts
ivarnakken: hmm, well I think it's always a `string`, but I wasn't sure if it somehow could be `undefined`. I believe I got some error, and I therefore added it. But I think it can be removed ..

app/routes/users/components/UserSettings.tsx
erlingfn: I see, this is good to know!

app/utils/validation.ts
```diff
@@ -53,6 +53,16 @@ export const timeIsAfter = (otherField, message) => (value, context) => {
   return [true];
 };
 
+export const isValidAllergy =
+  (
+    message = 'La feltet stå tomt hvis du ikke har noen allergier/preferanser'
+  ) =>
+  (value: string) => {
+    const notValidAnswers = ['ingen', 'ingenting', 'nei', 'nope', 'nada'];
```
Arashfa0301, line 61: I believe we should also include "nay", "nah" and "negative" :)

app/utils/validation.ts
```diff
@@ -53,6 +53,16 @@ export const timeIsAfter = (otherField, message) => (value, context) => {
...
+
+    return [!notValidAnswers.includes(value.toLowerCase()), message];
```
Arashfa0301, line 63: Could you explain how this syntax works! It looks interesting :)

app/utils/validation.ts
erlingfn: Are there any examples of people entering them in the allergies field?

app/utils/validation.ts
ivarnakken: Not sure which part you're referring to. The validator needs an array with two values; a boolean and a message. The boolean indicates if the validation has passed or not, and the message is what is given to the user. What we're actually creating here is a function returning a function which returns the array.

PR #3574, opened 2023-02-14
Add information regarding rights as an eventmaker

tirilsys:
# Description
A request from the big o Ledelsen that we include information about the rights you have (and don't have) when making an event. 

# Result
**before**
image
**after**
(after trying to submit an event without checking the box, and when hovering the text) 
image


ivarnakken (approved):
Code looks solid! 💯 Thanks a lot!! 🤗 

I've run this through our CI, and it's failing on linting. It seems like you might be running a different `prettier` version than us? Simply run `yarn prettier` and we should be all good! 😄 

Also, you'll need to rebase this due to a tiny conflict. 
_EDIT: it looks like the conflict is due to the formatting, so you'll just need to format 😬_

PR #3568, opened 2023-02-13
Sort upcoming events on profile page

eikhr:
# Description

It has annoyed me for too long that the events on the profile page are shown in a seemingly random order. This sorts them by start-date.
I simply replaced `selectEvents` with `selectSortedEvents` in `selectUpcomingEvents`. 

# Result

Before:
Screenshot 2023-02-13 at 22 24 52
After:
Screenshot 2023-02-13 at 22 25 04

# Testing

- [x] I have thoroughly tested my changes.

PR #3577, opened 2023-02-15
Sort incomingEvents on user page

Arashfa0301:
Made use of the orderBy function to sort event by startTime on user page. I have thoroughly tested my changes. 

----
before:

image

After: 

image


ivarnakken:
Is this not a duplicate of #3568?

Arashfa0301:
> Is this not a duplicate of #3568?

It seems so. But I seem to have produced the same results by a totally different method. I am a little bewildered by which way is the more optimal way of handling this to be frank.

jonasdeluna (approved):
Nice

PR #3584, opened 2023-02-16
Make the validator actually validate

LudvigHz:
# Description

Previously it only checked if you had a user, and the qr did nothing at
all. With this change, we validate that the user is a member of Abakus
and the QR scanner also displays success or error messages for
validation status.

On error, it just gives an alert. I would make it nicer if I had time. But the alert works great tbh. It makes sure that the person validating needs to confirm the message before continuing. Not very fancy but effective.

All the other changed files are just related typing improvements.

# Result

### User is not member of Abakus:

[Screencast from 2023-02-16 18-11-01.webm](https://user-images.githubusercontent.com/42850232/219440522-c90ffca6-06e8-4f4f-bb27-05f95cab1797.webm)

### User does not exist (only makes sense for QR):
 
[Screencast from 2023-02-16 18-16-03.webm](https://user-images.githubusercontent.com/42850232/219440533-a2949997-f10c-435b-b3b6-b363e5cbdb5c.webm)

### Successful validation:

[Screencast from 2023-02-16 18-24-01.webm](https://user-images.githubusercontent.com/42850232/219441108-c7e1030e-3e4f-41d1-9c1a-25d71c0b0027.webm)

# Testing


app/components/UserValidator/index.tsx
```diff
@@ -1,53 +1,59 @@
...
+    (result: UserWithUsername) => {
       clearSearch();
       return handleSelect(result)
         .then(
-          () => {
-            const sound = new window.Audio(goodSound);
-            sound.play();
-            showCompleted();
+          (user: User) => {
+            console.log(user);
```
norbye, line 43: I assume this is only here for testing?

norbye (approved):
Reviewed on mobile and the videos didn't play in the app - so do with this as you'd like, but the code looked fine to me:)

app/components/UserValidator/index.tsx
LudvigHz: Ooops. Removed, thanks :)

juniwbjerde (approved):
lgtm!

erlingfn (approved):
LGTM

PR #3588, opened 2023-02-18
Small fixes on infopage design

norbye:
# Description

Key points:
* Prevent overflow
* Fill available space
* Add more responsive breakpoints.
* Prevent padding on top caused by a not specific enough selector

# Result

It requires a loot of pictures to showcase every edge case of this change, so to check every screen size you'd rather have to pull this and test it locally.

Before: unnecessary padding on top + overflow on the right side
image

After
image

Before: doesn't fill the entire page
image

After: equal paddings on the right and left side
image

Before: the breakpoints are either desktop or mobile, where mobile is a single row even if its pretty wide
image

After: three breakpoints, with 3, 2 and 1 columns depending on the screen size
image

# Testing

### DIFF
--- a/app/routes/pages/components/subcomponents/TextWithTitle/index.tsx
+++ b/app/routes/pages/components/subcomponents/TextWithTitle/index.tsx
@@ -1,3 +1,4 @@
+import cx from 'classnames';
 import styles from './TextWithTitle.css';
 import type { ReactNode } from 'react';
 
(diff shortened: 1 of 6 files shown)


app/routes/pages/components/subcomponents/TextWithTitle/index.tsx
```diff
@@ -1,24 +1,31 @@
...
       <p className={styles.text}>{text}</p>
     </div>
   );
 };
 
-const TextWithTitle = ({ title, text, extraStyle }: Props) => {
+const TextWithTitle = ({ title, text, extraStyle, extraClassName }: Props) => {
   return (
-    <div className={styles.container} style={extraStyle}>
+    <div className={`${styles.container} ${extraClassName}`} style={extraStyle}>
```
LudvigHz, line 28: `cx`?

LudvigHz (approved):
Awesome! Much needed👌👌

ivarnakken (approved):
Yay💯💯

app/routes/pages/components/subcomponents/TextWithTitle/index.tsx
norbye: Ah yes, was testing moving back and forth to see if the css selectors specificity would change (it didn't:/) - moved back to cx now

PR #3590, opened 2023-02-18
Create TextWithIcon component

ingraso:
# Description

Text is displayed with icons throughout the whole webapp. However, there is no component or common way of doing this. With the `<TextWithIcon />` component there is now a generalized component that can be used for this purpose. An example of usage can be seen here:
image

There should be no visual changes to the text with icons.

Also fixes some TS errors and moves the tooltip on `<EventDetail />` to be on the icon rather than the text.

# Result
**Before**
image

**After**
image

# Testing

Tested by clicking through where changes have been done.

---

Resolves ABA-293

### DIFF
--- a/app/routes/events/components/EventDetail/index.tsx
+++ b/app/routes/events/components/EventDetail/index.tsx
@@ -212,7 +212,7 @@ export default class EventDetail extends Component<Props, State> {
     const color = colorForEvent(event.eventType);
 
     const onRegisterClick = event.following
-      ? () => unfollow(event.id, event.id)
+      ? () => unfollow(event.following, event.id)
       : () => follow(currentUser.id, event.id);
 
     const currentMoment = moment();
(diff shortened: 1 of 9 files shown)


app/routes/events/components/EventDetail/index.tsx
```diff
@@ -211,7 +212,7 @@ export default class EventDetail extends Component<Props, State> {
     const color = colorForEvent(event.eventType);
 
     const onRegisterClick = event.following
-      ? () => unfollow(event.following, event.id)
+      ? () => unfollow(event.id, event.id)
       : () => follow(currentUser.id, event.id);
```
ollfkaih, line 216: Types were wrong here, but these lines are correct (#3592)

ollfkaih (changes requested):

LudvigHz (approved):
LGTM

ivarnakken (approved):
Nice to standardize stuff like this 💯 Just fix up in Olav's comment and we should be good here

ollfkaih (approved):
Looks good now!

PR #3594, opened 2023-02-19
Make company descriptions collapsible

PeterJFB:
# Description

Make company descriptions over a certain length collapsible, making the succeeding components more accessible.

# Result

**Before**
Screenshot from 2023-02-19 14-47-36

**After**
Long description: 
Screencast from 02-19-2023 02:46:45 PM.webm

Forgot to record in light theme, but here's how that looks^
image

Short descriptions will not have the option to expand the content
Screenshot from 2023-02-19 14-48-17

# Testing

Ensured that the component does not perform any unnecessary animation while loading into the page with short or long content. Made sure the interactions works nicely on mobile as well:)

---

Resolves [ABA-294](https://linear.app/abakus-webkom/issue/ABA-294)

### DIFF
--- a/app/components/CollapsibleDisplayContent/CollapsibleDisplayContent.css
+++ b/app/components/CollapsibleDisplayContent/CollapsibleDisplayContent.css
@@ -1,0 +1,55 @@
+@import url('~app/styles/variables.css');
+
+.collapse {
+  position: relative;
+  overflow: hidden;
+  transition: height var(--easing-slow);
+}
+
+.base {
+  word-break: break-word;
+
+  div[data-type='todo'] > input {
+    cursor: pointer;
+    float: left;
+    position: relative;
+    top: 4px;
+    left: -4px;
+  }
+
+  ul {
+    margin-left: 15px;
+  }
+
+  ol {
+    margin-left: 15px;
+  }
+
+  figure {
+    text-align: center;
+    font-size: 13px;
+  }
+}
+
+.showMore {
+  display: flex;
+  justify-content: center;
+  height: 50px;
+  width: 100%;
+  position: absolute;
+  bottom: 0;
+  background-image: linear-gradient(
+    rgba(0, 0, 0, 0%),
+    var(--lego-card-color),
+    var(--lego-card-color)
+  );
+  cursor: pointer;
+}
+
+.showMoreIcon {
+  padding: 3px;
+  margin: 0;
+  border-radius: 50%;
+  background-color: var(--lego-card-color);
+  box-shadow: 0 0 5px var(--lego-card-color);
+}
--- a/app/components/CollapsibleDisplayContent/index.tsx
+++ b/app/components/CollapsibleDisplayContent/index.tsx
@@ -1,0 +1,86 @@
+import Editor from '@webkom/lego-editor';
+import '@webkom/lego-editor/dist/style.css';
+import 'react-image-crop/dist/ReactCrop.css';
+import { type CSSProperties, useRef, useState } from 'react';
+import Icon from 'app/components/Icon';
+import styles from './CollapsibleDisplayContent.css';
+
+type Props = {
+  content: string;
+  id?: string;
+  className?: string;
+  style?: CSSProperties;
+  placeholder?: string;
+  collapsedHeight?: number;
+};
+
+/**
+ * Renders `content` produced by the Editor component in a collapsible read-only format.
+ */
+function CollapsibleDisplayContent({
+  content,
+  id,
+  style,
+  className,
+  placeholder,
+  collapsedHeight = 250,
+}: Props) {
+  let domParser = null;
+
+  if (!__CLIENT__) {
+    const JSDOM = require('jsdom').JSDOM;
+
+    domParser = (val) => new JSDOM(val).window.document;
+  }
+
+  const [isOpened, setIsOpened] = useState(false);
+
+  const ref = useRef<HTMLDivElement>(null);
+  const useCollapse =
+    collapsedHeight < ref.current?.clientHeight || !ref.current;
+
+  return (
+    <div
+      className={styles.collapse}
+      style={{
+        height: !useCollapse
+          ? null
+          : isOpened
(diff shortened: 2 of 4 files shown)


app/components/CollapsibleDisplayContent/CollapsibleDisplayContent.css
```diff
@@ -0,0 +1,43 @@
+@import url('~app/styles/variables.css');
+
+.collapse {
+  position: relative;
+  overflow: hidden;
+  transition: 0.5s var(--easing) height;
```
ivarnakken, line 6: Use `transition: height var(--easing-slow) instead`, please. It's important to not use different speeds everywhere.

app/components/CollapsibleDisplayContent/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  /** The content to be displayed - the text */
+  content: string;
+
+  /** The id of the div wrapping the content - the id */
+  id?: string;
+
+  /** The classname of the div wrapping the content - the className */
+  className?: string;
+
+  /** Any style tp be added to the div wrapping the content - the style */
```
ivarnakken, line 18: Although comments are often nice, I'm more a fan of not commenting on props, as I feel they are self explanatory. If you need to explain the props, that just means the naming is bad.

Also there's a typo on the last one; `tp -> to`.

app/components/CollapsibleDisplayContent/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  style?: CSSProperties;
+  placeholder?: string;
+  collapsedHeight?: number;
+};
+
+/**
+ * Renders `content` produced by the Editor component in a read-only format.
+ */
+function CollapsibleDisplayContent({
```
ivarnakken, line 27: I'd prefer a short name, like `CollapsibleContent`. Not sure why "display" is needed (I get that it was named that from before).

app/components/CollapsibleDisplayContent/index.tsx
```diff
@@ -0,0 +1,92 @@
+import Editor from '@webkom/lego-editor';
+import '@webkom/lego-editor/dist/style.css';
+import 'react-image-crop/dist/ReactCrop.css';
```
ivarnakken, line 3: Why do these two need to be imported?

app/components/CollapsibleDisplayContent/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { type CSSProperties, useRef, useState } from 'react';
+import Icon from '../Icon';
```
ivarnakken, line 5: Absolute import, please. I'm sure it was just an auto import though.

ivarnakken (changes requested):
Very cool! 💯💯 

I have some minor comments, but they aren't that important. What bothers me more is the trigger icon on top of the text. Since it is outlined, it sort of blends in together with the text, and I'm not a big fan of it. With "show more / expanding content triggers" I prefer flat text buttons, like these:

image

But, whatever you end up with, please don't put it on top of the text.

LudvigHz:
Nice! :100: 
But why not just add a `collapse` prop on displayContent instead of making a nearly identical one?

app/components/CollapsibleDisplayContent/index.tsx
PeterJFB: I agree. Removed them from `DisplayContent as well`

app/components/CollapsibleDisplayContent/index.tsx
PeterJFB: I understand the use of `Display` as the content being "read-only", though there might exist better names for it ...

app/components/CollapsibleDisplayContent/index.tsx
PeterJFB: Yikes, no. Thanks for spotting :eyes:

PeterJFB:
> Very cool! 100100
> 
> I have some minor comments, but they aren't that important. What bothers me more is the trigger icon on top of the text. Since it is outlined, it sort of blends in together with the text, and I'm not a big fan of it. With "show more / expanding content triggers" I prefer flat text buttons, like these:
> 
> image
> 
> But, whatever you end up with, please don't put it on top of the text. I like the "fade" a lot though!

Yeah i agree the outline was clashing with the text too much. 

The updated look is on the light theme picture. I moved it and added a smooth background to it, think it looks way cleaner now^

PeterJFB:
> Nice! 100 But why not just add a `collapse` prop on displayContent instead of making a nearly identical one?

Honest answer, I was lazy:P

Looking over the other pages, I didn't really think it was worth the time to make the more logic-intensive component. I also think this solution will have the least performance-impact, since it will only add the `UseRef`-code to components that don't need it.

app/components/CollapsibleDisplayContent/index.tsx
ivarnakken: I'm not following you on the "read-only" part. `CollapsibleContent` is much cleaner and more precise.

app/components/CollapsibleDisplayContent/index.tsx
PeterJFB: Yikes yes***. Turns out they add styling-properties to the lego-editor :sweat_smile:

app/components/CollapsibleDisplayContent/index.tsx
PeterJFB: It makes more sense if you look at how it is implemented: It uses the `Editor`-component to display the content, but since you cannot edit it, it is only a "display".

I'd prefer keeping the name as it is such that it is easy to see it is related to the `DisplayContent` and not the `Collapse`-wrapper (which we also have but isn't flexible enough for this purpose unfortunately).

ivarnakken (approved):
:shipit:

Arashfa0301 (approved):
Looks good to me :)

ingraso (approved):
Nice work! Looks very nice :rocket:

PR #3626, opened 2023-02-28
Use correct symbol on EvenStatus-component

PeterJFB:
# Description

Someone pointed out an event with infinite number of places showed x/0 in the listrings-views

# Result

Used code from the same file (line 78)

Before:
Screenshot from 2023-02-28 11-32-09

After:
Screenshot from 2023-02-28 11-31-57

# Testing

- [x]  I have thoroughly tested my changes.

Inspected images above, and made sure the other event-items looked the same:)

---


ivarnakken (approved):

LudvigHz (approved):

ivarnakken:
Would it be better to just say "65 påmeldte"? I feel like the infinite symbol might be redundant.

PR #3631, opened 2023-02-28
Redesign CompanyDetail v2

ingraso:
# Description
After the redesign of `<CompanyDetail />` there are some small issues to address. As can be seen in the image below, the fonts for "Ingen kommende arrangementer" and "Ingen tilgjengelige jobbannonser" are different. Also, for companies without info in the sidebar, the vertical line can be seen without any content on the right side. These issues are now solved, as well as moving the `<LoadingIndicator />` inside `<EventListCompact />` instead of wrapping the loading indicator around all of the event lists separately. Not that many visual changes, other than the font and the vertical line.

In addition, the company description is not longer italic.

Screenshot from 2023-02-28 14-52-59

# Result
**Before**
Screenshot from 2023-02-28 14-52-59

**After**
image

**Before**
image

**After**
image

# Testing

Changes have been tested visually both with light and dark mode.

---

Resolves ABA-310

### DIFF
--- a/app/components/CollapsibleDisplayContent/index.tsx
+++ b/app/components/CollapsibleDisplayContent/index.tsx
@@ -51,15 +51,17 @@ function CollapsibleDisplayContent({
       }}
     >
       <div key={content} id={id} style={style} className={className} ref={ref}>
-        <Editor
-          onChange={() => {}}
-          onBlur={() => {}}
-          onFocus={() => {}}
-          value={content}
-          placeholder={placeholder}
-          disabled
-          domParser={domParser}
-        />
+        <i>
+          <Editor
+            onChange={() => {}}
+            onBlur={() => {}}
+            onFocus={() => {}}
+            value={content}
+            placeholder={placeholder}
+            disabled
+            domParser={domParser}
+          />
+        </i>
       </div>
       {useCollapse && (
         <div
--- a/app/components/CollapsibleDisplayContent/index.tsx
+++ b/app/components/CollapsibleDisplayContent/index.tsx
@@ -51,17 +51,15 @@ function CollapsibleDisplayContent({
       }}
     >
       <div key={content} id={id} style={style} className={className} ref={ref}>
-        <i>
-          <Editor
-            onChange={() => {}}
-            onBlur={() => {}}
-            onFocus={() => {}}
-            value={content}
-            placeholder={placeholder}
-            disabled
-            domParser={domParser}
-          />
-        </i>
+        <Editor
+          onChange={() => {}}
+          onBlur={() => {}}
+          onFocus={() => {}}
+          value={content}
+          placeholder={placeholder}
+          disabled
+          domParser={domParser}
+        />
       </div>
       {useCollapse && (
         <div
--- a/app/components/EventListCompact/EventListCompact.css
+++ b/app/components/EventListCompact/EventListCompact.css
@@ -1,6 +1,0 @@
-.emptyState {
-  color: var(--secondary-font-color);
-  font-size: 20px;
-  font-weight: lighter;
-  font-style: italic;
-}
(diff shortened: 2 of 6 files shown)


app/components/EventListCompact/index.tsx
```diff
@@ -1,41 +1,39 @@
...
-      </EmptyState>
-    )}
-  </>
-);
-
+  loading,
+}: Props) => {
+  return loading ? (
+    <LoadingIndicator margin="20px auto" loading />
+  ) : events && events.length ? (
```
ivarnakken, line 22: I'm more fan of returning this earlier. Turneries can easily become hard to read.

```js
if (loading) {
  return ...
}

return (
  ...
)
```

app/components/EventListCompact/EventListCompact.css
```diff
@@ -1,6 +0,0 @@
-.emptyState {
-  color: var(--secondary-font-color);
-  font-size: 20px;
-  font-weight: lighter;
-  font-style: italic;
-}
```
ivarnakken, line 6: I think it looked good with the `var(--secondary-font-color)` though.

ivarnakken (approved):
💯 Thanks for iterating!

app/components/EventListCompact/index.tsx
ollfkaih: Partially agree, but this is simple enough to be a style preference.

ollfkaih (approved):

app/components/EventListCompact/index.tsx
ingraso: Sure, for me it doesn't matter so I'll just change it to returning earlier as you're saying

app/components/EventListCompact/index.tsx
```diff
@@ -1,41 +1,42 @@
...
-import { Flex } from '../Layout';
-import styles from './EventListCompact.css';
 
 type Props = {
   events: Array<Event>;
   noEventsMessage: string;
-  loggedIn: boolean;
   eventStyle?: EventStyle;
+  loading?: boolean;
+  classNameErrorMessage?: string;
```
PeterJFB, line 12: the `classNameErrorMessage` is used to style the `noEventsMessage`. Might be something i missed, but i think something like `classNameNoEventsMessage` makes the purpose of the field more clear.

app/components/CollapsibleDisplayContent/index.tsx
```diff
@@ -51,15 +51,17 @@ function CollapsibleDisplayContent({
...
-        <Editor
-          onChange={() => {}}
-          onBlur={() => {}}
-          onFocus={() => {}}
-          value={content}
-          placeholder={placeholder}
-          disabled
-          domParser={domParser}
-        />
+        <i>
```
LudvigHz, line 54: I'm a bit negative to "hardcoding" this to always be italic

ivarnakken: Why not just set `font-style: italic` on the `.noEventsMessage` class?

app/routes/company/components/Company.css
```diff
@@ -6,3 +6,7 @@
   border-bottom: 1px solid var(--border-gray);
   margin-bottom: 20px;
 }
+
+.errorMessage {
```
ivarnakken, line 10: I feel like the naming shouldn't contain "error", as it's not an error.

app/components/CollapsibleDisplayContent/index.tsx
ingraso: Yeah, but on second thought I don't really think this text needs to be italic anyways

ingraso:
Fixed up in your comments and changed the company description to not be italic now. What do you guys think about the changes @LudvigHz @ivarnakken?

PR #3632, opened 2023-02-28
Circumvent uBlock origin

LudvigHz:
# Description

Circumvent uBlcok's newest filters. For some reason there is a filer that blocks https://\*/script.js.

# Result

Should not be blocked now.

# Testing

This worked on webkom.dev


ivarnakken (approved):
snoot snoot

PR #3665, opened 2023-03-11
Stop event editor from breaking in dev

ivarnakken:
# Description

There was problems with storing non-serializable `moment` objects in the redux store. Issue is fixed by turning event dates into ISO strings.

Multiple places used the dates directly as `moment` objects, which now needs to be wrapped in a `moment()` function. This is the way it's done elsewhere in our codebase.

# Result

https://user-images.githubusercontent.com/69514187/224502008-9c1cf9a8-9e89-4462-ae7c-f39358eac162.mov

# Testing

See video above. 

More testing is needed, as I was experiencing some weird behaviour.

Pages will crash if you try to perform `moment` related functions on a string, so I need to make sure there's no such thing left.

### DIFF
--- a/app/reducers/events.ts
+++ b/app/reducers/events.ts
@@ -176,11 +176,11 @@ export default createEntityReducer({
 function transformEvent(event: EventType) {
   return {
     ...event,
-    startTime: moment(event.startTime),
-    endTime: moment(event.endTime),
+    startTime: event.startTime && moment(event.startTime).toISOString(),
+    endTime: event.endTime && moment(event.endTime).toISOString(),
     activationTime:
-      event.activationTime !== null ? moment(event.activationTime) : null,
-    mergeTime: event.mergeTime && moment(event.mergeTime),
+      event.activationTime && moment(event.activationTime).toISOString(),
+    mergeTime: event.mergeTime && moment(event.mergeTime).toISOString(),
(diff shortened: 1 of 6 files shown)


app/reducers/events.ts
```diff
@@ -208,7 +208,9 @@ export const selectUpcomingEvents = createSelector(selectEvents, (events) =>
   events.filter((event) => event.isUsersUpcoming)
 );
 export const selectSortedEvents = createSelector(selectEvents, (events) =>
```
erlingfn, line 210: This might be out of scope, but could it be possible to add some types to these functions, to make it easier to maintain in the future?

app/routes/events/utils.ts
```diff
@@ -236,8 +237,7 @@ export const transformEvent = (data: TransformEvent) => ({
   unregistrationDeadline: calculateUnregistrationDeadline(data),
   unregistrationDeadlineHours: calculateUnregistrationDeadlineHours(data),
   pools: calculatePools(data),
-  useCaptcha: true,
-  // always use Captcha, this blocks the use of CLI
+  useCaptcha: config.environment === 'ci' ? false : data.useCaptcha,
```
erlingfn, line 240: How did this work before?

erlingfn (approved):
LGTM

eikhr (approved):
nice!

LudvigHz (approved):

PR #3678, opened 2023-03-14
Fix bug where select blurred on select for mobile

norbye:
# Description

When on mobile react-select would call onBlur on select. This caused the value to change, but very soon after that it would change back to the original value (empty for new fields/events)

The actual fix is only on the new line 141, the rest is improved types.

# Result

Now it doesn't blur on select (which it didn't do on desktop and now doesn't on mobile either)

# Testing

I have only tested it on the edit event view, but I assume that it will work the same way everywhere it is used.

It is not actually tested on mobile, only on chrome devtools, but the bug was there to begin with so I hope this will also resolve it for mobile.

---

Resolves ABA-328

### DIFF
--- a/app/components/Form/SelectInput.tsx
+++ b/app/components/Form/SelectInput.tsx
@@ -1,9 +1,11 @@
-import Select, { type ThemeConfig } from 'react-select';
+import Select from 'react-select';
 import Creatable from 'react-select/creatable';
 import mazemapAutocomplete from '../Search/mazemapAutocomplete';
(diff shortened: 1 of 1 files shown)


LudvigHz (approved):
LGTM

ingraso (approved):
:rocket:

ivarnakken (approved):
Looks good! 🚀 Thanks a lot!!!

You may use [`ngrok`](https://ngrok.com/) to test if it works.

PR #3687, opened 2023-03-15
Standardize the use of `border-radius`

ivarnakken:
# Description

It is now set globally by three variables.

Makes the webapp look less random, and it's now much easier to change the radius if we want to.

# Testing

Went through ALL uses of `border-radius`.

### DIFF
--- a/app/components/InfoBubble/InfoBubble.css
+++ b/app/components/InfoBubble/InfoBubble.css
@@ -15,7 +15,7 @@
   width: 70px;
   height: 70px;
   background-color: #d13c32;
-  border-radius: 35px;
+  border-radius: 50%;
 }
 
 .data,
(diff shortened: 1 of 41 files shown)


app/components/InfoBubble/InfoBubble.css
```diff
@@ -15,7 +15,7 @@
   width: 70px;
   height: 70px;
   background-color: #d13c32;
-  border-radius: 35px;
+  border-radius: 50%;
```
LudvigHz, line 18: Lol nice :medal_sports:

LudvigHz (approved):
LGTM

Super nice :100: 

I like how we're slowly approaching our own version of a tiny tailwind :nerd_face:

ivarnakken:
> I like how we're slowly approaching our own version of a tiny tailwind 🤓

haha yes 🙈 
It's so satisfying standardizing our codebase. Now, by simply changing some few variables you can change the entire look and feel of the webapp.

PR #3679, opened 2023-03-14
General changes to the interest form

ivarnakken:
# Description

[Minor changes to the interest form](https://github.com/webkom/lego-webapp/commit/08e55cc7bb4036cd624174749fac45ebda5b2337) 

- Add a helmet
- Make header look like other headers
- readmify some text
- Minor changes to some labels (especially the English translations)

---
  
[Migrate interest form to react-final-form](https://github.com/webkom/lego-webapp/commit/ef674cde678f87efd6735d1d567c79c3a9954248)

# Testing

- [x] I have thoroughly tested my changes.

Tried to create an interest, also with "extra comments".

---

Resolves ABA-58

PR #3689, opened 2023-03-15
Map over list of eventComment

Arashfa0301:
# Description

Improved companyInterestPage by mapping over a list of eventComment field instead of directly having the code snippets for each comment

# Testing

### DIFF
--- a/app/routes/companyInterest/components/CompanyInterestPage.tsx
+++ b/app/routes/companyInterest/components/CompanyInterestPage.tsx
@@ -482,6 +482,16 @@ type Props = {
   officeInTrondheim: boolean;
   initialValues: any;
 };
+type EventDetailType = {
+  norwegian: string;
+  english: string;
+};
+type EventCommentFieldType = {
+  name: string;
+  eventTitle: EventDetailType;
+  eventDescription: EventDetailType;
+  eventComment: EventDetailType;
+};
 
 const validate = createValidator({
   company: [required()],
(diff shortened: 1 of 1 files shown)


ivarnakken:
haha sorry, but you'll have to change up the logic here with the [migration over to `react-final-form`](https://github.com/webkom/lego-webapp/pull/3679). The values need to be spied on.

app/routes/companyInterest/components/CompanyInterestPage.tsx
```diff
@@ -882,139 +935,27 @@ const CompanyInterestPage = (props: Props) => {
...
-              name="companyToCompanyComment"
-              component={TextEditor.Field}
-              rows={10}
-              className={styles.textEditor}
-            />
-          </div>
-        )}
+        {eventCommentFields.map((event, i) => {
+          return (
+            event.active && (
```
LudvigHz, line 941: It makes more sense to filter the list instead. Either here or in the declaration.

app/routes/companyInterest/components/CompanyInterestPage.tsx
```diff
@@ -882,139 +935,27 @@ const CompanyInterestPage = (props: Props) => {
...
+              <div className={styles.topline} key={i}>
```
LudvigHz, line 942: You should use another property as the key here. It looks like `event.name`  is unique so use that f.ex.

app/routes/companyInterest/components/CompanyInterestPage.tsx
Arashfa0301: This code snippet is now updated. Could you check and tell me if filtering is needed now or not.

Arashfa0301: I could try to put `showComment` and the `spyValue` part in the `eventCommentFields` list, but maybe filtering isn't necessary now??

app/routes/companyInterest/components/CompanyInterestPage.tsx
```diff
@@ -899,215 +945,32 @@ const CompanyInterestPage = (props: Props) => {
...
-                      className={styles.textEditor}
-                    />
-                  </div>
-                )
-              );
+            {eventCommentFields.map((event) => {
+              return spyValues((values: CompanyInterestFormEntity) => {
+                const showComment = values.events?.some(
+                  (e) => e.name === event.name && e.checked === true
+                );
```
norbye, line 952: Haven't looked into how these values work exactly, so I'm just throwing it out there so you can decide if it makes sense.

As you're checking the array for events that are .checked (I assume that you have covered all the event names in the array above - so that the `e.name === event.name` check would only make sure that specific event is checked), could it be an option to instead do something like the following?
```javascript
spyValues((values: CompanyInterestFormEntity) => 
    values.events?.filter(event => event.checked).map(event => {
        const eventFields = eventCommentFields[event.name]
        return (...);
    })
)
```
For it to make sense, `eventCommentFields` would need to be an object with what is currently the `name: ""` value as the key.

Just an alternative approach to do the same thing, but it would shorten it down a little bit more, and remove the need for the `some()` calls and boolean values making it a little bit more readable(:

eikhr: ➕

eikhr (approved):
Huge improvement! I like @norbye's suggestion, but I think it's fine as it is too:)

ivarnakken:
@Arashfa0301 Any updates on this? Would be nice to eventually get it in

eikhr:
@Arashfa0301 can we get this merged?

ivarnakken:
@Arashfa0301

ivarnakken:
Closing this since it the branch is stale and too far behind `master` (a rewrite will be much easier).

PR #3693, opened 2023-03-17
Add average to survey results

juniwbjerde:
# Description
Add the average result to questions where the options start with a number as requested by Fagkom.

PS: I am aware that just adding the average based on wheter the first part of the question text is a number is a kind of hacky way to do this. I considered adding a field in the question which mapped the answers to a number and then show average when this field was set or something like that, but I imagine it would just cause more bugs as that opens up for the user to add wrong mapping etc. So yes, this is not the cleanest way to do it, but i figured it would be the most idiot-proof way.

# Result
The average shows up right under the grapgh for the question results. The number is in a pill with a color indicating the size of the number.

Before:
image

After:
image

The pill colors are in this range:
image
image
image
image

# Testing

---

Resolves ABA-303


app/routes/surveys/components/Submissions/Results.tsx
```diff
@@ -166,6 +175,11 @@ const Results = ({
                           dataKey="selections"
                         />
                       )}
+                      <div>
+                      {questionIsNumberic &&
+                        <p>Gjennomsnitt: {getAverage(graphData[question.id])}</p>
+                      }
+                      </div>
```
ivarnakken, line 182: Is the `<div>` necessary?

ivarnakken (approved):
Kind of hacky, yes, but this is still a nice MVP 🚀 
Might consider adding an info tooltip to clarify that the number may be wrong, idk.

PeterJFB (approved):
Nice and simple. Couple of small type fixes + linting and it should be good 👍

app/routes/surveys/components/Submissions/Results.tsx
```diff
@@ -138,6 +146,7 @@ const Results = ({
           const graphType = graphOptions.find(
             (a) => a.value === question.displayType
           );
+          const questionIsNumberic = question.options.reduce((result, option) => result && /^\d+/.test(option.optionText), true);
```
PeterJFB, line 149: typo: numberic => numeric or number :)

app/routes/surveys/components/Submissions/Results.tsx
juniwbjerde: Nope, absolutely not. Will remove!

app/routes/surveys/components/Submissions/Results.tsx
juniwbjerde: Ah good catch!

ivarnakken (approved):
Woah very cool with the colors! Makes it soo much easier to scroll through the results quickly! 😍

It's kind of difficult to read the number on the yellow pill on dark mode, but I think that can be iterated on later, as we're probably want to introduce more yellow shades anyways.

You just need to lint this and we're good to go 🚀

PR #3694, opened 2023-03-18
Filter out unnecessary dash at the start of a quote source

haukkagu:
# Description
People keep adding a dash at the beginning of a quote source, even though a dash is added automatically when submitting. This PR will resolve this issue by removing a dash if it's the first non-space character of the source.

# Result
**Before**
image

**After**
image

# Testing

I've tested for numerous sources. Notably, you can still have a source like "Ola-Nordmann" without the "Ola-" part being cut out. Again, it will only remove the dash if there are ONLY spaces in front of the dash.

---

Resolves ABA-338

### DIFF
--- a/app/routes/quotes/components/AddQuote.tsx
+++ b/app/routes/quotes/components/AddQuote.tsx
@@ -30,7 +30,22 @@ const validate = createValidator({
 });
 
 const AddQuote = ({ addQuotes, actionGrant }: Props) => {
-  const onSubmit = withSubmissionErrorFinalForm(addQuotes);
+  const removeUnnecessaryDash = (source: string) => {
+    if (source === undefined) return undefined;
+
+    const dashIndex = source.indexOf('-');
+    if (source.slice(0, dashIndex).match(/^ *$/)) {
+      source = source.slice(dashIndex + 1).trim();
+    }
+
+    return source;
+  };
+
+  const onSubmit = (quote: { text: string; source: string }) =>
+    withSubmissionErrorFinalForm(addQuotes)({
+      text: quote.text,
+      source: removeUnnecessaryDash(quote.source),
+    });
 
   return (
     <div className={styles.root}>
@@ -86,7 +101,7 @@ const AddQuote = ({ addQuotes, actionGrant }: Props) => {
                   currentQuote={{
                     id: 1,
                     text: values.text || 'Det er bare å gjøre det',
-                    source: values.source || 'Esso',
+                    source: removeUnnecessaryDash(values.source) || 'Esso',
                     approved: true,
                     contentTarget: '',
                     reactionsGrouped: [],


app/routes/quotes/components/AddQuote.tsx
```diff
@@ -30,7 +30,23 @@ const validate = createValidator({
 });
 
 const AddQuote = ({ addQuotes, actionGrant }: Props) => {
-  const onSubmit = withSubmissionErrorFinalForm(addQuotes);
+  const removeInitialDash = (source: string) => {
+    if (source == undefined) return undefined;
```
ivarnakken, line 34:
```suggestion
    if (source === undefined) return undefined;
```

app/routes/quotes/components/AddQuote.tsx
```diff
@@ -30,7 +30,23 @@ const validate = createValidator({
...
+
+    return source;
+  };
+
+  const onSubmit = (quote: { text: string; source: string }) => {
+    return withSubmissionErrorFinalForm(addQuotes)({
+      text: quote.text,
+      source: removeUnnecessaryDash(quote.source),
+    });
+  };
```
ivarnakken, line 49: When all the function does is return something, you may use an implicit return. Not necessary, rather just nice to know.

```suggestion
  const onSubmit = (quote: { text: string; source: string }) => (
    withSubmissionErrorFinalForm(addQuotes)({
      text: quote.text,
      source: removeUnnecessaryDash(quote.source),
    })
  );
```

ivarnakken (approved):
Awesome! 💯 I've wanted this for a long time!

ingraso (approved):
Nice work! Have also wanted this feature for a while :rocket:

app/routes/quotes/components/AddQuote.tsx
```diff
@@ -30,7 +30,22 @@ const validate = createValidator({
 });
 
 const AddQuote = ({ addQuotes, actionGrant }: Props) => {
-  const onSubmit = withSubmissionErrorFinalForm(addQuotes);
+  const removeUnnecessaryDash = (source: string) => {
+    if (source === undefined) return undefined;
+
+    const dashIndex = source.indexOf('-');
+    if (source.slice(0, dashIndex).match(/^ *$/)) {
+      source = source.slice(dashIndex + 1);
```
norbye, line 38: Could you also add a trim for spaces after the dash to prevent double spacing if someone inserts " -   Esso", to make the text consistent?

norbye (approved):

app/routes/quotes/components/AddQuote.tsx
haukkagu: I also thought about this! However, it seems like it does that already.
image

norbye: Ahh right, the browser does that for you when displaying it. Still wouldn't hurt to remove it to have cleaner data in the database, but that makes it less important

haukkagu: Good point. I'll fix it right away

PR #3711, opened 2023-03-23
Add slugs to events

ollfkaih:
# Description

To make links prettier, show the slug of the event in the url bar instead of the id. Currently, /events/id will be redirected to /events/slug and /events/slug will work and the url will not change.

Feedback appreciated, backend is here: https://github.com/webkom/lego/pull/3260

# Result

### After: 

https://user-images.githubusercontent.com/64247965/227374605-57846b4a-7429-41e8-be51-19988787e905.mov  
  
###
Before, url would be `http://localhost:3000/events/14`

# Testing

Seems to work fine 🤷 

---

### DIFF
--- a/app/routes/events/EventEditRoute.ts
+++ b/app/routes/events/EventEditRoute.ts
@@ -13,6 +13,7 @@ import { uploadFile } from 'app/actions/FileActions';
 import { LoginPage } from 'app/components/LoginForm';
 import {
   selectEventById,
+  selectEventBySlug,
   selectPoolsWithRegistrationsForEvent,
   selectRegistrationsFromPools,
   selectWaitingRegistrationsForEvent,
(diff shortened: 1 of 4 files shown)


Arashfa0301 (approved):
Fantastico

ivarnakken (approved):
Cool 💯

LudvigHz (approved):
Nice. Seems like the best solution to a difficult issue like this :ok_hand: 

LGTM

app/routes/events/EventEditRoute.ts
```diff
@@ -29,10 +30,13 @@ import {
 } from './utils';
 
 const mapStateToProps = (state, props) => {
-  const eventId = props.match.params.eventId;
-  const event = selectEventById(state, {
-    eventId,
-  });
+  const eventIDorSlug = props.match.params.eventId;
```
eikhr, line 33: I feel like this should be `eventIdOrSlug`. With `eventIDorSlug` I read it as "event i dor slug".

eikhr, line 33:
```suggestion
  const eventIDorSlug = props.match.params.eventIDorSlug;
```
Hasn't this changed because of path change in the `RouteWrapper`?

ollfkaih: Yes

PR #3723, opened 2023-03-28
Extend Card component with a `danger` prop

ivarnakken:
# Description

Replaces our current use of the `eventWarning` css class across the webapp.

Probably going to iterate on the styling later when extending the component further soon with `info`, `success` and `warning`.

# Result

Top card is the current look, while the bottom is with the change.
<p align="center">
	Screenshot 2023-03-29 at 00 12 04
	Screenshot 2023-03-29 at 00 11 45
</p>

# Testing

Went through all uses of `styles.eventWarning`. The danger card might be applicable to other places, but that can come later.

### DIFF
--- a/app/actions/GalleryPictureActions.ts
+++ b/app/actions/GalleryPictureActions.ts
@@ -138,7 +138,7 @@ function uploadGalleryPicturesInTurn(files, galleryId, dispatch) {
   };
 
   const uploadPictureWithErrorhandler = (file) =>
-    uploadPicture(file).catch((error) => {
+    uploadPicture(file).catch(() => {
       dispatch({
         type: GalleryPicture.UPLOAD.FAILURE,
         error: true,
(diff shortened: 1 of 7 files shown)


LudvigHz:
Live it!

But the dark mode component is a lot less legible now. Any way you can change it up so it's easier to read?

LudvigHz (approved):

ivarnakken:
> Also, I still feel like it's "off". Not just the contrast, but it's a bit too much IMO. It kinda screams "SOMETHING IS HORRIBLY WRONG YOU'RE ABOUT TO DIE" and not just "you have a prikk". That is subjective though, but just my thought. The light mode one is much "nicer" in that sense.

I specifically asked you about this earlier. I'm planning on extending it with a `warning` prop as well, which might be fitting. As I've said, we can iterate on the styling later.

ivarnakken:
@LudvigHz I'll iterate on the styling together with the introduction of the other card types, but that'll be separated into a new PR.

PR #3732, opened 2023-03-30
Only calculate penalty reg time if heedPenalties

LudvigHz:
# Description

We don't check if the event uses penalties before calculating the
activation time. Also the calculation subtracted hours instead of adding
for some reason.

Also fixes random type errors in the mix here.

# Result

The correct time is shown

# Testing

- [x] I have thoroughly tested my changes. I have not tested anything :)). Probably flawless :tm:.

---

Resolves ABA-356

### DIFF
--- a/app/routes/events/components/JoinEventForm.tsx
+++ b/app/routes/events/components/JoinEventForm.tsx
@@ -17,6 +17,8 @@ import type {
   User,
   EventRegistration,
   EventRegistrationStatus,
+  Penalty,
+  Event,
 } from 'app/models';
 import { selectPenaltyByUserId } from 'app/reducers/penalties';
 import { selectUserByUsername } from 'app/reducers/users';
@@ -34,7 +36,6 @@ import styles from './Event.css';
 import withCountdown from './JoinEventFormCountdownProvider';
 import PaymentRequestForm from './StripeElement';
 
-type Event = Record<string, any>;
 export type Props = {
   title?: string;
   event: Event;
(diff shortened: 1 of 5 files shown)


app/routes/events/components/JoinEventForm.tsx
```diff
@@ -477,7 +476,6 @@ const JoinEventForm = (props: Props) => {
                       component={TextInput.Field}
                       label={feedbackLabel}
                       className={styles.feedbackText}
-                      fieldClassName={styles.feedbackField}
```
ivarnakken, line 480: Did the css class not exist, or did you forget to remove it? Or is it used anywhere else? I'm on my phone so I can't check sorry

ivarnakken (approved):
LGTM! Thanks for cleaning up the other stuff as well🎖️

ivarnakken:
> I have thoroughly tested my changes. I have not tested anything :)). Probably flawless ™️.

💀

app/routes/events/components/JoinEventForm.tsx
LudvigHz: yeah it doesn't exist. TS gives you an error when you use styles that don't exist :))

LudvigHz:
I have now tested the changes, and it was not flawless :upside_down_face:.

I realized the subtraction is to "remove" the penalty hours from LEGO, so we display the actual activation time. I guess this might be more clear as well. Why we're doing it in the frontend I don't know.

ivarnakken (approved):

PR #3737, opened 2023-03-31
Add a special component for an special event

Arashfa0301:
Special component for special days 🥸


LudvigHz (changes requested):
Fix date

Didn't look any at the special component tho but I'm hoping it's good

PR #3738, opened 2023-04-02
Fix check for special day

erlingfn:
# Description

The current check only checks if the time is after 11:00 April 1st. This means that april fools is active after April 1st. This commit makes it so that "special day" is between 11:00 and 18:00 on April 1st.

# Result

No visual changes

# Testing

---


ingraso (approved):
Nice catch!

Arashfa0301 (approved):
Looks good to me. But shouldn't we just delete this special component? A little lame using this next year too isn't it. We can always make a new special component each year.

PR #3739, opened 2023-04-03
add more padding to bottom of quicklinks

emirdero:
readme overflows on the quicklinks dropdown menu on my iphone 12 pro (in google chrome):
337782690_1253395138885401_2802626941147422091_n

This pull request adds a bit more padding to the bottom of the menu (32px->48px), a safe margin is ~50px https://developer.apple.com/documentation/apple_news/margin


ivarnakken (approved):
Thanks 🚀

PR #3740, opened 2023-04-03
Revert "Add a special component for an special event"

LudvigHz:
Reverts webkom/lego-webapp#3737

It's not a special day anymore :grimacing:


erlingfn (approved):
Yeet

PR #3749, opened 2023-04-06
Bump `ionicons` to v7

ivarnakken:
# Description

Two major versions up, but no breaking changes that affect us. Nice to do every now and then.

Mostly `ariaHidden` -> `aria-hidden`, but read on the changes here: 
v6: https://github.com/ionic-team/ionicons/releases/tag/v6.0.0
v7: https://github.com/ionic-team/ionicons/releases/tag/v7.0.0

# Result

No visual changes.

# Testing

Icons still work as expected.

### DIFF
--- a/server/pageRenderer.ts
+++ b/server/pageRenderer.ts
@@ -158,8 +158,8 @@ export default function pageRenderer({
            window.__IS_SSR__ = ${isSSR};
            ${preparedStateCode}
         </script>
-        <script type="module" src="https://unpkg.com/ionicons@5.5.2/dist/ionicons/ionicons.esm.js"></script>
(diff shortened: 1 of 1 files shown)


norbye (approved):

norbye:
Might be an idea to add our own aria-label attribute in the future as it's no longer auto-added (and we could probably do better than the non-descriptive auto-added labels), but not necessarily in the scope of this PR

ivarnakken:
> Might be an idea to add our own aria-label attribute in the future as it's no longer auto-added (and we could probably do better than the non-descriptive auto-added labels), but not necessarily in the scope of this PR

Yes, definitely. Out of scope here, but I'll write it up as a todo!

LudvigHz (approved):

PR #3754, opened 2023-04-07
Show correct status on registration pill

ivarnakken:
# Description

Currently, it would always show "Venteliste" when being admin
registered by someone in Webkom. All the if/else statements made
it difficult to work with. This is better.

# Result

| Before | After |
:------------------:|:-----------------------------:
image | image

# Testing

Tested all different statuses, with and without "Webkom register".


ingraso (approved):
Good catch! :tada:

SmithPeder (approved):

LudvigHz (approved):
LGTM

PR #3755, opened 2023-04-07
Align field labels appropriately

ivarnakken:
# Description

CheckBox and RadioButton fields are given inline labels, which makes
them align appropriately.

# Result

**Before**
image

**After**
image

---

| Before | After |
:--------------------:|:--------------------------:
image | image
# Testing

Testes the functionality of checkboxes and radio buttons, both inside and outside forms.

The spacing on radio buttons outside forms are a bit weird, but I can now get going on redesign its look.

---

Resolves https://github.com/webkom/lego/issues/3060


app/components/Form/Field.tsx
```diff
@@ -102,55 +102,81 @@ export function createField(Component: ComponentType<any>, options?: Options) {
...
+    );
+
+    const content = inlineLabel ? (
+      <Flex gap={7}>
+        {component}
+        {labelComponent}
+      </Flex>
+    ) : (
+      <>
+        {labelComponent}
```
ingraso, line 168: Why are these in the opposite order as opposed to line 163-164?

ingraso (approved):
Very nice changes! Visually the changes look great, I only left one question regarding the code :rocket:

app/components/Form/Field.tsx
ivarnakken: It's because I think it looks better to have the label on the right side.  It's also the most normal way to do it, ref.
the search result of "checkboxes":
image

On line 163-164 the label is above the component, so that's why it comes before.

ollfkaih (approved):
LGTM 🚀

app/components/Form/Field.tsx
ingraso: Okay! Yeah, I totally agree with having the checkbox to the left and label to the right!

LudvigHz (approved):
LGTM

Finally the most annoying issue fixed :heart_eyes:

itsisak (approved):
LGTM

app/routes/bdb/components/BdbPage.tsx
```diff
@@ -3,6 +3,7 @@ import { Component } from 'react';
 import { Helmet } from 'react-helmet-async';
 import { Content } from 'app/components/Content';
 import TextInput from 'app/components/Form/TextInput';
+import Flex from 'app/components/Layout/Flex';
```
eikhr, line 6:
```suggestion
```

unused

eikhr (approved):
LGTM, have no idea what is causing the "Text content does not match server-rendered HTML."-error

ivarnakken:
@eikhr It passes cypress now ... I suppose it was the unused import, even though that was in the bdb, which should not have affected the interest form 🤔 Thanks!

PR #3756, opened 2023-04-07
Stop default card content from being flex aligned

ivarnakken:
# Description

Kind of an ugly fix imo., but it'll be cleaned up in the soon-to-be iteration. Nevertheless good to get this fixed in the meantime.

# Result

| Before | After |
image | image
image | image

# Testing

`danger` card still works.


SmithPeder (approved):
💩

PR #3757, opened 2023-04-08
Support changing group roles in the membership table

ivarnakken:
# Description

[Minor changes to (interest) groups page](https://github.com/webkom/lego-webapp/pull/3757/commits/e469ee2c8ce41b4f67b04a0058efcef50b47208e)

---

[Support changing group roles in the membership table](https://github.com/webkom/lego-webapp/pull/3757/commits/e9956ced3c931e9b7aa378679fd69e5f521f4216) 
Previously, you had to remove the user from the group and re-add them
back with a different role. You may now simply change the user's role
directly in the membership table.

---

[Type group roles correctly](https://github.com/webkom/lego-webapp/pull/3757/commits/00dff1856cbb61f7d178a7e359ee77cf7c47ef71)

# Result

**Before**
image

**After**

https://user-images.githubusercontent.com/69514187/230737973-c19d4003-72e3-4ec2-bf9d-64e00d22c788.mov

# Testing

I have tested it thoroughly, but I plan on testing it even more ... Might even write an E2E spec for it 😨

---

Resolves https://github.com/webkom/lego/issues/3051

### DIFF
--- a/app/actions/GroupActions.ts
+++ b/app/actions/GroupActions.ts
@@ -2,12 +2,13 @@ import { push } from 'connected-react-router';
 import callAPI from 'app/actions/callAPI';
 import { groupSchema, membershipSchema } from 'app/reducers';
 import type MembershipType from 'app/store/models/Membership';
+import type { ID } from 'app/store/models/index';
 import type { Thunk } from 'app/types';
 import { Group, Membership } from './ActionTypes';
 
 export type AddMemberArgs = {
-  groupId: number;
-  userId: number;
+  groupId: ID;
+  userId: ID;
   role: string;
 };
 export function addMember({
--- a/app/actions/GroupActions.ts
+++ b/app/actions/GroupActions.ts
@@ -1,6 +1,7 @@
 import { push } from 'connected-react-router';
 import callAPI from 'app/actions/callAPI';
 import { groupSchema, membershipSchema } from 'app/reducers';
+import type MembershipType from 'app/store/models/Membership';
 import type { Thunk } from 'app/types';
 import { Group, Membership } from './ActionTypes';
 
@@ -30,7 +31,7 @@ export function addMember({
     },
   });
 }
-export function removeMember(membership: Record<string, any>): Thunk<any> {
+export function removeMember(membership: MembershipType): Thunk<any> {
   return callAPI({
     types: Membership.REMOVE,
     endpoint: `/groups/${membership.abakusGroup}/memberships/${membership.id}/`,
--- a/app/actions/GroupActions.ts
+++ b/app/actions/GroupActions.ts
@@ -4,12 +4,13 @@ import { groupSchema, membershipSchema } from 'app/reducers';
 import type MembershipType from 'app/store/models/Membership';
 import type { ID } from 'app/store/models/index';
 import type { Thunk } from 'app/types';
+import type { RoleType } from 'app/utils/constants';
 import { Group, Membership } from './ActionTypes';
 
 export type AddMemberArgs = {
   groupId: ID;
   userId: ID;
-  role: string;
+  role: RoleType;
 };
 export function addMember({
   groupId,
--- a/app/routes/admin/groups/components/GroupMembersList.tsx
+++ b/app/routes/admin/groups/components/GroupMembersList.tsx
@@ -1,8 +1,16 @@
+import cx from 'classnames';
 import qs from 'qs';
+import { useState } from 'react';
 import { Link } from 'react-router-dom';
+import Select from 'react-select';
+import type { AddMemberArgs } from 'app/actions/GroupActions';
+import Icon from 'app/components/Icon';
+import Flex from 'app/components/Layout/Flex';
 import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
 import Table from 'app/components/Table';
+import { isCurrentUser as checkIfCurrentUser } from 'app/routes/users/utils';
 import type Membership from 'app/store/models/Membership';
+import type { CurrentUser } from 'app/store/models/User';
 import { ROLES } from 'app/utils/constants';
 import styles from './GroupMembersList.css';
 
@@ -11,6 +19,7 @@ type Props = {
   hasMore: boolean;
   groupId: number;
   memberships: Membership[];
+  addMember: (arg0: AddMemberArgs) => Promise<any>;
   removeMember: (membership: Membership) => Promise<void>;
   showDescendants: boolean;
   groupsById: Record<
@@ -31,11 +40,13 @@ type Props = {
   search: string;
   query: Record<string, any>;
   filters: Record<string, any>;
+  currentUser: CurrentUser;
 };
 
 const GroupMembersList = ({
   memberships,
   groupId,
+  addMember,
   removeMember,
   showDescendants,
   fetch,
@@ -47,22 +58,17 @@ const GroupMembersList = ({
   search,
   query,
   filters,
+  currentUser,
 }: Props) => {
-  const GroupMembersListColumns = (fullName, membership) => {
-    const { user, abakusGroup } = membership;
+  // State for keeping track of which memberships are being edited
+  const [membershipsInEditMode, setMembershipsInEditMode] = useState({});
+
+  const GroupMembersListColumns = (fullName, membership: Membership) => {
+    const { user } = membership;
     return (
-      <>
-        <ConfirmModalWithParent
-          title="Bekreft utmelding"
-          message={`Er du sikker på at du vil melde ut "${user.fullName}" fra gruppen "${groupsById[abakusGroup].name}"?`}
-          onConfirm={() => removeMember(membership)}
-        >
-          <i key="icon" className={`fa fa-times ${styles.removeIcon}`} />
-        </ConfirmModalWithParent>
-        <Link key="link" to={`/users/${user.username}`}>
-          {user.fullName} ({user.username})
-        </Link>
-      </>
+      <Link key={user.id} to={`/users/${user.username}`}>
+        {user.fullName} ({user.username})
+      </Link>
     );
   };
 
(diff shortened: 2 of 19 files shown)


app/actions/GroupActions.ts
```diff
@@ -43,6 +44,25 @@ export function removeMember(membership: Record<string, any>): Thunk<any> {
     },
   });
 }
+export function editMembership(
+  membership: MembershipType,
+  role: string
```
eikhr, line 49: Do we have an enum for this?

app/actions/GroupActions.ts
```diff
@@ -43,6 +44,25 @@ export function removeMember(membership: Record<string, any>): Thunk<any> {
...
+    types: Membership.UPDATE,
+    endpoint: `/groups/${membership.abakusGroup}/memberships/${membership.id}/`,
+    method: 'PATCH',
+    body: {
+      role,
+    },
+    schema: membershipSchema,
+    meta: {
+      id: membership.id,
+      groupId: membership.abakusGroup,
```
eikhr, line 61: Are these used?

app/routes/admin/groups/components/GroupMembersList.tsx
```diff
@@ -47,21 +54,20 @@ const GroupMembersList = ({
   query,
   filters,
 }: Props) => {
+  // State for keeping track of which memberships are being edited
+  const [editState, setEditState] = useState(
+    memberships.reduce((acc, membership) => {
+      acc[membership.id] = false;
+      return acc;
+    }, {})
```
eikhr, line 62: Is this necessary, or could we just use an empty object where all ids would be undefined (still falsy)?

eikhr (approved):
Had a few questions, but it looks very good. A much needed feature! 🙌

app/actions/GroupActions.ts
ivarnakken: Ah yes you're right!

ivarnakken: well, it's actually a `const`. Would you like something like `role: keyof typeof ROLES`? I see that `role: string` is the way that it's done elsewhere, so I would need to rewrite it like that multiple places.

app/actions/GroupActions.ts
ivarnakken: Nope, I just blindly copied the `removeMember` function above 🙈

app/routes/admin/groups/components/GroupMembersList.tsx
ivarnakken: But is it not the same? How would you write it?

ivarnakken: haha I started to doubt myself so I had to check in with the boss
image

SmithPeder: When looking at how you use `editState` and `setEditState` I actually think this shouldn't be modeled this way at all 😬 . This logic creates an object with key-value pairs that, as you say: "_State for keeping track of which memberships are being edited_".  I feel like this is overly complex and it has a couple of bad effects. 

Firstly, you need this weird `reduce` to create your initial value and it needs to hold all memberships when you're probably only going to edit a couple.

Secondly, when you're using the `setEditState` you need to do the following 👇🏻 
```js
setEditState({
  ...editState,
  [id]: false,
});
// or
setEditState({
  ...editState,
  [membership.id]: !editState[membership.id],
})
```
which is more complex and error-prone than it should be. 

1. It requires you to always have the entire `editState` in scope, when in theory you're just toggling an id to `true` or `false`.
2. It requires you to know how to use `setEditState`. People might even update the value of a state using setState (_based on the previous state_) without taking the previous state via a function 😬 This is bad practice. You should always do `setEditState((prev) => {...prev, etc...})`

My suggestion is to just create a list of the `id's` gettings edited (_as this is basically what you're modeling here anyways_). Then you can use the `toggleEditId` method without any side effects and you don't need the state in scope to use it.

```ts
const [editState, setEditState] = useState([]) 
const toggleEditId = (id) => { 
  editState.includes(id)
    ? setEditState((current) => current.filter(filterId => filterId !== id))
    : setEditState((current) => ([...current, id]))
}

// Then later in the code whenever you check if we're editing an id
if (editState.includes(membership.id))

// And when you toggle something you just need the toggleEditId(id). 
onChange={async (value: { label: string; value: string }) => {
  toggleEditId(id)
  await editMembership(membership, value.value);
}}
```

ivarnakken: > Firstly, you need this weird reduce to create your initial value and it needs to hold all memberships when you're probably only going to edit a couple.

But the state does not hold all memberships? It only hold the memberships which are being edited. 

image

I want to rename it from `editState` to `membershipsInEditMode` or something to make this clear.


---


> Secondly, when you're using the setEditState you need to do the following 👇🏻

That's literally how I do it? The previous state is taken in. I'm not sure if I'm following you here.

app/routes/admin/groups/components/GroupMembersList.tsx
```diff
@@ -71,8 +77,63 @@ const GroupMembersList = ({
     </Link>
   );
 
-  const RoleRender = (role: string) =>
-    role !== 'member' && <i>{ROLES[role] || role} </i>;
+  const RoleRender = (fullName, membership: Membership) => {
```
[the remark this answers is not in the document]
ivarnakken: It's the same for `GroupMemberListColumns` above.  To be honest, I have no idea why and how the function gets its parameters and at this point I'm too afraid to ask. Removing `fullName` just breaks it. Feel free to enlighten me.

app/routes/admin/groups/components/GroupMembersList.tsx
SmithPeder: If it starts as an empty list this is even more dumb! Why reduce nothing into nothing?

app/routes/admin/groups/components/GroupMembersList.tsx
SmithPeder, line 80: Does this work at all? The `dataIndex` for this is `role` and you just randomly pass it a fullName and Membership
image

ivarnakken: Again, no idea how `dataIndex` actually works. It just works 🤷🏼 I removed `dataIndex: 'edit'` though.

ivarnakken: It is only used for the filtering btw.

app/routes/admin/groups/components/GroupMembersList.tsx
SmithPeder: > That's literally how I do it? The previous state is taken in. I'm not sure if I'm following you here.

No, you just spread `...editState` directly. You should take it as a function `(editState) => [...editState]`

app/routes/admin/groups/components/GroupMembersList.tsx
SmithPeder: Me neither 🤷🏻 🙃 But I remember now. The `dataIndex` is used to basically keep track of everything for that column. Basically whatever you do on that column will only work if the `dataIndex` is set.

image

app/actions/GroupActions.ts
ivarnakken: I typed it correctly now 👍🏼

app/routes/admin/groups/components/GroupMembersList.tsx
ivarnakken: > No, you just spread ...editState directly. You should take it as a function (editState) => [...editState]

ah I misinterpreted your comment. I changed it.

app/routes/admin/groups/components/GroupMembersList.tsx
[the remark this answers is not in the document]
SmithPeder: I belive the render method always gets the full data for the row based on the data sent to the table, which in this case it memberships

LudvigHz:
Very cool, but I think the way this is implemented here needs to be discussed.

I big question around this is that I'm quite certain it is the way it is currently (at least partially) due to the fact that we want to keep membership history, and if you have had one role for a year and a second the year after, that's two different memberships. With this change, you are practically "deleting" the previous membership and overriding it.

I don't think this is the right way to do this. I think the feature in itself is good, but you should instead just to a delete+create, OR a patch IF the date is less than some timedelta f.ex. This will preserve the history but still allow people to easily update group roles in the frontend.

ivarnakken:
> due to the fact that we want to keep membership history,

>  but you should instead just to a delete+create

You do realize these contradict each other, right? How is deleting memberships any better? That's not how you preserve the history.

A membership relates to a group, and you should never have more than 1 membership at the same time as that makes it difficult to know which membership to show in the membership overview table.

I'll point out; the beauty of this solution is that the membership stays intact, which means your position in the table stays the same. I wouldn't want to delete and create memberships, since that would cause users to drop to the bottom of the list (at least without a bunch of extra logic). If that is the case, then you might as well just delete and re-add them like before.

Also, why exactly do you want to preserve this history? It's not like we currently do?

LudvigHz:
> You do realize these contradict each other, right? How is deleting memberships any better? That's not how you preserve the history.

Sorry, I'll be more clear. When you "delete" a membership (i.e. remove a person from a group) you just set the membership end date such that it is not active any more. So by delete, I mean performing a `DELETE` to the API, not deleting the row from the DB.

> A membership relates to a group, and you should never have more than 1 membership at the same time as that makes it difficult to know which membership to show in the membership overview table.

Yes of course. This will not change.

> I'll point out; the beauty of this solution is that the membership stays intact, which means your position in the table stays the same. I wouldn't want to delete and create memberships, since that would cause users to drop to the bottom of the list (at least without a bunch of extra logic). If that is the case, then you might as well just delete and re-add them like before.

Well yes, but you change the history. I don't think anyone cares where in the list the user is? 
You are absolutely correct in that it is the same as deleting and re-adding. But that's my point. This is what we want. This is not a feature issue, it's just a UX issue. The logic of deleting and re-adding is good, but it just doesn't make sense from a user perspective, which is why we want this change :) What happens in the background is not important as long as it does what the user expects, and that is that the user has a new role.

> Also, why exactly do you want to preserve this history? It's not like we currently do?

Yes, this is exactly what we do currently. When you delete and create a new membership with a new role, you create a new membershipHistory object and a new membership. My point here is that there is basically no difference to the user what we do here, and it's just a matter of changing the `edit()` to a `delete().then(create())`. So the tradeoff is that we don't discard all the logic and functionality we have for membership history vs basically nothing.

LudvigHz (approved):
LGTM

PR #3765, opened 2023-04-10
Fix usage of query params in quotes-list

norbye:
# Description

In the quotes-list, the query params are overridden every time the state of the ordering-select changes it's values. As this happens not only when the user changes it, but also on page load, any other query params were overridden.

Currently this was a problem because the useEffect only updated one of the two possible query params.

# Result

This still happens, but now the useEffect that does the overriding takes both the query params into account, enabling filtering in the list of non-approved quotes and navigating freely between the three pages and directly to the non-approved quotes page.

Also removed the reactions prop as it was not used.

# Testing

Tested locally and with staging, navigating through the pages and using sorting on all pages.

---

Resolves ABA-272

### DIFF
--- a/app/routes/quotes/components/QuoteList.tsx
+++ b/app/routes/quotes/components/QuoteList.tsx
@@ -12,7 +12,6 @@ type Props = {
   actionGrant: ActionGrant;
   currentUser: any;
   loggedIn: boolean;
-  reactions: Record<string, any>;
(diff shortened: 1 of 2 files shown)


ivarnakken (approved):
Interesting! 
Must have taken some time to debug - good job 🏅

ingraso (approved):
Good job! Also, nice descriptions of the solution🙌

LudvigHz (approved):

PR #3772, opened 2023-04-11
Fix bug causing unnecessary rerender of entire page

eikhr:
# Description

copilot:summary

copilot:walkthrough
​
copilot:poem

# Result

An easy-to-reproduce example was loading the notifications while on the front page. The colored bars are components that were re-rendered.

Before:
Screenshot 2023-04-11 at 23 27 39

After:
Screenshot 2023-04-11 at 23 27 13

# Testing

There aren't really any functional changes. Everything seems to work exactly the same.

---

Resolves ABA-370


LudvigHz (approved):
LGTM :shipit:

PS: the copilot description kinda sucks. Cool gimmick but would prefer a clear and consice actual description.

LudvigHz:
> Which avoids merging, and thus changing, the byId-arrays unless something has actually changed.

Shouldn't immer do that for you? Or is there some difference in positions or something?

eikhr:
> > Which avoids merging, and thus changing, the byId-arrays unless something has actually changed.
> 
> Shouldn't immer do that for you? Or is there some difference in positions or something?

Yeah, not sure why it doesn't, but I had to add the if-statement to avoid the re-rendering. It might be that immer is also using shallow equality, which would make the new array output by `mergeObjects` a different object and thus not equal.

PR #3788, opened 2023-04-14
Replace pill in survey result with tag

ivarnakken:
# Description

The colors in the tags are more "supported", and the design on the pills are outdated.

"orange" had to be introduced as a tag color.

Also fix an issue where it would print "NaN" during fetching.

# Result

| Before | After |
image | image image
image | imageimage
image | image image
image | image image

# Testing

See images above.

### DIFF
--- a/app/styles/variables.css
+++ b/app/styles/variables.css
@@ -97,9 +97,6 @@
 
   --color-yellow: #fcd703;
 
-  --color-select-hover: rgb(178, 212, 255);
-  --color-select-multi-bg: rgb(230, 230, 230);
-
   --color-red-hover-alpha: 5%;
 
   --font-size: 16px;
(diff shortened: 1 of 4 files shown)


LudvigHz (approved):

app/styles/variables.css
```diff
@@ -97,9 +97,6 @@
 
   --color-yellow: #fcd703;
 
-  --color-select-hover: rgb(178, 212, 255);
-  --color-select-multi-bg: rgb(230, 230, 230);
```
ivarnakken, line 101: These were mistakenly added [here](https://github.com/webkom/lego-webapp/commit/68051176d2f2c080ebeb2aa45940aa8a5a9acdc1#diff-c5b3e2c872c7cd711552043894959047495cb3f36501d86d1bd76a0706b6c358R100-R211).

ingraso (approved):
Much easier to read the numbers🙌

PR #3789, opened 2023-04-14
Render fields correctly on the server

ivarnakken:
# Description

On the server, the component name would be undefined, resulting in the label not being inline. But, on the client side this would actually work, resulting in an error stating that the rendered content did not match what was server-rendered.

# Result

The labels still align as expected.
<p align="center">

</p>

# Testing

For some reason I don't get to test this through `yarn ssr`, but the cypress test passed without errors so we're all good.

### DIFF
--- a/app/components/Form/CheckBox.tsx
+++ b/app/components/Form/CheckBox.tsx
@@ -48,7 +48,7 @@ const CheckBox = ({
   );
 };
 
-const RawField = createField(CheckBox);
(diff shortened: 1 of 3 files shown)


eikhr (approved):
💯

LudvigHz (approved):
Cool :shipit: 
Although I don't understand why the component name wouldn't be defined on the server :thinking:

PR #3790, opened 2023-04-14
Refactor public frontpage

eikhr:
# Description

Refactor `PublicFrontpage` component, splitting it into smaller components and reusing shared components instead of implementing the same thing in multiple places.

Notable changes:
- Public frontpage is now only composed of functional components
- Make use of `Card` component
- Create a shared component for the login/register/forgot-password section used in frontpage, header and login-page
- Use `LatestReadme` component (from logged-in frontpage) instead of custom code on public frontpage
- Use article component from articles page on public frontpage
- Remove facebook page widget

# Result

Mostly just refactoring, but some small visual changes can be seen here:

Before:
Screenshot 2023-04-15 at 18 34 35

After:
Screenshot 2023-04-15 at 18 32 04

# Poem

### <samp>🤖 Generated by Copilot at e8d8947</samp>

> _We're breaking down the walls of code_
> _Refactoring the components of doom_
> _We're simplifying the logic and UI_
> _Using `AuthSection` and `OverviewItem` to rule_

# Testing

The frontpage still loads, and shows the same information as before. The login forms still work on both front page, in header and on login-page.

### DIFF
--- a/app/components/AuthSection/AuthSection.css
+++ b/app/components/AuthSection/AuthSection.css
@@ -1,0 +1,4 @@
+.toggleButton {
+  color: var(--lego-link-color);
+  font-size: var(--font-size-small);
+}
--- a/app/routes/articles/components/Overview.tsx
+++ b/app/routes/articles/components/Overview.tsx
@@ -14,26 +14,35 @@ import styles from './Overview.css';
 
 const HEADLINE_EVENTS = 2;
 
-const OverviewItem = ({ article }: { article: ArticleWithAuthorDetails }) => (
+export const OverviewItem = ({
+  article,
+}: {
+  article: ArticleWithAuthorDetails;
+}) => (
   <div className={styles.item}>
     <Link to={`/articles/${article.id}`} className={styles.imageLink}>
-      <Image src={article.cover} placeholder={article.coverPlaceholder} />
+      <Image
+        src={article.cover}
+        alt="Article cover"
+        placeholder={article.coverPlaceholder}
+      />
     </Link>
     <h2 className={styles.itemTitle}>
       <Link to={`/articles/${article.id}`}>{article.title}</Link>
     </h2>
 
     <span className={styles.itemInfo}>
-      {article.authors.map((e) => {
-        return (
-          <span key={e.username}>
-            <Link to={`/users/${e.username}`} className={styles.overviewAuthor}>
-              {' '}
-              {e.fullName}
-            </Link>{' '}
-          </span>
-        );
-      })}
+      {article.authors.map((author) => (
+        <span key={author.username}>
+          <Link
+            to={`/users/${author.username}`}
+            className={styles.overviewAuthor}
+          >
+            {' '}
+            {author.fullName}
+          </Link>{' '}
+        </span>
+      ))}
 
       <Time time={article.createdAt} format="DD.MM.YYYY HH:mm" />
       <Tags className={styles.tagline}>
--- a/app/routes/overview/components/PublicFrontpage.tsx
+++ b/app/routes/overview/components/PublicFrontpage.tsx
@@ -70,9 +70,6 @@ const PublicFrontpage = ({ frontpage, readmes }: Props) => {
         <Card style={{ gridArea: 'links' }}>
           <UsefulLinks />
         </Card>
-        <Card style={{ gridArea: 'facebook' }}>
-          <Facebook />
-        </Card>
       </Container>
     </Container>
   );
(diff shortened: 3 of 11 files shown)


app/routes/overview/components/PublicFrontpage.tsx
```diff
@@ -1,256 +1,187 @@
...
+      <iframe
+        src="https://www.facebook.com/plugins/page.php?href=https%3A%2F%2Fwww.facebook.com%2FAbakusNTNU%2F&amp;small_header=false&amp;adapt_container_width=true&amp;hide_cover=false&amp;show_facepile=true&amp;appId=1717809791769695"
+        className={styles.facebookIframe}
+        title="facebook"
+        scrolling="no"
+        frameBorder="0"
+      />
+    </div>
+  </>
+);
```
ivarnakken, line 185: In my opinion, we could probably just remove this.

app/components/AuthSection/AuthSection.css
```diff
@@ -0,0 +1,4 @@
+.toggleButton {
+  color: var(--lego-link-color);
+  font-size: 14px;
```
ivarnakken, line 3: Could we please use one of the `font-size` variables? I really don't like using a bunch of different sizes across the webapp. Planning on standardizing everything in ABA-358.

Currently we only have support for `12px`, `16px` and `18px`. Would `var(--font-size-small)` be too small? I'm considering making `14px` the new "small" and then call `12px` `extra-small` / `xs`, as I don't see us really needed something as small as `12px`.

I now also realize that you simply copied the code, but oh well, my opinion stays the same.

app/routes/articles/components/Overview.tsx
```diff
@@ -24,16 +29,20 @@ const OverviewItem = ({ article }: { article: ArticleWithAuthorDetails }) => (
...
-            <Link to={`/users/${e.username}`} className={styles.overviewAuthor}>
-              {' '}
-              {e.fullName}
-            </Link>{' '}
-          </span>
-        );
-      })}
+      {article.authors.map(
+        (author) =>
+          typeof author === 'object' && (
```
ivarnakken, line 34: This seems a bit hacky.

app/routes/overview/components/PublicFrontpage.tsx
```diff
@@ -1,256 +1,187 @@
...
 
-    const isArticle = (item) => item.documentType === 'article';
+const HspInfo = () => (
+  <div className={styles.hsp}>
+    <h3>
+      <a href="https://www.netcompany.com/no" target="blank">
+        <Image className={styles.hspImage} src={netcompany} alt="NETCOMPANY" />
+      </a>
+    </h3>
```
ivarnakken, line 97: I'll write it up as a todo that we should support a dark theme hsp logo, because this one is very difficult to read on dark theme.

Also, why is this inside of an `h3` header? For universal accessibility?

ivarnakken (approved):
Awesome! Very much needed 🚀 🚀

app/routes/overview/components/PublicFrontpage.tsx
ollfkaih: I think this might just be due to adblocker or similar, @eikhr ?
Disabling enhanced tracking protection in firefox makes it load (means also that by default the embed is not loaded at all):
Skjermbilde 2023-04-15 kl  11 54 04

Also, with 7-8 posts/year, I agree that we might want to drop this.

ollfkaih (approved):
Much better code here now 💯

app/routes/overview/components/PublicFrontpage.tsx
ivarnakken: It's also very ugly 😬

eikhr: Yea, maybe it's best to just drop it and give more space to the useful links

app/routes/overview/components/PublicFrontpage.tsx
eikhr: Nice.
The `h3` tag was mostly to get some padding, but I suppose it helps for accessibility as well:)

app/routes/articles/components/Overview.tsx
eikhr: It was. I've fixed it now so that it shows authors on the public front page.

app/components/AuthSection/AuthSection.css
eikhr: I've changed it to `var(--font-size-small)` and I think it works fine. Making 14px the "small" seems like a good idea though.

LudvigHz (approved):
LGTM :shipit:

PR #3791, opened 2023-04-14
Give our RadioButton a retouch

ivarnakken:
# Description

It now shares the same transition time and "bounce" as the
CheckBox, which makes them resemble each other far more.

Its label now aligns itself better as well.

# Result

**Before**

https://user-images.githubusercontent.com/69514187/232163632-6b5f5248-a62a-4a29-897d-a0b6ef509847.mov

**After**

https://user-images.githubusercontent.com/69514187/232163455-a8f69e3d-6613-492c-baff-459f4293b989.mov

https://user-images.githubusercontent.com/69514187/232163444-4a2bdfb1-d0a4-4fe2-ba7d-0ad65b8a0b1d.mov

# Testing

Tested radio buttons in and outside of forms.

### DIFF
--- a/app/components/Form/RadioButton.css
+++ b/app/components/Form/RadioButton.css
@@ -1,90 +1,95 @@
-.box {
-  display: inline-flex;
-  width: 21px;
-  height: 21px;
+@import url('~app/styles/variables.css');
+
+.radioButton {
+  --border-hover: var(--color-gray-4);
+  --border-active: var(--lego-red-color);
+  --inner-circle: white;
+
+  position: relative;
 }
 
-/* Base for label styling */
-.input {
-  cursor: pointer;
-  position: absolute;
-  z-index: 1;
-  opacity: 0;
+.radioButton input,
+.radioButton svg {
+  width: 1.1rem;
+  height: 1.1rem;
+  display: block;
 }
 
 .label {
-  position: relative;
-  display: inline-block;
   cursor: pointer;
-  padding: 4px 26px;
+  user-select: none;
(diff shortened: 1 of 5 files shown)


app/components/Form/RadioButton.css
```diff
@@ -1,90 +1,95 @@
-.box {
```
ivarnakken, line 1: The entire file is rewritten, so there's no need to compare changes.

ollfkaih (approved):
Looks much more like the checkboxes, I like it :))
Cleaner code also in Joblisting, which is always nice

app/routes/joblistings/components/JoblistingRightNav.tsx
```diff
@@ -150,64 +150,44 @@ const JoblistingsRightNav = (props: Props) => {
...
-          htmlFor="deadline"
-          style={{
-            cursor: 'pointer',
+        <RadioButton
+          name="sort"
+          id="deadline"
+          label="Frist"
+          checked={order.deadline}
+          onChange={() => {
+            props.history.push({
```
LudvigHz, line 161: Would be nice if you could replace these with `replace` as well :smile:

LudvigHz (approved):
Awesome!

app/routes/joblistings/components/JoblistingRightNav.tsx
ivarnakken: Out of scope here. Write it up as a todo and we can go over all of these

PR #3794, opened 2023-04-16
Fix readme sizing on public frontpage

LudvigHz:
# Description

Fix readme sizing on public frontpage.

# Result

**before**

image

**after**
image

# Testing

Tested public and logged in pages on desktop and different mobile sizes.

---

Resolves ABA-397

### DIFF
--- a/app/routes/overview/components/PublicFrontpage.css
+++ b/app/routes/overview/components/PublicFrontpage.css
@@ -65,3 +65,9 @@
 .linkDescription {
   margin-left: 30px;
 }
+
+.latestReadme {
+  min-height: 410px;
+  max-width: 330px;
+  width: 100%;
+}
--- a/app/routes/overview/components/LatestReadme.tsx
+++ b/app/routes/overview/components/LatestReadme.tsx
@@ -1,3 +1,4 @@
+import cx from 'classnames';
 import { useEffect, useState } from 'react';
 import { Collapse } from 'react-collapse';
 import Card from 'app/components/Card';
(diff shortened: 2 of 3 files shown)


app/routes/overview/components/PublicFrontpage.css
```diff
@@ -65,3 +65,8 @@
 .linkDescription {
   margin-left: 30px;
 }
+
+.latestReadme {
+  min-height: 442px;
+  width: 330px;
```
ivarnakken, line 71: Why not just use percentages

haukkagu (approved):
LGTM :fire:

ivarnakken (approved):
Not a fan of the very large image, but this is still much better than the whitespace.

app/routes/overview/components/PublicFrontpage.css
LudvigHz: I guess I can use percentage for the width, but I still need some fixed size for the height, otherwise you'd end up with layout shifting when loading.

LudvigHz: Although, with `witdh: 100%` the image is going to be super huge for "middle" sized screens where you still have one column. Not super likely, but it is a drawback. Kinda feel the best might be to just set the size fixed and you shouldn't have issues like this.

LudvigHz: I changed this to `max-width`, `min-width` and `width: 100%`. That should cover all the points for responsiveness, no layout shift and reasonably sized image.

PR #3814, opened 2023-04-18
Improve typing on createEntityReducer.

LudvigHz:
The RootStore now correctly infers the type from all reducers :partying_face: 

I've also typed some reducers:
- Article reducer
- Quote reducer, and entity reducer mixin
- comment reducer mixin

aand some more random type fixes here and there.

I also set `strictMode` to `true`. This helps when fixing the types, cause we actually fix the strict mode errors as well.

# Result

No visual changes. We now have... even more TS errors :upside_down_face:. The consequence of fixing types is often that it catches more errors and incorrect types. This is for the better though.

# Testing

---

### DIFF
--- a/app/reducers/allowed.ts
+++ b/app/reducers/allowed.ts
@@ -1,5 +1,5 @@
-import { type AnyAction } from '@reduxjs/toolkit';
 import { Meta } from '../actions/ActionTypes';
+import type { AnyAction, Reducer } from '@reduxjs/toolkit';
 
 const initialState = {
   events: false,
@@ -20,7 +20,10 @@ const initialState = {
   polls: false,
 };
 
-export default function allowed(state = initialState, action: AnyAction) {
+const allowed: Reducer<typeof initialState> = (
+  state = initialState,
+  action: AnyAction
+) => {
   switch (action.type) {
     case Meta.FETCH.SUCCESS:
       return action.payload.isAllowed;
@@ -28,4 +31,6 @@ export default function allowed(state = initialState, action: AnyAction) {
     default:
       return state;
   }
-}
+};
+
+export default allowed;
--- a/app/reducers/allowed.ts
+++ b/app/reducers/allowed.ts
@@ -1,3 +1,4 @@
+import { type AnyAction } from '@reduxjs/toolkit';
 import { Meta } from '../actions/ActionTypes';
 
 const initialState = {
@@ -18,11 +19,8 @@ const initialState = {
   users: false,
   polls: false,
 };
-// export type Allowed = {
-//   [$Keys<typeof initialState>]: boolean
-// };
-export type Allowed = $ObjMap<typeof initialState, <V>(v: any) => boolean>;
-export default function allowed(state: Allowed = initialState, action: any) {
+
+export default function allowed(state = initialState, action: AnyAction) {
   switch (action.type) {
     case Meta.FETCH.SUCCESS:
       return action.payload.isAllowed;
--- a/app/reducers/quotes.ts
+++ b/app/reducers/quotes.ts
@@ -7,6 +7,11 @@ import createEntityReducer from 'app/utils/createEntityReducer';
 import joinReducers from 'app/utils/joinReducers';
 import { Quote } from '../actions/ActionTypes';
 
+/**
+ * @deprecated Do not use me
+ *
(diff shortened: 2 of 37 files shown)


app/reducers/quotes.ts
```diff
@@ -2,11 +2,17 @@ import { produce } from 'immer';
...
 import joinReducers from 'app/utils/joinReducers';
 import { Quote } from '../actions/ActionTypes';
+import type { AnyAction } from 'redux';
 
+/**
+ * !!!! DONT USE ME
+ */
 export type QuoteEntity = {
   id: ID;
   text: string;
```
norbye, line 18: Is there any reason why this is not deleted? Tried to search for it in the codebase but I couldn't find any usages of it.
Or if it is needed, could the comment be changed to use `@deprecated` to make it more visible in the places that is used that it shouldn't be used any more? (this would at least for me, make the type appear as strikethrough and give a warning when using it)
```typescript
/**
 * @deprecated Do not use
 */
export type QuoteEntity = {
  id: ID;
```

app/routes/surveys/AddSurveyRoute.ts
```diff
@@ -15,19 +16,20 @@ import SurveyEditor, {
...
 import { defaultActiveFrom } from './utils';
+import type { DeepPartial } from 'utility-types';
 
 const loadData = (props, dispatch) => {
   const { templateType, event } = qs.parse(props.location.search, {
     ignoreQueryPrefix: true,
   });
 
-  if (event) {
+  if (event && typeof event == 'string') {
```
norbye, line 26: `===`?

app/routes/surveys/AddSurveyRoute.ts
```diff
@@ -15,19 +16,20 @@ import SurveyEditor, {
...
     return dispatch(fetchEvent(event)).then((result) =>
       dispatch(fetchTemplate(result.payload.entities.events[event].eventType))
     );
   }
 
-  if (templateType) {
+  if (templateType && typeof templateType == 'string') {
```
norbye, line 32: `===`?

norbye (approved):
Beauutiful! A little big PR so it took some time to get through it, but it looks pretty nice as far as I can tell:)

app/reducers/quotes.ts
LudvigHz: hehe I didn't actually check. I just didn't want to remove it before we're sure. The idea is to leave these and go through once we have fixed up more types. Although It might be better just to remove them.

The downside is that if people don't fix them up as we go along, it'll be worse for a long time until someone fixes it.

The deprecated docstring is a lot better :upside_down_face:.

norbye: Ahh right, sounds fair enough..
If we're keeping it it would also be nice to include the reason it shouldn't be used - and when it should be deleted. 

Right now it's not very informative so I wouldn't really know what to do with it if I found it like that in a couple of weeks.

app/utils/fetchJSON.ts
```diff
@@ -2,7 +2,7 @@ import 'isomorphic-fetch';
 
 export type HttpMethod = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
 export class HttpError extends Error {
-  response: Response;
+  response: Response | undefined;
```
ivarnakken, line 5: When is the http response undefined?

app/reducers/allowed.ts
```diff
@@ -1,4 +1,5 @@
 import { Meta } from '../actions/ActionTypes';
+import type { AnyAction, Reducer } from '@reduxjs/toolkit';
```
ivarnakken, line 2: I see that you're mixing between importing `AnyAction` from `redux` and `@reduxjs/toolkit`. Probably doesn't matter, but still nice to be consistent?

app/store/models/Comment.d.ts
```diff
@@ -3,12 +3,14 @@ import type User from 'app/store/models/User';
...
-  author: User;
+  text: string | null;
+  author: User | null;
   contentTarget: ContentTarget;
   createdAt: Dateish;
   updatedAt: Dateish;
   parent: ID | null;
 }
+
+export default Comment;
```
ivarnakken, line 16: Double export

eikhr (approved):

PR #3828, opened 2023-04-23
Extend Icon component with more props

ingraso:
# Description
There is a lot of custom styling for different looks of icons. Now we have the option of using the danger, edit or success prop, which adds common styling for the icons with this prop.

# Result
Mostly, the changes are not visual except some changes of sizes and the fact that the icons in FontAwesome and IonIcons are different. Some of the visual changes are provided below, but check out the branch if you want to see them all, or let me know in the comments.

The following change affects Abacard:
**Before**
image

**After**
image

**Before**
image

**After**
image

Examples of changes because of the icon library used:
**Before**
image

**After**
image

**Before**
image

**After**
image

This pencil icon is not the nicest imo., but the only other possible option I find on IonIcons is this one:
image

# Testing

I have visually compared all the changes with prod.
---

Resolves ABA-367

### DIFF
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -74,20 +74,12 @@ type TimeStampProps = {
 const TimeStamp = ({ event }: TimeStampProps) => {
   return (
     <div className={styles.eventTime}>
-      <Flex alignItems="center">
-        <Icon
-          name="calendar-number-outline"
-          size={20}
-          style={{ cursor: 'pointer', marginRight: '10px' }}
-        />
+      <Flex alignItems="center" gap={10}>
+        <Icon name="calendar-number-outline" size={20} />
         <Time time={event.startTime} format="ll" />
       </Flex>
-      <Flex alignItems="center">
-        <Icon
-          name="time-outline"
-          size={20}
-          style={{ cursor: 'pointer', marginRight: '10px' }}
-        />
+      <Flex alignItems="center" gap={10}>
+        <Icon name="time-outline" size={20} />
         <Time time={event.startTime} format="HH:mm" />
       </Flex>
     </div>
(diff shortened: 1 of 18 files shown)


app/components/EventItem/index.tsx
```diff
@@ -78,16 +78,12 @@ const TimeStamp = ({ event }: TimeStampProps) => {
         <Icon
           name="calendar-number-outline"
           size={20}
-          style={{ cursor: 'pointer', marginRight: '10px' }}
```
ingraso, line 81: These were removed because it doesn't make sense to be able to click on these icons.

app/routes/bdb/components/BdbDetail.tsx
```diff
@@ -266,7 +255,7 @@ export default class BdbDetail extends Component<Props, State> {
           </tr>
         ));
     const title = (
-      <span>
+      <Flex alignItems="center">
```
ivarnakken, line 258: Could probably benefit from a bit of `gap` as well, but still a good change!

app/components/EventItem/index.tsx
```diff
@@ -78,16 +78,12 @@ const TimeStamp = ({ event }: TimeStampProps) => {
...
+          style={{ marginRight: '10px' }}
```
ivarnakken, line 81: Could replace the inline styling with a `gap` prop on `Flex` if you'd like, but not super important

app/routes/users/components/UserSettingsOAuth2.css
```diff
@@ -5,14 +5,3 @@
 .copied {
   color: var(--color-green-6);
 }
```
ivarnakken, line 7: While you're at it, you could remove this and add `success={copied}` to the icon in `app/routes/users/components/UserSettingsOAuth2.tsx`.

ivarnakken (approved):
Awesome!!! 😍 Looks much cleaner now!

app/components/EventItem/index.tsx
ivarnakken: We should probably add a `pointer` prop or something, since I see inline styling like this everywhere. But that can be done later!

ingraso: Yeah, I agree! Thought about it, but I'll rather create a separate issue for it :ok_hand:

PR #3839, opened 2023-04-25
Move meeting end when setting start time after end

erlingfn:
# Description

When creating a new meeting one often have to set the date multiple times, both for the end time field and start time field. This commit changes the form to work similar to a lot of other meeting forms in changing the end time if one moves the start time past the end time. Also changes the default time to something more probable.

# Result
Peek 2023-04-25 20-12

# Testing

---

Resolves ABA-438

### DIFF
--- a/app/routes/meetings/MeetingCreateRoute.ts
+++ b/app/routes/meetings/MeetingCreateRoute.ts
@@ -26,8 +26,8 @@ const mapStateToProps = (state, props) => {
   return {
     user: props.currentUser,
     initialValues: {
-      startTime: time(17, 15),
-      endTime: time(20),
+      startTime: time(16, 15),
+      endTime: time(18),
       report: EDITOR_EMPTY,
       useMazemap: true,
     },
--- a/app/routes/meetings/components/MeetingEditor.tsx
+++ b/app/routes/meetings/components/MeetingEditor.tsx
@@ -1,5 +1,6 @@
 import { unionBy } from 'lodash';
-import { Field } from 'react-final-form';
+import moment from 'moment-timezone';
+import { Field, FormSpy } from 'react-final-form';
 import { Helmet } from 'react-helmet-async';
 import { useHistory } from 'react-router-dom';
 import { Content } from 'app/components/Content';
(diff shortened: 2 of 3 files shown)


app/routes/meetings/MeetingCreateRoute.ts
```diff
@@ -26,8 +26,8 @@ const mapStateToProps = (state, props) => {
   return {
     user: props.currentUser,
     initialValues: {
-      startTime: time(17, 15),
-      endTime: time(20),
+      startTime: time(16, 15),
+      endTime: time(18),
```
ivarnakken, line 30: I see that both Google Calendar and Outlook sets the "default" time to the current time (or rather the closest half hour). Could we perhaps do something similar?

app/routes/meetings/components/MeetingEditor.tsx
```diff
@@ -183,6 +184,23 @@ const MeetingEditor = ({
...
+                  onChange={({ values }) => {
+                    const startTime = moment(values.startTime);
+                    const endTime = moment(values.endTime);
+                    if (endTime.isBefore(startTime)) {
+                      form.change(
+                        'endTime',
+                        startTime
+                          .clone()
+                          .add(1, 'hour')
+                          .add(45, 'minutes')
```
ivarnakken, line 198: Any specific reason for choosing 1 hour and 45 minutes and not just 2 hours or something?

ivarnakken (approved):
Awesome! Much needed feature 😍

app/routes/meetings/components/MeetingEditor.tsx
erlingfn: Because a lot of student meetings last 1 hour and 45 minutes :)

app/routes/meetings/MeetingCreateRoute.ts
erlingfn: Sure, I think 15 minutes past the next hour would be appropriate for our users. Although I am very much open to suggestions.

erlingfn: After discussing it at the office we concluded that a hard coded time suited our needs better

ivarnakken: Why? Simply because _we_ have meetings at 16?

app/routes/meetings/components/MeetingEditor.tsx
ivarnakken: Sure, with akademisk kvarter ™️ that makes sense

PR #3840, opened 2023-04-26
Implement Plausible analytics into event statistics

ivarnakken:
# Description

[Implement Plausible analytics into event statistics](https://github.com/webkom/lego-webapp/pull/3840/commits/134dfc9e8f77ec1ea58ed28ab8765717f25e116c)

Very useful for publicity.

Considering removing "event duration" later since it doesn't really make sense.
Might also implement "bounce rate", but not sure how useful that is.

# Result

image

image

# Testing

Since it's a new feature, it should not break anything else.

### DIFF
--- a/app/actions/EventActions.ts
+++ b/app/actions/EventActions.ts
@@ -7,6 +7,7 @@ import {
   eventAdministrateSchema,
   followersEventSchema,
 } from 'app/reducers';
+import type { ID } from 'app/store/models';
 import type { Thunk, Action } from 'app/types';
 import createQueryString from 'app/utils/createQueryString';
 import { Event } from './ActionTypes';
@@ -415,6 +416,16 @@ export function isUserFollowing(eventId: number): Thunk<any> {
     method: 'GET',
     meta: {
       errorMessage: 'Henting av interesse feilet',
+    },
+  });
+}
+export function fetchAnalytics(eventId: ID): Thunk<Promise<Action>> {
+  return callAPI({
+    types: Event.FETCH,
+    endpoint: `/events/${String(eventId)}/statistics/`,
+    method: 'GET',
+    meta: {
+      errorMessage: 'Henting av analyse feilet',
     },
   });
 }
--- a/app/routes/events/components/EventAdministrate/Statistics.tsx
+++ b/app/routes/events/components/EventAdministrate/Statistics.tsx
@@ -1,12 +1,13 @@
-import type { Event, EventRegistration, Group } from 'app/models';
+import type { EventRegistration, Group } from 'app/models';
 import EventAttendeeStatistics from 'app/routes/events/components/EventAttendeeStatistics';
+import type { DetailedEvent } from 'app/store/models/Event';
 
 interface Props {
   committees: Group[];
   revueGroups: Group[];
   registered: EventRegistration[];
   unregistered: EventRegistration[];
-  event: Event;
+  event: DetailedEvent;
 }
 
 const Statistics = ({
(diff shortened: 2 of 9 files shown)


app/routes/events/components/EventAdministrate/Statistics.tsx
```diff
@@ -1,12 +1,13 @@
...
 import EventAttendeeStatistics from 'app/routes/events/components/EventAttendeeStatistics';
+import type { DetailedEvent } from 'app/store/models/Event';
 
 interface Props {
   committees: Group[];
   revueGroups: Group[];
   registered: EventRegistration[];
   unregistered: EventRegistration[];
-  event: Event;
+  event: DetailedEvent;
```
LudvigHz, line 10: :star_struck:

LudvigHz: Would be nice if you could do this for the other types here as well :grimacing:

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -216,19 +223,147 @@ const isEventFromPreviousSemester = (eventStartTime: Dateish): boolean => {
...
+        {Object.values(metrics).map((metric, i) => (
+          <>
+            <Flex column justifyContent="center" key={metric.title}>
+              <span className={styles.metricHeader}>{metric.title}</span>
+              <span className={styles.metricNumber}>
+                {metric.value.toLocaleString('no-NO')}
+              </span>
+            </Flex>
+
+            {i !== Object.values(metrics).length - 1 && (
```
LudvigHz, line 286: Could you write this as `i < Object.keys(metrics).length - 1`? Just a bit easier to understand :smile:

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -216,19 +223,147 @@ const isEventFromPreviousSemester = (eventStartTime: Dateish): boolean => {
...
+        </AreaChart>
+      </ResponsiveContainer>
+    </>
+  );
+};
+
+type Props = {
+  eventId: ID;
+  registrations: EventRegistration[];
+  unregistrations: EventRegistration[];
```
LudvigHz, line 353: Please use the updated types here. `app/store/models/Registration`

app/actions/EventActions.ts
```diff
@@ -418,3 +419,13 @@ export function isUserFollowing(eventId: number): Thunk<any> {
     },
   });
 }
+export function fetchAnalytics(eventId: ID): Thunk<Promise<Action>> {
```
LudvigHz, line 422: Since you're using the data from this directly, it would be nice if you could type this properly :)

LudvigHz:
I also think you can just remove the visit time thing. It's not very useful at all.

app/actions/EventActions.ts
ivarnakken: I typed it now, but to do so I had to make a bunch of changes to the call api. Take a look.

app/store/createStore.ts
```diff
@@ -72,4 +74,8 @@ const createStore = (
 export default createStore;
 
 export type Store = ReturnType<typeof createStore>;
-export type AppDispatch = Store['dispatch'];
+export type AppDispatch = ThunkDispatch<
+  RootState,
+  { getCookie: GetCookie },
+  AnyAction
+>;
```
LudvigHz, line 81: No. This is just bypassing the incorrect types we have elsewhere. This is not correct either.

app/actions/EventActions.ts
ivarnakken: I removed the typing commit since it didn't work and got out of scope here. You can now find it on branch `type-actions`.

PR #3860, opened 2023-05-02
Add dummy Plausible api key to drone step

ivarnakken:
# Description

Caused the `api` step to fail


ollfkaih (approved):
if it works, it works

PR #3871, opened 2023-05-05
Add slugs to joblistings and articles and use them in common places

ollfkaih:
# Description

- Fix the infinite loop when redirecting from /{id}
- Add slugs to joblistings and articles
- Use slugs in common links and on frontpage

# Result

No visual changes except for url

# Testing

Seems to work fine now

---

Resolves ABA-212

### DIFF
--- a/app/routes/articles/ArticleDetailRoute.ts
+++ b/app/routes/articles/ArticleDetailRoute.ts
@@ -25,7 +25,6 @@ type Params = {
 
 const mapStateToProps = (state, props) => {
   const { articleIdOrSlug } = props.match.params;
-  console.log('articleIdOrSlug', articleIdOrSlug);
   const article = !isNaN(articleIdOrSlug)
     ? selectArticleById(state, {
         articleId: articleIdOrSlug,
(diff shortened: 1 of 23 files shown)


app/routes/articles/ArticleDetailRoute.ts
```diff
@@ -36,6 +44,11 @@ const mapStateToProps = (state, props) => {
     });
   });
   const emojis = selectEmojis(state);
+  setTimeout(() => {
+    if (article?.slug && articleIdOrSlug !== article.slug) {
+      props.history.replace(`/articles/${article.slug}`);
+    }
+  }, 0);
```
ivarnakken, line 51: Did you try out `useHistory`?

This is quite hacky, but still better than nothing. I'd like a comment here to explain why the timeout is needed though.

ivarnakken (approved):
Cool! 🏅

app/routes/articles/ArticleDetailRoute.ts
ollfkaih: Read the rules of hooks (See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem).
useHistory cannot be used here.

I can add an explainer.

ivarnakken: Yes it wouldn't work in the route of course, but why not move it to the page component?

ollfkaih: Well, that might work just as well, I can do some experimentation with it

ivarnakken: Since it's just a `replace` I don't really mind where it is

ollfkaih: I've moved it for Joblistings and Articles, Events will have to stay like this until EventDetail is rewritten to a function component

LudvigHz (approved):

PR #3873, opened 2023-05-06
Fix survey template picker

eikhr:
# Description

This was very buggy. It was calling a function `destroy()` that didn't exist, that I couldn't even find it in old code, so idk how that has ever worked. The template picker dropdown also has a hidden button, that it was still possible to hit, causing the dropdown to be toggled twice and instantly close again. This is a super hacky workaround and should definitely be refactored later on, but for now I set the height and width to 0 so that it is impossible to hit.

# Result

It is actually possible to use the survey templates again.

# Testing

Seems to work locally:)
---

Resolves ABA-465

### DIFF
--- a/app/routes/surveys/components/SurveyEditor/SurveyEditor.tsx
+++ b/app/routes/surveys/components/SurveyEditor/SurveyEditor.tsx
@@ -7,14 +7,14 @@ import { Link } from 'react-router-dom';
 import Button from 'app/components/Button';
 import { Content } from 'app/components/Content';
 import Dropdown from 'app/components/Dropdown';
-import { TextInput, SelectInput, DatePicker } from 'app/components/Form';
+import { DatePicker, SelectInput, TextInput } from 'app/components/Form';
 import LegoFinalForm from 'app/components/Form/LegoFinalForm';
 import Icon from 'app/components/Icon';
 import Flex from 'app/components/Layout/Flex';
 import { ConfirmModalWithParent } from 'app/components/Modal/ConfirmModal';
 import Time from 'app/components/Time';
 import type { EventType } from 'app/models';
-import { eventTypeToString, EVENT_CONSTANTS } from 'app/routes/events/utils';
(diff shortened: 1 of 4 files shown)


juniwbjerde (approved):
LGTM!

app/routes/surveys/components/SurveyEditor/SurveyEditor.tsx
```diff
@@ -64,21 +60,22 @@ function TemplateTypeDropdownItems({
...
-              e.stopPropagation();
-              destroy();
-              push(link(eventType));
-            }}
-          >
-            {eventTypeToString(eventType)}
-          </Link>
-        </Dropdown.ListItem>
-      ))}
+      {(Object.keys(EVENT_CONSTANTS) as (keyof typeof EVENT_CONSTANTS)[]).map(
```
ivarnakken, line 63:
```suggestion
      {(Object.keys(EVENT_CONSTANTS) as EventType[]).map(
```
Is this not the same?

app/routes/surveys/components/surveys.css
```diff
@@ -191,8 +191,11 @@
 }
 
 .templateDropdown {
-  position: absolute;
   color: transparent;
+  position: absolute;
+  bottom: 0;
+  height: 0;
+  width: 0;
```
ivarnakken, line 198: Wouldn't hurt with a comment but okay

ivarnakken (approved):
Nice! 💯 Much cleaner!

Works fine for me btw

app/routes/surveys/components/SurveyEditor/SurveyEditor.tsx
eikhr: They should be equal, but it wouldn't actually check that the keys are the same strings as EventType

app/routes/surveys/components/surveys.css
eikhr: It wouldn't hurt, but I'm planning on refactoring it soon™️

PR #3875, opened 2023-05-07
Fix theme toggle button bugs

Bestem0r:
# Description

- Adds background hover effect to the theme change button in the drop down menu.
- Fixes clicking on the toggle theme button in the header not changing the icon of the button in the drop down and vice versa.
- Adds debounce to theme changing to prevent "glitchy" behaviour.

# Result

| Before | After |
| -- | -- |
| image | image |
| Without background hover | With background hover |
| image | image |
| Not synchronised icons | Synchronised icons

# Testing

1. Click on your profile picture to open the drop down menu.
2. Hover over the change theme button and verify that there's a hover effect present.
3. Click on the change theme button and verify that the icon in the header also changes.
4. Spam click the header change theme button and verify that the site doesn't change theme after spamming has ended.

Resolves: [#2926](https://github.com/webkom/lego/issues/2926)

### DIFF
--- a/app/components/Header/Header.css
+++ b/app/components/Header/Header.css
@@ -21,9 +21,13 @@
   justify-content: space-between;
   align-items: center;
   cursor: pointer;
+  transition: background-color var(--linear-medium);
 
   & button {
     padding: 0;
+  }
+  &:hover {
+   background-color: var(--color-gray-1);
   }
 }
 
--- a/app/components/Header/Header.css
+++ b/app/components/Header/Header.css
@@ -27,7 +27,7 @@
     padding: 0;
   }
   &:hover {
-   background-color: var(--color-gray-1);
+    background-color: var(--color-gray-1);
   }
 }
 
--- a/app/components/Header/Header.css
+++ b/app/components/Header/Header.css
@@ -26,6 +26,7 @@
   & button {
     padding: 0;
   }
+
   &:hover {
     background-color: var(--color-gray-1);
   }
--- a/app/components/Header/ToggleTheme.tsx
+++ b/app/components/Header/ToggleTheme.tsx
@@ -1,4 +1,5 @@
 import cx from 'classnames';
+import { debounce } from 'lodash';
 import { useCallback, useState, useEffect } from 'react';
 import { applySelectedTheme, getTheme } from 'app/utils/themeUtils';
 import Icon from '../Icon';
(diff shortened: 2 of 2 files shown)


app/components/Header/ToggleTheme.tsx
```diff
@@ -32,6 +33,10 @@ const ToggleTheme = ({
     setIcon(getIcon());
   }, []);
 
+  window.addEventListener('themeChange', () => {
+    setIcon(getIcon());
+  });
+
```
LudvigHz, line 39: This will unfortunately have some unintended consequences (server-side rendering will break & it will register multiple eventListeners on every re-render).

If you move this into the `useEffect` above, and add a [cleanup function](https://react.dev/reference/react/useEffect#examples-connecting) to remove the listener it should work fine though :)

LudvigHz (changes requested):
Hi! Thanks for your contribution! :star: 

Just move the `eventListener` to the `useEffect` and I'm happy to merge this in!

app/components/Header/ToggleTheme.tsx
Bestem0r: Thanks for the quick review! Should be done now! 😃

LudvigHz (approved):
Awesome!

The CI is just failing on `lint` (you're missing a newline in the css :upside_down_face:) so when CI passes this can go in :+1:

eikhr (approved):
Awesome! Just come by the webkom-office in F252 if you want a LEGO-pin:) There is almost always someone there

app/components/Header/Header.css
```diff
@@ -21,10 +21,15 @@
...
   align-items: center;
   cursor: pointer;
+  transition: background-color var(--linear-medium);
 
   & button {
     padding: 0;
   }
+
+  &:hover {
+    background-color: var(--color-gray-1);
```
ivarnakken, line 31: I'd like this to match the rest of the webapp.
```suggestion
    background-color: var(--additive-background);
```

ivarnakken (changes requested):
Thanks a lot! 🚀🚀

Thinks looks nice, but I have a small nitpick 😅

app/components/Header/Header.css
Bestem0r: You're welcome! Fixed! 👍

itsisak (approved):
Awesome!💯

LudvigHz:
@Bestem0r Another slight issue here :sweat_smile:. We had a small issue with the CI that was fixed a few days ago, so you need to rebase this on master for CI to work properly.

ivarnakken (approved):

haukkagu (approved):
LGTM 🚀 Remember to squash before you merge!

PR #3876, opened 2023-05-07
Refactor confirm modal

eikhr:
# Description

Replaces the `ConfirmModalWithParent`-component which used a lot of confusing higher-order-component trickery with a simpler functional component. The refactored component uses the [Function as Child Component](https://reactpatterns.js.org/docs/function-as-child-component) pattern to pass a `openConfirmModal`-function to the children, allowing f.ex. a button to open the modal. This makes for way more readable code imo.

Also rewrote the modal logic from class-component to function with hooks.

Aaand I also did the same thing for `AttendanceModal`

# Result

No visual or functional changes, just cleaner code 🧼

# Testing

I have tested multiple confirm modals to ensure no changes in behaviour. I have also double-checked that I set the onClick correctly in all the usages of the new component.

---


app/components/Modal/ConfirmModal.tsx
```diff
@@ -63,130 +60,71 @@ type WithModalProps = {
...
+        if (closeWhenDone) {
+          setWorking(false);
+          setModalVisible(false);
+          setErrorMessage(undefined);
+        }
+        return result;
+      } catch (error) {
+        setWorking(false);
+        setErrorMessage(
+          (error as any)?.meta?.errorMessage || 'Det skjedde en feil...'
```
LudvigHz, line 102: hmmm, not a huge fan :|

(Also, wont this give a TS error with strict mode?)

LudvigHz (approved):
LGTM

app/components/Modal/ConfirmModal.tsx
eikhr: Yes I know, I hate it too. Do you know what type we are expecting here? I suspect it might be an API-error of some sort, but I didn't want to spend too much time looking into it

LudvigHz: Yeah that is what it looks like. Could you make it generic that contains the expected parameters?

app/components/UserAttendance/AttendanceModalContent.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const generateAmendedPools = (pools: Pool[]) => {
+    if (pools.length === 1) return setAmendedPools(pools);
+
+    const registrations = flatMap(pools, (pool) => pool.registrations);
+    const summaryPool = {
+      name: 'Alle',
+      registrations,
+    };
+    return setAmendedPools([summaryPool, ...pools]);
+  };
```
LudvigHz, line 75: Not related to this change directly but since you are cleaning up.. :grimacing: 

This should be a `useMemo` and the `generateAmemdedPools` should be a pure function defined outside the component. Should be faster, alot cleaner and more reacty :)

app/components/UserAttendance/AttendanceModalContent.tsx
```diff
@@ -0,0 +1,144 @@
...
+            </Flex>
+          </li>
+        ))}
+      </ul>
+
+      <Flex justifyContent="space-between" className={styles.nav}>
+        {amendedPools.map((pool, i) => (
+          <Tab
+            name={pool.name}
+            key={i}
```
LudvigHz, line 133: Hmm, bad key :worried:

LudvigHz (commented):
Cool. Some missed low-hanging fruits in the refactoring of the AttendanceModal but overall super clean. This pattern is a lot better than the old HOCs :100:

ivarnakken (approved):
Looks better! 💯 
Our modals have been a mess for a long time, so this was much needed!!

app/routes/events/components/EventAdministrate/Attendees.tsx
```diff
@@ -156,14 +156,18 @@ const Attendees = ({
...
               onConfirm={createInfoCSV}
             >
-              <Button>Eksporter deltakere til csv</Button>
-            </ConfirmModalWithParent>
+              {({ openConfirmModal }) => (
+                <Button onClick={openConfirmModal}>
+                  Eksporter deltakere til csv
+                </Button>
+              )}
+            </ConfirmModal>
```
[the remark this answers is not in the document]
LudvigHz: Because there is some important information about exporting registrations that is detailed in the modal. So it emphasizes that the user known what he is doing and agrees to the terms.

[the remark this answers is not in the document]
ivarnakken: Ye that makes sense. For some reason I thought this was for exporting company interests to csv hehe

app/components/UserAttendance/AttendanceModalContent.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export type Registration = {
+  id: ID;
+  user: PublicUser;
+};
+
+export type Pool = {
+  name: string;
+  registrations: Registration[];
+};
```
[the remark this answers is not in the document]
eikhr: No, we add "fake" a fake pool for "All" that only contains these properties:)

app/components/Modal/ConfirmModal.tsx
eikhr: Out of scope

PR #3900, opened 2023-05-13
Add support for dark theme source on images.

jonasdeluna:
# Description

Add support for dark theme source on the custom Image component using the new "darkThemeSource" argument.
Implement this for the Public Frontpage on the main sponsor.

# Result

image

image

# Testing

---

Resolves ABA-390

### DIFF
--- a/app/components/Image/Image.tsx
+++ b/app/components/Image/Image.tsx
@@ -1,5 +1,6 @@
 import cx from 'classnames';
 import { useEffect, useState } from 'react';
+import { getTheme } from 'app/utils/themeUtils';
 import styles from './Image.css';
 import type { ImgHTMLAttributes, StyleHTMLAttributes } from 'react';
 
@@ -9,6 +10,7 @@ type Props = {
   className?: string;
   alt: string;
   style?: StyleHTMLAttributes<HTMLImageElement>;
+  darkThemeSource?: string;
 } & ImgHTMLAttributes<HTMLImageElement>;
 
 const EMPTY_IMAGE =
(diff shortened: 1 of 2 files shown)


app/components/Image/Image.tsx
```diff
@@ -37,20 +45,28 @@ const ImageComponent = (props: Props) => {
...
-
+    window.addEventListener('themeChange', () => {
+      setThemedSource(
+        getTheme() && props.darkThemeSource
+          ? (getTheme() ?? 'light') === 'light'
+            ? src
+            : props.darkThemeSource
+          : src
+      );
+    });
```
eikhr, line 56:
```suggestion
    const onThemeChange = () => {
      setThemedSource(
        getTheme() && props.darkThemeSource
          ? (getTheme() ?? 'light') === 'light'
            ? src
            : props.darkThemeSource
          : src
      );
    window.addEventListener('themeChange', onThemeChange});
	return () => window.removeEventListener('themeChange', onThemeChange);
```
	
Remember to unregister the event listener i the cleanup (returned) function. If not a new event listener will be added on each render but never removed, causing the app to slow down after a while.

eikhr: Oh, and also add an empty dependency array to avoid running the effect unnecessarily. No dep-array means it runs on every rerender.

eikhr: This should also be in a separate useEffect with `[src, props.darkThemeSrc]` as dependencies, to make sure it works if src changes after initializing the component.

eikhr (approved):
Nice!

PeterJFB (approved):
lgtm :100:

LudvigHz:
Considering all the places we do this ugly logic, we _really_ should have the theme in a context (or the store)

PR #3932, opened 2023-05-23
Standardize font sizes

ivarnakken:
# Description

We've long been using a wide variety of font sizes. This makes the webapp look random and pages less "connected".

All font sizes are now set by variables.

# Result

A lot of minor changes. Feel free to try it out yourself and see if you notice any difference. It should be extremely subtle, but everything is now consistent - at least to a further extent.

# Testing

Went through all changes ..

---

This also resolves ABA-247


juniwbjerde (approved):
LGTM! This is a lot cleaner than before!

PR #3941, opened 2023-05-27
Refactor DatePicker- and TimePicker-component

eikhr:
# Description

Small refactor of the DatePicker thingy. Shouldn't change too much, but rewrote to functional components and simplified the logic a bit. ~100 fewer lines of code:)

# Result

Some minor changes in appearance:

<table>
<tr>
 <td>Before</td>
 <td>After</td>
<tr>
 <td>Screenshot 2023-05-27 at 18 54 41</td>
 <td>Screenshot 2023-06-10 at 10 25 07</td>
</table>

~~I have changed the behaviour slightly to simplify the logic. In the meeting editor we would previously automatically move the end time forward if it was set before the start time, but this didn't play well with the simplified component. Now it will simply show a form validation error (which is how it works in f.ex. the event-editor).~~
Screenshot 2023-06-10 at 09 54 58

The end-time auto-moving is kept intact! It now moved the end time to ~two hours after the start time when blurring (deselecting) the start time field.

# Testing

I've gone through every usage of the component in the webapp, and checked manually that it still works as expected.

### DIFF
--- a/app/components/Form/DatePicker.css
+++ b/app/components/Form/DatePicker.css
@@ -6,13 +6,14 @@
   display: flex;
   width: 100%;
   flex-wrap: wrap;
-  margin: 0;
+  margin: 0 0 10px;
 }
 
 .dropdown {
   width: auto;
   max-width: 320px;
-  padding: 0 20px;
+  padding: 0 20px 10px;
+  flex: 1;
 }
 
 .inputField {
--- a/app/components/Form/DatePicker.tsx
+++ b/app/components/Form/DatePicker.tsx
@@ -1,7 +1,8 @@
 import cx from 'classnames';
 import moment from 'moment-timezone';
-import { Component } from 'react';
+import { useState } from 'react';
 import Dropdown from 'app/components/Dropdown';
+import Icon from 'app/components/Icon';
 import Flex from 'app/components/Layout/Flex';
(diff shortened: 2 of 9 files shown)


ivarnakken (approved):
Code looks solid, and the component looks nicer!

But I do like when the end time automatically updates after changing the start time.

erlingfn:
Excellent rewrite, yet the feature moving the end time forward will be sorely missed. Is it not possible to make this work with functional components?

erlingfn (approved):
Excellent rewrite, yet the feature moving the end time forward will be sorely missed. Is it not possible to make this work with functional components?

ivarnakken:
> Excellent rewrite, yet the feature moving the end time forward will be sorely missed. Is it not possible to make this work with functional components?

I'd like to add that we should extend the date picker to have a range functionality. Like this: https://ui.shadcn.com/docs/components/date-picker#date-range-picker

ivarnakken:
@eikhr What's the status on this? Is see there are some unresolved questions.

eikhr:
> @eikhr What's the status on this? Is see there are some unresolved questions.

Yep, I am going to add back the functionality for automatically moving end date forwards. Just haven't gotten around to it yet.

ivarnakken (approved):

PR #3942, opened 2023-05-28
Replace react-hot-loader with react-refresh

eikhr:
# Description

Faster, better and crashes less:)

As recommended by the author of `react-hot-loader`: https://github.com/gaearon/react-hot-loader#moving-towards-next-step

# Result

Better dev experience

# Testing

---


juniwbjerde (commented):
lgtm:)

config/webpack.server.js
```diff
@@ -39,6 +40,7 @@ module.exports = (env, argv) => {
 
     plugins: [
       !isProduction && new webpack.HotModuleReplacementPlugin(),
+      !isProduction && new ReactRefreshWebpackPlugin(),
```
LudvigHz, line 43: Does this work?

ivarnakken (approved):
Looks good, I guess. Near impossible to actually understand webpack conf 

Perhaps deploy this branch to prod at night time to see if it actually works? :)

config/webpack.server.js
eikhr: I think so? Any reason why it wouldn't?

LudvigHz: Haven't looked too much at this but I would think fast refresh needs some we apis to send the updated code? Don't know how it would do this in node? But if it's tested and its documented then that's great!

PR #3977, opened 2023-06-07
Update the chart and password bar colors

ivarnakken:
# Description

[Update the chart colors](https://github.com/webkom/lego-webapp/pull/3977/commits/c1f5d38efe8de906c847c8eaf79a221ed7b15bc3)

They are now based entirely on our color palette, and will therefore both match the rest of the webapp better and work better on dark mode.

Also introduce a yellow and purple color palette. I did not spend too much time picking out these colors, so they might need to be iterated on. Nevertheless they seem fine.

---

[Update colors on the password strength bar](https://github.com/webkom/lego-webapp/pull/3977/commits/14ba5d8ab60d1ad7dde835e66dc8b1e1fa103fe6)

# Result

Overall I think it looks better - here's an average example:

| Before | After |
image | image
<video src="https://github.com/webkom/lego-webapp/assets/69514187/d0bb36e5-ff32-41cd-9ef0-6099b2714e94" /> | <video src="https://github.com/webkom/lego-webapp/assets/69514187/797b379c-cdd9-465f-a249-bc9f407318ad" />

# Testing

They are only used on event statistics and survey results. Both were tested.

### DIFF
--- a/app/components/Chart/utils.tsx
+++ b/app/components/Chart/utils.tsx
@@ -4,18 +4,15 @@ export interface DistributionDataPoint {
 }
 
 export const CHART_COLORS = [
-  'var(--lego-red-color)',
-  'var(--color-blue-4)',
-  'var(--color-orange-7)',
-  'var(--color-green-8)',
-  'var(--lego-chart-purple)',
-  'var(--lego-chart-yellow)',
-  '#ff87eb',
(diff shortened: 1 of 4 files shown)


ollfkaih (approved):
LGTM

eikhr (approved):
LGTM

PR #3978, opened 2023-06-07
Give the card component an info property

ivarnakken:
# Description

[Give the card component an info property](https://github.com/webkom/lego-webapp/commit/3cc4dc46d7942043df2c6040bd21056f6b9ecb72)

---
  
[Utilize the info card](https://github.com/webkom/lego-webapp/commit/755ba4a654521799790a5883e1778581b352aae8)

# Result

image

image

---

Also resolves ABA-222

### DIFF
--- a/app/components/Card/index.tsx
+++ b/app/components/Card/index.tsx
@@ -13,6 +13,34 @@ const CardHeader = ({ children, className }: CardHeaderProps) => (
   <div className={cx(styles.header, className)}>{children}</div>
 );
 
+type CardContentProps = {
+  children: ReactNode;
+  danger?: boolean;
+  info?: boolean;
+};
+
+const CardContent = ({ children, danger, info }: CardContentProps) => {
+  let icon;
+  if (danger) {
+    icon = <Icon name="alert-circle-outline" className={styles.warningIcon} />;
+  }
+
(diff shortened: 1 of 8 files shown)


eikhr (approved):
Looks good:)

app/components/Card/index.tsx
```diff
@@ -13,13 +13,42 @@ const CardHeader = ({ children, className }: CardHeaderProps) => (
...
+  let icon;
+  if (danger) {
+    icon = <Icon name="alert-circle-outline" className={styles.warningIcon} />;
+  }
+
+  if (info) {
+    icon = (
+      <Icon name="information-circle-outline" className={styles.infoIcon} />
+    );
+  }
```
itsisak, line 32: I think maybe `if (info)` should come before `if (danger)` that way if both are true, it is going to be a danger card

itsisak (approved):
LGTM

app/components/Card/index.tsx
ivarnakken: That's a valid point. I'm planning on adding `warning` as well, and when I do I think I'll merge these into one prop so you'll never actually get to pass multiple severities.

PR #3987, opened 2023-06-09
Greatly improve the tooltip component

ivarnakken:
# Description

It is now no longer super buggy!

Firstly it will not overflow out of view when there is no space.
Through the `popper` library, it will find the most suitable postion.
No more need for the `renderDirection` prop and hacky calculations.

Another nice feature is that it will point to the right direction even
when the trigger component is flexed, which it often is. This is most
noticeable on tables.

Some minor styling changes as well, but that's not really the important
part of this change. Worth noting that there is no longer an annoying
buggy line between the arrow and the tooltip whenever there's a bad
render.

# Result

https://github.com/webkom/lego-webapp/assets/69514187/9a7480c8-7a5c-4220-8008-8252171098a0

# Testing

Went through all uses of tooltips. Tried all viewports, hard refreshes and other "stress testing".

---

Resolves https://github.com/webkom/lego/issues/2849

### DIFF
--- a/app/components/Tooltip/index.tsx
+++ b/app/components/Tooltip/index.tsx
@@ -1,5 +1,5 @@
 import cx from 'classnames';
-import { Component } from 'react';
+import { useState, useRef, useEffect } from 'react';
 import styles from './Tooltip.css';
 import type { ReactNode, CSSProperties } from 'react';
 
@@ -13,10 +13,7 @@ type Props = {
   renderDirection?: string;
   pointerPosition?: string;
 };
-type State = {
-  hovered: boolean;
-  childrenContainerWidth: number;
-};
+
 /**
  * A tooltip that appears when you hover over the component placed within.
  * The tooltip will by default be centered, however it supports a 'renderDirection'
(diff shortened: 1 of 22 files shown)


app/components/Tooltip/index.tsx
```diff
@@ -9,127 +9,72 @@ type Props = {
...
-      renderDirection,
-      pointerPosition,
-    } = this.props;
-    let renderDirectionClass = styles.renderFromCenter;
-    let startPointChildren = 2;
+const Tooltip = ({ children, content, className, style, onClick }: Props) => {
+  const [hovered, setHovered] = useState(false);
+  const tooltipRef = useRef<HTMLDivElement | null>(null);
+  const triggerRef = useRef<HTMLDivElement | null>(null);
+  const arrowRef = useRef<HTMLDivElement | null>(null);
```
eikhr, line 18:
```suggestion
  const tooltipRef = useRef<HTMLDivElement>(null);
  const triggerRef = useRef<HTMLDivElement>(null);
  const arrowRef = useRef<HTMLDivElement>(null);
```
Don't need the null in type arg here

eikhr (approved):
I don't love adding more dependencies, but it might be worth it in this case. It works significantly better than before, and the code is way simpler. Looks great!

I think it might be worth keeping the possibility to override the popover position though, f.ex. on the event attendees I think it looks better with the popover over the profile picture.

app/components/Tooltip/index.tsx
ivarnakken: Good point

ollfkaih (approved):
Looks good!

PR #3989, opened 2023-06-10
Remove broken webpack hot reload from server

eikhr:
# Description

The main point of this is to remove the ReactRefreshWebpackPlugin that I unnecessarily added to the webpack server config a while back, but I figured I could also remove the hot reloading code from the server as it doesn't seem to be working. I removed instead of fixing it as I don't really see the need to have hot reload on the server, it is rarely changed much.

# Result

Removed some code from webpack config and server code

# Testing

Hot reload still works in client-code.


ivarnakken (approved):
Sure that makes sense

PR #3991, opened 2023-06-11
Slightly improve the public overview

ivarnakken:
# Description

Small changes here and there that I believe improves it. Firstly it resembles the "private" overview page a lot more by using the same components. It feels less "square" without the heavy use of cards. I do not think this is perfect, but I thought the page needed some love, especially before next semester. I don't want to spend too much time on it right now.

# Result

| Before | After |
image | image

# Testing

Tested on different viewports. It now actually works better on mobile imo.

### DIFF
--- a/app/components/Header/Header.css
+++ b/app/components/Header/Header.css
@@ -82,7 +82,10 @@
   align-items: center;
   flex-direction: row;
   height: var(--lego-header-height);
-  padding: 0 20px;
+
+  @media (--mobile-device) {
+    padding: 0 10px;
+  }
 }
 
 .logo {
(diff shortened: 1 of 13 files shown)


Arashfa0301:
The "Velkommen til Abakus" part looks very naked to me. But that is the only thing that stands out. Great job 😍.

haukkagu (approved):
Love the new look :heart_eyes: 
Is the "Les mer om oss"-button necessary when there already is a "Om Abakus"-tab?

ivarnakken:
> Is the "Les mer om oss"-button necessary when there already is a "Om Abakus"-tab?

Not really, but it doesn't hurt to add more paths. It's also kind of "normal" to have a link on the frontpage.

ivarnakken:
> The "Velkommen til Abakus" part looks very naked to me. But that is the only thing that stands out.

Sure that's a valid point. But like I said, this is far from perfect and I'd like to iterate on it more later. The front page has _a lot_ of potential.

PR #4003, opened 2023-06-18
Add banner and update public frontpage for fp-23

norbye:
# Description

Add banner for fp-23, and change public fb-card from linking to an article to linking to ny.abakus.no.

I believe that the site is now good enough that it is better to publish it than wait, and rather iterate on the minor stuff later. Also starting to be about time that new students can find some sort of info.

# Result

Banner is now visible

# Testing

---


ivarnakken (approved):
🚼

LudvigHz (approved):

erlingfn (approved):
LGTM

PR #4013, opened 2023-06-27
Animate the modal appearance

ivarnakken:
# Description

And add a backdrop filter to make it pop out more.

# Testing

The modal still works.. even on mobile!

### DIFF
--- a/app/components/Modal/Modal.css
+++ b/app/components/Modal/Modal.css
@@ -1,12 +1,36 @@
+@keyframes fade-in {
+  from {
+    opacity: 0;
+  }
+
+  to {
+    opacity: 1;
+  }
+}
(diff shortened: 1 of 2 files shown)


app/components/Modal/Modal.css
```diff
@@ -43,7 +66,3 @@ html[data-theme='dark'] .backdrop {
     color: var(--color-gray-5);
   }
 }
-
-.confirmContainer {
-  text-align: center;
-}
```
ivarnakken, line 49: Not used.

norbye (approved):
Looks good(: haven't tested it locally so idk how it feels - but let's just get it in and tweak it if it feels weird

PR #4018, opened 2023-07-04
Give comments a retouch

ivarnakken:
# Description

[Improve the styling of comments](https://github.com/webkom/lego-webapp/commit/0ef2005375e97d2d8c693618119c9e0c7303eaeb) 

Reposition and align elements, slightly change some styling and also greatly clean up the css.

---

[Replace editor in comment form with text input](https://github.com/webkom/lego-webapp/commit/73ad4225258954c79237087d79953294bfdfb0b8) 

It's less buggy, and will probably work on mobile .. once and for all.

---

[Allow LegoFinalForm to only validate on submit](https://github.com/webkom/lego-webapp/commit/60fa7989c8d9ec3b3fcd0a21272d896a7c6c8649) 

Useful for forms that don't really need to give users real-time
feedback on inputs, such as on comments.

---

[Ability to add icons to tags through props](https://github.com/webkom/lego-webapp/commit/edc4b20a1f6d93225902fbf1bb87c23b4f6dbcf7) 

I still want the `tag` prop to only be used for strings. Better to
separate these than allow all kinds of types to be passed into the tag.

---

[Highlight content author (creator) on comments](https://github.com/webkom/lego-webapp/commit/86ed9ea0dd1d04c4a9fbfd14bef11ff403b98f62) 

Not _that_ useful, but a nice feature imo.

Implemented for events, articles and meetings. BDB and gallery also
has comments, but this feature doesn't really make sense on the BDB and the gallery creator is unknown on the client side.

# Result

The styling isn't perfect, but it's an improvement in my opinion. The comments never looked good as cards. I tried to follow industry norms, such as moving the comment form to the top instead of bottom.

<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	<tr>
		<td>image</td>
		<td>image</td>
</table>

# Testing

Comments (including nested comments) still look "good" on smaller viewports.

Form validation still works like normal for other forms that don't use the new `validateOnSubmitOnly` prop.

Author highlighting has been tested for all implementations.

---

Resolves ABA-504

### DIFF
--- a/app/components/Card/Card.css
+++ b/app/components/Card/Card.css
@@ -4,10 +4,6 @@
   border-radius: var(--border-radius-lg);
 }
 
-.tight {
-  padding: 0;
-}
-
 .shadow {
   box-shadow: var(--shadow-sm);
 }
--- a/app/components/Card/index.tsx
+++ b/app/components/Card/index.tsx
@@ -43,7 +43,6 @@ const CardContent = ({ children, danger, info }: CardContentProps) => {
 
 type Props = {
   className?: string;
-  tight?: boolean;
   shadow?: boolean;
   hideOverflow?: boolean;
   isHoverable?: boolean;
@@ -54,7 +53,6 @@ type Props = {
 function Card({
   children,
   className,
-  tight = false,
   shadow = true,
   hideOverflow = false,
   isHoverable = false,
@@ -67,7 +65,6 @@ function Card({
       className={cx(
         className,
         styles.card,
-        tight && styles.tight,
         shadow && styles.shadow,
         isHoverable && styles.isHoverable,
         danger && styles.danger,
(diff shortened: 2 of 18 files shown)


LudvigHz:
> Replace editor in comment form with text input

Nice, but did you give any though to only doing this on mobile and keeping the other ok desktop? Not necessarily any huge difference, but might be nice with formatting and history?

ivarnakken:
> Nice, but did you give any though to only doing this on mobile and keeping the other ok desktop? Not necessarily any huge difference, but might be nice with formatting and history?

@LudvigHz. 
I never really considered formatting and history, but it seems to work fine:
image

Do you think the editor is really needed at all on desktop? Currently it's an editor so it should as far as I've understood support all `lego-editor` features, but there is no GUI to utilize them? I somehow made some bullet points as seen on the image above though heh

PR #4019, opened 2023-07-04
Refactor Sidebar

haukkagu:
# Description
- Refactor the Sidebar component to utilize a hook
- The previous JSX and CSS was a mess. This looks much cleaner imo <3
- Make the sidebar fixed in mobile view in order to remove dead space.
- Fix a bug where the sidebar in mobile view would overflow vertically if too many tabs where opened

This sidebar looks basically the same for desktop view.

# Result

## Desktop view

Proof that they look about the same:

**Before**
before

**After**
after

## Mobile view

By making the sidebar fixed, you remove the "dead space" that would appear if a page was too long. This makes it so that the user doesn't have to scroll to the top of the page when opening the sidebar. It preserves the scroll of the page between opening and closing the sidebar.

**Before**

[Screencast1](https://github.com/webkom/lego-webapp/assets/66320400/c7b820ee-d9e4-4405-9bbe-5be017478d12)

**After**

[Screencast2](https://github.com/webkom/lego-webapp/assets/66320400/79c0df30-28c2-427e-8d69-983a7b1b4ecb)

It also fixes this overflow bug as described earlier:
Screenshot_20230704_221345

# Testing

I have tested it with many different window sizes and tabs in the sidebar collapsed.

---

### DIFF
--- a/app/routes/pages/components/Sidebar.css
+++ b/app/routes/pages/components/Sidebar.css
@@ -1,108 +1,49 @@
 @import url('~app/styles/variables.css');
 
-.side {
-  height: 100%;
-  display: flex;
-  justify-content: center;
-  box-sizing: border-box;
-  background-color: rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 2%);
-  margin-bottom: 5em;
-  width: 290px;
-}
-
 .sidebar {
-  margin: 0 1.1em;
-  width: 100%;
-  padding-top: 20px;
   display: flex;
   flex-direction: column;
-
-  @media (--mobile-device) {
-    margin: 0;
-    padding-right: 1.5rem;
-  }
+  width: 290px;
+  flex-shrink: 0;
+  padding: 1.1em;
+  padding-bottom: 4.4em;
+  background-color: rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 2%);
+  border-radius: var(--border-radius-lg) 0 0 var(--border-radius-lg);
+  z-index: 11;
 }
 
 .sidebarHeader {
-  margin-top: 20px;
-  width: 95%;
+  margin: 1em 0 1.4em;
+  align-self: center;
   text-align: center;
-  border-bottom: 2px solid var(--border-gray);
   padding-bottom: 0.6em;
+  width: 93%;
+  border-bottom: 2px solid var(--border-gray);
 }
 
-.sidebarPicture {
-  margin-top: 30px;
+.sidebarCloseButton {
   display: none;
-  flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


app/routes/pages/components/Sidebar.css
```diff
@@ -1,108 +1,49 @@
...
-    height: 100%;
-    background-color: rgba(var(--rgb-min), var(--rgb-min), var(--rgb-min), 7%);
-    z-index: 10;
+  .sidebarClosed {
+    display: none;
+  }
+
+  .sidebarCloseButton {
+    display: block;
+    align-self: flex-end;
```
ivarnakken, line 46: [`align-self` should in theory not apply to block-level boxes](https://developer.mozilla.org/en-US/docs/Web/CSS/align-self) - so these two lines sort of contradict each other.

app/routes/pages/components/Sidebar.tsx
```diff
@@ -16,49 +12,24 @@ type Props = {
...
-          <aside className={styles.sidebar}>
-            <div className={styles.sidebarTop}>
-              <button className={styles.sidebarCloseBtn} onClick={handleClose}>
-                <Icon name="close" size={50} />
-              </button>
-              <h2 className={styles.sidebarHeader}>Om Abakus</h2>
-            </div>
+const Sidebar = (props: Props) => {
+  const { categorySelected, currentUrl, pageHierarchy, isOpen, handleClose } =
+    props;
```
ivarnakken, line 17: You can destruct these props directly and then implicitly return the JSX below. Doesn't really matter, it changes nothing and readability isn't really improved either, but still nice to know. 

```suggestion
const Sidebar = ({ categorySelected, currentUrl, pageHierarchy, isOpen, handleClose }: Props) => {

```

app/routes/pages/components/Sidebar.tsx
```diff
@@ -16,49 +12,24 @@ type Props = {
...
-class Sidebar extends Component<Props, State> {
-  render() {
-    const {
-      categorySelected,
-      currentUrl,
-      pageHierarchy,
-      isOpen,
-      handleClose,
-    }: Props = this.props;
-    // const pictureLabel = 'Hemmelig bilde';
```
ivarnakken, line 28: 👀 core

ivarnakken (approved):
A big improvement! 😍 I love PRs like these!!

app/routes/pages/components/Sidebar.css
haukkagu: That makes sense, but it works for some weird reason. I need to set the display to something (in order to override the `display: none`) though. What should I set it to? `display: flex`?

PR #4020, opened 2023-07-04
Update the look of buttons

ivarnakken:
# Description

[Update the look of buttons](https://github.com/webkom/lego-webapp/commit/07de4885dd82f81c9a29f376ea67bd7e4b8b14e9) 

They were a bit outdated.

`flat` buttons now look (work) somewhat different.

The new buttons are flatter with less shades and depth, and most importantly more round <3 Was inspired by YouTube and Twitter's buttons. 

---

[Clean up surveys code and make small improvements](https://github.com/webkom/lego-webapp/commit/f9daf66364993bf61ea7b36f5f81548e7b6e633a) 

Small step in the right direction. A lot of bad code here to be dealt with.

This is definitely out of scope of this PR - but I didn't bother to do it in a separate PR as it sort of relies on the button changes.

# Result

<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	<tr>
		<td>
			image
			image
</td>
		<td>
			image
			image
		</td>
</table>

# Testing

All different types still work as expected. Went through most uses of `flat`.

### DIFF
--- a/app/components/Header/index.tsx
+++ b/app/components/Header/index.tsx
@@ -10,7 +10,6 @@ import type { UserEntity } from 'app/reducers/users';
 import type { ID } from 'app/store/models';
 import utilStyles from 'app/styles/utilities.css';
 import { applySelectedTheme, getOSTheme, getTheme } from 'app/utils/themeUtils';
-import Button from '../Button';
 import Dropdown from '../Dropdown';
 import NotificationsDropdown from '../HeaderNotifications';
 import Icon from '../Icon';
@@ -53,13 +52,7 @@ function AccountDropdownItems({
   return (
     <Dropdown.List>
       <Dropdown.ListItem>
-        <NavLink
-          to="/users/me"
-          onClick={onClose}
-          style={{
-            color: 'var(--lego-font-color)',
-          }}
-        >
+        <NavLink to="/users/me" onClick={onClose}>
           <strong>{username}</strong>
           <Icon name="person-circle-outline" size={24} />
         </NavLink>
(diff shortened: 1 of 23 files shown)


app/components/CommentForm/CommentForm.css
```diff
@@ -70,6 +70,6 @@
 }
 
 .submit {
-  max-width: 150px;
-  margin-top: 30px;
+  /* max-width: 150px; */
```
erlingfn, line 73: Any specific reason to comment these out instead of removing them?

app/components/Header/index.tsx
```diff
@@ -92,16 +92,17 @@ function AccountDropdownItems({
 
       <Dropdown.Divider />
       <Dropdown.ListItem>
-        <Button
-          flat
+        <Flex
```
erlingfn, line 95: While this keeps the styling consistent, did you try having it as an actual button? I feel like having it as a danger-button might fit well.

erlingfn (approved):
Great work!

app/components/CommentForm/CommentForm.css
ivarnakken: oops. It is removed in [this PR](https://github.com/webkom/lego-webapp/pull/4018) so I just wanted to quickly test that I didn't break anything without creating any conflicts. Forgot to uncomment though - thanks!

app/components/Header/index.tsx
ivarnakken: Very good point! It looks really nice as a danger button 😍 

And yes, it should have just been a `button` for it to keep some important properties such as pointer cursor etc. - I actually realized this before going to bed yesterday.

haukkagu:
Very nice!

PR #4032, opened 2023-07-12
Random bits and pieces 4

ivarnakken:
# Description

[Retouch the poll component](https://github.com/webkom/lego-webapp/commit/5cd5f66fa334a4b2cee9c2ad8d9516824b1298f3) 

It just really looked like it had been created by someone external, and
stood out from the rest of the webapp.

A lot of technical debt.

---

[Fix random bits and pieces 4](https://github.com/webkom/lego-webapp/commit/00a67b74c49dc5f7ed3819346ddd518ef69e178d) 

* Bilder -> Album
* Fix a bug where the event title in the compact event component would
  overflow
* Reduce line-height some places
* "Ikke aktive" -> "Ikke-aktive"

# Result

It's not that visible, but the code is also a bit cleaner.

<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	<tr>
	<tr>
		<td>image</td>
		<td>image</td>
	</tr>
    <tr>
		<td>image</td>
		<td>image</td>
	</tr>
    <tr>
		<td>image</td>
		<td>image</td>
	</tr>
</table>

# Testing

Poll component was tested on light and dark theme, and on smaller viewports. Also tested with short and long poll titles.

Other changes should not be breaking anything.

### DIFF
--- a/app/components/Poll/Poll.css
+++ b/app/components/Poll/Poll.css
@@ -35,7 +35,6 @@
   background-color: var(--lego-red-color);
   padding: 5px;
   border-radius: 0 var(--border-radius-sm) var(--border-radius-sm) 0;
-  font-style: italic;
   color: var(--color-absolute-white);
   height: 30px;
 }
@@ -60,8 +59,7 @@
   }
 }
 
-.pollHeader {
-  margin-bottom: 10px;
+.pollHeader a {
   font-weight: 500;
   color: var(--lego-font-color);
   line-height: 1.3;
@@ -70,7 +68,6 @@
 .voteButton {
   width: 100%;
   margin: 1px !important;
-  max-width: 400px;
 }
 
 .moreOptionsLink {
@@ -92,13 +89,14 @@
   justify-content: center;
   width: 100%;
   height: 100%;
+  cursor: pointer;
 }
 
 .blurOverlay {
   position: absolute;
   z-index: 2;
-  color: var(--color-black);
-  margin-top: 25px;
+  font-weight: 500;
+  color: var(--color-absolute-white);
 }
 
 .optionWrapper:hover .blurContainer {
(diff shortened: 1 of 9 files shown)


haukkagu (approved):
LGTM. Very nice :100: 

I've noticed that you're turning many components that were previously red to black. I'm personally a fan of keeping things red, since red is kind of the Abakus-colour.

ivarnakken:
> I've noticed that you're turning many components that were previously red to black. I'm personally a fan of keeping things red, since red is kind of the Abakus-colour.

Yes, and that is very much intentional.

In web design, and really design in general, there is a rule called the 60-30-10 rule. I can recommend [this video on it](https://www.youtube.com/watch?v=UWwNIMHFdW4).

In short, you ideally want to use your primary color (in our case red) only 10% on a page. Some very good examples of this is YouTube and Twitter, which only use their primary colors, red and blue respectively, to a minimal extent and really only when they need something to stick out. YouTube pretty much only uses red in their logo on their landing page, but that doesn't mean we forget that red is "their" color. Twitter uses their black secondary color far more than their primary, but I still don't associate Twitter with black. The same applies for Abakus; just because we don't use red everywhere we can, people won't forget that red is the Abakus color and associate us with something else. It's worth noting that this only works since our secondary colours are very neutral. I think using the dark black works as a nice secondary color (30%) and makes the webapp look both modern and sharp.

Besides, you don't want an icon like on the poll component to actually stick out since it serves no purpose but being a visual touch.

ivarnakken:
> Oh i see. Thanks for the explanation!

It's very nice that you pointed it out though! And of course, I'm all ears and we may discuss it further if there's any dissent.

haukkagu:
While you're at it, could you fix this styling issue: 
image

It feels quite pointless to create a new PR just to fix that...

haukkagu:
I see. Great job!

PR #4033, opened 2023-07-12
Let users set event location before registration type

ivarnakken:
# Description

I don't see how the location has anything to do with the registration type. I have personally created an event where I knew the location but not the registration type, so this restriction is just annoying.

# Testing

Tested that it works to create an event with a location but without any specified registration type.

### DIFF
--- a/app/routes/events/components/EventEditor/index.tsx
+++ b/app/routes/events/components/EventEditor/index.tsx
@@ -228,6 +228,41 @@ function EventEditor({
               fieldClassName={styles.metaField}
               className={styles.formField}
             />
+            <Field
+              label="Bruk MazeMap"
+              name="useMazemap"
+              component={CheckBox.Field}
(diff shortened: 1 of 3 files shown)


app/routes/events/components/EventEditor/index.tsx
```diff
@@ -256,48 +291,6 @@ function EventEditor({
...
-              />
-            )}
-            {['NORMAL', 'OPEN', 'INFINITE'].includes(
-              event.eventStatusType && event.eventStatusType.value
-            ) &&
-              (!event.useMazemap ? (
-                <Field
-                  label="Sted"
-                  name="location"
-                  placeholder="Den gode nabo, R5, ..."
```
ivarnakken, line 278: Removed `R5` from the placeholder, since that's part of MazeMap.

ollfkaih (approved):
LGTM

PR #4064, opened 2023-07-30
Fix a bunch of stuff

ivarnakken:
# Description

Picked up by this crawler:

image

The content authors object is used for comments to highlight the creator
/ author of the content in the comment section. But this would ultimately
break the event page when it would try to read id off of null.

---

[Fix misalignment of anonymous attendance status](https://github.com/webkom/lego-webapp/commit/d038eba8bc16ac29b1c499d2f860c68356b37b0b) 

It was the tooltip that caused the misalignment, so just repositioning
it fixed the bug. A maximum width was also added to prevent tooltips
with a lot of content text from just becoming one very long line.

Some minor additional styling was added by utilizing the button component,
and disabling it when the user is not authenticated.

---

[Type the event page more thoroughly](https://github.com/webkom/lego-webapp/commit/1380d5febfc298f6e40afe4c6039d911ffdd4e86) 

A lot of the store types were wrong.

# Result

<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	</tr>
	<tr>
		<td>
			image
		</td>
		<td>
			image
		</td>
	</tr>
</table>

# Testing

Highlighting content authors still works:
image

### DIFF
--- a/app/components/Image/ProfilePicture.tsx
+++ b/app/components/Image/ProfilePicture.tsx
@@ -1,28 +1,27 @@
-import type { User } from 'app/models';
+import type { UnknownUser } from 'app/store/models/User';
 import CircularPicture from './CircularPicture';
 import type { ComponentProps } from 'react';
 
-type Props = Omit<
-  {
-    user: Pick<
-      User,
-      'profilePicture' | 'username' | 'profilePicturePlaceholder'
-    >;
-    size: number;
-    className?: string;
-  } & ComponentProps<typeof CircularPicture>,
-  'placeholder' | 'src'
->;
+type Props = {
+  user: Pick<
+    UnknownUser,
+    'profilePicture' | 'username' | 'profilePicturePlaceholder'
+  >;
+  size: number;
+  className?: string;
+  alt?: string;
+} & Omit<ComponentProps<typeof CircularPicture>, 'src' | 'alt'>;
 
 const ProfilePicture = ({
   user,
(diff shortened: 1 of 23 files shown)


LudvigHz (approved):

ivarnakken:
This PR ended up fixing some other stuff as well heh sowwy

app/components/Image/ProfilePicture.tsx
```diff
@@ -1,28 +1,27 @@
...
+  size: number;
+  className?: string;
+  alt?: string;
+} & Omit<ComponentProps<typeof CircularPicture>, 'src' | 'alt'>;
 
 const ProfilePicture = ({
   user,
   size = 100,
   className,
+  alt = `${user.username}'s profile picture`,
```
LudvigHz, line 19: Why bother with the alt here when this is the only one that makes sense? Just remove the alt prop alltogether and use this one always

LudvigHz (approved):
LGTM

eikhr (approved):

app/components/Image/ProfilePicture.tsx
ivarnakken: Idk. I just thought it was nice to let users override _if_ they wanted to - but yes never should heh

LudvigHz:
@ivarnakken Can I/you merge this now? I need my branch without this flaky test :grimacing:

PR #4072, opened 2023-08-02
Fix overflowing text in tooltip on event creator page

ollfkaih:
# Result
Now looks like this  
Screenshot 2023-08-02 at 18 52 56
instead of this  

Screenshot 2023-07-22 at 18 20 45

# Testing

Looks good on various screen sizes

---

Resolves ABA-517

### DIFF
--- a/app/routes/events/components/EventEditor/index.tsx
+++ b/app/routes/events/components/EventEditor/index.tsx
@@ -555,12 +555,18 @@ function EventEditor({
               <>
                 Jeg er kjent med at jeg kun kan bruke rettighetene mine til å
                 opprette et Abakusarrangement som er i tråd med{' '}
-                <Link to="/pages/arrangementer/86-arrangementskalender">
+                <Link
+                  style={{ display: 'contents' }}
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):
🫶🏼

eikhr (approved):
Nice!
You could instead wrap the content in a `<div>` to avoid the flex positioning. I don't really love inline css, but it's up to you. I think this is fine too.

ollfkaih:
> Nice!
> You could instead wrap the content in a `<div>` to avoid the flex positioning. I don't really love inline css, but it's up to you. I think this is fine too.

Yeah, I do agree about inline css, but using a div as a wrapper would add more lines (depends on line length and prettier, I guess). Lots of divs make the document tree really messy also (which I guess it already is).   
I was thinking about maybe looking for other uses of the <Tooltip> where this could be an issue and maybe just add a prop to it, but decided not to dig into it.

haukkagu (commented):
NICE WORK

PR #4087, opened 2023-08-09
Fix production docker image

eikhr:
# Description

Soo, yarn workspaces creates a symlink from `lego-bricks` to `node-modules`. This doesn't work if we don't copy the actual files into the docker image.

I don't love this solution, but I couldn't find a way to make yarn actually copy the files. I think perhaps a good solution would be to just build `lego-bricks` into the bundle using webpack, but it wasn't suuper simple, so I might do that later.

# Result

It works now:)

# Testing

Built the image locally, and ran it successfully locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,6 +51,8 @@ ENV RELEASE ${RELEASE}
 
 COPY --from=builder /app/dist dist
 COPY --from=builder /app/dist-client dist-client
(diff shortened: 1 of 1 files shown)


LudvigHz (approved):
LGTM

I don't see anything wrong with this solution tbh.

ivarnakken (approved):

ivarnakken (approved):

PR #4099, opened 2023-08-21
Fix incorrect stripe input color

Bestem0r:
# Description

Previously, the content of the card number, expiration date and cvc input fields were hard too see when using dark mode. This PR makes the content color white when using dark mode to improve contrast.

# Result

| Before: Dark text on dark background | After: Light text on dark background |
| -- | -- |
| image | image |

# Testing

Checked that the contrast is correct in both dark and light mode.

### DIFF
--- a/app/routes/events/components/StripeElement.tsx
+++ b/app/routes/events/components/StripeElement.tsx
@@ -57,7 +57,9 @@ type CompleteStatus =
 const StripeElementStyle = {
   style: {
     base: {
-      color: '#0d0d0d',
+      color: window.matchMedia('prefers-color-scheme: dark')
+        ? '#f2f2f2'
+        : '#0d0d0d',
       letterSpacing: '0.025em',
       fontFamily: 'Source Code Pro, monospace',
       '::placeholder': {
--- a/app/routes/events/components/StripeElement.tsx
+++ b/app/routes/events/components/StripeElement.tsx
@@ -71,8 +71,8 @@ function StripeElementStyle(fontColor) {
         color: '#c81917',
       },
     },
-  }
-};
+  };
+}
 
 const CardForm = (props: CardFormProps) => {
   const [paymentStarted, setPaymentStarted] = useState(false);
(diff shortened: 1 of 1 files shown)


app/routes/events/components/StripeElement.tsx
```diff
@@ -57,7 +57,9 @@ type CompleteStatus =
 const StripeElementStyle = {
   style: {
     base: {
-      color: '#0d0d0d',
+      color: window.matchMedia('prefers-color-scheme: dark')
+        ? '#f2f2f2'
+        : '#0d0d0d',
```
ivarnakken, line 62: `color` should just be set equal to `var(--lego-font-color)` (which will also fix the bug).

ivarnakken (changes requested):
Thanks! Although this works - it's better to just use the predefined css variable

app/routes/events/components/StripeElement.tsx
Bestem0r: This will unfortunately not work since the input is actually inside an `<iframe>`. It uses a `CardNumberElement` from `react-stripe`, which handles the styling.
image

haukkagu (approved):
Great work!

app/routes/events/components/StripeElement.tsx
ivarnakken: You can still override the styling. You may just add it to the classname though, that's better

Bestem0r: The `className` prop only applies the class to the element container (outside the `<iframe>`).
Refering to this comment: https://github.com/stripe-archive/react-stripe-elements/issues/51#issuecomment-320147662

Bestem0r: @ivarnakken

Bestem0r:
Replaced `matchMedia` with `useTheme`. Is that better? @ivarnakken

ivarnakken (approved):
Thanks! 😄

PR #4100, opened 2023-08-22
Fix unset theme on initialization

haukkagu:
# Description
The theme value accessed through `getTheme()` didn't get set on initialization, so the value was set to the default value (light). On dark mode, this would cause certain bugs, like the graph artwork in the background to disappear and the netcompany logo to use its light color variant.

# Result
**before**
before

**after**
after

# Testing

It just works

---

Resolves ABA-550

### DIFF
--- a/app/utils/themeUtils.tsx
+++ b/app/utils/themeUtils.tsx
@@ -24,7 +24,7 @@ export const ThemeContextListener = () => {
 
   useEffect(() => {
     const handleThemeChange = () => dispatch(setTheme(getTheme()));
-
+    handleThemeChange();
     window.addEventListener('themeChange', handleThemeChange);
     return () => window.removeEventListener('themeChange', handleThemeChange);
   }, [dispatch]);


app/utils/themeUtils.tsx
```diff
@@ -22,13 +22,16 @@ export const ThemeContext = createContext<'dark' | 'light'>(getTheme());
 export const ThemeContextListener = () => {
   const dispatch = useAppDispatch();
 
+  const handleThemeChange = () => dispatch(setTheme(getTheme()));
   useEffect(() => {
-    const handleThemeChange = () => dispatch(setTheme(getTheme()));
-
```
norbye, line 27: Is there any reason why you didn't run `handleThemeChange` inside this useEffect? I haven't checked into how many times this useEffect is re-rendered so idk if it's a good idea - but seeing that `handleThemeChange` relies on `dispatch`, it could be prudent to call it when it changes (in case it's not properly initialized on the first run)

norbye (approved):
Idk if what I commented is a good idea, but this is an improvement either way you do it:) 👍

ivarnakken (approved):
Wondering about the same as @norbye 

Still a fix imo

app/utils/themeUtils.tsx
haukkagu: I made it it's own useEffect since I only figured it should run once on startup, but it might be safer to run it inside this useEffect? Not sure how often it rerenders either. I can investigate...

norbye: Did you find out any thing interesting?(:

PR #4105, opened 2023-08-26
Properly type actions

ivarnakken:
# Description

[Properly type the callAPI and promise middleware](https://github.com/webkom/lego-webapp/commit/7814e68d4752729ef70e559b169d571bea2e18b3) 

It is now possible to type API request!

---

[Type announcement, article, comment and company actions](https://github.com/webkom/lego-webapp/commit/6996a217d16b060daf41de4ea2a6e1fb95bd76ef)

# Result

No visual changes except this tiny fix:
<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	</tr>
	<tr>
		<td>
			image
		</td>
		<td>
			image
		</td>
</table>

# Testing

Tested all actions that were typed, even though that should not have broken anything.

### DIFF
--- a/app/actions/AnnouncementsActions.ts
+++ b/app/actions/AnnouncementsActions.ts
@@ -1,10 +1,15 @@
 import callAPI from 'app/actions/callAPI';
 import { announcementsSchema } from 'app/reducers';
 import { Announcements } from './ActionTypes';
-import type { Thunk } from 'app/types';
+import type { AppDispatch } from 'app/store/createStore';
+import type { ID } from 'app/store/models';
+import type {
+  DetailedAnnouncement,
+  ListAnnouncement,
+} from 'app/store/models/Announcement';
 
-export function fetchAll(): Thunk<any> {
-  return callAPI({
+export function fetchAll() {
+  return callAPI<ListAnnouncement[]>({
     types: Announcements.FETCH_ALL,
     endpoint: '/announcements/',
     schema: [announcementsSchema],
@@ -14,6 +19,7 @@ export function fetchAll(): Thunk<any> {
     propagateError: true,
   });
 }
+
 export function createAnnouncement({
   message,
   users,
@@ -22,11 +28,10 @@ export function createAnnouncement({
   meetings,
   fromGroup,
   send,
-}: Record<string, any>): /*AnnouncementModel*/
-Thunk<any> {
-  return (dispatch) =>
+}: Record<string, any>) {
+  return (dispatch: AppDispatch) =>
     dispatch(
-      callAPI({
+      callAPI<DetailedAnnouncement>({
         types: Announcements.CREATE,
         endpoint: '/announcements/',
         method: 'POST',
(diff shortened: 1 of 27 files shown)


eikhr:
Hmm, there's something strange here. It seems like the provided T to `callApi` gets used to set the type of `action.type`, which is a string id (ish) for that specific action. But in the typed actions you provide the entity type fetched? This seems unintentional.

Also I think the `Action`-type you have changed here is used for other things beside just API-calls, meaning the types will be wrong. A new API-call specific type would probably be a good idea

ivarnakken:
> Hmm, there's something strange here. It seems like the provided T to callApi gets used to set the type of action.type, which is a string id (ish) for that specific action. But in the typed actions you provide the entity type fetched? This seems unintentional.

Yes, that's strange. The provided `T` is the type of the entities returned. It should be typed like this:
```ts
export type Action<T = unknown> = {
  type: unknown;
  payload?: {
    actionGrant?: ActionGrant;
    entities?: T;
    next?: string;
    previous?: string;
    result?: ID;
  };
};
```

---

> Also I think the Action-type you have changed here is used for other things beside just API-calls, meaning the types will be wrong. A new API-call specific action-type would probably be a good idea

Not sure if I'm following you here. The `Action` type  is only used to type actions?

ivarnakken:
> > Not sure if I'm following you here. The `Action` type is only used to type actions?
> 
> 
> 
> Yes, but not only API-call actions. It is f.ex. used in `ToastActions.ts` for the `removeToast` action, which has a payload consisting of only an id. This doesn't fit your new `Action` type, but did fit the previous one which allowed for any payload.
> 
> 
> 
> I'm just saying you should take your new type, call it f.ex. `APIAction` and use it for all the API-actions, while letting other types of actions keep on using the old type.


Yes but all actions are currently wrong and have to be refactored.

This PR just makes it possible to actually type them

eikhr:
> > > Not sure if I'm following you here. The `Action` type is only used to type actions?
> > 
> > 
> > Yes, but not only API-call actions. It is f.ex. used in `ToastActions.ts` for the `removeToast` action, which has a payload consisting of only an id. This doesn't fit your new `Action` type, but did fit the previous one which allowed for any payload.
> > I'm just saying you should take your new type, call it f.ex. `APIAction` and use it for all the API-actions, while letting other types of actions keep on using the old type.
> 
> Yes but all actions are currently wrong and have to be refactored.
> 
> This PR just makes it possible to actually type them

The `deleteToast` action is perfectly fine? Why would it need entities in it, it simply removes the toast with the given id from the redux state

eikhr:
> > Hmm, there's something strange here. It seems like the provided T to callApi gets used to set the type of action.type, which is a string id (ish) for that specific action. But in the typed actions you provide the entity type fetched? This seems unintentional.
> 
> Yes, that's strange. The provided `T` is the type of the entities returned. It should be typed like this:
> 
> ```ts
> export type Action<T = unknown> = {
>   type: unknown;
>   payload?: {
>     actionGrant?: ActionGrant;
>     entities?: T;
>     next?: string;
>     previous?: string;
>     result?: ID;
>   };
> };
> ```

Regarding this part, I think you may have forgotten to push the latest changes? In the diff right now the action type looks like this:
```
export type Action<T = unknown> = {
  type: T;
  payload?: {
    actionGrant?: ActionGrant;
    entities?: Record<string, unknown>;
    next?: string;
    previous?: string;
    result?: ID;
  };
};
```

It should be more like what you posted over, except that entities is actually a map of entity type names and a map of entityId to the entity:
Screenshot 2023-10-18 at 16 31 12

It you look in `app/store/models/entities.ts` I have actually created a helper type for this. `NormalizedEntityPayload<EntityKeys>` will give you the type of the `entities`-field containing all the entity types specified in `EntityKeys`.

eikhr:
I think this works well now! I suggest we merge this, and then add types to the remaining actions as they are needed. They will default to `unknown`.

eikhr (approved):
I think this is good now! 
But maybe someone else should review as well, as I wrote quite a bit of it:)

PR #4118, opened 2023-09-05
Move Card, Icon and Layout components into lego-bricks

ivarnakken:
# Description

I initially only needed the card component, but since that uses the layout and icon component I had to add those as well. 

# Result

No visual changes.

# Testing

All components still work.


eikhr (approved):
Great work! Definitely some of the most important components💯

I would love some storybook stories though:)

ivarnakken:
> I would love some storybook stories though:)

heh that was in my todos. I'll get it in before merging

danielyanghansen (approved):
Looks good to me.
I assume the linter is already complaining about the import order some places 🙃

ivarnakken:
> I assume the linter is already complaining about the import order some places 🙃

What do you mean? Nothing wrong with the import order here

PR #4131, opened 2023-09-15
Migrate to `react-router` v6

ivarnakken:
# Description

A "stable" branch for react-router v6

# ⚠️ Known issues

- [x] Problems with fetching quotes when switching between the quotes list page and `/quotes?approved=false` after doing a hard refresh on the latter
- [x] Unable to fetch memberships on both `/admin/groups/{id}/members?descendants=false` and e.g. `/pages/komiteer/{id}`
- [x] Infinite fetch loop on `/events`
- [x] Infinite fetch loop on `/meetings` when the user has no prior meetings
- [x] All fields in `/admin/email/restricted/new` are disabled
- [x] Unable to view previous events by clicking on the arrow icon on e.g. `/companies/4587`

---

Resolves ABA-116


PeterJFB (approved):
Tried to run locally, works like a charm! :100:

eikhr (approved):
Not sure what the plan is here, but if this works without the rest of the PRs I guess we should get this commit in as well: https://github.com/webkom/lego-webapp/commit/21deeb2dd5aa5d954708e577ca44fa8631b18ee6

Other than that, seems good 👌

ivarnakken:
> Not sure what the plan is here, but if this works without the rest of the PRs I guess we should get this commit in as well: https://github.com/webkom/lego-webapp/commit/21deeb2dd5aa5d954708e577ca44fa8631b18ee6

I think I'll merge the batch PRs into this, and let this be the "stable" branch.

eikhr:
Screenshot 2024-01-11 at 19 50 22
Seems to be a problem with the styling on the events page. There is also something weird with ssr here, it sometimes doesn't load the events.

ivarnakken:
> Seems to be a problem with the styling on the events page. There is also something weird with ssr here, it sometimes doesn't load the events.

Yes I'm aware - I will fix that when I rebase. There was a fix for this recently merged into master

PR #4132, opened 2023-09-17
[1st batch] Start migrating to `react-router` v6

ivarnakken:
# Description

[Setup router compatibility package](https://github.com/webkom/lego-webapp/commit/656907c66410cae9ed9d9ee9e535cc3068cd0622)

Currently this does not work on the server side, so I guess we'll have to refactor everything before deploying ..
 
---

[Properly extract store from createStore on server side](https://github.com/webkom/lego-webapp/commit/21deeb2dd5aa5d954708e577ca44fa8631b18ee6)

---

[Refactor quotes route to react-router v6](https://github.com/webkom/lego-webapp/commit/a7fc5ad1af6a6fa797e91a447dd65fcf3c7ec774)

Goodbye `connect`!

# Result

No visual changes.

# Testing

I think I tested all features on the quotes route. As mentioned, this does not work on ssr because of `react-router-dom-v5-compat`.


app/routes/quotes/components/QuotePage.tsx
```diff
@@ -1,96 +1,110 @@
...
+    }
+
+    return selectQuotes(state, { pagination });
+  });
+
+  const fetching = useAppSelector((state) => state.quotes.fetching);
+
+  const query = qs.parse(location.search, {
+    ignoreQueryPrefix: true,
+  });
```
eikhr, line 79: The "load more" button doesn't work, because this qs-parsing doesn't add `approved=true`. Use the `qsParamsParser`-function:)

```suggestion
  const query = qsParamsParser(location.search);
```

eikhr: This also allows you to reuse the parsed qs on line 99

app/routes/quotes/components/QuotePage.tsx
```diff
@@ -1,96 +1,110 @@
...
+      dispatch(fetchQuote(quoteId));
+    } else {
+      dispatch(
+        fetchAll({
+          query: qsParamsParser(location.search),
+        })
+      );
+    }
+    dispatch(fetchEmojis());
+  }, [quoteId, location.search, dispatch]);
```
eikhr, line 104:
```suggestion
  usePreparedEffect(
    '',
    () => {
      if (quoteId) {
        dispatch(fetchQuote(quoteId));
      } else {
        dispatch(
          fetchAll({
            query,
          })
        );
      }
      dispatch(fetchEmojis());
    },
    [quoteId, location.search, dispatch]
  );
  ```
 Use `usePreparedEffect` to eventually support SSR:)

File: app/routes/index.tsx [hunk 3]
```diff
@@ -68,7 +69,7 @@ const AppWrapper = (props) => (
           <Route path="/pages" component={Pages} />
           <Route path="/photos" component={Photos} />
           <Route path="/polls" component={Polls} />
-          <Route path="/quotes" component={Quotes} />
+          <CompatRoute path="/quotes" component={Quotes} />
```
eikhr, line 72: Does this need to be CompatRoute? Seems to work fine without

app/routes/quotes/components/QuotePage.tsx
```diff
@@ -111,6 +125,10 @@ export default function QuotePage({
     });
   }, [history, ordering, query.approved]);
 
+  if (!loggedIn) {
+    return LoginPage;
+  }
```
eikhr, line 130: This doesn't work. We go to the 401-page because we attempt to fetch. Not sure how best to fix that...

eikhr (changes requested):
This will be a great improvement! I love how much prop-drilling you have gotten rid of already.
Some small things to fix, and SSR is a problem but you have a plan for that maybe?

app/components/Form/utils.ts
```diff
@@ -69,8 +70,9 @@ export const withSubmissionError = <Args extends unknown[], Return>(
  * withSubmissionErrorFinalForm(onSubmit)
  */
 export const withSubmissionErrorFinalForm = <Args extends unknown[], Return>(
+  dispatch: Dispatch<any>,
   onSubmit: (...args: Args) => Promise<Return>
```
eikhr, line 74: Types are wrong
```suggestion
  dispatch: AppDispatch,
  onSubmit: (...args: Args) => Thunk<Promise<Return>>
```

app/routes/quotes/components/QuotePage.tsx
ivarnakken: ah ye you're right .. I guess we should have to check before fetching 🤷🏼

app/routes/index.tsx
ivarnakken: You're right, and it is already fixed in the second batch heh

app/routes/quotes/components/QuotePage.tsx
ivarnakken: Yes, but I'll fix this in the fifth batch because I don't want to rebase and fix 4 different branches. There's some other ssr related stuff that needs to be delt with

app/routes/quotes/components/QuotePage.tsx
ivarnakken: Very nice catch!

app/components/Form/utils.ts
ivarnakken: Nice catch

app/routes/quotes/components/QuotePage.tsx
eikhr: Yea, the only way to avoid running the fetch-hook at all would be to wrap the whole page in a component that checks the login. I think just checking in the hook will be easiest.

eikhr (approved):

PR #4137, opened 2023-09-21
[4th batch] Refactor (primarily) article routes to support `react-router` v6

ivarnakken:
# Description

[Replace article editor's `redux-form` with `react-final-form`](https://github.com/webkom/lego-webapp/commit/33e3d55432d07301980da56b8795af7ad064f93b) 

Had to introduce our own `<Fields>` component, since it is not shipped with
`react-final-form`.

---

[Implement a custom `useHelmet` hook](https://github.com/webkom/lego-webapp/commit/4dbd4de049bbd026db40a636af092800f57edcd9) 

The hook encapsulates the same logic as the `helmet` hoc, but can be be
employed inside of components as a hook. It will manage the Helmet
properties as a side effect.

---

[Refactor article routes to `react-router` v6](https://github.com/webkom/lego-webapp/commit/cc611b3b6fd5c1714801dba2411eb3c676bedf26) 

Also some other minor refactoring to reactions. Less props passed around
and less unnecessary fetching!

# Result

No visual changes. However, I've been routing around a lot and I genuinely feel an improvement!

# Testing

All routes were tested. This PR also fixes a bug where the LEGO editor did not render properly on /admin/groups/

I was, however, not able to fully test the helmet changes, but it seems to be related to the broken SSR so we'll have to note this for later.


app/routes/articles/components/ArticleDetail.tsx
```diff
@@ -1,53 +1,129 @@
...
+        content: 'article',
+      },
+      {
+        property: 'og:image:width',
+        content: '500',
+      },
+      {
+        property: 'og:image:height',
+        content: '500',
+      },
```
norbye, line 103: Isn't this referencing the articles cover? Wouldn't it then be wrong to say that it's a square??

norbye (approved):
Great stuff! Love to see the props go away

app/routes/articles/components/ArticleDetail.tsx
ivarnakken: No idea - I just copied it 🤷🏼‍♂️ I'll look into it though

PR #4155, opened 2023-09-28
Feat event organizer field

jonasdeluna:
# Description

Adds event organizer field to specify responsible users.

image

image

# Result

If you've made visual changes, please include **before** and **after** images, preferably with a description. Make sure they do not contain any real user information.

# Testing

---

Depends on: https://github.com/webkom/lego/pull/3496
Resolves: ABA-570

### DIFF
--- a/app/routes/articles/ArticleEditRoute.ts
+++ b/app/routes/articles/ArticleEditRoute.ts
@@ -24,7 +24,7 @@ const mapStateToProps = (state, props) => {
   });
 
   const currentUser = selectCurrentUser(state);
-  const authors: Array<DetailedUser> = article?.authors?.length
+  const authors: DetailedUser[] = article?.authors?.length
     ? article.authors.map((e) => selectUserById(state, { userId: e }))
     : [currentUser];
 
--- a/app/routes/articles/ArticleEditRoute.ts
+++ b/app/routes/articles/ArticleEditRoute.ts
@@ -11,6 +11,7 @@ import { LoginPage } from 'app/components/LoginForm';
 import { selectArticleById } from 'app/reducers/articles';
 import { selectCurrentUser } from 'app/reducers/auth';
 import { selectUserById } from 'app/reducers/users';
+import type { DetailedUser } from 'app/store/models/User';
 import loadingIndicator from 'app/utils/loadingIndicator';
 import replaceUnlessLoggedIn from 'app/utils/replaceUnlessLoggedIn';
 import withPreparedDispatch from 'app/utils/withPreparedDispatch';
@@ -23,7 +24,7 @@ const mapStateToProps = (state, props) => {
   });
 
   const currentUser = selectCurrentUser(state);
-  const authors = article?.authors?.length
+  const authors: Array<DetailedUser> = article?.authors?.length
     ? article.authors.map((e) => selectUserById(state, { userId: e }))
     : [currentUser];
 
--- a/app/routes/events/EventCreateRoute.ts
+++ b/app/routes/events/EventCreateRoute.ts
@@ -10,7 +10,6 @@ import {
   uploadFile,
 } from 'app/actions/FileActions';
 import { LoginPage } from 'app/components/LoginForm';
-import { selectCurrentUser } from 'app/reducers/auth';
 import {
   selectEventById,
   selectPoolsWithRegistrationsForEvent,
--- a/app/routes/events/EventCreateRoute.ts
+++ b/app/routes/events/EventCreateRoute.ts
@@ -10,6 +10,7 @@ import {
   uploadFile,
 } from 'app/actions/FileActions';
 import { LoginPage } from 'app/components/LoginForm';
+import { selectCurrentUser } from 'app/reducers/auth';
 import {
   selectEventById,
   selectPoolsWithRegistrationsForEvent,
@@ -25,6 +26,8 @@ import {
   transformEventStatusType,
   EVENT_CONSTANTS,
 } from './utils';
+import { current } from 'immer';
+import { DetailedUser } from 'app/store/models/User';
 
 const mapStateToProps = (state, props) => {
   const actionGrant = state.events.actionGrant;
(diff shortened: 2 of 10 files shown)


app/routes/events/EventCreateRoute.ts
```diff
@@ -68,6 +70,8 @@ const mapStateToProps = (state, props) => {
     },
     pools: valueSelector(state, 'pools'),
   };
+  const currentUser: DetailedUser = selectCurrentUser(state);
```
ivarnakken, line 73: `EventCreateRoute` (and therefore also `EventEditor`) has already access to `currentUser` through its props. See app/routes/events/index.tsx

Also, the correct type is `CurrentUser`, not `DetailedUser`, but that doesn't matter since you can just remove this.

app/routes/events/components/EventEditor/index.tsx
```diff
@@ -306,6 +306,15 @@ function EventEditor({
               component={SelectInput.AutocompleteField}
               placeholder="Ansvar for arrangement"
             />
+            <Field
+              name="responsibleUsers"
+              label="Ansvarlige Brukere"
```
ivarnakken, line 311:
```suggestion
              label="Ansvarlige brukere"
```

app/routes/events/utils.ts
```diff
@@ -85,6 +85,7 @@ export type EditingEvent = Event & {
   hasFeedbackQuestion: boolean;
   isClarified: boolean;
   authors: Option[];
+  responsibleUsers: Option[];
```
ivarnakken, line 88: And here it's a completely new type

app/store/models/Reaction.d.ts
```diff
@@ -13,4 +13,5 @@ export interface ReactionsGrouped {
   count: number;
   hasReacted: boolean;
   reactionId?: ID;
+  user?: string;
```
ivarnakken, line 16: Is this related to this PR? We haven't merged in the backend changes for this yet

app/routes/articles/ArticleEditRoute.ts
```diff
@@ -23,7 +24,7 @@ const mapStateToProps = (state, props) => {
   });
 
   const currentUser = selectCurrentUser(state);
-  const authors = article?.authors?.length
+  const authors: Array<DetailedUser> = article?.authors?.length
```
ivarnakken, line 27: Please be consistent on this
```suggestion
  const authors: DetailedUser[] = article?.authors?.length
```
It's not _super_ important since it's equivalent, but still nice to be consistent

ivarnakken (commented):
This looks mostly good, but I do have a question:

As I see it now, if the user has not selected any responsible users - it will default to the creator. But can't we make this optional instead? There are usually multiple "responsible people" for an event (sometimes many), and I'd hate for the detailed event page to get even more crowded with names. 

I interpreted the issue to be that authors would receive all questions related to events, instead of the responsible group. Therefore, wouldn't it be clean to make this optional? In other words; to only show responsible users if they want to be shown, or else show nothing (meaning only the responsible group). I think that would make the detailed event page a tiny bit more organized.

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -333,7 +334,7 @@ export default class EventDetail extends Component<Props, State> {
               <Link to={groupLink}>{event.responsibleGroup.name}</Link>
             ) : (
               event.responsibleGroup.name
-            )}{' '}
+            )}
```
norbye, line 337: I'm on mobile so might've missed it, but have you added styling to replace this space between the group name and the email?

norbye (commented):
Agree with @ivarnakken that it would be nice to **not** show the author if the field is not populated. IMO that would make more people send emails to the entire group when a persons name is not visible.

I'm not sure how many people are normally responsible for each event - but an alternative could also be to call the field "Kontaktperson" and make it singular (or "Kontaktpersoner" and plural). In that way it wouldn't feel as natural for the event creator to add all the responsible people to the list and add more clutter to the sidebar.

jonasdeluna:
I've made it so that it's now optional and will display the author if the responsible users are not set. The responsible group will be shown at all times. Is this the behavior we want?

app/routes/events/EventCreateRoute.ts
```diff
@@ -114,6 +116,7 @@ const mapStateToProps = (state, props) => {
       }),
       registrationDeadlineHours: 2,
       unregistrationDeadlineHours: 2,
+      responsibleUsers: [],
```
norbye, line 119: To avoid making the process of creating new events more tiresome, we could set the default value of responsibleUsers to be the author here. In that way, if the responsible user is creating the event they wouldn't have to change anything, but they have the option to do so if they desire.

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -342,20 +343,46 @@ export default class EventDetail extends Component<Props, State> {
...
+                </Link>
+              ),
+            },
+          ]
+        : [
+            {
+              key: 'Forfatter',
+              value: 'Anonym',
+            },
+          ]),
```
norbye, line 384: So our thought was to basically just remove this side of the ternary and keep the previous lines

norbye (approved):
> I've made it so that it's now optional and will display the author if the responsible users are not set. The responsible group will be shown at all times. Is this the behavior we want?

I think that what @ivarnakken and I meant would be the following;  
Always display the responsible group
Display responsible users if they are set, if not display nothing (alas, never display the author)

Anyways, imo this is pretty good, so I'd say it's good to go if you want it to be, then we can always iterate.

app/routes/events/components/EventDetail/index.tsx
jonasdeluna: Ok great I'll implement the changes and get it merged 👍

app/routes/events/EventCreateRoute.ts
jonasdeluna: This was how it was before the latest commit so I can revert it back to that 👍

PR #4157, opened 2023-10-02
Update navlinks to use gap instead of margin

norbye:
Give navlinks vertical distance when they wrap.

# Description

Smol visible change

# Result

### Before
image

### After:
image

# Testing

Visited a few places they were used on mobile and desktop sizes - seems okidoks to me

---


ivarnakken (approved):
Muy bien!❤️

norbye:
> Muy bien!❤️

image

PR #4158, opened 2023-10-02
Replace abascan popups with a list of recent scans

norbye:
# Description

Changes in how abacard handles responses
* Avoids using `alert()` or other elements that require the user to interact with them (`Modal` is used, but it is set to close by itself)
* Remove custom overlay code and replace it with `Modal` component
* Search by username: Displays responses using Modal and Toast
* Scan by ABAID: Displays responses in a list displaying the 5 most recent responses
* Displays a different feedback when it is used through `/validator` route than through `Abacard` route to emphasize what the result actually means.

This does nothing to solve the double-scanning, but it handles it quite a bit more gracefully - which was the main problem to be solved. If someone wants to ratelimit the amount of times requests for the same user is sent - feel free, it's a pretty minimal change.

# Result

<table>
<tr>
 <th>
 <th> Before
 <th> After
<tr>
 <td>Before scanning
 <td>image
 <td>image
<tr>
 <td>After registering (with abaid)
 <td>image
image
Both happen pretty much at the same time
 <td>image
If the last message received matches the new message, the counter for that message updates. If not it displays as a new row.
<tr>
 <td>After registering (with search)
 <td>image
 <td>image
<tr>
 <td>After faulty register (with search)
 <td>image
 <td>image
Using toasts
</table>

# Testing

---

Resolves ABA-449

### DIFF
--- a/app/components/UserValidator/index.tsx
+++ b/app/components/UserValidator/index.tsx
@@ -1,14 +1,15 @@
 import { Button } from '@webkom/lego-bricks';
-import cx from 'classnames';
 import { get } from 'lodash';
 import { useCallback, useRef, useState } from 'react';
 import { QrReader } from 'react-qr-reader';
+import type { addToast } from 'app/actions/ToastActions';
 import goodSound from 'app/assets/good-sound.mp3';
 import Icon from 'app/components/Icon';
 import Modal from 'app/components/Modal';
 import SearchPage from 'app/components/Search/SearchPage';
 import type { User } from 'app/models';
 import type { UserSearchResult } from 'app/reducers/search';
+import { Flex } from '../Layout';
 import styles from './Validator.css';
 import type { ComponentProps } from 'react';
 import type { Required } from 'utility-types';
@@ -19,10 +20,16 @@ type Res = {
   payload: unknown;
 };
 
+type ScanResult = {
+  message: string;
+  count: number;
+};
+
 type Props = Omit<
   ComponentProps<typeof SearchPage<UserSearchResult>>,
   'handleSelect'
 > & {
+  addToast: typeof addToast;
   clearSearch: () => void;
   handleSelect: (arg0: UserWithUsername) => Promise<User | Res>;
   onQueryChanged: (arg0: string) => void;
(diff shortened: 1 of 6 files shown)


app/components/UserValidator/index.tsx
```diff
@@ -38,38 +43,50 @@ const isUser = (user: User | Res): user is User => {
...
+    setScanResults((prevScanResults) => {
+      if (
+        prevScanResults.length !== 0 &&
+        prevScanResults[0].message === result
+      ) {
+        prevScanResults[0].count++;
+        return prevScanResults;
+      }
+      return [{ message: result, count: 1 }, ...prevScanResults];
+    });
```
danielyanghansen, line 60: Just an idea: Pushing the message whose count was most recently updated to the top.
(Solution: Always remove/slice from list on match, increase count, and then push to list head)

danielyanghansen (approved):
Nice! This willl be useful!

I just have one comment :)

app/components/UserValidator/index.tsx
norbye: I'm not sure that I understand exactly what you mean - or if it's just that the code could be clearer as to what it does.

Currently new scans are added to the start of the list, so that when the list is printed on the website, the most recently scanned item is displayed first, and the older items overflow into nothingness.  

Did you want to make a change to that, or just change the way it is implemented in the code?

app/components/UserValidator/index.tsx
```diff
@@ -127,6 +132,17 @@ const Validator = (props: Props) => {
             facingMode: 'environment',
           }}
         />
+        <h3>Nylig scannede ABA-IDer</h3>
+        <Flex column justifyContent="center" gap={5}>
+          {scanResults.map(({ message, count }) => (
+            <span key={message + count}>
+              {message} (x{count})
+            </span>
+          ))}
```
ivarnakken, line 141: Am I reading the code wrong, or won't this list just keep growing? Won't that look weird?

app/components/UserValidator/index.tsx
```diff
@@ -127,6 +132,17 @@ const Validator = (props: Props) => {
...
+          {scanResults.length === 0 && (
+            <span>De siste ABA-IDene du scanner vises her.</span>
```
ivarnakken, line 143:
```suggestion
            <span>De siste ABA-IDene du scanner vises her</span>
```
Not sure if I dislike the use of periods or not. I'm sort of on the fence

app/components/UserValidator/index.tsx
norbye: You're reading it absolutely right. Personal preference I guess - and I don't really mind doing it either way - but I tested it by generating an overflow of messages and it looked a-ok to me.

The behaviour currently is that the modal grows in height after scans - but as it approaches the max height of the screen (the maxheight attribute is set to leave some space on the top and bottom - which is nice), the oldest results at the bottom overflow and are not seen unless the user scrolls down to view them.  
If we dont want it to resize I could set a maxHeight to the Flex div that displays the result

app/components/UserValidator/index.tsx
norbye: Hmm I think I agree that the period should be removed there.. The thing I know that I dislike is that the responses that are shown directly from the backend contain periods, which doesn't really fit in so I'll see if I can find a way to handle that neatly

app/components/UserValidator/index.tsx
norbye: Updated it now to only display the 5 last results

app/components/UserValidator/index.tsx
```diff
@@ -127,6 +192,17 @@ const Validator = (props: Props) => {
...
         />
+        <h3>Nylig scannede ABA-IDer</h3>
+        <Flex column justifyContent="center" gap={5}>
+          {scanResults.map(({ message, count }) => (
+            <span key={message + count}>
+              {message} (x{count})
+            </span>
+          ))}
+          {scanResults.length === 0 && (
+            <span>De 5 siste ABA-IDene du scanner vises her</span>
```
ivarnakken, line 203:
```suggestion
            <span className="secondaryFontColor">De 5 siste ABA-IDene du scanner vises her</span>
```

app/components/UserValidator/index.tsx
```diff
@@ -145,4 +221,8 @@ const Validator = (props: Props) => {
   );
 };
 
-export default Validator;
+const mapDispatchToProps = {
+  addToast,
+};
+
+export default connect(null, mapDispatchToProps)(Validator);
```
ivarnakken, line 228: Preferably add this together with the rest in `app/routes/userValidator/ValidatorRoute.tsx`.

app/components/UserValidator/index.tsx
```diff
@@ -36,40 +44,95 @@ const isUser = (user: User | Res): user is User => {
...
+  const displayResult = useCallback(
+    (result: string, success = false) => {
+      if (showScanner) {
+        addScanResult(result);
+        return;
+      }
+      if (success) {
+        showSuccessModal(result);
+      } else {
+        console.log('Show me the toast');
```
ivarnakken, line 95: Intentional?

ivarnakken (approved):
Good job! 🏅

app/components/UserValidator/index.tsx
```diff
@@ -81,36 +144,38 @@ const Validator = (props: Props) => {
...
-        <Icon name="checkmark" success size={160} />
-      </div>
+        <Flex
+          alignItems="center"
+          column
+          justifyContent="center"
+          padding={'2rem 0'}
+        >
+          <h3>{successMessage}</h3>
+          <Icon name="checkmark" success size={160} />
```
ivarnakken, line 169: By the way, the `success` here does not do anything. I see why though, so just keep it and I will fix it shortly

PR #4159, opened 2023-10-02
Mention lego-bricks in main README.md

norbye:
# Description

Add stuff to readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,17 @@
 ## Quick access
 
 1. [Quick Start](#quick-start)
-2. [Development](#development)
-3. [Deployment (webkom/lego#deployment)](https://github.com/webkom/lego#deployment)
+2. [LEGO-BRICKS](#lego-bricks)
+3. [Development](#development)
+4. [Deployment (webkom/lego#deployment)](https://github.com/webkom/lego#deployment)
 
 ## Quick Start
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,16 @@
 
 ## Quick Start
 
+Before running LEGO-WEBAPP, you have to compile LEGO-BRICKS
+
+```bash
+$ cd packages/lego-bricks
```
ivarnakken, line 22: You can build from the ropt as well, no? I believe you build both webapp and bricks then

README.md
```diff
@@ -16,6 +16,16 @@
 
 ## Quick Start
 
+Before running LEGO-WEBAPP, you have to compile LEGO-BRICKS
```
ivarnakken, line 19:
```suggestion
Before running LEGO-WEBAPP, you have to compile LEGO-BRICKS.
```
Perhaps we should also describe what lego-bricks actually is🧐

ivarnakken (commented):
Very nice to have!

README.md
norbye: Ahh that's probably true, will check tomorrow

README.md
norbye: Updated with a little bit more of a description now

danielyanghansen (approved):
Looks good. For some reason, it's only necessary to manually build bricks once, and then it builds automatically later.
Don't know if that's necessary to mention

PR #4164, opened 2023-10-04
Analytics date filter

Bestem0r:
# Description

Adds filtering options, `Start date` and `End date`, for registration statistics.

# Result

[Screencast from 04. okt. 2023 kl. 20.21 +0200.webm](https://github.com/webkom/lego-webapp/assets/33326578/c40aafd0-2381-42a6-89ac-271eae782b30)

# Testing

1. Navigate to an event and click on "Se påmeldinger". 
2. Click on "Statistics".
3. Try different values in the `Startdato for påmelding` and `Sluttdato for påmelding`. 

---

Resolves [ABA-499](https://linear.app/abakus-webkom/issue/ABA-499/)


ivarnakken:
The linear issue was regarding the analytics, not the statistics, but this is interesting in its own way as well. However, is there really any reason to filter on the statistics? Also, if so, it should come after the analytics.

Bestem0r:
> The linear issue was regarding the analytics, not the statistics, but this is interesting in its own way as well. However, is there really any reason to filter on the statistics? Also, if so, it should come after the analytics.
> 
> We really should separate statistics and analytics into different tabs or something, because they are really two different things.

Ah, so this is the the "analytics"?
Screenshot from 2023-10-04 21-07-05

ivarnakken:
> Ah, so this is the the "analytics"? @ivarnakken 

Jupp. I know it is confusing. It should not be next to the statistics

Bestem0r:
Updated to filter the page view statistics instead. See the new video in the description @ivarnakken

danielyanghansen (approved):
Looks good, I just have one small comment about memoization.

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -256,10 +262,22 @@ const Analytics = ({ eventId }: { eventId: ID }) => {
   useEffect(() => {
     eventId &&
       dispatch(fetchAnalytics(eventId)).then((res) => {
+        if (viewStartTime) {
+          res.payload = res.payload.filter(
+            (item) => new Date(item.date) >= new Date(viewStartTime as string)
+          );
+        }
```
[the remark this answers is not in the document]
Bestem0r: Let me know how it looks now @danielyanghansen

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -253,13 +259,36 @@ const Analytics = ({ eventId }: { eventId: ID }) => {
...
-        setData(res.payload);
-        setMetrics(calculateMetrics(res.payload));
+    if (eventId) {
+      memoizedData().then((data) => {
+        setData(data);
+        setMetrics(calculateMetrics(data));
       });
-  }, [eventId, dispatch]);
+    }
+  }, [eventId, memoizedData]);
```
danielyanghansen, line 291: Because you are memoizing the data based on `eventId`, you don't need this `useEffect` to also depend on `eventId` :smile: 
```suggestion
  }, [memoizedData, viewStartTime, viewEndTime]);
```

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -253,13 +259,36 @@ const Analytics = ({ eventId }: { eventId: ID }) => {
...
+
   useEffect(() => {
-    eventId &&
-      dispatch(fetchAnalytics(eventId)).then((res) => {
-        setData(res.payload);
-        setMetrics(calculateMetrics(res.payload));
+    if (eventId) {
+      memoizedData().then((data) => {
+        setData(data);
+        setMetrics(calculateMetrics(data));
```
danielyanghansen, line 288: You can do the filtering here. Something like this. Feel free to change up the syntax and convention of this
```suggestion
      memoizedData().then((data) => {
        const filteredData = data.filter((item => {return /*Calculation here*/}))
        setData(filteredData);
        setMetrics(calculateMetrics(filteredData));
```

app/routes/events/components/EventAttendeeStatistics.tsx
```diff
@@ -253,13 +259,36 @@ const Analytics = ({ eventId }: { eventId: ID }) => {
...
+
+        if (viewEndTime) {
+          data = data.filter(
+            (item) => new Date(item.date) <= new Date(viewEndTime as string)
+          );
+        }
+        return data;
+      }
+    };
+  }, [eventId, viewStartTime, viewEndTime, dispatch]);
```
danielyanghansen, line 282: The API returns complete data, regardless of what the user has set has `viewStartTime` and `viewEndTime`.

I recommend moving all of the filtering logic into your `useEffect` statement below, and removing these fields from the `useMemo`'s dependency array
```suggestion
  }, [eventId, dispatch]);
```

danielyanghansen (changes requested):
I have suggested some changes for the memoization. Feel free to ask any questions if there's something I should explain or clarify :grinning: 

Once that's fixed, all is good :sunglasses:

app/routes/events/components/EventAttendeeStatistics.tsx
Bestem0r: Seems like I have to for some reason... CI fails when I remove `eventId` with this message:
```
React Hook useEffect has a missing dependency: 'eventId'. Either include it or remove the dependency array
```

danielyanghansen (approved):
Looks good!

PR #4167, opened 2023-10-04
Use default cursor on disabled checkboxes and radio buttons

ivarnakken:
# Description

A pointer cursor indicates to the user that the input fields can be interacted with, and should therefore be changed accordingly when they are disabled.

# Result

Cursor now behaves as expected.

# Testing

Tested both input fields as both disabled fields and normal fields.

### DIFF
--- a/app/components/Form/CheckBox.css
+++ b/app/components/Form/CheckBox.css
@@ -18,8 +18,11 @@
 }
 
 .label {
-  cursor: pointer;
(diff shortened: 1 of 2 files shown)


ollfkaih (approved):
Lgtm

danielyanghansen (approved):

norbye (approved):

PR #4176, opened 2023-10-08
Add wait to fix cypress comment test

eikhr:
# Description

I don't know why it didn't work, but waiting a bit fixes it.

# Result

# Testing

Cypress no logner fails. It consistently failed without the wait.

---

### DIFF
--- a/cypress/e2e/event_visit_spec.js
+++ b/cypress/e2e/event_visit_spec.js
@@ -51,6 +51,8 @@ describe('View event', () => {
   it('Should be possible to comment', () => {
     cy.visit('/events/20');
 
+    cy.wait(500);
+
     cy.contains('button', 'Kommenter').should('not.be.visible');
     cy.get(c('CommentForm')).find('input').first().click();
     cy.focused().type('This event will be awesome');


ivarnakken (commented):
I get that it works, but I'm highly against this as a fix😬

eikhr:
Yea, it would be better if someone figured out how to wait until it is clickable. Do you have any idea what might be causing this? I think the most recent change to the comment field is to replace lego-editor with a normal input-field.

norbye:
> Yea, it would be better if someone figured out how to wait until it is clickable. Do you have any idea what might be causing this? I think the most recent change to the comment field is to replace lego-editor with a normal input-field.

tl;dr: I don't think the problem is waiting until it is clickable, as it is clickable all along, but rather that the input resets both the focus and content when other parts of the page are rendered

I tried running the event page on a simulated slow connection (chrome slow 3G), and every time another API call was fetched successfully the comment area was re-rendered, and the focus on the element was lost along with the content that was written before the re-render.

My initial thought was to try to write directly into the field as below (writing before focus is lost), but this might not be sufficient as it mostly avoids the real problem..

```diff
-cy.get(c('CommentForm')).find('input').first().click();
-cy.focused().type('This event will be awesome');
+cy.get(c('CommentForm')).find('input').first().type('This event will be awesome')
```

It might be better to investigate the element itself to make sure the input gets saved in the state when it gets edited so the value doesn't get reset when the page re-renders.

PR #4177, opened 2023-10-10
Remove random key from comment field

norbye:
# Description

Removes the wait introduced in #4176 

`CommentField`s were set to have a random key (why I do not know), that would make it re-render excessively (and thus clear the focus and content of the new comment). This removes the key, and rather handles form resetting by using the `form.change()` method.

Changes done to `TextInput.ts` were done to avoid bugs that were discovered while trying another approach.
* When `inputRef` was used as a parameter in `useRef()`, it did not update `inputRef` when `newInputRef` was used. Thus it was not possible to pass a ref through to the component.
* Small typescript errors

# Result

When run on my machine this works quite well, and I can start writing comments before all the content of the page has been loaded - which was not the case previously.

To see if this resolves the cypress-bug, we'll probably have to wait and see. This passed cypress on the first attempt, but lately it has worked some times and some times not, so it's no guarantee.

# Testing

Have been testing that the comment form works as expected in the event view.

### DIFF
--- a/app/components/CommentForm/index.tsx
+++ b/app/components/CommentForm/index.tsx
@@ -48,21 +48,21 @@ const CommentForm = ({
     <Card>
       <LegoFinalForm
         validateOnSubmitOnly
-        initialValues={{
-          commentKey: Math.random(),
-        }}
         validate={validate}
-        onSubmit={({ text }) =>
+        onSubmit={({ text }, form) => {
+          // Clear the form value
+          form.change('text', undefined);
+
           dispatch(
             addComment({
               contentTarget,
               text,
               parent,
             })
-          )
-        }
+          );
+        }}
       >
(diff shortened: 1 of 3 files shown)


app/components/CommentForm/index.tsx
```diff
@@ -48,29 +48,28 @@ const CommentForm = ({
...
       <LegoFinalForm
         validateOnSubmitOnly
-        initialValues={{
-          commentKey: Math.random(),
-        }}
         validate={validate}
-        onSubmit={({ text }) =>
+        onSubmit={({ text }, form) => {
+          // Clear the form value
+          form.change('text', undefined);
```
ivarnakken, line 54: Clever!

danielyanghansen (approved):

PR #4182, opened 2023-10-12
Add registration activationTime to default EventItem card in eventlist

danielyanghansen:
# Description

Update `EventItem` component to also include the time the event registration opens.

Also includes a tooltip

## Note:
Due to some styling differences, it was better to not combine the `TimeStamp` and `TimeStartAndRegistration` components

# Result
### Calendar Tooltip
image

<hr/>
<hr/>
<hr/>

 ## After
Screenshot from 2023-10-12 19-17-54

### Tooltip:
Screenshot from 2023-10-12 19-22-47

### Calendar Tooltip:
image

# Testing

---

Resolves ABA-588

### DIFF
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -75,13 +75,45 @@ const TimeStamp = ({ event }: TimeStampProps) => {
   return (
     <div className={styles.eventTime}>
       <Flex alignItems="center" gap={10}>
-        <Icon name="calendar-number-outline" size={20} />
+        <Tooltip content={"Arrangementsdato"}>
+          <Icon name="calendar-number-outline" size={20} />
+        </Tooltip>
         <Time time={event.startTime} format="ll" />
       </Flex>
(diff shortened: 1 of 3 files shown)


eikhr (approved):
nice

eikhr:
I think they wanted the posibility to sort by registration date/time, is that easy enough to be a part of this PR or should we do that separately?

eikhr:
Oh! And the calendar view! We should add it to the popups in the calendar as well

danielyanghansen:
@eikhr 

1: I think we can do sorting in a separate PR 
2: I can definitely update the Calendar Component in this PR
3: bOiLeRpLaTe?

norbye:
> I think they wanted the posibility to sort by registration date/time, is that easy enough to be a part of this PR or should we do that separately?

In a way we already have it I think, except that it's a bit hard to notice right now.. If you sort based on "Åpner i framtiden", it seems to do exactly that - which will look a bit more obvious after this PR as deployed

PR #4183, opened 2023-10-12
Fix wrong color in text inputs

ivarnakken:
# Description

Also fix wrong styling on prefix icon given by the recent changes to the Icon component.

# Result

<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	<tr>
		<td>
			image
		</td>
		<td>
			image
		</td>
</table>

# Testing

- Text inputs now have a correct text color. 
- Tested with and without a `prefix`, and `centered` inputs.
- `prefix` still works as expected, by focusing on the input when clicked.

---

Resolves ABA-602

### DIFF
--- a/app/components/Form/TextInput.css
+++ b/app/components/Form/TextInput.css
@@ -15,9 +15,12 @@
   text-align: center;
 }
 
-.spacing:not(.centered) input,
 .prefix {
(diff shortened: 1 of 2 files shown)


eikhr (approved):
Nice, this was super-annoying!

PR #4192, opened 2023-10-15
Remove extra div from settings navtab

norbye:
# Description

I missed a spot when I updated the styling of the navtab component previously - so with this, I hope there are no more occurrences that are not working..

# Result

NavLink elements missed spacing in user profile page, now it has spacing.

# Testing


ivarnakken (approved):
<3

PR #4201, opened 2023-10-16
Give unverified users a warning on overview page

ivarnakken:
# Description

There are currently `966` unverified users out of `1209` in grades 1 to 5 (including PhD). This means only ~20% have verified themselves ..

# Result

image

# Testing

The warning only appears if the is a current user and it is not a student (not verified). Closing the card works fine.

### DIFF
--- a/app/components/Footer/Footer.css
+++ b/app/components/Footer/Footer.css
@@ -82,7 +82,6 @@ html[data-theme='dark'] .bottomContent a {
   height: auto;
   margin-left: 1px;
   transition: transform var(--easing-slow);
-  transform-style: preserve-3d;
 }
 
 .gitHubLink:hover .octocat {
--- a/app/routes/overview/components/Overview.css
+++ b/app/routes/overview/components/Overview.css
@@ -63,3 +63,27 @@
     display: grid;
   }
 }
+
+.verificationWarning {
+  position: fixed;
+  width: 400px;
+  margin: 40px;
+  bottom: 0;
+  left: 0;
+
+  @media (--small-viewport) {
+    width: 100%;
+    margin: 10px 0;
+  }
+}
+
+.verificationWarning .close {
+  position: absolute;
+  top: 7px;
+  right: 7px;
+  cursor: pointer;
+}
+
+.verificationWarning .header {
+  width: 80%;
+}
--- a/app/routes/overview/components/Overview.tsx
+++ b/app/routes/overview/components/Overview.tsx
@@ -1,16 +1,19 @@
 import moment from 'moment-timezone';
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { Helmet } from 'react-helmet-async';
 import { Link } from 'react-router-dom';
+import Card from 'app/components/Card';
 // import Banner from 'app/components/Banner';
 import Icon from 'app/components/Icon';
 import { Container, Flex } from 'app/components/Layout';
 import Poll from 'app/components/Poll';
 import RandomQuote from 'app/components/RandomQuote';
 import type { Event, Readme } from 'app/models';
+import { selectCurrentUser } from 'app/reducers/auth';
 import type { WithDocumentType } from 'app/reducers/frontpage';
 import { isArticle, isEvent } from 'app/reducers/frontpage';
 import type { PollEntity } from 'app/reducers/polls';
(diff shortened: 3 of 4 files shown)


haukkagu (approved):
LGTM! A very useful feature!

However, this would probably be quite annoying for certain users who can't verify their account. E.g. students who have switched to indøk or students who studies something else but are going to switch to data/komtek. There's no way to manually verify students in the frontend yet, so we should probably add that as well.

ivarnakken:
> However, this would probably be quite annoying for certain users who can't verify their account. E.g. students who have switched to indøk or students who studies something else but are going to switch to data/komtek. There's no way to manually verify students in the frontend yet, so we should probably add that as well.

Yes it will be annoying, but then again, they won't be able to attend any events or basically use the site at all so I reckon they won't be regular visitors.
Besides, if they want to use the site, they must verify themselves either way or another.

eikhr:
> > I think the text should be changed a bit temporarily as most of the unverified students are simply because the verification system was not in place when they registered. It should inform them that we recently changed the verification system and that they should therefore re-verify, otherwise I think a lot of people will just assume it is a bug.
> 
> Okay. I just copied the existing text from the verification page. Should that be updated as well? I don't really want to put too much text in the card - could it be sufficient to add it to the verification page? I suppose they won't think it's a bug when it shows up every time they visit the page, or?

What if you just add a short sentence like "Vi har nylig endret verifiseringssystem" to the beginning of the popup?

ivarnakken:
> What if you just add a short sentence like "Vi har nylig endret verifiseringssystem" to the beginning of the popup?

I accept

norbye:
> but then again, they won't be able to attend any events or basically use the site at all so I reckon they won't be regular visitors.

Am I missing something here? As far as I've seen, the `isStudent` attribute is not used at all

I mean yes it is used for the student verification, but as soon as you have verified yourself, the only permission that is used for events are group memberships - and if you verified your user previously you will still be a member of the groups "Abakus" and whichever class you're in..  
And the only place users can really see if they have the isStudent attribute set or not is if you go into your settings to see if you're a member of abakus, but nobody does that and it doesn't change any functionality or permissions.

If I'm not mistaken, it would make quite a lot of sense that so few users are validated as students (with the isStudent attribute that is), because everyone that registered this year would be verified - and everyone that verified previous years are already part of the groups so it doesn't really make a difference for them..

norbye:
So I guess my main question is - why do we want to do this?  
Especially if it has no practical implications for the individual users. If we do want to force them through the revalidation each year even if they have previously verified their student status - we should have some sort of mechanism in place to not use your membership in the class groups (1. klasse komtek, etc.), so you can't go to events that are limited to students if you're not registered as a student..

ivarnakken:
@haukkagu @norbye

Yes you're both right. `isStudent` currently does nothing. However, I suppose we _should_ add a check for this during event registration, no? If you're not actually allowed to register for events, you shouldn't be able to either.
I sort of blindly assumed this was already done based on the text given to users, but you're right, previously verified users who have been given a grade (Abakus group) are able to register for events.

---

This is especially counterproductive given that users must verify themselves every year. ABA-534 would give more meaning to it though.

norbye:
> However, I suppose we _should_ add a check for this during event registration, no? 

I agree, but it might should probably be optional in some way right? I would assume that some events would be open to alumni as well (genfors(?),etc.) - so I guess it's ideal if it's either linked to membership of the student groups or a checkable field in the event admin page (yayy more clutter, but at least I guess it could be applied by default)

I think the user-experience could be sorta smooth if we require isStudent to be set for memberships of subgroups of the Student group tho, could give the same experience and usage as we have now just with the added step of validation.  
Would also give event admins the freedom of still choosing the members with quite a lot flexibility.

> This is especially counterproductive given that users must verify themselves every year. [ABA-534](https://linear.app/abakus-webkom/issue/ABA-534) would give more meaning to it though.

Hmm I don't immediately see what difference this would make.. would decrease the amount of students we have to manually bump yeah, but after the initial error the years are mostly correct by bumping every summer

ch0rizo (commented):
Nice!

danielyanghansen:
This popped into my notifications. Are we re-opening this discussion?

ivarnakken:
> This popped into my notifications. Are we re-opening this discussion?

Probably not. First we'll have to actually use the is_verified field ... which we currently don't really do 🙃 I might get around to it one day

PR #4202, opened 2023-10-16
Fix TextWithIcon icon alignment

jepunnerud:
# Description

Fixed wrapping of TextWithIcon component, thus also fixing the alignment of icons. Did this by wrapping the content (the text) in a span, and adding a div wrapping the span with "width: max-content".

# Result

**before**
image

**after**
image

# Testing

- I have thoroughly tested my changes.

Tested locally on both laptop, desktop and mobile views.

---

### DIFF
--- a/app/components/TextWithIcon/index.tsx
+++ b/app/components/TextWithIcon/index.tsx
@@ -15,13 +15,14 @@ type Props = {
 
 const TextWithIcon = ({
   iconName,
+  className,
   content,
   tooltipContentIcon,
   iconRight = false,
   size,
 }: Props) => {
   return (
-    <Flex alignItems="center" gap={5}>
+    <Flex alignItems="center" gap={5} className={className}>
       <div className={styles.textContainer}>
         {iconRight ? <span>{content}</span> : <></>}
       </div>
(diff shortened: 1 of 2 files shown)


app/components/TextWithIcon/index.tsx
```diff
@@ -15,15 +15,16 @@ type Props = {
...
 const TextWithIcon = ({
   iconName,
-  className,
   content,
   tooltipContentIcon,
   iconRight = false,
   size,
 }: Props) => {
   return (
-    <Flex alignItems="center" className={className}>
```
ivarnakken, line 25: The `className` is used other places than on the event page, so the removal of this will break all of those. Any reason for removing?

ivarnakken (changes requested):
Thanks for fixing this! It has been bugging me for ages! <3

Please look into my comment though.

app/components/TextWithIcon/index.tsx
jepunnerud: Not really. Added it back now.

Arashfa0301 (approved):
Looks good. Thank for the contribution ❤️

ivarnakken (approved):
Assuming you also checked the removal of the `min-width` for the other use cases of the component 😄 

Thanks again!

PR #4211, opened 2023-10-18
Change default survey activation time to 14:30

eikhr:
# Description

This was requested by the arranging komitees a few weeks ago to avoid conflicts with event registrations, which are usually on the hour. It sets default activation time to 14:30 the day after the event end time, or tomorrow at 14:30 if no event is selected.

# Result

See how it says 14:30(the event is 23. Oct):
Screenshot 2023-10-18 at 15 03 49

# Testing

Have tested both with and without a connected event, and with templates.
---

Resolves ABA-612


ollfkaih (approved):
good change

Bestem0r (approved):
Looks good to me!

ivarnakken (approved):

PR #4218, opened 2023-10-19
Replace jest with vitest

eikhr:
# Description

I was trying to fix the tests in #4118, and I figured maybe vitest would help as we use vite in lego-bricks. It did! But in the process I realized a much simpler fix that would make it work in jest, so I added that to #4118. 

Then I was left with this code... It fixes all the issues I encountered when trying to use vitest, so everything works fine, but it also doesn't really improve anything. The tests run slightly faster, but not really noticably (4.9s vs 4.4s). It does fix some compatibility issues with vite though, so if we want to migrate to vite in the future, those would be nice to have already fixed.

# Result

No visual changes, tests work almost exactly the same as before.

I did remove some workarounds we used for jest, seems like they are not needed in vitest. 

It also reduces the amount of dev dependencies needed, as we don't need jest anymore and already needed vitest for lego-bricks.

# Testing

The tests run fine, and seem to fail when something is wrong in the code.

### DIFF
--- a/app/components/Comments/CommentTree.tsx
+++ b/app/components/Comments/CommentTree.tsx
@@ -42,7 +42,7 @@ function CommentTree({
 
     if (comment.children.length) {
       return (
-        <div key={comment.id} className={className}>
+        <div key={comment.id} data-ischild={isChild} className={className}>
           <Comment
             comment={comment}
             commentFormProps={commentFormProps}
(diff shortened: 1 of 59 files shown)


app/components/Comments/CommentTree.tsx
```diff
@@ -42,7 +42,7 @@ function CommentTree({
 
     if (comment.children.length) {
       return (
-        <div key={comment.id} className={className}>
+        <div key={comment.id} data-ischild={isChild} className={className}>
```
ivarnakken, line 45: Clean!

ivarnakken (approved):
Looks clean!

PR #4220, opened 2023-10-20
Allow for admin registering multiple people at once

eikhr:
I replaced redux-form with react-final-form and cleaned up some of the routing and redux data fetching in eeae32cad0d1ff7943f9ed6d824a0868a6045c48, then implemented the multi-select thing in 1946e98d5e0f5cb7cbab75c9651a06539dc1e3f2

# Result

I changed the order of the form and added some required-marks as well:

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>Screenshot 2023-10-20 at 19 32 52
 <td>Screenshot 2023-10-20 at 19 32 03
</table>

# Testing

I have tested admin-registering single people and multiple people. Have also tested that error messages show up as intended, both from local validation and server-errors.

---

Resolves ABA-616


app/routes/events/EventAdminRegisterRoute.ts
ivarnakken, line 1: Thanks for getting rid of these at the same time 🙌🏼

ivarnakken (approved):

PR #4223, opened 2023-10-21
Rewrite autocomplete higher order components to use hooks

eikhr:
# Description

I found this in an old branch:) I rebased and have tested a bit, and it seems pretty much done to me. Not sure why I haven't created a PR on this earlier.

Basically just rewrites the autocomplete field-logic (normal and mazemap) from higher order components to hooks. The hooks are then used in higher-order-component wrappers for compatibility. The functionality should be exactly the same, but the code is typed a bit better and hopefully a bit more readable for someone who has learned react using function components and hooks.

# Result

No visible or functional changes.

# Testing

- [kind of] I have thoroughly tested my changes.

The autocomplete fields still work, but I haven't tested any edge cases or verified that the types are correct.

---


app/components/Form/Field.tsx
```diff
@@ -61,18 +67,38 @@ export const RenderWarningMessage = ({
 
   return <FieldWarning warning={warning} key={warning} />;
 };
+
+// TODO: replace with imported FieldRenderProps once we have migrated to react-final-form
```
danielyanghansen, line 71: We have migrated to react final form :sunglasses:

app/components/Form/SelectInput.tsx
```diff
@@ -126,14 +106,13 @@ function SelectInput({
...
           options={options}
           isLoading={fetching}
           styles={selectStyle ?? selectStyles}
           theme={selectTheme}
           onInputChange={(value) => {
-            if (props.onSearch) {
-              props.onSearch(value);
+            if (onSearch) {
+              onSearch(value);
             }
```
danielyanghansen, line 116: Is option chaining better (faster) here?
```suggestion
            onSearch?.(value)
```

Same for below (line 137)

app/components/Search/mazemapAutocomplete.tsx
```diff
@@ -1,112 +1,120 @@
...
+  WrappedComponent,
+}: {
+  WrappedComponent: ComponentType<P & InjectedProps>;
+}) => {
+  const Component = (props: P) => {
+    const { options, onSearch, fetching } = useMazemapAutocomplete();
+
+    return (
+      <WrappedComponent
+        {...(props as P)}
```
danielyanghansen, line 106: A bit confused here, is there a reason why typecasting is necessary here?

danielyanghansen (approved):
Looks good :sunglasses:

app/components/Form/Field.tsx
ivarnakken: We are not done

app/components/Search/mazemapAutocomplete.tsx
eikhr: Yes

eikhr: Actually no, I'll remove it

PR #4224, opened 2023-10-22
Reimplement createEntityReducer in the style of redux-toolkits createEntityAdapter

eikhr:
# Description

This was a project I was playing with a bit this summer. It's basically a full rewrite of `createEntityReducer` with the goal of making it more modular and easier to integrate with [Redux Toolkit](https://redux-toolkit.js.org/).

The [`createEntityAdapter`](https://redux-toolkit.js.org/api/createEntityAdapter) function in Redux Toolkit is basically a way to add some of the most basic `createEntityReducer` into a redux slice created using the Redux Toolkit syntax. It minimizes boilerplate, while still being typed perfectly. I therefore built the new `createEntityReducer` as an extended version of `createEntityAdapter`, calling it `createLegoAdapter`. Each separate function of the adapter is defined separately (f.ex. updating `fetching`-state), well typed and unit-tested.

I also decided on keeping our weird `generateStatuses` `ActionTypes`, as it would be a breaking change to remove them. It doesn't match Redux Toolkit's standard way of doing actions, so we are missing out on a lot of typing stuff there, but I think it will be easier to make the transition once all entity reducers are updated to `createLegoAdapter`, and we no longer need backwards compatibility with `createEntityReducer`.

There are a couple of things missing compared to the old `createEntityReducer`. This only supports `paginationNext`, not the other system just called `pagination`, meaning some routes will need some refactoring. This also doesn't support optimistic updates on creation, but this is barely ever used so it can be implemented at a later date.

# Result

It will be suuper easy to rewrite existing `createEntityReducer`-reducers to Redux Toolkit slices using `createLegoAdapter` to keep the same functionality.

It will in general reduce a lot of the boilerplate code in reducers. F.ex. we have a lot of selectors that simply select an entity by id, which looks something like this:
```ts
export const selectCompanyInterestById = createSelector(
  (state) => state.companyInterest.byId,
  (state, props) => props.companyInterestId,
  (companyInterestById, companyInterestId) =>
    companyInterestById[companyInterestId]
);
```
This selector is provided for free with `createLegoAdapter`, you can simply get it using:
```ts
export const { 
	selectById: selectCompanyInterestById 
} = legoAdapter.getSelectors((state: RootState) => state.companyInterest);
```

# Testing

Everything is quite thoroughly unit tested, but all functionality should still be tested thourougly when migrating slices. 

---

Resolves ABA-659

### DIFF
--- a/app/utils/legoAdapter/buildDeleteEntityReducer.ts
+++ b/app/utils/legoAdapter/buildDeleteEntityReducer.ts
@@ -1,0 +1,23 @@
+import { isAsyncApiActionSuccess } from 'app/utils/legoAdapter/asyncApiActions';
+import type { EntityState } from '@reduxjs/toolkit/src/entities/models';
+import type { ActionReducerMapBuilder } from '@reduxjs/toolkit/src/mapBuilders';
+import type { AsyncActionType } from 'app/types';
+import type { DeleteMeta } from 'app/utils/legoAdapter/asyncApiActions';
+
+/**
+ * Handles optimistic and actual deletion of entities
+ */
+const buildDeleteEntityReducer = (
+  builder: ActionReducerMapBuilder<EntityState<unknown>>,
+  deleteActions: AsyncActionType[]
+) => {
+  builder.addMatcher(
+    isAsyncApiActionSuccess.matching<DeleteMeta>(deleteActions),
+    (state, action) => {
+      state.ids = state.ids.filter((id) => id !== action.meta.id);
+      delete state.entities[action.meta.id];
+    }
+  );
+};
+
+export default buildDeleteEntityReducer;
--- a/app/actions/ActionTypes.ts
+++ b/app/actions/ActionTypes.ts
@@ -1,6 +1,6 @@
 import type { AsyncActionType } from 'app/types';
 
-const generateStatuses = (name: string): AsyncActionType => ({
+export const generateStatuses = (name: string): AsyncActionType => ({
   BEGIN: `${name}.BEGIN`,
   SUCCESS: `${name}.SUCCESS`,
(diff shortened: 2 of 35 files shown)


danielyanghansen (commented):
Looks good so far. I've only skimmed over the code so far, but I have a decent understanding of what you're doing.
Do you want more detailed reviews while this is still in draft? You already have TODO notes for yourself

eikhr:
> Looks good so far. I've only skimmed over the code so far, but I have a decent understanding of what you're doing. Do you want more detailed reviews while this is still in draft? You already have TODO notes for yourself

I think this is ready enough to merge now, so I would love a review!

app/utils/legoAdapter/buildDeleteEntityReducer.ts
```diff
@@ -0,0 +1,38 @@
...
+    (state, action) => {
+      state.ids = state.ids.filter((id) => id !== action.meta.id);
+    }
+  );
+  builder.addMatcher(
+    isAsyncApiActionFailure.matching<DeleteMeta>(deleteActions),
+    (state, action) => {
+      state.ids.push(action.meta.id);
+    }
+  );
```
PeterJFB, line 29: yeet

juniwbjerde (approved):
I haven't looked at the tests yet (and I'm not sure if I'll bother since there's almost 1500 lines of just tests), but all the other code looks good to me:)

app/utils/legoAdapter/buildDeleteEntityReducer.ts
eikhr: Yeah, I agree optimistic delete is not really necessary. Probably better to simplify it

PR #4226, opened 2023-10-23
Fix registration form

PeterJFB:
# Description

It used field which were only supported by React-Final-Form, which made it crash.

# Result

Solved by migrating it to LegoFinalForm and adding async validation to createValidator, as RFF supports this.

Code could probably be more simplified, but works as solution while we are migrating to RFF.

# Testing

- Checked that each form submits with expected payload
---

### DIFF
--- a/app/routes/users/components/ChangePassword.tsx
+++ b/app/routes/users/components/ChangePassword.tsx
@@ -1,71 +1,79 @@
 import { Button } from '@webkom/lego-bricks';
-import { Field } from 'redux-form';
-import { TextInput, Form, legoForm } from 'app/components/Form';
+import { Field, FormSpy } from 'react-final-form';
+import { TextInput } from 'app/components/Form';
+import LegoFinalForm from 'app/components/Form/LegoFinalForm';
 import type { UserEntity } from 'app/reducers/users';
-import { createAsyncValidator } from 'app/utils/asyncValidator';
 import { createValidator, required, sameAs } from 'app/utils/validation';
 import { validPassword } from '../utils';
 import PasswordField from './PasswordField';
-import type { FormProps } from 'redux-form';
 
 type PasswordPayload = {
   newPassword: string;
   password: string;
   retype_new_password: string;
 };
-type Props = FormProps & {
+type Props = {
   push: (arg0: string) => void;
   changePassword: (arg0: PasswordPayload) => Promise<void>;
   user: UserEntity;
 };
 
-const ChangePassword = ({
-  handleSubmit,
-  invalid,
-  pristine,
-  submitting,
-  user,
-}: Props) => {
-  const disabledButton = invalid || pristine || submitting;
+const ChangePasswordForm = ({ user, changePassword, push }: Props) => {
+  const onSubmit = (data: PasswordPayload) =>
+    changePassword(data).then(() => push('/users/me'));
   return (
-    <Form onSubmit={handleSubmit}>
-      <Field
-        label="Gammelt passord"
-        name="password"
-        type="password"
-        autocomplete="current-password"
-        component={TextInput.Field}
-      />
-      <PasswordField user={user} label="Nytt passord" name="newPassword" />
-      <Field
-        label="Nytt passord (gjenta)"
-        name="retypeNewPassword"
-        autocomplete="new-password"
-        type="password"
-        component={TextInput.Field}
-      />
-      <Button disabled={disabledButton} submit danger>
-        Endre passord
-      </Button>
-    </Form>
+    <LegoFinalForm onSubmit={onSubmit} validate={validate}>
+      {({ handleSubmit }) => (
+        <form onSubmit={handleSubmit}>
+          <Field
+            label="Gammelt passord"
+            name="password"
+            type="password"
+            autocomplete="current-password"
+            component={TextInput.Field}
+          />
+          <PasswordField user={user} label="Nytt passord" name="newPassword" />
+          <Field
+            label="Nytt passord (gjenta)"
+            name="retypeNewPassword"
+            autocomplete="new-password"
+            type="password"
+            component={TextInput.Field}
+          />
+          <FormSpy
+            subscription={{
+              pristine: true,
+              submitting: true,
+              invalid: true,
+            }}
+          >
+            {({ pristine, submitting, invalid }) => (
+              <Button
+                submit
+                danger
+                disabled={pristine || submitting || invalid}
+              >
+                Endre passord
+              </Button>
+            )}
+          </FormSpy>
+        </form>
+      )}
+    </LegoFinalForm>
   );
 };
 
-const validate = createValidator({
-  password: [required()],
-  newPassword: [required()],
-  retypeNewPassword: [
-    required(),
-    sameAs('newPassword', 'Passordene er ikke like'),
-  ],
-});
-const asyncValidate = createAsyncValidator({
-  newPassword: [validPassword()],
-});
-export default legoForm({
-  form: 'changePassword',
-  validate,
-  asyncValidate,
-  onSubmit: (data, dispatch, { changePassword, push }: Props) =>
-    changePassword(data).then(() => push('/users/me')),
-})(ChangePassword);
+const validate = createValidator(
+  {
+    password: [required()],
+    newPassword: [required(), validPassword()],
+    retypeNewPassword: [
+      required(),
+      sameAs('newPassword', 'Passordene er ikke like'),
+    ],
+  },
+  undefined,
+  true
+);
+
+export default ChangePasswordForm;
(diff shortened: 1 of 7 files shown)


app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
...
+                .filter(([isValid]) => !isValid)
+                .map(([, message]) => message || 'not valid') || 0;
+
+            if (fieldErrors.length) errors[field] = fieldErrors;
+            return errors;
+          },
+          {}
+        );
+
+        return merge(fieldValidatorErrors, rawValidatorErrors);
```
PeterJFB, line 153: This part is the most interesting to review. It's a rewrite of the old validate which unfolds Promises if they exist:)))

app/routes/users/components/UserConfirmation.tsx
```diff
@@ -89,91 +95,94 @@ const UserConfirmation = ({
...
-          </MultiSelectGroup>
-          <Field
-            name="allergies"
-            placeholder="Allergier"
-            label="Matallergier/preferanser"
-            component={TextInput.Field}
-          />
+        <LegoFinalForm onSubmit={onSubmit} validate={validate}>
+          {({ handleSubmit }) => {
+            return (
```
ivarnakken, line 100: A bigger fan of just letting it be a implicit return
```suggestion
          {({ handleSubmit }) => (
```

app/routes/users/components/UserConfirmation.tsx
```diff
@@ -89,91 +95,94 @@ const UserConfirmation = ({
...
-  createUser(token, data);
+const validate = createValidator(
+  {
+    username: [required()],
+    password: [required(), validPassword()],
+    retypePassword: [required(), sameAs('password', 'Passordene er ikke like')],
+    gender: [required()],
+  },
+  undefined,
+  true
```
ivarnakken, line 185: What is this?

app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
     context[field] ? ([true] as const) : validator(value, context);
 
 export function createValidator(
-  fieldValidators: { [field: string]: ReturnType<Validator>[] },
-  rawValidator?: (input) => ValidatorResult
+  fieldValidators: {
+    [field: string]: (ReturnType<Validator> | ReturnType<AsyncValidator>)[];
+  },
+  rawValidator?: (input) => ValidatorResult,
+  makeAsync = false
```
ivarnakken, line 120:
```suggestion
  async = false
```
a so called smør-på-flesk-variabel

app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
...
-            .map(([, message]) => message || 'not valid') || 0;
-        if (fieldErrors.length) errors[field] = fieldErrors;
-        return errors;
-      },
-      {}
-    );
-    return merge(fieldValidatorErrors, rawValidatorErrors);
-  };
+  return makeAsync
+    ? async function validate(input) {
```
ivarnakken, line 123: can we just do `if (makeAsync)` instead? Turneries are just more difficult to read when they get this big

app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
...
+          (errors: ValidatorResult, field) => {
+            const fieldErrors =
+              fieldValidators[field]
+                .map((validator) => validator(input[field], input))
+                .filter(([isValid]) => !isValid)
+                .map(([, message]) => message || 'not valid') || 0;
+            if (fieldErrors.length) errors[field] = fieldErrors;
+            return errors;
+          },
+          {}
```
ivarnakken, line 167: Empty object? what's happening here

app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
...
+            })
+          );
+        }
+
+        const fieldValidatorErrors = Object.keys(fieldValidators).reduce(
+          (errors: ValidatorResult, field) => {
+            const fieldErrors =
+              validationResults[field]
+                .filter(([isValid]) => !isValid)
+                .map(([, message]) => message || 'not valid') || 0;
```
ivarnakken, line 145: The `|| 0` part seems out of place. If the intention was to provide a default value, I believe this won't work as expected because the map function will always return an array, making the `|| 0` part redundant.

app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
...
+      }
+    : function validate(input) {
+        const rawValidatorErrors: ValidatorResult = rawValidator?.(input) || {};
+        const fieldValidatorErrors = Object.keys(fieldValidators).reduce(
+          (errors: ValidatorResult, field) => {
+            const fieldErrors =
+              fieldValidators[field]
+                .map((validator) => validator(input[field], input))
+                .filter(([isValid]) => !isValid)
+                .map(([, message]) => message || 'not valid') || 0;
```
ivarnakken, line 163: See comment above

app/utils/validation.ts
```diff
@@ -109,23 +113,59 @@ export const ifNotField =
...
+            if (fieldErrors.length) errors[field] = fieldErrors;
```
ivarnakken, line 164: Please just write this as multiple lines. Less lines does not equal better

ivarnakken (commented):
Cool!

I have some comments though. Also, from a quick glance there seems to be some duplicate code in the createValidator function which might benefit from being broken down into smaller reusable functions. Would probably improve readability. Right now it's a difficult function to read. Also, get rid of the turneries and unnecessary one-liners - this is not itgk

app/utils/validation.ts
PeterJFB: Its how reducers work, the last argument is the initial state, an empty object (updated by each `return`)

I find it quite unreadable myself so I've rewritten into something more readable/maintainable

app/utils/validation.ts
PeterJFB: Agreed. It is also technically unreachable the way it is written now but I've refactored it into something that makes more sense.

app/utils/validation.ts
PeterJFB: Gonna avoid refactoring this old implementation, don't want this bug-fix to add more errors:P

app/routes/users/components/UserConfirmation.tsx
PeterJFB: apparently some validators does not exist in a single field, so rawValidators are ran without a field argument (but is unused here)

app/utils/asyncValidator.ts
```diff
@@ -48,6 +48,9 @@ const getFieldErrorArray = async <T>(
   ).filter(([, fieldErrors]) => fieldErrors.length);
 };
 
+/**
+ * DEPRECATED, use createAsyncValidator (with async=true) instead
```
eikhr, line 52:
```suggestion
 * DEPRECATED, use createValidator (with async=true) instead
```

PeterJFB:
Tests are failing due to PasswordField as it is used in other forms. Seem simple enough to migrate the other forms as well, though i don't not have time to do it tonight :worried:

PeterJFB:
> Tests are failing due to PasswordField as it is used in other forms. Seem simple enough to migrate the other forms as well, though i don't have time to do it tonight 😟

sike :eyes:

app/routes/users/components/ChangePassword.tsx
norbye, line 1: If the component is exported and imported as `ChangePasswordForm` instead of `ChangePassword`, can you also change the filename to match?

app/routes/users/components/UserConfirmation.tsx
```diff
@@ -89,91 +89,92 @@ const UserConfirmation = ({
...
+                label="Telefonnummer"
+                name="phoneNumber"
+                autocomplete="tel"
+                component={PhoneNumberInput.Field}
+              />
+              {spySubmittable((submittable) => (
+                <Button submit disabled={!submittable}>
+                  Registrer bruker
+                </Button>
+              ))}
```
norbye, line 160: Potential use for `SubmitButton`?

app/routes/users/components/UserResetPassword.tsx
```diff
@@ -34,49 +25,54 @@ const UserResetPassword = ({
...
+                name="retypeNewPassword"
+                type="password"
+                component={TextInput.Field}
+              />
+
+              {spySubmittable((submittable) => (
+                <Button danger submit disabled={!submittable}>
+                  Tilbakestill passord
+                </Button>
+              ))}
```
norbye, line 55: `<SubmitButton>Tilbakestill passord</SubmitButton>`?

PeterJFB:
Tested all forms, it should be good to go :shipit:

norbye (approved):
noice stuff(: skimmed the updated validation thingy quite quickly - but looked a bit more thoroughly into the rest of it

danielyanghansen (approved):
Looks good! While I agree with the suggested changes from @norbye , I think we need to just ship this out soon. We have one or two people waiting to register

PR #4232, opened 2023-10-24
Changed backgroundcolor for interested on bdb interestpage

ShaileshS1702:
reason: problems with darkmode

ABA-615

### DIFF
--- a/app/routes/bdb/components/bdb.css
+++ b/app/routes/bdb/components/bdb.css
@@ -114,7 +114,7 @@
 }
 
 .interested {
-  background-color: #f5f5d5;
+  background-color: var(--color-yellow-3);
 }
 
 .notInterested {


app/routes/bdb/components/bdb.css
```diff
@@ -114,7 +114,7 @@
 }
 
 .interested {
-  background-color: #f5f5d5;
+  background-color: #B33939;
```
ivarnakken, line 117: Can we use a variable instead?😄

ShaileshS1702: > Can we use a variable instead?😄

We thought about it. If we only change .interested and not the other classes: interested will change color when we change between light/dark mode, but the others will stay same

ShaileshS1702: > > Can we use a variable instead?😄
> 
> We thought about it. If we only change .interested and not the other classes: interested will change color when we change between light/dark mode, but the others will stay same

one solution is to change multiple classes, but then the color at the page will be different and can maybe cause problems for bedroom

Should we still do it?

app/styles/variables.css
```diff
@@ -19,6 +19,7 @@
 
   --lego-max-width: 1100px;
   --lego-default-padding: 2rem;
+  --bdb-interested: #B33939;
```
ivarnakken, line 22: I think I prefer to use existing colours rather than introducing new ones

ShaileshS1702: The question is then what is the best way:
If we do choose to use variables, the color will change with the lightning mode:
image
image

While in light mode "ikke interessert" is darker than "Interessert", in dark mode the opposite can be observed. While this may not be a big issue, if the user of the site remember states based on color, it may be difficult if they change lightning. 
  We can either go through with or we can use a different color over all. The reason we chose to use red was that we thought it was smart that states that are similar should also have similar color. 
As we are new to this, we are of course open to suggestions, so what do you think we should do?

haukkagu: @ivarnakken What's your opinion on adding color variables that don't change based on theme? It'd make the solution to issues such at these a lot easier.

ivarnakken: "Ikke interessert" should be the danger color. "Interessert" shouldn't really be red at all, but rather a green successful color.

The colors here in the bdb were picked randomly by a webkomer many many years ago. Feel free to change.

PR #4233, opened 2023-10-25
Update Tag alignment

norbye:
# Description

Ivar found out that some of the tags were slightly unaligned, so with this, all the tags should be perfectly aligned(:

When the linkSpacing thingy was inline, it added some more space on top of itself, which it does not when it is inline-flex(:

# Result

TODO

# Testing

I've tested this checking the alignment in;
- article overview taglist
- bottom of article
- admin > email > lists
- admin > email > users
- pinned article on overview

### DIFF
--- a/app/components/Tags/Tag.css
+++ b/app/components/Tags/Tag.css
@@ -4,11 +4,11 @@
   margin: 20px 0;
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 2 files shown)


danielyanghansen (approved):
LGTM

ivarnakken (approved):
😍😍

PR #4242, opened 2023-10-26
Listen to changes in current page to expand category

norbye:
# Description

The correct page group was only expanded when the component was rendered (e.g. when navigating from another part of the webapp). When navigating through hyperlinks within the flatpages however, the component was already loaded and as such did not change. Now it does.

For a video showcasing the issue, check out the linear issue

# Result

Works on my machine ™️ 

# Testing

---

Resolves ABA-606

### DIFF
--- a/app/routes/pages/components/PageHierarchy.tsx
+++ b/app/routes/pages/components/PageHierarchy.tsx
@@ -1,5 +1,5 @@
-import { useState } from 'react';
-import { Link, useLocation } from 'react-router-dom';
+import { useEffect, useState } from 'react';
+import { Link } from 'react-router-dom';
 import Icon from 'app/components/Icon';
 import { readmeIfy } from 'app/components/ReadmeLogo';
 import styles from './PageHierarchy.css';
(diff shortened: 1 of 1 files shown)


app/routes/pages/components/PageHierarchy.tsx
```diff
@@ -85,15 +85,22 @@ type AccordionProps = {
   currentCategory: string;
 };
 
-const AccordionContainer = ({ title, children }: AccordionProps) => {
-  const location = useLocation();
-  const currentCategory = location.pathname.split('/')[2];
+const AccordionContainer = ({
+  title,
+  children,
+  currentCategory,
```
ivarnakken, line 91: lol did it have access to this all along 💀

ivarnakken (approved):
:shipit:

app/routes/pages/components/PageHierarchy.tsx
norbye: Heheh it was even typed already(: so why it wasn't used is quite the mystery to me

PR #4256, opened 2023-11-03
Make checkboxes focusable

ivarnakken:
# Description

With the changes, it is possible to tab through checkboxes and select them.

Turned out to be impossible to do for radio buttons, as essentially according to W3C; a radio button is a group that functions as a single element since it retains only a single value. Tabbing to a radio group will bring you to the first item and then using the arrow keys you navigate within the group.

However, I did add the similar logic for radio buttons in case there is no default value to the radio button group (e.g. in the company interest form). Despite this being bad practice.

# Result

Notice how the cursor does not move:

<video src="https://github.com/webkom/lego-webapp/assets/69514187/293907e1-13a8-46bb-a0cb-19b52a8483ff" />

# Testing

Checkboxes still behaves like before, but you can now focus on them by tabbing 🎉

### DIFF
--- a/app/components/Form/CheckBox.css
+++ b/app/components/Form/CheckBox.css
@@ -44,7 +44,8 @@
   }
 }
 
-.checkbox input:hover:not(:disabled) {
+.checkbox input:hover:not(:disabled),
+.checkbox input:focus:not(:disabled) {
   --s: 2px;
   --b: var(--border-hover);
 }
--- a/app/components/Form/CheckBox.tsx
+++ b/app/components/Form/CheckBox.tsx
@@ -3,7 +3,7 @@ import cx from 'classnames';
 import styles from './CheckBox.css';
 import { createField } from './Field';
 import type { FormProps } from './Field';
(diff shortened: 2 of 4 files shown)


norbye (approved):
Looks good!

Don't know if the handleKeyDown methods are necessary tho. At least on chrome (and I would expect it to be the same on other browsers), the default key to press to select checkboxes is the spacebar - and that already works on the webapp. But it wasnt that visible that it worked until you added the styling for focused elements as you couldn't really see which element you would activate beforehand..

But then again it's not really a problem to enable the same functionality with both space and enter soo might as well 🚀

ivarnakken:
> Don't know if the handleKeyDown methods are necessary tho. At least on chrome (and I would expect it to be the same on other browsers), the default key to press to select checkboxes is the spacebar - and that already works on the webapp. But it wasnt that visible that it worked until you added the styling for focused elements as you couldn't really see which element you would activate beforehand..

I actually didn't know I could toggle with the spacebar haha

PR #4258, opened 2023-11-03
Use param instead of inconsistent variable for accordion expansion

norbye:
# Description

Found out that the variable I relied on for my last fix was incorrectly typed :'( so it turns out the variable wasn't set for groups or main pages, so now it's rather fetched straight from the sauce.

# Result

Should work better now

# Testing

---

Resolves ABA-646


eikhr (approved):
looks good!

ivarnakken (approved):
Haha I guess it was too good to be true😆 Thanks for the quick fix!

danielyanghansen (approved):

PR #4266, opened 2023-11-07
Fix validation bug in CompanyEditor and improve types

eikhr:
# Description

The CompanyEditor would freeze up if you attempted to submit with an invalid link. While fixing this I got a bit lost in the types 😅

# Result

It no longer freezes, and instead gives correct feedback.

# Testing

---

Resolves ABA-660

### DIFF
--- a/app/routes/bdb/components/CompanyEditor.tsx
+++ b/app/routes/bdb/components/CompanyEditor.tsx
@@ -13,27 +13,40 @@ import LegoFinalForm from 'app/components/Form/LegoFinalForm';
 import SubmissionError from 'app/components/Form/SubmissionError';
 import { SubmitButton } from 'app/components/Form/SubmitButton';
 import InfoBubble from 'app/components/InfoBubble';
+import { AutocompleteContentType } from 'app/store/models/Autocomplete';
 import { createValidator, required, isEmail } from 'app/utils/validation';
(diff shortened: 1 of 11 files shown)


app/routes/bdb/components/CompanyEditor.tsx
```diff
@@ -13,28 +14,40 @@ import LegoFinalForm from 'app/components/Form/LegoFinalForm';
...
+  logo?: string;
+  name?: string;
+  description?: string;
+  companyType?: string;
+  website?: string;
+  address?: string;
+  studentContact?: AutocompleteEntity<AutocompleteContentType.User>;
+};
+
+const TypedLegoForm = LegoFinalForm<FormValues>;
```
juniwbjerde, line 49: I don't like that this name is so generic, when it's typed just to fit this form. A bit nitpicky, but still

juniwbjerde (approved):

app/routes/bdb/components/CompanyEditor.tsx
eikhr: I think it makes sense like this. The file it is in gives context about what form it is typed to fit, just like `Props`.

ivarnakken (approved):

danielyanghansen (approved):
Looks good :+1:

PR #4267, opened 2023-11-07
Fix issue with import sorting in CI

eikhr:
# Description

As you may have noticed, CI would fail at the linting step about 2/3 of all times it was run, regardless of code changes. Not 100% sure what caused it, but my theory is that something is the webpack config regarding lego-bricks confused the import resolver in eslint. This fixes it by overriding the import group of lego-bricks imports to always be external.

I also added a rule to ignore import group overrides for type impors, which is the cause of the numerous code changes. This ensures that type imports always come after the other imports, even if they are from "app/**" or "@webkom/lego-bricks":)

`.eslintrc.js` is the only interesting change in the PR.

# Result

CI doesn't fail as much, type imports are always last.

# Testing

I have run this PR in CI 3 times, and it passed linting every time!

---


.eslintrc.js
```diff
@@ -70,7 +70,12 @@ module.exports = {
             pattern: 'app/**',
             group: 'internal',
           },
+          {
+            pattern: '@webkom/lego-bricks',
+            group: 'external',
+          },
         ],
+        pathGroupsExcludedImportTypes: ['type'],
```
ivarnakken, line 78: Nice this has been bugging me for a while

ivarnakken (approved):
🙌🏼

PR #4272, opened 2023-11-08
Make text editor input similar to other text inputs

ivarnakken:
# Description

* Fix wrong font
* Fix bad padding
* Fix bad font-size
* Prevent editor from growing out of its parent (you can currently drag and resize it out of your screen in all directions if you want)
* Add min-height

# Result

<table>
	<tr>
		<td>Before</td>
		<td>After</td>
	<tr>
		<td>
			image
</td>
		<td>
			image
		</td>
</table>

# Testing

The added min-height does not break editors where the height has been specified, such as on the company interest form.

### DIFF
--- a/app/components/Form/TextEditor.css
+++ b/app/components/Form/TextEditor.css
@@ -1,5 +1,17 @@
 .input {
   composes: input from './TextInput.css';
   width: 100%;
-  padding: 5px;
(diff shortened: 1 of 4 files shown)


eikhr (approved):
Nice!

PR #4279, opened 2023-11-09
Fix sidebar position on mobile

eikhr:
# Description

This changes the sidebar positioning from using flex-wrap to changing layout direction at a set breakpoint `--medium-viewport`. The reason for this change is that there is no way to create the border line between main content and sidebar, only when the content is not wrapped. (I don't think this has ever happened entierly in sync).

# Result

Before:
Screenshot 2023-11-09 at 14 18 08
After:
Screenshot 2023-11-09 at 14 19 14

# Testing

Have clicked through most of the pages using sidebar, and checked that they look good on mobile.

---

Resolves ABA-672

### DIFF
--- a/app/components/Content/ContentSidebar.css
+++ b/app/components/Content/ContentSidebar.css
@@ -2,13 +2,12 @@
 
 .sidebar {
   flex: 1;
-  padding: 20px;
-  padding-top: 0;
+  padding: 0 0 0 2em;
   border-left: 1px solid var(--border-gray);
-  margin-left: 3em;
+  margin-left: 2em;
   line-height: 1.3;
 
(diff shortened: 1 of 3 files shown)


app/components/Content/ContentSidebar.css
```diff
@@ -2,13 +2,12 @@
 
 .sidebar {
   flex: 1;
-  padding: 20px;
-  padding-top: 0;
+  padding: 0 0 0 2em;
   border-left: 1px solid var(--border-gray);
```
ollfkaih, line 6: sorta random but I guess this is good

ollfkaih (approved):

PR #4281, opened 2023-11-10
Fix bug where meeting icon in header does not appear after logging in

eikhr:
# Description

The users upcoming meetings were only fetched once when loading the page, so if you weren't logged in already they never get fetched, and the icon did not appear. 

The fix is to add a dependency to the meeting fetch call, so that it is run again every time the `loggedIn` state changes.

# Result

The icon now appears a few seconds after logging in:)

# Testing

---

Resolves ABA-674

### DIFF
--- a/app/routes/app/AppRoute.tsx
+++ b/app/routes/app/AppRoute.tsx
@@ -214,20 +214,19 @@ export default compose(
     () => [],
     { serverOnly: true }
   ),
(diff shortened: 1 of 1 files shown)


juniwbjerde (approved):
Nice catch, lgtm!

ivarnakken (approved):

PR #4282, opened 2023-11-10
Ensure array query parameters are always parsed as arrays

eikhr:
# Description

The new query parameter parsing I added in #4259 would parse parameters like "?grades=2" as a string, even if the defaultValue, and thus the typescript type is an array of strings.

# Result

It now parses them correctly by checking the type of the default value, and ensuring the parsed value is an array if the default value is.

I also changed the array param format to drop the square brackets, because I think they are ugly. Array parameters will look like this "?grades=1&grades=2" now instead of this "?grades[]=1&grades[]=2".

# Testing

Have tested with multiple different array parameter formats on the joblisting page. Filtering works well with all of them:)

---

Resolves ABA-675

### DIFF
--- a/app/utils/__tests__/useQuery.spec.ts
+++ b/app/utils/__tests__/useQuery.spec.ts
@@ -23,7 +23,7 @@ describe('parseQueryString', () => {
     const initialValues = {
       foo: ['fizz'],
     };
-    const queryString = '?foo[]=bar&foo[]=baz';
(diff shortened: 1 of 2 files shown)


jonasdeluna (approved):

haukkagu (approved):

danielyanghansen (approved):
Looks good

PR #4311, opened 2023-11-16
Refactor group form and related routes

eikhr:
# Description

Extracted from the react-router refactorings, this changes the form from a `redux-form` to `react-final-form`. It also changes the routes that use it to rely mostly on hooks instead of props. This is needed for the migration to react-router v6.

# Result

No visual changes

# Testing

Have tested the form in interest group create and edit pages as well as the admin group edit page.


ivarnakken (approved):
Difficult to see what has changed, but it looks alright

eikhr:
There were some issues with interestgroups-creation, other than that it has not changed much from the react-router branch.

PR #4315, opened 2023-11-17
Make it possible to set a referent when creating a meeting

ivarnakken:
# Description

The select input field's options did not previously include the memberships of invited groups. Therefore you had to create the meeting first and then edit it to add referents (or alternatively you could invite specific users, but that's just a workaround).

I don't think this is the "cleanest" code, but I'd say it's still nice to have and it doesn't break anything so 🤷🏼‍♂️ 

# Result

https://github.com/webkom/lego-webapp/assets/69514187/88f99c0a-a5b9-4f96-8967-04ad1d7c7d25

# Testing

Tested adding and removing multiple groups, and groups without members as well.

### DIFF
--- a/app/routes/meetings/components/MeetingEditor.tsx
+++ b/app/routes/meetings/components/MeetingEditor.tsx
@@ -6,9 +6,11 @@ import {
 } from '@webkom/lego-bricks';
 import { unionBy } from 'lodash';
 import moment from 'moment-timezone';
+import { useState } from 'react';
 import { Field, FormSpy } from 'react-final-form';
 import { Helmet } from 'react-helmet-async';
 import { useHistory } from 'react-router-dom';
+import { fetchMemberships } from 'app/actions/GroupActions';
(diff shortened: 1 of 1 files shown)


app/routes/meetings/components/MeetingEditor.tsx
```diff
@@ -248,30 +278,13 @@ const MeetingEditor = ({
...
-                  [currentUserSearchable],
-                  invitedUsersSearchable,
-                  invitingUsers,
-                  'value'
-                );
-                return (
-                  <Field
-                    name="reportAuthor"
-                    label="Referent"
-                    placeholder="La denne stå åpen for å velge deg selv"
```
ivarnakken, line 263: Removed this placeholder, since it's a lie

eikhr (approved):
Looks good!

Bestem0r (approved):
Nice!

PR #4347, opened 2023-11-28
Bump react-prepare to v1.0.1

eikhr:
# Description

See https://github.com/webkom/react-prepare/pull/33, almost fixes ABA-686

Forms will acutally still usually not render, because `lego-editor` cannot be rendered on the server, but the bug with `react-prepare` is fixed:)

# Testing


PeterJFB (approved):
yes :shipit:

PR #4348, opened 2023-11-28
Refactor AppRoute and related components to use redux hooks

eikhr:
# Description

Refactor global app-components like `<App>` and `<Header>` to use hooks instead of connect and props. I think this makes the code more readable and hopefully easier to maintain.

# Result

No visual changes.

# Testing

Have tested a bit, with and without ssr. Stuff still works.


ollfkaih (approved):
lgtm

PR #4374, opened 2023-12-17
Fix incorrect tooltip for reactions

Bestem0r:
# Description

Fixes tooltip on reactions showing that "<Your name> reagerte med <emoji>" when reacting with an emoji on articles and quotes.

# Testing

Resolves [#ABA709](https://linear.app/abakus-webkom/issue/ABA-709/fix-incorrect-reaction-tooltip-label)

### DIFF
--- a/app/components/Reactions/Reaction.tsx
+++ b/app/components/Reactions/Reaction.tsx
@@ -28,6 +28,7 @@ type Props = {
   canReact: boolean;
   reactionId: ID;
   contentTarget: ContentTarget;
+  showPeople?: boolean;
 };
 // Note: Most use cases won't want to use this class directly. Instead, use
 // app/components/LegoReactions.
@@ -45,6 +46,7 @@ const Reaction = ({
   canReact,
   reactionId,
   contentTarget,
+  showPeople,
 }: Props) => {
   const classes = [
     className ? className : styles.reaction,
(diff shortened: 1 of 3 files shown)


app/routes/meetings/components/MeetingDetail.tsx
```diff
@@ -257,6 +257,7 @@ const MeetingDetails = ({
                   deleteReaction={deleteReaction}
                   parentEntity={meeting}
                   loggedIn={loggedIn}
+                  showPeople={true}
```
ivarnakken, line 260:
```suggestion
                  showPeople
```

app/components/Reactions/Reaction.tsx
```diff
@@ -60,7 +62,7 @@ const Reaction = ({
   }
 
   let tooltipContent = '';
-  if (users && users.length > 0) {
+  if (showPeople && users && users.length > 0) {
```
ivarnakken, line 65: Isn't `users` `null` on articles and quotes etc.? Why isn't this enough? `showPeople` seems overkill

Bestem0r: The problem is that redux makes `users` not null (it containts your reaction) when you react on an article/quote I think. So that will make the tooltip show only your reaction when you react in those cases.

Bestem0r: I am quite new to Redux and React in general, so let me know if you have a better solution for this @ivarnakken 😊

ivarnakken: ah yes you're right, it makes sense that Redux will store your own reaction 👍🏼

ivarnakken (approved):

PR #4377, opened 2023-12-25
Final steps in the `react-router` v6 migration! 🎉

ivarnakken:
# Description

[Remove redux-first-history](https://github.com/webkom/lego-webapp/commit/e092e4cc7610dc06d436dd3d6ad635dec2ee7e13) 

It is not compatible with `react-router` v6.4, and is not something we
"really" need.

---

[Remove compatabillity package and bump react-router to v6](https://github.com/webkom/lego-webapp/commit/be406e874940ae1bcceb399e4902ce41d7486e76) 

All uses of `useHistory` had to be replaced with `useNavigate`.

---

[Fix ssr routing with context](https://github.com/webkom/lego-webapp/commit/936c6045ffd563dc62a115bb55c438e60b5dfba8) 

`react-router` v6 does not allow context to be passed as a prop to
`StaticRouter`, so a simple workaround was to just create our own React
Context.

---

[Make lego-bricks compatible with webapp's upgraded router](https://github.com/webkom/lego-webapp/commit/82835c36864e81f57bb57bc030e49fadd6632a37) 

Also remove a bunch of stub https://github.com/types packages. Some are already provided in
the packages themselves, and the `react-router` specific ones caused some
compatabillity issues for some reason.

And fix a small ts error in the ConfirmModal code preventing build to complete.

# Testing

SSR seams to work completely fine. Tried to visit most routes, no problems.

However, I did notice some problems with building on node 16, so we ought to bump that very soon.


haukkagu:
Great job :partying_face: I'll try to review your code, although that's quite a daunting task :sweat_smile: I'll, at the very least, run the code and test if everything works alright :))

haukkagu (approved):
LGTM :fire:

ivarnakken:
This migration actually removes 305 ts errors!

norbye (approved):
Awesome stuff! Haven't run it to test it, but I've looked through all of the code changes and they look nice 💯

app/actions/UserActions.ts
```diff
@@ -123,44 +113,35 @@ export function updateUser(
...
-          successMessage: 'Oppdatering av bruker fullført',
-          errorMessage: 'Oppdatering av bruker feilet',
-        },
-      })
-    ).then((action) => {
-      if (!action || !action.payload) return;
-
-      if (!options.noRedirect) {
-        dispatch(push(`/users/${username}`));
-      }
```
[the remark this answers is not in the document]
ivarnakken: So this had to be refactored since the navigation should not be a part of the action function (because then you end up with nested dispatches). And yes all places that called this action was updated accordingly.

ivarnakken:
We're finally passing CI! I never thought this day would come when I started this migration back in August hehe

ivarnakken:
@haukkagu @norbye Made some few additions that you may want to look over 😄

app/routes/contact/components/ContactForm.tsx
```diff
@@ -54,7 +54,8 @@ const ContactForm = () => {
     () =>
       Promise.all([
         dispatch(fetchAllWithType(GroupType.Committee)),
-        dispatch(fetchGroup(REVUE_BOARD_GROUP_ID)),
+        // The revue board group does not exist in local dev environment
```
norbye, line 57: Wouldn't it be better to add the fixture to have the revue board group? And if you do not intend to do so in this PR, could you change/add the comment to something along the lines of;
```
// TODO Remove `, { propagateError: false } ` after adding the revue board group to the fixtures
```

app/routes/events/components/EventEditor/index.tsx
```diff
@@ -91,9 +91,10 @@ const validate = createValidator({
   title: [required('Tittel er påkrevd')],
   description: [required('Kalenderbeskrivelse er påkrevd')],
   eventType: [required('Arrangementstype er påkrevd')],
-  location: [
-    atLeastOneFieldRequired(
-      ['location', 'mazemapPoi'],
+  location: [required('Sted er påkrevd')],
```
norbye, line 94: I might be wrong - but isn't `location` empty if mazemap is used? And thus ideally speaking not required

app/routes/events/components/EventEditor/index.tsx
```diff
@@ -594,13 +604,45 @@ const EventEditor = () => {
...
+                      ];
+                    }
+                  } else {
+                    // Removing all pools so that they are not validated on submit
+                    if (values.pools.length > 0) {
+                      values.pools = [];
+                    }
+                  }
+
+                  return (
```
norbye, line 635: Stonks

app/components/NavigationTab/NavigationLink.tsx
```diff
@@ -1,32 +1,44 @@
...
-import { NavLink } from 'react-router-dom-v5-compat';
+import { NavLink, useLocation } from 'react-router-dom';
 import styles from './NavigationLink.css';
-import type { ReactNode } from 'react';
+import type { MouseEventHandler, ReactNode } from 'react';
 
 type Props = {
-  to: string;
-  onClick?: (e: Event) => void;
+  to: string | string[];
```
norbye, line 7: Initially didn't find it that intuitive that the remainder of the `to`-array is not used for navigating and only to decide `isActive`. Could it be considered to change the params to something like the following?
```suggestion
  to: string;
  activePaths?: string[];
```

Otherwise clean solution for the link 💯

norbye (approved):
Nice stuff(:

app/components/NavigationTab/NavigationLink.tsx
ivarnakken: Yes that's probably more intuitive

app/routes/events/components/EventEditor/index.tsx
ivarnakken: haha yes the code here is awful. I'm oblivious as to why this magically worked before, and I'd like to refactor it - but I'm not going to spend more hours (read 5) on this now.

app/routes/events/components/EventEditor/index.tsx
ivarnakken: Yes you're right! I just copied the existing validation when refactoring, so this is actually a bug in prod. However, it is rarely caught as `event.location` has a default value of `TBA`. 

https://github.com/webkom/lego-webapp/pull/4131/commits/52cb725b459ab8fdecd1eb9590cb62c408cadbf6#diff-9a787b7f7873cbf26136947b17f0a926272d1ce799f81ac1b7c0b0bab7f5c4efL707-L709

app/routes/contact/components/ContactForm.tsx
ivarnakken: Sure :)

app/components/NavigationTab/NavigationLink.tsx
ivarnakken: I'm thinking of calling it `additionalActivePaths` though and let it accept "additional" paths so that you don't have to write up the "main" path twice - less error prone I suppose.

norbye: Absolutely, I was thinking the exact same thing - but your wording is better

PR #4378, opened 2023-12-25
Fix pagination loading endlessly when there's nothing left to load

haukkagu:
# Description

The pagination would load endlessly when there's nothing left to load. Can for example be encountered [when searching for an email that doesn't exist](https://abakus.no/admin/email/users?enabled=true&userFullname=test)

# Result

It is fixed

# Testing

Resolves ABA-725


ivarnakken (approved):
This has been bugging me for a long time! Good stuff 💯

Bestem0r (approved):
Nice!

PR #4390, opened 2024-01-14
Resolve some router-v6 bugs

norbye:
Undefined objects for non-authorized users.
Missing location invocations.

# Result

Now it doesn't crash for those specific issues.

# Testing

Went through the webapp in most routes as both authenticated and non-authenticated user. Now these two issues appear to be resolved.

---

Resolves ABA-737

### DIFF
--- a/app/reducers/events.ts
+++ b/app/reducers/events.ts
@@ -321,7 +321,7 @@ export const selectMergedPoolWithRegistrations = createSelector(
               pool.permissionGroups
             );
             const registrations = total.registrations.concat(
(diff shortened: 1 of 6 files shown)


ivarnakken (approved):
Thanks so much for fixing!

norbye:
No worries:) uncovered another issue I haven't fixed just yet - but detailed it in linear.

And forgot to mention in the text description that after your PRs, the board,committee,etc. pages were visible to non-authenticated members - and now they are not (but then again they don't crash either)

PR #4400, opened 2024-01-18
Validate allergies in user registration form

ivarnakken:
# Description

We actually had 4 users with one of the "invalid" allergies, which I have now removed. But while snooting around I also found a bunch of other bad allergies, so I added those to the list 

# Result

image

# Testing

See image above.

### DIFF
--- a/app/utils/validation.ts
+++ b/app/utils/validation.ts
@@ -160,7 +160,19 @@ export const isValidAllergy =
     message = 'La feltet stå tomt hvis du ikke har noen allergier/preferanser'
   ) =>
   (value: string) => {
-    const notValidAnswers = ['ingen', 'ingenting', 'nei', 'nope', 'nada'];
+    const notValidAnswers = [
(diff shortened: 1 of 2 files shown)


app/utils/validation.ts
```diff
@@ -160,7 +160,19 @@ export const isValidAllergy =
...
+      'ingen',
+      'ingenting',
+      'nei',
+      'no',
+      'nope',
+      'none',
+      'n/a',
+      'null',
+      'nada',
+      'indøk',
```
haukkagu, line 173: :sunglasses:

haukkagu (approved):
Looks good.

However, isn't this a bit overcomplicated? What if we instead always display the text "La feltet stå tomt hvis du ikke har noen allergier/preferanser" above the input box (or as placeholder text). People will always find creative ways to circumvent the validation anyways...

norbye (approved):

PR #4401, opened 2024-01-18
Allow non-logged in users to view interest group details

ivarnakken:
# Description

Page will not crash on load, and it looks like it should.

This is not actually related to the react-router v6 migration, as this is actually a bug in prod as well.

# Result

Technically the user would only see the second before photo for half a second, but this illustrates the changes made - such as the "Admin" header always being shown to the user despite not being an admin at all ..

<table>
	<tr>
		 <td>Before</td>
		 <td>After</td>
	<tr>
	 <td>
		image
	</td>
	 <td>
		image
	</td>
   </tr>
   <tr>
	 <td>
		image
	</td>
	 <td>
		image
	</td>
	<tr>
</table>		

# Testing

See images above.

---

Related to ABA-744


haukkagu (approved):
LGTM

norbye (approved):
Nice!

PR #4403, opened 2024-01-18
Fix bug where search was performed on undefined string causing crash

norbye:
# Description

The check was for `null` values, but after the refactor the value was set to undefined - which is correct typewise, but fucked up the search function.  
Removed the null-value check as a null-value should not be allowed with the current type (`string | undefined`).

# Result

No crash

# Testing

---

Resolves ABA-746

### DIFF
--- a/app/components/Reactions/ReactionPicker.tsx
+++ b/app/components/Reactions/ReactionPicker.tsx
@@ -133,7 +133,7 @@ const searchEmojis = (
 
 const ReactionPicker = ({ emojis, contentTarget }: Props) => {
   const [activeCategory, setActiveCategory] = useState<string>();
-  const [searchString, setSearchString] = useState<string>();
+  const [searchString, setSearchString] = useState<string>('');
   const categories = useMemo(() => {
     if (!emojis) {
       return {};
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):
Hahah this was driving me nuts! Thanks for just fixing❤️

ivarnakken:
@norbye 
Have you looked at the code in [`ReactionPickerContent.tsx`](https://github.com/webkom/lego-webapp/blob/master/app/components/Reactions/ReactionPickerContent.tsx) as well? Shouldn't the type and the two checks be wrong here as well now?

norbye:
> @norbye Have you looked at the code in [`ReactionPickerContent.tsx`](https://github.com/webkom/lego-webapp/blob/master/app/components/Reactions/ReactionPickerContent.tsx?rgh-link-date=2024-01-19T17%3A52%3A45Z) as well? Shouldn't the type and the two checks be wrong here as well now? EDIT: I now see that the `useMemo` returns `null`, so this is actually correct. I misread.

ReactionPickerContent.tsx doesn't have access to `searchString`, so that should be good - at least from any changes in this PR.

> But, [this check](https://github.com/webkom/lego-webapp/blob/master/app/components/Reactions/ReactionPicker.tsx?rgh-link-date=2024-01-19T17%3A52%3A45Z#L186) in `ReactionPicker.tsx` seems wrong, no?

You're absolutely right, will fix:) while we're on it, it always checks for both undefined and empty string, so I might as well just set it to default to an empty string and as such the type won't allow undefined.

Would've been so nice if the linter picked up comparisons between invalid values

eikhr (approved):

PR #4405, opened 2024-01-20
Optimistically update reactions

ivarnakken:
# Description

This has been bugging me for a long time - having to wait for the promise to resolve each time I react to something. This is the normal way of doing it, but it's not foolproof since the state update itself is async as well. I guess we'll have to wait for `useOptimistic`.

# Result

**Before**
The gray screen is due to an uncaught error

https://github.com/webkom/lego-webapp/assets/69514187/5248e8c1-6322-4c84-9dbe-5336ce8be735

**After**

https://github.com/webkom/lego-webapp/assets/69514187/0588eb6f-722c-4e95-b4d3-d65cbc411718

# Testing

Reactions still work fine, both adding and removing. Also tested that users are still shown as expected on meetings.

### DIFF
--- a/app/components/LegoReactions/index.tsx
+++ b/app/components/LegoReactions/index.tsx
@@ -1,7 +1,6 @@
 import Reactions from 'app/components/Reactions';
 import Reaction from 'app/components/Reactions/Reaction';
 import { selectEmojis } from 'app/reducers/emojis';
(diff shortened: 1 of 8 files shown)


Bestem0r (approved):
Nice!

haukkagu (approved):

PR #4410, opened 2024-01-25
Refactor table component to functional

norbye:
# Description

Refactor Table component to functional and split it over multiple files

* Change checkbox to radio button
    * The behaviour has always been corresponding to Radio buttons, so I have no idea why they were checkboxes in the first place.
    * I have a PR out that was intended to change the behaviour to match checkboxes (#4257) - but on second though I'd rather refactor first (this PR) - then touch up the code in the other PR to support checkbox/radiobutton depending on the table and column, as not all filters are suited to be checkboxes
* Prevent autoscroll on open searchbar
    * If you go to prod/staging and scroll a little bit down before you click the search-icon on tables, you will be scrolled to the top of the page. This was annoying. This is gone.
* Members list seemed to have a bug in dev where the group be accessed before being loaded - `?` to the rescue

# Result

If you've made visual changes, please include **before** and **after** images, preferably with a description. Make sure they do not contain any real user information.

## Before
https://github.com/webkom/lego-webapp/assets/13599770/17d2fa3c-3e4a-4f7d-b55e-fa88a698a8db

## After
https://github.com/webkom/lego-webapp/assets/13599770/ce2a502b-f568-4b51-abe0-a2c37285b526

# Testing

Tested a on a few different tables with local backend and staging

---

Resolves ABA-763

### DIFF
--- a/app/components/Table/BodyCell.tsx
+++ b/app/components/Table/BodyCell.tsx
@@ -1,0 +1,44 @@
+import { get } from 'lodash';
+import type { ColumnProps, ShowColumn } from '.';
+
+type CellProps = {
+  column: ColumnProps;
+  data: Record<string, any>;
+  index: number;
+  showColumn?: ShowColumn;
+};
+
+const BodyCell: React.FC<CellProps> = ({ column, data, index, showColumn }) => {
+  if (column.columnChoices) {
+    if (!showColumn) {
+      return null;
+    }
+    const columnIndex: number = showColumn[column.dataIndex];
+    column = column.columnChoices[columnIndex];
+  }
+
+  const cellData = get(data, column.dataIndex);
+
+  const {
+    render = (cellData) => cellData,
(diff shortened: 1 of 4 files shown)


haukkagu (approved):
Nice!

itsisak (approved):
LGMT!

ivarnakken (approved):
I suppose some types here and there could be better, but this is nevertheless a big improvement! Thanks for refactoring this nasty component! <3

norbye:
> I suppose some types here and there could be better, but this is nevertheless a big improvement! Thanks for refactoring this nasty component! <3

Ah yeah - those anys where there from before and this PR felt big enough. I've fixed most of them in the next Table-PR that is coming with improved filtering - coming soon to a repo near you

I'm keeping your comments unresolved to make sure all of them are addressed in the next iteration(:

PR #4412, opened 2024-01-28
Fix group table being cleared after adding a new member

haukkagu:
# Description

A bug was recently implemented where the table over all the members of a group under `/admin/groups` would be cleared after adding a new member. Only a hard refresh could fix that problem. This has now been fixed.

# Result

**before**
Screencast from 2024-01-28 14-59-34.webm

**after**
Screencast from 2024-01-28 15-00-52.webm

# Testing

---

Resolves ABA-751


ivarnakken (approved):
Fetching the memberships again is kinda overkill but this is still a fix 🤷🏼‍♂️  
I believe there's an issue with the selector

PR #4416, opened 2024-01-29
Filter seenQuotes to only allow numbers

norbye:
# Description

It appears that https://github.com/webkom/lego-webapp/commit/b959dfb7ea094e7ba4d5601c5109c01ee3e48984 might have introduced a possible state where null/undefined values are loaded into the `seen` array that is passed to the backend - rendering arrays such as `[,254]` - which the backend very much does not like.

This should resolve that.

Errors can be found in sentry. Seemed to only affect a few people for some weird reason, so I couldn't reproduce - but this seems to fix it.

Could probably also be resolved with a
```
if (!quoteId) return;
```
but I don't have that big of a preference.

# Result

I tried inserting invalid values locally and they were stripped - so I'm assuming it works ™️ 

# Testing

---

Fixes LEGO-Q1, LEGO-2M

### DIFF
--- a/app/components/RandomQuote/RandomQuote.tsx
+++ b/app/components/RandomQuote/RandomQuote.tsx
@@ -24,6 +24,8 @@ const RandomQuote = ({ dummyQuote, useReactions = true }: Props) => {
   useEffect(() => {
     const quoteId = randomQuote.id;
 
+    if (!quoteId) return;
+
     if (!seenQuotes.current.includes(quoteId)) {
       seenQuotes.current = [...seenQuotes.current, quoteId];
     }


app/components/RandomQuote/RandomQuote.tsx
```diff
@@ -25,7 +25,9 @@ const RandomQuote = ({ dummyQuote, useReactions = true }: Props) => {
     const quoteId = randomQuote.id;
 
     if (!seenQuotes.current.includes(quoteId)) {
-      seenQuotes.current = [...seenQuotes.current, quoteId];
+      seenQuotes.current = [...seenQuotes.current, quoteId].filter(
+        (seenQuote) => !isNaN(parseInt(seenQuote))
+      );
```
ivarnakken, line 30:
```suggestion
      seenQuotes.current = [...seenQuotes.current, quoteId].filter(Boolean);
```
Is this simpler?

ivarnakken (approved):
If I understand it correctly, it's `quoteId` that can be nullish - not `seenQuotes` that contain a nullish item? If that's the case then I'm more fan of just checking `quoteId` instead of filtering the entire array

This is still a nice fix though, so I approve it. Good job of catching it! 🏅

norbye:
> If I understand it correctly, it's `quoteId` that can be nullish - not `seenQuotes` that contain a nullish item? If that's the case then I'm more fan of just checking `quoteId` instead of filtering and parsing the entire array
> 
> This is still a nice fix though, so I approve it. Good job of catching it! 🏅

True - updated to just verify the `quoteId` before manipulating the list

ivarnakken (approved):

PR #4418, opened 2024-01-29
Check if allergies are undefined in validator

norbye:
# Description

This error caused the validator on the register page to fail when the user did not provide an allergy (just like we want them to), making it impossible to register a new user.

Also added validators for first and last name, as they are required by the backend but weren't validated in the frontend.

# Result

The validator works

# Testing

---

Resolves ABA-771


ivarnakken (approved):
hmm why is this only a problem during registration and not in the settings? 🤨

norbye:
> hmm why is this only a problem during registration and not in the settings? 🤨

Because in the settings the form is loaded with the initial value from the backend, which is "" if I'm not mistaken. But on registration, initial values are never set, making the value undefined.

Ergo, it would also resolve the issue to set initialvalues in the registration form - but this feels more foolproof.

PR #4419, opened 2024-01-30
Store registrations as undefined if user has no access

norbye:
# Description

It would seem we missed another bug in the react-router migration.

As can be seen in LEGO https://github.com/webkom/lego/blob/master/lego/apps/events/models.py#L131, it's not enough to be logged in to be able to see registrations, you have to be either a member of abakom, creator of the event, or the event has to be in the future and you are allowed to sign up.

This became a problem when using the hook - as the return value would never become `undefined`, just an empty array - which made it seem like the user had permission even though it didn't -> crash

First reported by ludde in slack, but now it's been reported to sentry by 214 times by 94 users in the last 7 days - so it's about time to get this out of here

# Result

I've tested it locally through the staging backend where I got to reproduce it by removing myself from abakom, and it seems to work.

# Testing

---

Resolves ABA-773


PeterJFB (approved):
:100: :100: :100:

PR #4422, opened 2024-01-31
Remove `redux-form`

ivarnakken:
# Description

Finally!

# Result

No visual changes

# Testing

Tested all checkboxes and radio buttons, and all refactored forms. The join event form was pretty thoroughly tested.


app/components/Form/Field.tsx
norbye, line 1: Is the warning functionality you have removed here implemented somewhere else, or are warnings simply not supported any more?

app/components/LoginForm/ForgotPasswordForm.tsx
```diff
@@ -1,86 +1,56 @@
...
+});
 
-class ForgotPasswordForm extends Component<Props, State> {
-  mounted = false;
-  state = {
-    submitted: false,
-  };
+const ForgotPasswordForm = () => {
+  const [mounted, setMounted] = useState(false);
+  const [submitted, setSubmitted] = useState(false);
```
norbye, line 14: Might be possible to get these values from `react-final-form` through the formState https://final-form.org/docs/final-form/types/FormState

app/components/LoginForm/RegisterForm.tsx
```diff
@@ -1,96 +1,72 @@
...
+});
 
-class RegisterForm extends Component<Props, State> {
-  mounted = false;
-  state = {
-    submitted: false,
-  };
+const RegisterForm = () => {
+  const [mounted, setMounted] = useState(false);
+  const [submitted, setSubmitted] = useState(false);
```
norbye, line 22: Same here

app/components/LoginForm/RegisterForm.tsx
```diff
@@ -1,96 +1,72 @@
...
+        <Form onSubmit={handleSubmit} onClick={(e) => e.stopPropagation()}>
+          <Field
+            name="email"
+            component={TextInput.Field}
+            placeholder="E-post"
+          />
+          <Field
+            name="captchaResponse"
+            fieldStyle={{
+              width: 304,
```
norbye, line 59: Doesn't feel too dynamic(?)

norbye (approved):
Awesome stuff, good job!!

My comments are just small stuff - so do with it as you want

app/components/LoginForm/RegisterForm.tsx
ivarnakken: Yup but I didn't want to bother improving css here

app/components/LoginForm/ForgotPasswordForm.tsx
ivarnakken: Yes probably - I'll look into it tomorrow

ivarnakken: Looked into it now. It is possible to look at the submitted state, but it's not possible to get and use these outside of the form.

PR #4453, opened 2024-02-10
Enforce custom flex styling is prioritized

norbye:
# Description

Dirty fix - but it resolves the problem. Deployed it to https://webapp-staging.abakus.no/ to preview.

# Result

Werks

# Testing

---


ivarnakken:
https://github.com/webkom/lego-webapp/blob/master/packages/lego-bricks/src/components/Layout/Flex.module.css#L35

ivarnakken:
Deploying to staging first to check that it works on iOS - since that's the reason the `!important`s were added in the first place

ivarnakken:
Works fine 👍🏼

PR #4464, opened 2024-02-14
Properly reject promises on server

ivarnakken:
# Description

`Promise.resolve()` will not reject upon any of the input promises rejecting, which can cause promises to never stop on the client if the server sends unhandled promises. Simply replacing with a `Promise.all()` fixes this issue.

# Result

No visual changes.

# Testing

Spamming hard refreshes does not put page in limbo 👍🏼 

https://github.com/webkom/lego-webapp/assets/69514187/d271c68e-f5ad-450a-9775-0dd1d0726691

### DIFF
--- a/app/components/HeaderNotifications/index.tsx
+++ b/app/components/HeaderNotifications/index.tsx
@@ -92,7 +92,7 @@ const NotificationsDropdown = () => {
   usePreparedEffect(
     'fetchNotificationDropdownData',
     () =>
-      Promise.all([
+      Promise.allSettled([
         dispatch(fetchNotificationFeed()),
         dispatch(fetchNotificationData()),
       ]),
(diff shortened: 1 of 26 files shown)


ollfkaih (approved):
Probably fine, but wouldn't this potentially not run some callbacks if one of them fails?

ivarnakken:
> Probably fine, but wouldn't this potentially not run some callbacks if one of them fails?

Yes you're correct, but I can't think of a better solution?

ollfkaih:
> > Probably fine, but wouldn't this potentially not run some callbacks if one of them fails?
> 
> @ollfkaih Yes you're correct, but I can't think of a better solution

Allsettled should be the same but ensure all promises are settled?
Obviously only relevant where there are multiple callbacks.

itsisak (approved):
👍

ivarnakken:
> Allsettled should be the same but ensure all promises are settled?
Obviously only relevant where there are multiple callbacks.

ah yes you're right - this actually just seems like a better options for us here. Replaced them now

PR #4472, opened 2024-02-19
Implement overview-styling with grid

norbye:
# Description

Implemented the styling of the frontpage through CSS grid. The goal of this implementation is twofold;

1.  The overview both looks like a grid and smells like a grid - so imo it should be a grid
2. Plus this separates the components from the styling between them - facilitating moving components around and replacing them with improved versions without affecting the rest of the page.

Removed the two containers used for desktop and mobile and rather merged them into one responsive container.

Also updated some of the styling to add a little bit more space to the page, and make the spacing more consistent (e.g. adding padding to the right side of the screen and not only the left).

# Result

> [!CAUTION]

## Before

**NOTE: what looks like a right-overflow in the video only appears when you change the viewport. When you keep the viewport the same, it fills the screen**

https://github.com/webkom/lego-webapp/assets/13599770/68f552f7-8087-4565-b4df-1dbd50d3adc4

## After

https://github.com/webkom/lego-webapp/assets/13599770/4672269c-679b-4781-b242-4243f6b3acca

# Testing

---

Resolves ABA-832, ABA-423, ABA-727

### DIFF
--- a/packages/lego-bricks/src/components/Layout/Flex.tsx
+++ b/packages/lego-bricks/src/components/Layout/Flex.tsx
@@ -16,6 +16,7 @@ type Props = {
 
   /* Container component*/
   component?: ElementType;
+  componentRef?: React.RefObject<HTMLElement>;
 
   /** Wrap elements */
   wrap?: boolean;
@@ -44,6 +45,7 @@ const Flex = ({
   className,
   column = false,
   component: Component = 'div',
+  componentRef,
   wrap = false,
   wrapReverse = false,
   justifyContent = 'flex-start',
@@ -72,6 +74,7 @@ const Flex = ({
       gap,
       ...style,
     }}
+    ref={componentRef}
     {...htmlAttributes}
   >
     {children}
--- a/app/components/Banner/Banner.css
+++ b/app/components/Banner/Banner.css
@@ -1,8 +1,17 @@
 @import url('~app/styles/variables.css');
 
+.link {
+  display: block;
+  margin: 0 var(--lego-default-padding);
+
+  @media (--mobile-device) {
+    margin: 0 1rem;
+  }
+}
+
 .header {
   text-align: center;
-  margin-bottom: 20px;
+  margin-bottom: var(--lego-default-padding);
   animation-name: wiggle;
   animation-duration: 4s;
   animation-iteration-count: infinite;
--- a/app/components/Banner/index.tsx
+++ b/app/components/Banner/index.tsx
@@ -27,9 +27,11 @@ const Banner = (props: Props) => {
 
   const LinkComponent = ({ children }: { children: ReactNode }) => {
     return internal ? (
-      <Link to={link}>{children}</Link>
+      <Link to={link} className={styles.link}>
+        {children}
+      </Link>
     ) : (
-      <a href={link} target="_blank" rel="noreferrer">
+      <a href={link} target="_blank" rel="noreferrer" className={styles.link}>
         {children}
       </a>
     );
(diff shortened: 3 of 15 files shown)


ivarnakken (approved):
Much much cleaner! 😍 

This definitely "smells like grid"!

itsisak (approved):
Smart!

haukkagu (commented):
I love it! The only thing I slightly dislike is this empty space that occurs when the poll-component is too short:
image

ivarnakken:
> The only thing I slightly dislike is this empty space that occurs when the poll-component is too short:

I agree, but I think we'll redesign this page quite soon so it won't really be a problem for long.

Nevertheless, now that this is refactored I believe it's easier to "fix this" if we want to.

norbye:
> I love it! The only thing I slightly dislike is this empty space that occurs when the poll-component is too short:

Absolutely agree. I've been holding on merging this as I was thinking about replacing the poll component with the weekly component - as that should always be the same height as the pinned article component.

Or I might make the left section into a sub grid to enable the right side to resize on the y-axis independently of the left side - but I'm more keen on promoting the weekly-section as a first iteration

norbye:
Moved the weekly component up to pinned articles now, and updated the styling to match the height of pinned articles.

Tested with the longest length a description can have, so this should keep the styling looking good regardless - with the presumption that they will continue keeping the title in the same format as it is now.

Also the padding around the image annoyed me, so I removed it(:

image

haukkagu (commented):
Wow! That looks a lot cleaner 😍 Very good idea!!

Another thing I noticed was that the readme component becomes very long:
image

norbye:
> Another thing I noticed was that the readme component becomes very long:

Ah yeah, that component has been poorly styled for a while - so that happens on a few screen sizes already, but it would appear to happen more frequently now that I increased the padding and gave the component less space.

Will see what I can do about it

norbye:
Updated the styling of the latestReadme component now so that it's always displayed as a two column grid.

Also found out that the public frontpage never fetches readme's, so thats just an empty component rn - which is now fixed.

Regarding ABA-423, I kept only two readmes on the public frontpage, as the neighbor component looks weird with 4 as the height difference is too big.

Resolves ABA-423, ABA-727

Now there shouldn't be any more minor issues with the frontpage - but if there is I think it's time to look at that in another PR (unless it's caused by this PR and looks horrible 🤷 )

packages/lego-bricks/src/components/Layout/Flex.tsx
ivarnakken, line 1: I don't see a problem with this - but is it used?

ivarnakken (approved):
Beautiful!!

packages/lego-bricks/src/components/Layout/Flex.tsx
norbye: > Edit: Oh and the componentRef attribute is something I needed while the styling was still using the Flex component - but it's something I've thought about having a use for in other PRs I've done so I want to keep it even though I ended up not using it

norbye:
> Sort of related to this; why do we even separate this? Why not just show the exact same component with 4 readmes? I suppose it's nice to show the component expanded on the public frontpage to showcase it clearer - but why not show 4 there as well?

I kept it that way because the extra height of the readmes affected the spacing on the rest of the page - and I didn't really want to have the extra space under the pinned thingy;

With two readmes;
image


With four readmes;
image

So imo the 4-readmes work smoother on the logged in overview, as it only affects the components underneath, and not the components on its side

PR #4475, opened 2024-02-20
Fix paging and members table becoming empty

Bestem0r:
# Description

Previously the pagination on the group members page was not working. This would also cause the members table to become empty when editing roles. 

Changes:
- Fixed paging on the group members table
- Fixed group members table becoming empty when editing user roles 

# Result

Previously:
[Screencast from 20. feb. 2024 kl. 13.18 +0100.webm](https://github.com/webkom/lego-webapp/assets/33326578/96dc148f-73be-4cb6-bdcc-1a3f8a61a9b5)

Fixed:
[Screencast from 20. feb. 2024 kl. 13.39 +0100.webm](https://github.com/webkom/lego-webapp/assets/33326578/804d8ade-44c3-4c5b-927d-f5bfa84957c1)

# Testing

- Navigate to `/admin/groups/11/members`
- Change the role of one of the members.
- Verify that the table does not become empty and that the role is correctly updated.
- Navigate to `/admin/groups/3/members`.
- Scroll down and verify that paging works.

Resolves ABA-793

### DIFF
--- a/app/routes/admin/groups/components/AddGroupMember.tsx
+++ b/app/routes/admin/groups/components/AddGroupMember.tsx
@@ -1,15 +1,10 @@
 import { Field } from 'react-final-form';
-import {
-  addMember,
-  fetchMembershipsPagination,
-} from 'app/actions/GroupActions';
+import { addMember } from 'app/actions/GroupActions';
 import { Form, LegoFinalForm, SelectInput } from 'app/components/Form';
 import SubmissionError from 'app/components/Form/SubmissionError';
 import { SubmitButton } from 'app/components/Form/SubmitButton';
-import { defaultGroupMembersQuery } from 'app/routes/admin/groups/components/GroupMembers';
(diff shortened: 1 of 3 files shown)


haukkagu (approved):
Works on my machine. Good job!

itsisak (approved):
Nice! Just curious; would it be possible to prevent the table from disappearing while it fetches after role is updated? It would be a bit more smuud, but I get if this would be tricky

magnusbrecke (approved):
Looks good to me. Good code

norbye (approved):
Good job!

Now that this works again, it might be worth testing if the changes made in https://github.com/webkom/lego-webapp/pull/4412 to solve the clearing of the table upon editing roles are really necessary:) if they aren't it would be sweet to not fetch the entire list manually after submitting

dobif (approved):
Works for me. Does the page need to load the entire table after updating the roles?

PR #4480, opened 2024-02-21
Fix ordering of questions on survey

haukkagu:
# Description

Previously, you couldn't change the ordering of questions on surveys. This is now fixed (in a slightly hacky way).

# Result

**before**
Screencast from 2024-02-21 19-09-16.webm

**after**
Screencast from 2024-02-21 19-10-12.webm

# Testing

I've tested for several surveys and it works.

---

Resolves ABA-792

### DIFF
--- a/app/routes/surveys/components/SurveyEditor/SurveyForm.tsx
+++ b/app/routes/surveys/components/SurveyEditor/SurveyForm.tsx
@@ -62,10 +62,11 @@ const SurveyForm = ({
     return onSubmit({
       ...surveyData,
       event: surveyData.event.value,
-      questions: surveyData.questions.map((question) => ({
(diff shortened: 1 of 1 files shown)


ivarnakken (approved):
I suppose you could use the question ids as well but I don't think using the indexes here makes a difference

Good stuff - very clean!

ivarnakken:
The videos you shared don't seem to work btw

itsisak:
The videos work for me..

itsisak (approved):

itsisak:
No, its the same for me then to, I thought you meant you could not view them at all

jonasdeluna (approved):
🛸🛸🛸🏎️

PR #4482, opened 2024-02-23
Accurately check checkboxes on /events filter

ivarnakken:
# Description

The default checked value of these checkboxes when visiting /events would be incorrect if you visit with query params, e.g. through the links on the frontpage "Bedpres og kurs" and "Arrangementer".

# Result

https://github.com/webkom/lego-webapp/assets/69514187/e10356c3-4c59-46e6-a682-0353dbccf61d

# Testing

The checkboxes are now given default values when visiting the page with query params.


Bestem0r (approved):
Looks good to me!

PR #4484, opened 2024-02-25
Clean up detailed event page

ivarnakken:
# Description

This all took more time than anticipated. There's still more to be done - but let's be agile

[Prevent user from accidentally unregistering by pressing enter](https://github.com/webkom/lego-webapp/commit/6ff070d0f36ce4344afec9d997e9c45547d24afe) 

Enter would cause the form to be submitted, which either registers or
unregisters the user.

This was reported to us by mail.

---

[Clean up detailed event page](https://github.com/webkom/lego-webapp/commit/9322008c382e025f81b360829844ee70a7361cba) 

Skeleton components have been implemented as much as I could given that most events are quite different. 

Minor clean difficult for the naked eye to spot have been made, such as spacing the sidebar components less random and more "correct".

Bugs have been fixed, e.g. this one 
image

# Result

**Before**
Previously, there was only a loading indicator

image

**After**
<video src="https://github.com/webkom/lego-webapp/assets/69514187/122d9da6-2a4a-4f6f-8b0f-fed0a3074a9f" />

# Testing

Tested registering and unregistering events, on as many different events as I could. Works fine when not logged in as well.

---

Resolves ABA-811
Resolves ABA-813
Resolves ABA-847

### DIFF
--- a/app/components/InfoList/index.tsx
+++ b/app/components/InfoList/index.tsx
@@ -1,3 +1,6 @@
+import { Flex } from '@webkom/lego-bricks';
+import { isNotNullish } from 'app/utils';
+import styles from './styles.css';
 import type { ReactNode } from 'react';
 
 type Item = {
@@ -15,21 +18,23 @@ type Props = {
  * Location <strong>Oslo</strong>
  * Time <strong>Yesterday</strong>
  */
-function InfoList({ items, className }: Props) {
+const InfoList = ({ items, className }: Props) => {
   return (
-    <table className={className}>
-      <tbody>
-        {items.filter(Boolean).map(({ key, keyNode, value }) => (
-          <tr key={key}>
-            <td>{keyNode ?? <span>{key}</span>}</td>
-            <td>
-              <strong>{value}</strong>
-            </td>
-          </tr>
-        ))}
-      </tbody>
-    </table>
+    <Flex column gap="0.5rem" className={className}>
+      {items.filter(isNotNullish).map(({ key, keyNode, value }) => (
+        <Flex
+          alignItems="center"
+          justifyContent="space-between"
+          gap="1rem"
+          key={key}
+        >
+          {keyNode ?? <span>{key}</span>}
+
+          <span className={styles.value}>{value}</span>
+        </Flex>
+      ))}
+    </Flex>
   );
-}
+};
 
 export default InfoList;
--- a/app/routes/events/components/Event.css
+++ b/app/routes/events/components/Event.css
@@ -1,35 +1,19 @@
 @import url('~app/styles/variables.css');
 
-.feedback {
-  margin-top: 1rem;
+.form {
+  gap: 1rem;
+}
+
+.feedbackLabel {
+  font-size: var(--font-size-lg);
 }
 
 .feedbackText {
   min-height: 40px;
-  min-width: 200px;
+  min-width: 190px;
   max-width: 893px;
   max-height: 300px;
   resize: vertical;
-}
-
-.registeredThumbnails {
-  flex-wrap: wrap;
-  max-height: 130px;
-  overflow: hidden;
-}
-
-.meta {
-  flex: 1;
-  padding: 20px;
-  padding-top: 0;
-  border-left: 1px solid var(--border-gray);
-  margin-left: 3em;
-
-  @media (--small-viewport) {
-    border-left: none;
-    margin-left: 0;
-    padding: 0;
-  }
 }
 
 .subHeader {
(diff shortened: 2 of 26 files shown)


app/routes/events/components/Event.css
```diff
@@ -1,56 +1,27 @@
...
-  border-left: 1px solid var(--border-gray);
-  margin-left: 3em;
-
-  @media (--small-viewport) {
-    border-left: none;
-    margin-left: 0;
-    padding: 0;
-  }
+.feedbackButton {
+  margin-top: 23.4px; /* The height of the field label */
```
Bestem0r, line 16: Is there no better way to do this?

Bestem0r (approved):
Really nice refactoring! And I love the skeleton! 😍

app/routes/events/components/Event.css
ivarnakken: hmm I can't think of a way other than updating the actual field component or moving the label outside - which at the time seemed excessive for this "edge case".

I totally agree that this is hacky and very bad practice - so it's a good thing you point it out!

ivarnakken: When I come to think of it - moving the label is probably cleaner. I'll look into it!

app/components/InfoList/index.tsx
```diff
@@ -15,21 +17,23 @@ type Props = {
...
+      {items.filter(isNotNullish).map(({ key, keyNode, value }) => (
+        <Flex
+          alignItems="center"
+          justifyContent="space-between"
+          gap="1rem"
+          key={key}
+        >
+          {keyNode ?? <span>{key}</span>}
+
+          <span style={{ fontWeight: '600', textAlign: 'right' }}>{value}</span>
```
norbye, line 32: could we pop this in to styles?

app/routes/events/components/JoinEventForm.tsx
```diff
@@ -484,25 +469,27 @@ const JoinEventForm = ({
...
                             component={TextInput.Field}
                             label={feedbackLabel}
                             className={styles.feedbackText}
+                            onKeyDown={(e) => {
+                              // Prevent user from registering / unregistering by pressing enter
+                              if (e.key === 'Enter') {
+                                e.preventDefault();
+                              }
+                            }}
+                            withoutMargin
```
norbye, line 486: Could it be an option to change the `<Form>` thingy instead? I just feel like this quickfix could easily be broken by any other browser-specific send functionality, and considering the pain of handling people that unregistered on purpose it would be nice to make it a little harder to avoid.

E.g. changing

```
<Form onSubmit={handleSubmit}>
```
to
```
<Form>
```

as handleSubmit is already called in the SubmitButton's onSubmit.

This should™️ remove the need of the preventDefault, but if you agree - please double check

norbye (approved):
Sweet stuff! Looks veri nice:) considering this is very styling heavy you might want to rebase with master to get in the refactored overview to ensure that the compactevents look good still

haukkagu (approved):
LGTM :fire:

app/components/InfoList/index.tsx
ivarnakken: There was no css file so I just didn't bother, but I moved it now 👍🏼

app/routes/events/components/Event.css
ivarnakken: Ended up moving the label outside of the Field component - worked like a charm!

app/routes/events/components/JoinEventForm.tsx
ivarnakken: Very good point!

However, I found a middle ground. Simply removing the `onSubmit` from `<Form />` did not stop the field from causing a submit (and registering / unregistering), so the `e.preventDefault()` is actually still needed, but I figured we should only prevent users from accidentally unregistering, while accepting registrations.

I ended up removing the `onSubmit` though, in case a user somehow finds a way around this and still manages to unregister ..

norbye: Ahhhh that makes sense - I guess it would be logical of browsers to look for the button[type=submit] if the onSubmit is not presented - so that wouldn't help. Although it might work to replace the onSubmit with `onSubmit={isRegistered && submit}`? (this is just me guessing)

Hmm although that may or may not mess up what happens when the button is actually clicked on purpose.. either way your solution solves the problem - so if anyone finds another problem at a later stage we can tackle it then

> I figured we should only prevent users from accidentally unregistering, while accepting registrations. Agree?

Very much agree - it's nice to make it as simple as possible to register

PR #4486, opened 2024-02-27
Removed button to show more events when there are no more events to s…

magnusbrecke:
# Description

Fixed the frontage so that when events displayed on the front-page has reached the event limit there is no longer a button to show more events. 

# Result
Before:

Skjermbilde 2024-02-27 kl  20 38 48

After:

Skjermbilde 2024-02-27 kl  20 38 21

- [x ] Changes look good on both light and dark theme.

# Testing

---

Resolves ABA-647

### DIFF
--- a/app/routes/overview/components/Overview.tsx
+++ b/app/routes/overview/components/Overview.tsx
@@ -126,7 +126,7 @@ const Overview = () => {
         <Articles articles={articlesShown} />
       </section>
 
-      {frontpage.length > 8 && (
+      {frontpage.length > 8 && events.length > eventsToShow && (
         <div className={styles.showMore}>
           <Icon onClick={showMore} name="chevron-down-outline" size={30} />
         </div>


app/components/Header/index.tsx
```diff
@@ -83,10 +83,12 @@ const AccountDropdownItems = ({ onClose }: AccountDropdownItemsProps) => {
   );
 };
 
+
 const UpcomingMeetingButton = () => {
+  
   const upcomingMeetingId = useAppSelector(selectUpcomingMeetingId);
   const navigate = useNavigate();
-
+
```
ivarnakken, line 91: `prettier` will complain about this. Run `yarn prettier` to auto-fix

ivarnakken (approved):
This has bugged me for so long! Awesome! 🏅

app/components/Header/index.tsx
magnusbrecke: Yep ran the command should be good now:)

Bestem0r (approved):
Very nice! 🚀

dobif (approved):
LGTM!

norbye:
Just dropping in with a quick comment on the checkboxes in the description;

if you want them to show as actual checkboxes you can either

1. Write `- [x] Tada`
2. Leave them as they are in the template (`- [ ] Tada`) until you create your pull request. After the pull request is created, you can then click on them to check them:)

That will make them show up as this;

- [x] Tada

Or this

- [ ] Tada

Instead of this;

- [x ] Whops

PR #4492, opened 2024-03-02
Clean up survey results

ivarnakken:
# Description

Bugs fixed, improved graphs, more responsive, better code and overall cleaner.

There's still much more to be done to our surveys, but I've spent enough time on it for now

# Result

I don't really want to show many before / after images of this for obvious reasons, but I've picked out some stuff here and there

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
			image
        </td>
        <td>
			image
        </td>
	</tr>
	<tr>
        <td>
			image
        </td>
        <td>
			image
        </td>
	</tr>
	<tr>
        <td>
			image
        </td>
        <td>
			image
        </td>
    </tr>
	<tr>
        <td>
			image
        </td>
        <td>
			image
        </td>
    </tr>
</table>

# Testing

Tested different surveys and different routes within. 
Tested on most viewports - big improvement.
Tested graphs on the event statistics page, the only other place using these

---

Resolves ABA-858
Resolves ABA-843

### DIFF
--- a/app/reducers/surveys.ts
+++ b/app/reducers/surveys.ts
@@ -55,7 +55,7 @@ export const selectSurveys = createSelector(
 
 export const selectSurveyById = createSelector(
   (state: RootState) => selectSurveys(state),
-  (_: RootState, surveyId: ID) => surveyId,
+  (_: RootState, surveyId?: ID) => surveyId,
   (surveys, surveyId) => {
     return surveys.find((survey) => survey.id === Number(surveyId));
   },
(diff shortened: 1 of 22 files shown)


app/reducers/surveys.ts
```diff
@@ -55,7 +55,7 @@ export const selectSurveys = createSelector(
 
 export const selectSurveyById = createSelector(
   (state: RootState) => selectSurveys(state),
-  (_: RootState, surveyId: ID) => surveyId,
+  (_: RootState, surveyId?: ID) => surveyId,
```
norbye, line 58: You've changed other surveyId's to `string`. Is it intentional that it is still `ID` here?

norbye (approved):
Mostly skimmed through it - but it looked pretti nice:))

app/reducers/surveys.ts
ivarnakken: They should actually be `ID` - I thought they needed to be `string` (like it actually is since it's retrieved from query) to fix a ts error, but setting it as optional was enough. Rewrote it now

PR #4494, opened 2024-03-02
Fix minor css issues caused by recent webpack bump

ivarnakken:
# Description

Let me know if you find any more

# Result

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>         
			image
        </td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>         
			image
        </td>
    </tr>
</table>

# Testing

Tested changes with ssr and on smaller viewports. Changes to attendance modal was also tested, but didn't bother to add a picture since the changes here were very minor (only 1px of spacing) because it has a bottom part (unlike the interest group modal) which makes the flex stretching less _sizeable_ (?).

### DIFF
--- a/app/routes/companyInterest/components/CompanyInterestPage.tsx
+++ b/app/routes/companyInterest/components/CompanyInterestPage.tsx
@@ -29,7 +29,6 @@ import {
 } from 'app/components/Form';
 import SubmissionError from 'app/components/Form/SubmissionError';
 import { SubmitButton } from 'app/components/Form/SubmitButton';
-import { Image } from 'app/components/Image';
 import { readmeIfy } from 'app/components/ReadmeLogo';
 import Tooltip from 'app/components/Tooltip';
 import { selectCompanyInterestById } from 'app/reducers/companyInterest';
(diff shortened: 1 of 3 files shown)


ollfkaih (approved):
Lgtm

app/routes/companyInterest/components/CompanyInterestPage.tsx
```diff
@@ -228,7 +227,7 @@ const LanguageFlag = ({ language }: { language: string }) => {
       break;
   }
 
-  return <Image src={flag} className={styles.flag} alt="country_flag" />;
+  return <img src={flag} className={styles.flag} alt="Country flag" />;
```
norbye, line 230: Why no `Image`?

norbye (approved):

app/routes/companyInterest/components/CompanyInterestPage.tsx
ivarnakken: Because the `Image` component comes with a `width: 100%`, which overrides the styling given by `styles.flag`. Besides, there's no real reason to use the `Image` component here because we don't utilize any of its features, like progressive or lazy loading, since the images are loaded directly from the assets.

PR #4514, opened 2024-03-05
Minor changes to event detail page

ivarnakken:
# Description

* Hide event type on mobile, as it took up unnecessary space
* Show skeleton components for registrations appropriately, as registrations are separated in the store
* Hide responsible group mail address in a tooltip
* Make admin buttons only clickable on the actual buttons
* Employ the actual text input component for the delete event field
* Clean up other minor stuff

# Result

# Testing

Things look as expected, and deleting events still work.

---

Resolves ABA-866
Resolves ABA-865
Resolves ABA-817
Resolves ABA-721

### DIFF
--- a/app/components/Content/ContentHeader.css
+++ b/app/components/Content/ContentHeader.css
@@ -1,3 +1,5 @@
+@import url('~app/styles/variables.css');
+
 .header {
   display: flex;
   justify-content: space-between;
@@ -12,4 +14,8 @@
   display: flex;
   justify-content: flex-end;
   flex-grow: 1;
+
+  @media (--mobile-device) {
+    display: none;
+  }
 }
--- a/app/routes/events/components/Admin.tsx
+++ b/app/routes/events/components/Admin.tsx
@@ -4,7 +4,9 @@ import { useState } from 'react';
 import { Link, useNavigate } from 'react-router-dom';
 import { deleteEvent } from 'app/actions/EventActions';
 import AnnouncementInLine from 'app/components/AnnouncementInLine';
+import { TextInput } from 'app/components/Form';
 import { useAppDispatch } from 'app/store/hooks';
+import styles from './Admin.css';
 import type { ActionGrant } from 'app/models';
 import type { ID } from 'app/store/models';
 import type { AuthUserDetailedEvent } from 'app/store/models/Event';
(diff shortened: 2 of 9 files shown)


app/routes/events/components/EventDetail/index.tsx
```diff
@@ -529,13 +524,13 @@ const EventDetail = () => {
...
                       currentRegistration={currentRegistration}
-                      skeleton={showSkeleton}
+                      skeleton={fetching && !registrations}
                     />
                     <AttendanceStatus
                       toggleModal={toggleModal}
                       pools={pools}
                       legacyRegistrationCount={event.legacyRegistrationCount}
-                      skeleton={showSkeleton}
+                      skeleton={fetching && !registrations}
```
ivarnakken, line 533: image

Here the event was retrieved from the store (because the event was already fetched from the overview page), while the registrations still need to be fetched

norbye (approved):
Nice stuffs, LGTM

Bestem0r (approved):
Looks very good!

app/components/Content/ContentHeader.css
```diff
@@ -12,4 +14,8 @@
   display: flex;
   justify-content: flex-end;
   flex-grow: 1;
+
+  @media (--mobile-device) {
+    display: none;
+  }
```
ShaileshS1702, line 20: What does media here do?? I am intrigued

app/routes/events/components/Admin.tsx
```diff
@@ -21,51 +23,47 @@ const DeleteButton = ({ eventId, title }: ButtonProps) => {
...
+                  navigate('/events');
+                })
+              }
+            >
+              {({ openConfirmModal }) => (
+                <Button onClick={openConfirmModal} danger>
+                  <Icon name="trash" size={19} />
+                  Slett
+                </Button>
+              )}
```
ShaileshS1702, line 61: I see that this an arrow key function. But for love of all that is holy, I can't understand how the first line is supposed to work. Could you clarify?

app/routes/events/components/Admin.tsx
```diff
@@ -21,51 +23,47 @@ const DeleteButton = ({ eventId, title }: ButtonProps) => {
...
-        )}
-      </ConfirmModal>
-    );
-  }
-
   return (
-    <div>
-      {show === false && (
+    <>
+      {!show ? (
```
ShaileshS1702, line 28: Nice 💸💸💸💸💸💸💸💸💸💸💸💸💸💸💸💸💸💸💸💸

app/routes/events/components/EventEditor/index.tsx
```diff
@@ -152,8 +153,17 @@ const validate = createValidator({
...
+  let { eventIdOrSlug } = useParams<{
+    eventIdOrSlug: string;
+  }>();
   const isEditPage = eventIdOrSlug !== undefined;
+
+  const { state } = useLocation();
+  // Look for event id in the location state given from the admin "copy event" button
+  if (!isEditPage && state?.id) {
+    eventIdOrSlug = state.id;
+  }
```
ShaileshS1702, line 165: Is it not possible to use conditional operator when initializing eventIdOrSlug. So moving the if statement predicates to the conditional operator. Conditional operators are a thing of beauty. So in short keeping eventIdOrSlug as a const variable. I am wondering this since you only change it's value once. Shouldn't you use let only if you intend to change the value many times?

ShaileshS1702 (changes requested):
Good work Nakken my man. Cheers. Just some minor things I am wodeimg about.

app/components/Content/ContentHeader.css
ivarnakken: Good question!

This is what we call [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries), and it is essentially how we achieve ["responsive layouts"](https://www.w3schools.com/html/html_responsive.asp). The _condition_ in this case is `--mobile-device`, which is a variable of `max-width: 992px` set in our `variables.css` file. This means that whenever the screen has a width of less than `992px`, the styling set within the brackets will "kick in" (which in this case removes some stuff with the `display: none`). Media queries are very interesting, and a lot more powerful than what we use it for in lego-webapp! It's something we should definitely use more.

app/routes/events/components/EventEditor/index.tsx
ivarnakken: > Conditional operators are a thing of beauty.

haha yes, at least sometimes.

I totally agree that using `let` is semi bad practice, especially if, as you say, it is not intended to change a lot.

Are you suggesting something like

```ts
  const params = useParams<{
    eventIdOrSlug: string;
  }>();
  const isEditPage = params.eventIdOrSlug !== undefined;
  const { state } = useLocation();
  // Fallback to a potential event id, e.g. given from the admin "copy event" button
  const eventIdOrSlug = params.eventIdOrSlug ?? state?.id;
```

I agree that this is both cleaner and better practice - so unless you were thinking of something else, I'll change it to this. Seems to work completely fine.

Thanks! 😄

app/routes/events/components/Admin.tsx
ivarnakken: It's nice that you're asking - I agree that this is kind of weird javascript.

The `ConfirmModal` takes in what we call "children" as a prop. React let's us pass this prop into the component by wrapping the `ConfirmModal` around the "children", which in this case is the delete button. This makes for much better readability, and you can see another example of this pattern [here](https://reactpatterns.js.org/docs/function-as-child-component/).

Now, what's happening on the line that you point out. Our `ConfirmModal` takes the `children` prop and puts it within itself, but in addition it also [passes a `openConfirmModal` function as a prop to its `children`](https://github.com/webkom/lego-webapp/blob/master/packages/lego-bricks/src/components/Modal/ConfirmModal.tsx#L118). Kind of weird, right? This means that the children of `ConfirmModal` (here, the code you see above) has access to this function `openConfirmModal`. But why the weird brackets and parentheses? Well, what we're doing here is ["destructing"](https://dev.to/angski/destructuring-props-in-react-4hfc) the function out of the props. This is sort of hacky javascript, as it lets us extract (destruct) things from objects and assign them to variables (in this case, `openConfirmModal`).

Does that make kinda sense? Good question!

haukkagu (approved):

ShaileshS1702 (approved):
LGTM!
Thanks for an in-depth answer to my question. I've gotten a much better understanding of some of the concept in JS, even though I'm not sure I understood everything. I probably just need to try it out

PR #4529, opened 2024-03-07
Warn users about using their student mail when registering

ivarnakken:
# Description

This has been talked about for far too long

# Result

https://github.com/webkom/lego-webapp/assets/69514187/08a57f14-7aea-4cc7-ab69-5465dc35ea3e

# Testing

You can still continue the registration with a student mail, and warnings are not given on normal mails.

### DIFF
--- a/app/components/LoginForm/RegisterForm.tsx
+++ b/app/components/LoginForm/RegisterForm.tsx
@@ -1,3 +1,4 @@
+import { Button, ConfirmModal, Flex, Icon } from '@webkom/lego-bricks';
 import { useEffect, useState } from 'react';
 import { Field } from 'react-final-form';
 import { sendRegistrationEmail } from 'app/actions/UserActions';
@@ -9,8 +10,22 @@ import {
   SubmissionError,
   LegoFinalForm,
 } from 'app/components/Form';
+import Tooltip from 'app/components/Tooltip';
 import { useAppDispatch } from 'app/store/hooks';
+import { spyValues } from 'app/utils/formSpyUtils';
 import { createValidator, required, isEmail } from 'app/utils/validation';
+import styles from './RegisterForm.css';
+
+const isStudMail = (email: string) =>
+  email.toLowerCase().endsWith('@stud.ntnu.no') ||
+  email.toLowerCase().endsWith('@ntnu.no');
+
+type FormValues = Partial<{
+  email: string;
+  captchaResponse: string;
+}>;
+
+const TypedLegoForm = LegoFinalForm<FormValues>;
 
 const validate = createValidator({
   email: [required(), isEmail()],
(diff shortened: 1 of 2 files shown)


app/components/LoginForm/RegisterForm.tsx
```diff
@@ -45,27 +57,42 @@ const RegisterForm = () => {
...
           <SubmissionError />
-          <SubmitButton dark>Registrer deg</SubmitButton>
+
+          {spyValues<FormValues>((values) => {
+            if (values.email && isStudMail(values.email)) {
+              return (
+                <ConfirmModal
+                  title="Pass på!"
+                  message="Det ser ut som du prøver å registrere deg med en student-e-post.
+                  Denne kommer du til å miste tilgang til når du går ut fra NTNU, og du vil da også miste brukeren din på abakus.no.
```
ivarnakken, line 78: I'd like some input on the wording here. It's not like the user will immediately lose the account - it's more that you won't see the "deleting account" mails, but I don't want to explain all that and make this text very long.

app/components/LoginForm/RegisterForm.tsx
```diff
@@ -10,8 +11,19 @@ import {
   LegoFinalForm,
 } from 'app/components/Form';
 import { useAppDispatch } from 'app/store/hooks';
+import { spyValues } from 'app/utils/formSpyUtils';
 import { createValidator, required, isEmail } from 'app/utils/validation';
 
+const isStudMail = (email: string) =>
+  email.endsWith('@stud.ntnu.no') || email.endsWith('@ntnu.no');
```
norbye, line 18: Could throw in a quick `.toLowerCase` in case people are idiots

app/components/LoginForm/RegisterForm.tsx
norbye:
```suggestion
                  Denne kommer du til å miste tilgang til når du går ut fra NTNU, og da vil du ikke lenger kunne lese e-post om oppdateringer som tilbakestilling av passord og automatisk sletting av brukeren din <Tooltip content={"For å respektere personvern og ikke lagre informasjon vi ikke trenger sletter vi brukere som ikke har vært logget inn på ett år. I perioden fram til dette skjer sender vi ut e-post for å advare deg i tilfelle du vil beholder brukeren din."}>InfoIcon</ToolTip>.
```
Not a 100% happy with my seggestion for the wording either, but imo it takes us very few words to add a few bits of context to make the text factually correct so imo we should do just that.

Rubs me the wrong way when people write stuff that's just wrong to avoid people doing something they don't want them to do

Initially my suggestion was without the info tooltip - so might fly without as well - but why not just add it

norbye (approved):
Sweet!

haukkagu (approved):

app/components/LoginForm/RegisterForm.tsx
ivarnakken: I agree that we shouldn't write anything "wrong", so I've updated it now. Good idea with the tooltip!

norbye (approved):
Beautiful

PR #4531, opened 2024-03-08
Refactor frontpage logic and improve types

eikhr:
# Description

I think a lot of our code for dealing with redux state is unnecessarily complicated. I have a longer term goal of simplifying our state management by doing the following:
- Removing uneccesary models
	- Decreading number of serializers in the backend (generally just one Detailed and one List serializer)
	- Not doing any transformations of data in redux selectors like adding author details to articles (this can be done in the component that needs author data)
- Moving logic down the component tree (into components instead of pages)
- Typing everything

This PR contains some refactorings of the frontpage, following some of the principles outlined above. Most components on the frontpage will now select their own relevant data and it's all typed (except some event stuff in redux). I also fixed a circular reference issue with the quotes reducer type that broke the redux-state `RootState` type.

# Result

One tiny visual change: The pinned box on the frontpage now has title "Oppslag" instad of "Festet oppslag" when nothing is pinned (in this case it will just show the first upcoming event/most recently posted article).

# Testing

I have visited both the public and authenticated frontpage and compared it to staging, everything looks identical:)
I have also tested all functionality on the Quotes pages to ensure the new quotes-reducer works as expected.

---


app/routes/overview/components/Pinned.tsx
```diff
@@ -20,18 +18,20 @@ const Pinned = ({ item, url, meta, style }: Props) => {
...
       state.frontpage.fetching ||
-      (isEvent(item) ? state.events.fetching : state.articles.fetching),
+      (item && isEvent(item) ? state.events.fetching : state.articles.fetching),
   );
 
   return (
     <Flex column style={style} className={styles.pinned}>
-      <h3 className="u-ui-heading">Festet oppslag</h3>
+      <h3 className="u-ui-heading">
+        {item?.pinned === false ? 'Oppslag' : 'Festet oppslag'}
```
ivarnakken, line 27: Not a big fan of boolean comparisons
```suggestion
        {item?.pinned ? 'Festet oppslag' : 'Oppslag'}
```

Any reason for it?

ivarnakken (approved):
Much much much cleaner! 😍 I love all the proper filtering

app/routes/overview/components/Pinned.tsx
eikhr: Yes, I wanted it to say "Festet oppslag" while loading as we usually will have something pinned. I realise this is a bit unclear, I should probably just do something like ```fetching || item?.pinned ? ...``` instead

PR #4537, opened 2024-03-08
Rename Overview to AuthenticatedFrontpage

eikhr:
# Description

This has annoyed me for so long, it's not at all obvious that the `Overview` component is the front page when logged in. Especially when the public front page is called `PublicFrontpage`. Here I have renamed it to `AuthenticatedFrontpage` to make it super obvious what the component is, and to reduce confusion between the frontpage compoenent and the two other components also named `Overview`.

# Result

- Folder previously named "overview" is now named "frontpage" (contains both public and authenticated frontpage)
- `Overview`-component is now called `AuthenticatedFrontpage`
- logic for showing the correct frontpage component is moved from `AuthenticatedFrontpage.tsx` to `index.ts` in "routes/frontpage"-folder

# Testing

Frontpage still works:)

---


ivarnakken (approved):
Agreed!

PR #4538, opened 2024-03-08
Fix "send announcement" link button

ivarnakken:
# Description

Forgot about this

State is now correctly passed

# Result

https://github.com/webkom/lego-webapp/assets/69514187/e500c964-9430-4623-8e44-2e2c1a712d0f

# Testing

Tested all uses of the link component; events, interest groups and meetings.

Verified that there are no more of these links with bad state passed

### DIFF
--- a/app/components/AnnouncementInLine/index.tsx
+++ b/app/components/AnnouncementInLine/index.tsx
@@ -12,28 +12,29 @@ type Props = {
   group?: UnknownGroup;
 };
 
-const TypedLink = Link<AnnouncementCreateLocationState>;
-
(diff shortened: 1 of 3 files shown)


eikhr (approved):

haukkagu (approved):
Great work!

I think it's good practice to avoid using æøå for branch names. GitHub complains about it

ivarnakken:
> I think it's good practice to avoid using æøå for branch names. GitHub complains about it

Yep - my branch names are auto-generated from linear so I didn't notice

Bestem0r (approved):

PR #4540, opened 2024-03-11
Refactor more redux code

eikhr:
# Description

Mostly refacoring redux slices to use `createSlice` and `legoAdapter` instead of `createEntityReducer`.
The following redux-slices are refactored:
- groups
- registrations
- pools
- users
- feedActivities
- feeds
- meetingInvitations
- toasts
- threads

I have also refactored some redux code, mostly selectors, in the following slices:
- articles
- emailLists
- emailUsers

I added a selector to `legoAdapter` for selecting entities based on some specified field in the entitiy. 
It basically allows us to write this:
```typescript
const { selectByField } = legoAdapter.getSelectors((state) => state.articles);
export const selectArticlesBySlug = selectByField('slug');
```
instead of this:
```typescript
export const selectArticlesBySlug = createSelector(
  selectArticles,
  (_: RootState, props: { slug: string }) => props.slug,
  (articles,slug) => articles.filter((article) => article.slug === slug),
);
```
especially useful to reduce duplicate code if we need to select by multiple different fields.

# Result

No visual or functional changes

# Testing

I have tested a bit, but I still want to do a thorough test of all the functionality I have touched just in case I made some silly mistake:)

Progress on ABA-688


jonasdeluna (commented):
This seems like a really cool change although I don't understand all the underlying redux code 👍👍

ivarnakken (approved):
I like! It's awesome that you typed the specs as well 😍

PR #4550, opened 2024-03-12
Remove 'ingen er påmeldte' from registrations when not logged in

magnusbrecke:
# Description

Removed {registrations} over the modal that shows how many are registered for an event when not logged. It just showed "ingen er påmeldt" no matter what. Now it is removed. Still works as normal when logged in. 

# Result

Before: 
Skjermbilde 2024-03-12 kl  20 41 50

After: 
Skjermbilde 2024-03-12 kl  20 42 29

# Testing

---

Resolves ABA-881

### DIFF
--- a/app/routes/events/components/EventDetail/index.tsx
+++ b/app/routes/events/components/EventDetail/index.tsx
@@ -519,12 +519,13 @@ const EventDetail = () => {
                 users={registrations?.slice(0, 14).map((reg) => reg.user)}
                 skeleton={fetching && !registrations}
               />
+
               <AttendanceModal key="modal" pools={pools} title="Påmeldte">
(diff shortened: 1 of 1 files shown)


falbru (approved):
LGTM

dobif (approved):
LGTM, but is it never supposed to say "Ingen påmeldte" when you are not logged in?

ivarnakken (approved):

ivarnakken:
> LGTM, but is it never supposed to say "Ingen påmeldte" when you are not logged in?

This is kind of fair. Could be an idea to just update the logic in `RegisteredSummary` to just not include any names when not logged in

eikhr (approved):

PR #4556, opened 2024-03-13
Improve spacing and etc. here and there

ivarnakken:
# Description

[Reposition components on user settings page](https://github.com/webkom/lego-webapp/commit/932a6879a8d9e8f23f30d4a39d05d4e3de9525f5) 

Less whitespace, and more coherent.

---

[Refactor error boundary](https://github.com/webkom/lego-webapp/commit/f99fea2f337e4c8c110eabb008d403d0a1847b68) 

Instead of a custom styling, it now uses the danger card

---

[Improve and clean up spacing in bdb](https://github.com/webkom/lego-webapp/commit/2d14c3c8a7a17b0486c0c534a620e6a4a7424d81) 

Still more to be done here

# Result

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
			image
        </td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
			Screenshot 2024-03-13 at 13 45 13
        </td>
    </tr>
</table>

# Testing

No "breaking" changes here - but everything was tested.

---

Resolves ABA-704


app/components/Comments/CommentView.tsx
```diff
@@ -37,12 +37,7 @@ const CommentView = (props: Props) => {
   return (
     <div style={style}>
       <Title displayTitle={displayTitle} />
-      <Flex
-        gap="var(--spacing-md)"
-        style={{
-          flexDirection: newOnTop ? 'column-reverse' : 'column',
```
ivarnakken, line 43: This causes the comment field to be put at the bottom, which in my eyes is just unpractical

norbye: I propose to keep it as it is, but rather move the commentform out of the flex component (or add a separate flex around the commenttree that can be reversed). If you check out e.g. https://abakus.no/bdb/4488 there are a lot of comments, and it would be useful to have the newest on top - with CommentForm above.

Update: I now noticed that there is actually some reversing going on in line 51 (`comments={newOnTop ? tree.reverse() : tree}`) when calling CommentTree, but that doesn't seem to work? as that should have made comments reverse twice..  
So remove either one of them but make sure the remaining one actually works

app/components/ErrorBoundary/ErrorBoundary.css
```diff
@@ -1,22 +1,4 @@
...
-
-  & img {
-    height: 100px;
-    width: 100px;
-  }
-}
-
-.message {
-  margin-left: 10px;
+  width: 40%;
```
norbye, line 2: Have you tested how smol this can get on phones? Obviously I haven't either - but I feel like this could benefit from a media query: width: 90% or something like that

app/routes/users/components/UserProfile.tsx
```diff
@@ -171,8 +171,9 @@ const UserProfile = () => {
 
   const params = useParams<{ username: string }>();
   const { currentUser } = useUserContext();
-  const isCurrentUser =
-    useIsCurrentUser(params.username) || params.username === 'me';
+  // const isCurrentUser =
+  //   useIsCurrentUser(params.username) || params.username === 'me';
```
norbye, line 175: Might as well delete this, eh?

norbye (approved):
Awesome stuff!! A few noncritical comments

app/components/Comments/CommentView.tsx
ivarnakken: hmm I assumed the `tree.reverse()` was enough. I'll look into it

app/components/ErrorBoundary/ErrorBoundary.css
ivarnakken: Yes you're right - didn't think of that hehe

app/routes/users/components/UserProfile.tsx
ivarnakken: 😬

app/components/Comments/CommentView.tsx
norbye: Lol my bad, `tree.reverse()` is definitely enough - ignore my previous comment(:

At first I thought the comments and the commentform were all in the same div, but they're evidently not

norbye: So I agree with the changes you made

app/routes/users/components/UserSettings.tsx
```diff
@@ -91,112 +96,116 @@ const UserSettings = () => {
...
-          <Form onSubmit={handleSubmit}>
-            <Field
-              placeholder="Brukernavn"
-              label="Brukernavn"
-              name="username"
-              readOnly
-              component={TextInput.Field}
-              disabled
-            />
+          <form onSubmit={handleSubmit}>
```
falbru, line 125: Quick question 😁 Why did this get turned into the default form?

falbru (approved):
Awesome! The new user page looks amazing :star:

itsisak (approved):
Only skimmed through, but it looks good!

app/routes/users/components/UserSettings.tsx
ivarnakken: Because the `Form` includes some flex styling, but it ended up not really mattering

PR #4557, opened 2024-03-13
La til en tidtaker på Abakus.no

ShaileshS1702:
# Description

La til en nedtelling i banner på abakus.no

# Result

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
        </td>
        <td>
image
        </td>
        <td>
image
        </td>
    </tr>
    <tr>
        <td>
        </td>
        <td>
        </td>
        <td>
image
        </td>
</tr>
</table>

# Testing

---


falbru (approved):
This works on my machine, although I think something like this would be cleaner: https://www.geeksforgeeks.org/how-to-use-setinterval-method-inside-react-components/

norbye (commented):
I'd preferably have this logic contained within the banner component - as it is not needed for anything else on the frontpage. If you only want this to happen for authenticated users it's simple enough to fetch `const { loggedIn } = useUserContext();` within the Banner component.

Throwing in a ➕1 on @falbru's comment. If you want a live example of it we currently use it in admissions https://github.com/webkom/admissions/blob/master/frontend/src/routes/LandingPage/CountDown.tsx

ShaileshS1702:
I didn't have enough time to look at it today. I will try to look at a solution tomorrow

ShaileshS1702:
> I'd preferably have this logic contained within the banner component - as it is not needed for anything else on the frontpage. If you only want this to happen for authenticated users it's simple enough to fetch `const { loggedIn } = useUserContext();` within the Banner component.
> 
> Throwing in a ➕1 on @falbru's comment. If you want a live example of it we currently use it in admissions https://github.com/webkom/admissions/blob/master/frontend/src/routes/LandingPage/CountDown.tsx

Isn't the AuthenticatedFrontpage only available for those who has loggedIn, or should I do a check on that?

norbye:
> Isn't the AuthenticatedFrontpage only available for those who has loggedIn, or should I do a check on that?

Yes it is, and if this is only a temporary one off thing there's no need to do it and you won't add the banner on the non-authenticated front page - I was just thinking more permanent

PR #4558, opened 2024-03-13
Use variables.css from lego-webapp in lego-bricks

norbye:
# Description

Seems a bit unnecessary to have two files with essentially the same content. There were a few differences, but that was only due to variables.css in lego-bricks not being updated every time someone changed variables.css in lego-webapp.

# Result

There should be no visual changes.

# Testing

Needs some more testing to verify that the variables work outside of lego-webapp.

---

Resolves ABA-780


ivarnakken:
Has this been tested with other packages using the variables file (currently just lego-editor)? Like how do you import this there? Would a solution be to have an empty variables.css file that only imports the lego-webapp variables (or perhaps the other way around)?

ivarnakken:
Also, did you check for inconsistencies between the two. I believe one of them is more "up-to-date"

norbye:
> Also, did you check for inconsistencies between the two? I believe one of them is more "up-to-date"

Yes

> There were a few differences, but that was only due to variables.css in lego-bricks not being updated every time someone changed variables.css in lego-webapp.


> Has this been tested with other packages that use the variables file (currently just lego-editor)? Like how do you import this there? Would a solution be to have an empty variables.css file that only imports the lego-webapp variables (or perhaps the other way around)?

Not yet, that's my main part of "testing-required"

> Needs some more testing to verify that the variables work outside of lego-webapp.

ivarnakken:
Perhaps I should read the descriptions hmmm

norbye:
> Perhaps I should read the descriptions hmmm

Now why would you do that when I could just copy all the parts that are not part of your template into the comments as well

ivarnakken:
Resolves ABA-780

eikhr:
I don't think this will work outside of lego-webapp, but if you just remove the one in lego-webapp and keep the one in lego-bricks instead, I think it should be fine.

norbye:
> I don't think this will work outside of lego-webapp, but if you just remove the one in lego-webapp and keep the one in lego-bricks instead, I think it should be fine.

I'll have to test this more - but I think it works all right. As far as I can see, the only place `lego-editor` imports styles from `lego-bricks` is in https://github.com/webkom/lego-editor/blob/master/src/Editor.css by importing the styles directly from the already built dist folder. And when I run `yarn build` on lego-webapp and bricks from this branch, the built style.css file still contains he variables because they are importet from `lego-webapp` into the styles used by bricks, and thus need to go in the compiled stylesheet.

eikhr:
Ahh, yeah you're absolutely right. All the styles will be added to the style.css when building, so as long as vite can find them during build we are fine!

eikhr (approved):

ivarnakken (approved):
:shipit:

PR #4560, opened 2024-03-14
Refactor and type feed rendering components

eikhr:
# Description

Notifications and the timeline page (announcements) show what LEGO calls _feeds_. They are basically a collection of activities, like a sent announcement or getting bumped up from a waiting list. The code for rendering these gets a bit complicated as we need to support different activitites that get rendered in slightly different ways. In this PR I have mostly just typed everything, and refactored it a bit to work better with typescript. I also renamed a lot of stuff so that it makes more sense to me (and hopefully you too).

# Result

No visual changes, but I fixed a small bug with feed-activities that mention users in the title:

**Before:**
Screenshot 2024-03-14 at 23 34 50
**After:**
Screenshot 2024-03-14 at 23 35 43

# Testing

I have tested with all the feed activities I was able to create. There are a few that are not possible to create using the webapp (or maybe at all?) that I have not tested, but they should still work fine if they even worked before.


ivarnakken (approved):
haha it's awesome that you actually refactored this

PR #4561, opened 2024-03-14
Migrate to RouterProvider

itsisak:
# Description

Migrating to RouterProvider in order to fully take advantage of the recent migration to `react-router-dom v6` by @ivarnakken. The new `createBrowserRouter()` (instead of `<BrowserRouter />` allows us to utilize the new DATA APIs.

If anyone knows more about, `loader`, `hydration`, `errorboundary` it seems like we can add such things to `createBrowserRouter`, however I am unsure if this would be necessary.

~~SSR is something that needs to be configured for this. Currently I just does not render the router if it is SSR... There seems to be good [docs](https://reactrouter.com/en/main/guides/ssr) for this.~~

# Result / Testing

As a start I primarily changed the "root" routing as backwards compatibility allows the rest to still work. I intend to change the remaining routes, the same way the routing for `admin/` has been changed atm. 

I have tested all the routes I have changed, now every single sub-route. 

---

Resolves ABA-883


jennymul (approved):
Great WORK!!

eikhr (commented):
This looks very cool! I think we might be able to completely get rid of `react-prepare` if we use the data routers:)

You will need to convert the ssr-router as well though, in `/server/ssr.ts`.

itsisak:
@eikhr Now SSR should work. I dont know how to extensively test it, but when running `yarn ssr:staging` the very first response (to file `/` in network tab) returns the same response as abakus.no, and I assume that means it works. 

Regarding `react-prepare`, is there any thing in specific you meant we can replace it with our just simply remove it?

eikhr:
Great! That is how I usually check SSR too:)

Regarding `react-prepare` I just mean that it seems like data-routes can provide the same functionality, giving us the ability to await data loading on the server before server-side rendering. It would require quite a lot of refactoring to change all out data loading to react-router loaders, though. But it might be worth looking into in the future, as `react-prepare` is a little bit hacky to put it mildly.

eikhr:
This looks really great! What's missing before this is ready for review?

itsisak:
All that is missing now is just migrating the remaining routes. I agrre further improvements can be done on a later occasion.

itsisak:
I have tested every single route and they still work! 

I noticed on the survey route the `SubmissionPublicResultsPage` doesn't seem to show anything. Is this currently unused and can be removed?

Maybe `loadable` can be replaced with `route.lazy` at some point, if we want to.

eikhr:
Nice! 
Regarding `SubmissionPublicResultsPage`, you're right that it doesn't show anything, but it is definitely supposed to, so don't remove it! It is used to share submission results with people who are not admins. I will look into why it doesn't work, but it doesn't seem to be related to this PR at all.

eikhr:
I have fixed `SubmissionPublicResultsPage` here: #4571 :)

app/routes/meetings/index.tsx
```diff
@@ -1,17 +1,17 @@
...
-    <Route path=":meetingId/edit" element={<MeetingEditor />} />
-    <Route path="*" element={<PageNotFound />} />
-  </Routes>
-);
+const MeetingList = loadable(() => import('./components/MeetingList'));
+const MeetingEditor = loadable(() => import('./components/MeetingEditor'));
+const MeetingDetailWrapper = loadable(() => import('./MeetingDetailWrapper'));
 
-export default MeetingRoute;
+const MeetingsRoute: RouteObject[] = [
```
eikhr, line 9:
```suggestion
const meetingsRoute: RouteObject[] = [
```
I think all the routes should start with a lower case letter, as they are not react components.

app/routes/surveys/components/Submissions/SubmissionsIndividual.tsx
```diff
@@ -1,15 +1,11 @@
 import cx from 'classnames';
+import { useOutletContext } from 'react-router-dom';
 import StaticSubmission from '../StaticSubmission';
 import styles from '../surveys.css';
-import type { SelectedSurvey } from 'app/reducers/surveys';
-import type { SurveySubmission } from 'app/store/models/SurveySubmission';
+import type { ContextType } from '../../index';
```
eikhr, line 5:
```suggestion
import type { ContextType } from 'app/routes/surveys';
```
Easier to read absolute imports, especially with multiple `../`s

app/routes/surveys/components/Submissions/SubmissionsSummary.tsx
```diff
@@ -1,24 +1,20 @@
...
+import { useOutletContext } from 'react-router-dom';
 import { hideAnswer, showAnswer } from 'app/actions/SurveySubmissionActions';
 import Tooltip from 'app/components/Tooltip';
 import { useAppDispatch } from 'app/store/hooks';
 import { isNotNullish } from 'app/utils';
 import styles from '../surveys.css';
 import Results from './Results';
 import type { GraphData } from './Results';
-import type { SelectedSurvey } from 'app/reducers/surveys';
+import type { ContextType } from '../../index';
```
eikhr, line 10:
```suggestion
import type { ContextType } from 'app/routes/surveys';
```

app/routes/surveys/index.tsx
```diff
@@ -1,45 +1,63 @@
...
+  () => import('./components/Submissions/SubmissionsSummary'),
+);
+const SubmissionsIndividual = loadable(
+  () => import('./components/Submissions/SubmissionsIndividual'),
+);
+const SubmissionPublicResultsPage = loadable(
+  () => import('./components/Submissions/SubmissionPublicResultsPage'),
+);
+
+export type ContextType = {
```
eikhr, line 33:
```suggestion
export type SurveysRouteContext = {
```
Maybe specify that this is specifically the surveys route context, not just any context

server/request.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Convert an incoming request to a fetch request
+ * Why? The docs did it
+ * https://reactrouter.com/en/main/guides/ssr
+ */
+const createFetchRequest = (req, res) => {
```
eikhr, line 6: Would be awesome if you typed this, but maybe not necessary

server/server.ts
```diff
@@ -89,4 +90,7 @@ app.use((err, req, res, next) => {
   res.statusCode = 500;
   res.end('Internal Error');
 });
+// app.get("*", (req, res) => {
+//   const fetchRequest = createFetchRequest(req, res);
+// });
```
eikhr, line 95: Remove?

eikhr (approved):
Looks great! Can't wait to experiment with the new data api features:)

ivarnakken (approved):
This looks good, and I like that you've added loadable to all the pages!

I'm sorry my review took so long 😅 

---

The only "bug" I could  find after spending some time navigating around was this:

https://github.com/webkom/lego-webapp/assets/69514187/4d6531a4-c17c-4750-8243-76f16c0f6e65

When clicking on /members, the query is lost. Re-clicking fixes it though? BUT, this is also in `master` - so let's go :shipit:

itsisak:
@ivarnakken  Thanks for the review, will be great to get this out now! 

Felt like adding loadable to all pages was the best option since we no longer can add loadable to a `route` they way we did before now that its just a list instead of a component. It might be worth looking into `router.lazy` at some point instead, but this works fine!

PR #4566, opened 2024-03-16
Add background color to mazemap embed

norbye:
# Description

Add a background to the MazeMap embed so you can see where it's supposed to appear while it is loading.

Right now there is an ugly empty area where the mazemap embed is supposed to load for a short while while it is loading. I think this would make it look a little smoother.

# Result

- [ ] Changes look good on both light and dark theme.
- [ ] Changes look good with different viewports (mobile, tablet, etc.).

&nbsp;

I'm not able to reproduce this locally, so we're gonna have to go for a "trust me bro" and rather revert it if it is not a success - or deploy it to staging before merging it

# Testing

---

### DIFF
--- a/app/components/MazemapEmbed/MazemapEmbed.css
+++ b/app/components/MazemapEmbed/MazemapEmbed.css
@@ -12,4 +12,10 @@
 
 .mazemapEmbed {
   --lego-font-color: var(--color-absolute-black);
+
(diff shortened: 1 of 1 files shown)


ShaileshS1702 (approved):
LGTM

ivarnakken (approved):

PR #4569, opened 2024-03-18
Add pool suggestions displaying frequently used groups

norbye:
# Description

Added some suggestions of frequently used group combinations to make life simpler for event organizers.

The group combinations disable if more comprehensive groups have been selected, ensuring the the minimum amount of groups required is stored, while maintaining the group overlap and so on.

The functionality does not at all affect the input, so the users can still use it as they used to, but I believe this could resolve the errors of adding the wrong groups as this is simpler than typing them in.

The suggestions listen to the content of the list, so even if it is not used to select the groups, it can be used as an indicator of what groups you've already selected.

# Result

## Preview

https://github.com/webkom/lego-webapp/assets/13599770/e68f0d6c-6af9-44f1-8782-8b5daed58dec

# Testing

---

Resolves ABA-848

### DIFF
--- a/app/routes/events/components/EventEditor/EventEditor.css
+++ b/app/routes/events/components/EventEditor/EventEditor.css
@@ -89,6 +89,14 @@
 
 .metaList > ul > li.poolBox {
   flex-basis: 32%;
+
+  @media (--medium-viewport) {
+    flex-basis: 49%;
+  }
+
+  @media (--small-viewport) {
+    flex-basis: 100%;
+  }
 }
 
 .metaList > span {
@@ -163,4 +171,24 @@
   margin: var(--spacing-md) 0;
   color: var(--secondary-font-color);
   text-align: center;
+}
+
+.groupSuggestionWrapper {
+  margin-bottom: var(--spacing-sm);
+
+  > .groupSuggestion {
+    cursor: pointer;
+    font-size: var(--font-size-xs);
+
+    &.selected {
+      background-color: var(--lego-font-color);
+      color: var(--inverted-font-color);
+    }
+
+    &.implicit {
+      cursor: not-allowed;
+      background-color: var(--color-gray-4);
+      color: var(--inverted-font-color);
+    }
+  }
 }
--- a/app/routes/events/components/EventEditor/PoolSuggestions.tsx
+++ b/app/routes/events/components/EventEditor/PoolSuggestions.tsx
@@ -1,0 +1,147 @@
+import { Flex } from '@webkom/lego-bricks';
+import cx from 'classnames';
+import { useState } from 'react';
+import Pill from 'app/components/Pill';
+import styles from './EventEditor.css';
+import type { SuggestionComponent } from 'app/components/Form/SelectInput';
+
+type GroupSuggestion = {
+  label: string;
+  values: { label: string; value: number }[];
+  children?: number[]; // Child group ids of the suggestion (these will be removed if this suggestion is selected)
+};
+
+const suggestions: GroupSuggestion[] = [
+  {
+    label: 'Alle studenter',
+    values: [{ label: 'Students', value: 14 }],
+    children: [15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26],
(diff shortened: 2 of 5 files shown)


app/routes/events/components/EventEditor/renderPools.tsx
```diff
@@ -15,11 +16,6 @@ type poolProps = {
   eventStatusType: EventStatusType;
 };
 
-const highWarning = (value) =>
-  value && value >= 500
-    ? 'Åpent arrangement gir uendelig plasser 😁'
-    : undefined;
-
```
eikhr, line 22: Why remove this?

app/routes/events/components/EventEditor/PoolSuggestions.tsx
```diff
@@ -0,0 +1,151 @@
...
+    values: [
+      { label: '5. klasse Datateknologi', value: 20 },
+      { label: '5. klasse Kommunikasjonsteknologi', value: 26 },
+    ],
+  },
+];
+
+type PoolSuggestionType = React.ComponentProps<
+  typeof SelectInput<{ label: string; value: number }>
+>['SuggestionComponent'];
```
eikhr, line 69: Could we maybe just export the `SuggestionComponent`-type from `SelectInput.tsx`?

eikhr (approved):
Looks great! This makes it a lot easier to select the typical event groups, and I think it will really help people to understand the implicit groups too.

app/routes/events/components/EventEditor/EventEditor.css
```diff
@@ -164,3 +172,31 @@
...
+
+    &.selected {
+      background-color: var(--color-green-6);
+    }
+
+    &.implicit {
+      cursor: not-allowed;
+      background-color: var(--color-green-3);
+
+      @media (prefers-color-scheme: dark) {
```
itsisak, line 191:
```suggestion
      [data-theme='dark'] {
```
`prefers-color-scheme: dark` is still true in light-mode when system/browser prefers dark

app/routes/events/components/EventEditor/EventEditor.css
```diff
@@ -164,3 +172,31 @@
...
+        background-color: var(--color-green-6);
+      }
+    }
+  }
+}
+
+@media (prefers-color-scheme: dark) {
```
itsisak, line 198:
```suggestion
[data-theme='dark'] {
```

itsisak (approved):
Great idea!

app/routes/events/components/EventEditor/renderPools.tsx
ivarnakken: Warning are currently not supported with react-final-form

ivarnakken (approved):
Very useful feature!

I am, however, not sure about the use of green. We don't really use that elsewhere to indicate an "active" state. Might I suggest the the `var(--lego-font-color)` color, like we utilize for navigation tabs and buttons?

app/routes/events/components/EventEditor/renderPools.tsx
norbye: Yeah, I figure we might re-add warnings where we want to at some point, but for now it's just dead code throwing console errors - so I'd rather remove it

app/routes/events/components/EventEditor/PoolSuggestions.tsx
norbye: Excellent idea, updated now

app/routes/events/components/EventEditor/EventEditor.css
norbye: Sweet, that explains why it wasn't working(: ended up removing it, but will add the data-theme stuff if the styling should require it

norbye:
> I am, however, not sure about the use of green. We don't really use that elsewhere to indicate an "active" state. Might I suggest the the `var(--lego-font-color)` color, like we utilize for navigation tabs and buttons?

Absolutely, changed to use lego-font-color now and updated the preview recording

PR #4570, opened 2024-03-18
Up z-index for toast and gallery dropdown

norbye:
# Description

Upping the z-index so that we have more numbers to play with to conf lower laying items. Setting it pretty high as these should always be on top.

# Result

# Testing

---

Resolves ABA-889


itsisak (approved):
Sweet!

PR #4575, opened 2024-03-19
Make event rules more visible

jennymul:
# Description

Remove how many spots there are left for an event and add concent for event rules

# Result

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            ...
        </td>
        <td>

        </td>
        <td>
        
        </td>
    </tr>
</table>

# Testing

---

Resolves ABA-276

### DIFF
--- a/app/routes/events/components/JoinEventForm.tsx
+++ b/app/routes/events/components/JoinEventForm.tsx
@@ -47,11 +47,6 @@ import type {
 import type Penalty from 'app/store/models/Penalty';
 import type { CurrentUser } from 'app/store/models/User';
 
-type SpotsLeftProps = {
-  activeCapacity: number;
-  spotsLeft: number;
-};
-
 /**
  *  Not using app/components/SubmitButton because that will "falsely"
  *  be disabled when the form is pristine, which it often is.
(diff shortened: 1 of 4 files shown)


app/routes/events/components/JoinEventForm.tsx
```diff
@@ -172,24 +167,6 @@ const PaymentForm = ({
   </div>
 );
 
-const SpotsLeft = ({ activeCapacity, spotsLeft }: SpotsLeftProps) => {
```
ivarnakken, line 175: In my opinion, we still need this. Any reason for removing?

jennymul: Don't agree, everyone can do the simple math when it says 23/30 places left..

falbru: It would also make the page more cluttered

falbru (approved):
Looks good!

app/routes/events/components/JoinEventForm.tsx
ivarnakken: hmm okay, fair enough hehe

ivarnakken (approved):
Much more visible! Nice 😄

ivarnakken:
> Also the mount() function, to my understanding, renders the element without the context of the rest of the route. It seems that the error message given in drone is related to the fact that the component uses <Link> without being wrapped in a router.

Yes this seems to be the issue. Wrapping the event form with a `MemoryRouter` seems to fix it though - at least on my computer™️ 

```diff
function renderJoinEventForm(props = {}) {
  return mount(
+    <MemoryRouter>
      <Provider store={store}>
        <JoinEventForm {...defaultProps} {...props} />
      </Provider>
+    </MemoryRouter>,
  );
}
```

PR #4591, opened 2024-03-22
Replace UserContext with redux state selectors

eikhr:
# Description

The UserContext was simply mirroring what was in the redux-state, and was completely unnecessary.

# Result

Only refactoring. I replaced all usage of `useUserContext` with either `useCurrentUser` or `useIsLoggedIn` which directly select from the redux state. I also did some minimal refactoring in the various components using `useCurrentUser`, because it now returns `undefined` instead of `{}` when the user is not yet loaded or not logged in.

Number of typescript errors reduced from 731 to 712:)

# Testing

I read through all my changes, and for all components where behaiviour might have changed, I did some manual testing of the component.

### DIFF
--- a/app/routes/events/components/EventDetail/index.tsx
+++ b/app/routes/events/components/EventDetail/index.tsx
@@ -26,6 +26,7 @@ import { AttendanceStatus } from 'app/components/UserAttendance';
 import AttendanceModal from 'app/components/UserAttendance/AttendanceModal';
 import UserGrid from 'app/components/UserGrid';
 import config from 'app/config';
+import { useCurrentUser, useIsLoggedIn } from 'app/reducers/auth';
 import {
   selectCommentsForEvent,
   selectEventByIdOrSlug,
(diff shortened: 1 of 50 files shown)


app/routes/events/components/EventDetail/index.tsx
```diff
@@ -162,6 +164,8 @@ const EventDetail = () => {
       : selectPoolsForEvent(state, { eventId }),
   );
 
+  const showSkeleton = (fetching && isEmpty(event)) || !currentUser;
```
ivarnakken, line 167: Won't this prevent unauthenticated users from viewing event details?

ivarnakken (approved):
This has bothered me for long! 🏅

app/routes/events/components/EventDetail/index.tsx
eikhr: Ahh, right! Good catch! That makes fixing the type-errors a bit harder...

eikhr: I think this page will need to be refactored more later to fix all the types

PR #4593, opened 2024-03-22
Clear error code when navigating to new page

eikhr:
# Description

Refactor routing slice to use `createSlice` and fix a bug where the error-code f.ex. 404 would not get cleared even after navigating to a different page.

# Result

Stronger typings for the routing slice in redux, and now error codes clear as expected:

Before:
https://github.com/webkom/lego-webapp/assets/8343002/43ac5e2d-73ab-47a8-bc30-1a68a7a186ef

After:
https://github.com/webkom/lego-webapp/assets/8343002/d4a63654-2fdb-4afe-9b9a-4877274d6666

# Testing

---

Resolves ABA-902


ivarnakken (approved):
This has annoyed me so long!

PR #4594, opened 2024-03-22
Refactor and type surveys state and components

eikhr:
# Description

Refactored `surveys` and `surveySubmissions`-slices in redux to use `legoAdapter` and `createSlice`.

Changed the survey-selector so that it no longer puts the relevant event into the survey. This can instead be selected separately if necessary. This is mainly to simplify and reduce the number of different types.

# Result

No visual or functional changes
Reduces number of typescript errors by 7

# Testing

I did all the survey things and they still work. 

Some survey things don't well for non-admins, but that's unrelated to this PR. F.ex. the page for reviewing your survey submission gets a 403 error from the backend. 

---

ABA-688

### DIFF
--- a/app/reducers/surveys.ts
+++ b/app/reducers/surveys.ts
@@ -1,3 +1,4 @@
+import { createSlice } from '@reduxjs/toolkit';
 import { usePreparedEffect } from '@webkom/react-prepare';
 import { omit } from 'lodash';
 import { createSelector } from 'reselect';
@@ -7,9 +8,10 @@ import {
   fetchWithToken,
 } from 'app/actions/SurveyActions';
 import { useAppDispatch, useAppSelector } from 'app/store/hooks';
-import createEntityReducer from 'app/utils/createEntityReducer';
+import { EntityType } from 'app/store/models/entities';
+import createLegoAdapter from 'app/utils/legoAdapter/createLegoAdapter';
 import { Survey } from '../actions/ActionTypes';
-import { selectEvents } from './events';
+import { selectEventById } from './events';
 import type { EntityId } from '@reduxjs/toolkit';
 import type { RootState } from 'app/store/createRootReducer';
 import type { EventForSurvey, EventType } from 'app/store/models/Event';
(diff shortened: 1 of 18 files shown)


app/routes/surveys/components/Submissions/Results.tsx
```diff
@@ -75,21 +78,24 @@ const Results = ({
   numberOfSubmissions,
 }: Props) => {
   const dispatch = useAppDispatch();
+  const event = useAppSelector((state) =>
+    selectEventById(state, { eventId: survey.event }),
+  ) as EventForSurvey;
```
ivarnakken, line 83:
```suggestion
  const event = useAppSelector((state) =>
    selectEventById<EventForSurvey>(state, survey.event),
  );
```

Do you still expect an object as param?

app/reducers/surveys.ts
```diff
@@ -18,126 +20,113 @@ import type {
...
     [surveyId],
   );
-  return useAppSelector((state: RootState) =>
+  const survey = useAppSelector((state: RootState) =>
     selectSurveyById(state, surveyId),
-  ) as SelectedSurvey | SelectedPublicResultsSurvey | undefined;
+  ) as UnknownSurvey | undefined;
+  const event = useAppSelector((state: RootState) =>
+    selectEventById(state, { eventId: survey?.event }),
+  ) as EventForSurvey | undefined;
```
ivarnakken, line 126: Same here

app/reducers/surveys.ts
```diff
@@ -18,126 +20,113 @@ import type {
...
     `useFetchedSurvey-${prepareId}`,
     () =>
-      surveyId
-        ? token
-          ? dispatch(fetchWithToken(surveyId, token))
-          : dispatch(fetchSurvey(surveyId))
-        : Promise.resolve(),
+      token
+        ? dispatch(fetchWithToken(surveyId, token))
+        : dispatch(fetchSurvey(surveyId))[surveyId],
```
ivarnakken, line 118:
```suggestion
        : dispatch(fetchSurvey(surveyId)),
```
Am I missing something here?

app/reducers/surveys.ts
```diff
@@ -18,126 +20,113 @@ import type {
...
+): TransformedSurveyTemplate | undefined => {
   const dispatch = useAppDispatch();
   usePreparedEffect(
     `useFetchedTemplate-${prepareId}`,
     () => templateType && dispatch(fetchTemplate(templateType)),
     [templateType],
   );
   return useAppSelector((state: RootState) =>
-    templateType ? selectSurveyTemplate(state, { templateType }) : undefined,
+    templateType ? selectSurveyTemplateByType(state, templateType) : undefined,
```
ivarnakken, line 90: This seems to be handled by `selectSurveyTemplateByType`

ivarnakken (approved):

app/routes/surveys/components/Submissions/Results.tsx
ivarnakken: nvm this is for the event - that's not refactored yet

app/reducers/surveys.ts
eikhr: hmm, that's weird

app/reducers/surveys.ts
eikhr: Ah, yes:)

PR #4599, opened 2024-03-23
Fix fetching of company contacts in joblisting editor

ivarnakken:
# Description

Two issues fixed here. When creating a new joblisting, the company contacts would not be fetched since the company id would be in `company.value` instead of `company.id`. Secondly, once you've managed to fetch, you would also get an error since it was trying to map over the payload instead of the results.

Also, 

# Result

No visual changes.

# Testing

Editing and creating joblistings now work without any errors.

### DIFF
--- a/app/routes/joblistings/components/JoblistingEditor.tsx
+++ b/app/routes/joblistings/components/JoblistingEditor.tsx
@@ -43,6 +43,9 @@ import {
 import { places, jobTypes, yearValues } from '../constants';
 import styles from './JoblistingEditor.css';
 import type { EntityId } from '@reduxjs/toolkit';
+import type { searchMapping } from 'app/reducers/search';
+import type { ListCompany } from 'app/store/models/Company';
+import type { DetailedJoblisting } from 'app/store/models/Joblisting';
 
 type SelectInputObject = {
   label: string;
@@ -82,7 +85,7 @@ const JoblistingEditor = () => {
   const isNew = joblistingId === undefined;
 
   const joblisting = useAppSelector((state) =>
-    selectJoblistingById(state, joblistingId),
+    selectJoblistingById<DetailedJoblisting>(state, joblistingId),
   );
 
   const navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


app/routes/joblistings/components/JoblistingEditor.tsx
```diff
@@ -223,16 +224,7 @@ const JoblistingEditor = () => {
...
               onChange={(event) => {
-                fetchContacts(event).then(() => {
-                  form.change('joblistingEditor', {
-                    label: 'Ingen',
-                    value: null,
-                  });
-                  form.change('responsible', {
-                    label: 'Ingen',
-                    value: null,
-                  });
```
ivarnakken, line 234: No idea why this was here

eikhr: I think we need the 'responsible' one. If you change the company, there will be a different set of responsible persons (Kontaktpersoner) to choose from. So if we have selected a responsible person from company A and then change to company B we need to clear the clear the field to not end up with an invalid responsible person.

ivarnakken: ah yes that actually makes sense haha

app/routes/joblistings/components/JoblistingEditor.tsx
```diff
@@ -124,13 +125,13 @@ const JoblistingEditor = () => {
         });
 
   const fetchContacts = useCallback(
-    (company: SelectInputObject) => {
+    (company: SelectInputObject | ListCompany) => {
       return dispatch(
         fetchCompanyContacts({
-          companyId: company.id,
+          companyId: 'value' in company ? company.value : company.id,
```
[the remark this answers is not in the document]
ivarnakken: Nice catch! The issue was `company` becoming an integer when submitting the form. Fixed this with a check

app/routes/joblistings/components/JoblistingEditor.tsx
```diff
@@ -124,13 +125,13 @@ const JoblistingEditor = () => {
...
         }),
-      ).then((action) => {
-        const responsibleOptions = action.payload.map((contact) => ({
+      ).then((res) => {
+        const responsibleOptions = res.payload.results.map((contact) => ({
```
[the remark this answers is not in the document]
ivarnakken: Fixed 😄

falbru (approved):
Awesome work!

ivarnakken:
Thanks for actually testing and checking the code! ❤️

PR #4604, opened 2024-03-27
Improve lego adaptor selectors

eikhr:
# Description

The entities contained in the redux-store are of unknown type, meaning a meeting might f.ex. either be from a meeting list or the detailed model used on the meeting page. When selecting them we usually know what type we are expecting, meaning we will need to typecast it (or do some insane type checking to make typescript infer it). With this new API this typecast is easier, and less error-prone as you can't forget to include the possibility of `undefined`!

```typescript
const detailedMeeting = useAppSelector((state) => selectMeetingById<DetailedMeeting>(state, meetingId));
const listMeetings = useAppSelector(selectAllMeetings<ListMeeting>);
```
These will get the following types:
```typescript
detailedMeeting: DetailedMeeting | undefined
listMeetings: ListMeeting[]
```

`selectById()`  now also allows `undefined` as Id, and in that case simply returns `undefined`.
Some `selectByIds`-functions now also allows undefined, which defaults to an empty list of ids.

# Result

Easier type overrides when using selectors, and less `undefined` checking in ids.

# Testing

Have tested a bit, don't see why this should break anything

### DIFF
--- a/app/store/models/entities.ts
+++ b/app/store/models/entities.ts
@@ -70,9 +70,6 @@ export enum EntityType {
   Tags = 'tags',
   Thread = 'threads',
   Users = 'users',
-  FollowersCompany = 'followersCompany',
-  FollowersUser = 'followersUser',
-  FollowersEvent = 'followersEvent',
 }
 
 // Most fetch success redux actions are normalized such that payload.entities is a subset of this interface.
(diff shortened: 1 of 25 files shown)


ivarnakken:
Without having looked at the code - what happens if I "accidentally" type it as a list (when I expect a list)?

```ts
const listMeetings = useAppSelector(selectAllMeetings<ListMeeting[]>);
```

Does it catch this? Or do I end up with `ListMeeting[][]`?

eikhr:
It should catch it. It only allows you to override the type as something that extends the relevant "UnknownSomething" type, which should limit it to only the different types of that entity

app/store/models/entities.ts
ivarnakken, line 1: imo they might as well be removed

ivarnakken (approved):
I like

app/store/models/entities.ts
eikhr: Yea, I guess

PR #4612, opened 2024-03-28
Add reactions to comments

falbru:
# Description

Adds reactions to comments.

I faced a small issue and came up with a "hacky" solution: Models that can be reacted on, like articles and meetings, have a contentTarget property that determines what model the reaction should be added to. This property points to the model itself, so that if you react to an article, the reaction is added to that article. I imagined that I could to the same thing for comments: adding a contentTarget property that points back to the comment itself. However, comments already have a contentTarget property that is used to link to the parent model, like an article, causing reactions to be added to the article instead of the comment (never really tested that this actually happened, but it's my guess). My workaround was to add a contentSelf property and pass it to LegoReactions as the contentTarget property. Do you guys think this implementation is alright?

# Result

Screencast from 2024-03-28 13-23-09.webm

# Testing

Adding and deletion works.

I haven't tested for comments that were made before these changes yet.

---

Resolves ABA-678

### DIFF
--- a/app/components/Comments/Comment.tsx
+++ b/app/components/Comments/Comment.tsx
@@ -15,6 +15,7 @@ import styles from './Comment.css';
 import type CommentType from 'app/store/models/Comment';
 import type { ContentAuthors } from 'app/store/models/Comment';
 import type { ContentTarget } from 'app/store/utils/contentTarget';
+import LegoReactions from '../LegoReactions';
 
 type Props = {
   comment: CommentType;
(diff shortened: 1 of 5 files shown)


eikhr (approved):
Code looks good, I think this solution is the best option. 
Not sure what I think about the name `contentSelf`. I feel like it should still contain "content target" as that is sort of the name of this concept in the code base. It should be easy to understand out of context what function the property has.

falbru:
@eikhr Thanks for your input 😁. Would you rather prefer changing the name of the existing property `contentTarget` for Comment to something else? What about `parentContent`?

eikhr:
Idk, it doesn't matter that much. I just meant I think it should be f.ex. "selfContentTarget" or "parentContentTarget" so that it still contains the entire phrase "content target". Because I don't feel like it's obvious what "content" in "selfContent" is referring to.

eikhr (approved):
Looks great! But you need to rebase and lint, and there seems to be a problem with the comment tests.

eikhr (approved):
Looks good!

PR #4627, opened 2024-04-09
Add localstorage saving to meeting report

jonasdeluna:
# Description

This adds localstorage saving to meeting reports.
Onchange, if the value is longer than the previous value it will override the localstorage value (or if the localstorage key does not exist).

This was re-opened as we decided at the meeting to go for this approach after consulting the router wiki.

# Result

image

image

- [ x] Changes look good on both light and dark theme.

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            ...
        </td>
        <td>
            ...
        </td>
        <td>
            ...
        </td>
    </tr>
</table>

# Testing

---

Resolves ABA-740

### DIFF
--- a/app/routes/meetings/components/MeetingEditor.tsx
+++ b/app/routes/meetings/components/MeetingEditor.tsx
@@ -124,9 +124,11 @@ const MeetingEditor = () => {
       : undefined,
   );
 
+  //For some reason calling the debounce function directly was wack.
+  //Also, we want to check if the new length is greater than the old one or if it doesn't exist already.
   const updateReportValue = useCallback(
     (newReportValue) => {
-      debounce(() => {
+      const debounceFunc = debounce(() => {
         const storedReport = sessionStorage.getItem(
           `meeting-${meetingId}-report`,
         );
@@ -134,11 +136,11 @@ const MeetingEditor = () => {
           setReportValue(newReportValue);
         }
       }, 4500);
+      debounceFunc();
     },
     [meetingId],
   );
 
-  //We only want to override the sessionStorage meeting if we add something, or if it does not exist already
   useEffect(() => {
     if (meeting && meetingId) {
       const storedReport = sessionStorage.getItem(
@@ -502,7 +504,7 @@ const MeetingEditor = () => {
                   sessionStorage.getItem(`meeting-${meeting.id}-report`) !==
                     null && (
                     <ConfirmModal
-                      title="Hente inn referat fra lokal lagring?"
+                      title="Hente inn referat fra sessionStorage?"
                       message={`Fant en lagret backup for dette møtet. Dette vil overskrive det nåværende referatet lokalt.`}
                       onConfirm={() => {
                         loadReportFromLocalStorage(form);
--- a/app/routes/meetings/components/MeetingEditor.tsx
+++ b/app/routes/meetings/components/MeetingEditor.tsx
@@ -7,7 +7,7 @@ import {
 import { usePreparedEffect } from '@webkom/react-prepare';
 import { debounce, unionBy } from 'lodash';
 import moment from 'moment-timezone';
-import { useCallback, useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


ShaileshS1702 (approved):
LGTM!

app/routes/meetings/components/MeetingEditor.tsx
```diff
@@ -451,6 +498,26 @@ const MeetingEditor = () => {
                     )}
                   </ConfirmModal>
                 )}
+                {isEditPage &&
+                  canDelete &&
```
norbye, line 502: Why do you need this line?

app/routes/meetings/components/MeetingEditor.tsx
```diff
@@ -105,6 +105,8 @@ type MeetingEditorParams = {
 };
 const MeetingEditor = () => {
   const { meetingId } = useParams<MeetingEditorParams>();
+  const [formDirtyFlag, setFormDirtyFlag] = useState(0);
```
norbye, line 108: Do you have any reasoning behind this naming? Didn't find it too intuitive

Also the method of refreshing the form is still kinda hacky - so might be nice with a comment about why this is needed.

```suggestion
  const [reportKey, setReportKey] = useState(0); // Needed to refresh the form component after loading the report content from localStorage
```

Now I'm just throwing out ideas, but I'm assuming this prompt is only supposed to be shown at page load. Is there any neat way that you could prevent loading the report component until the prompt has been adressed? If you could do that and set the correct value before rendering the component, I imagine you wouldn't need the hacky key-changing

app/routes/meetings/components/MeetingEditor.tsx
```diff
@@ -122,6 +124,34 @@ const MeetingEditor = () => {
...
 
+  //For some reason calling the debounce function directly was wack.
+  //Also, we want to check if the new length is greater than the old one or if it doesn't exist already.
+  const updateReportValue = useCallback(
+    (newReportValue) => {
+      const debounceFunc = debounce(() => {
+        const storedReport = sessionStorage.getItem(
+          `meeting-${meetingId}-report`,
+        );
+        if (!storedReport || storedReport.length < newReportValue.length) {
```
norbye, line 135: Thoughts on doing this based on a timestamp rather than the length of the report? Even though this probably works in the majority of the cases, a longer report doesn't always mean a newer one

itsisak:
For me it does not appear to work properly..

When typing something into the report and then leaving the page without saving, I can see that it is saved to sessionStorage, however when I re-enter the meeting or meeting/edit I am not getting the prompt to get the data from sessionStorage. 

I also found that if something is added to sessionStorage I don't seem to be able to replace this.

app/routes/meetings/components/MeetingEditor.tsx
```diff
@@ -122,6 +124,34 @@ const MeetingEditor = () => {
...
+    [meetingId],
+  );
+
+  useEffect(() => {
+    if (meeting && meetingId) {
+      const storedReport = sessionStorage.getItem(
+        `meeting-${meeting?.id}-report`,
+      );
+      if (!storedReport || storedReport.length < reportValue.length) {
+        sessionStorage.setItem(`meeting-${meeting?.id}-report`, reportValue);
```
PeterJFB, line 150: Minor/unlikely security concern. A user who has signed out after editing a report will still have this data in their session storage. A quick fix here is clearing it whenever a user signs out, i.e. add a `sessionStorage.clear()` [here](https://github.com/webkom/lego-webapp/blob/master/app/actions/UserActions.ts#L88)

ivarnakken: You can just clear it when this component unmounts

ivarnakken: Why add it to an action that has nothing to do with this?

PR #4630, opened 2024-04-11
Do not debounce local theme change

ivarnakken:
# Description

I hate to have to wait 200ms every time I want to update the theme. Only debouncing the requests to the server should be enough.

This also prevents unnecessary requests and responses, because now if the user spams the switch button, the user will only receive one toast feedback instead of 1000. I increased the debounce time to 300ms since the wait is fine now - could probably do more but meh

I'd say this is a bug fix

# Result

<details>
<summary>Before</summary>

https://github.com/webkom/lego-webapp/assets/69514187/fd2a5bcc-daea-4fbd-85ca-7ce2befd6171

</details>

<details>
<summary>After</summary>

https://github.com/webkom/lego-webapp/assets/69514187/c1e016f2-a35d-4b30-bac6-b29c98e039bf

</details>

# Testing

See videos above.

### DIFF
--- a/app/components/Header/ToggleTheme.tsx
+++ b/app/components/Header/ToggleTheme.tsx
@@ -23,15 +23,23 @@ const ToggleTheme = ({ className, children, isButton = true }: Props) => {
   const username = useCurrentUser()?.username;
   const icon = useIcon();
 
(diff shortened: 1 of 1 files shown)


eikhr (approved):

falbru (approved):
Works for me!

PR #4635, opened 2024-04-12
Remove followers slices from redux

eikhr:
# Description

Removed the following slices as they were not really used for anything. I also fixed the buggy behaiviour of following(starred) state of events when registering and unregistering. 

# Result

Following state is now updated correctly when registering/unregistering.

# Testing

I have looked through the code base for any usages of the removed stuff, and tested a lot of combinations of following, unfollowing, registering, unregistering and reloading.

---

ABA-688


ivarnakken:
> I also fixed the buggy behaiviour of following(starred) state of events when registering and unregistering.

Are you referring to [this](https://github.com/webkom/lego/pull/3254)? hehe

eikhr:
> > I also fixed the buggy behaiviour of following(starred) state of events when registering and unregistering.
> 
> Are you referring to [this](https://github.com/webkom/lego/pull/3254)? hehe

Yes! It didn't get updated correctly in frontend:)

app/actions/EventActions.ts
```diff
@@ -375,13 +370,14 @@ export function unfollow(followId: EntityId, eventId: EntityId) {
   });
 }
 
-export function isUserFollowing(eventId: EntityId) {
-  return callAPI<boolean>({
```
ivarnakken, line 379: I feel like the function name here is very misleading if it does not return a boolean

app/actions/EventActions.ts
```diff
@@ -375,13 +370,14 @@ export function unfollow(followId: EntityId, eventId: EntityId) {
...
+export function isUserFollowing(eventId: EntityId, currentUserId: EntityId) {
+  return callAPI({
     types: Event.IS_USER_FOLLOWING,
     endpoint: `/followers-event/?target=${eventId}`,
-    schema: [followersEventSchema],
     method: 'GET',
     meta: {
+      eventId,
+      currentUserId,
       errorMessage: 'Henting av interesse feilet',
```
ivarnakken, line 381: Also, this feedback seems misleading

ivarnakken (approved):
Nice to get rid of this!

app/actions/EventActions.ts
ivarnakken: Or when I think about it, I suppose following an event is showing "interesse". I was imagining a company interest, and assumed it was koked

norbye:
```suggestion
      errorMessage: 'Henting av stjernemarkering feilet',
```
```suggestion
      errorMessage: 'Henting av følgerstatus feilet',
```

app/actions/EventActions.ts
```diff
@@ -375,13 +370,14 @@ export function unfollow(followId: EntityId, eventId: EntityId) {
   });
 }
 
-export function isUserFollowing(eventId: EntityId) {
-  return callAPI<boolean>({
+export function isUserFollowing(eventId: EntityId, currentUserId: EntityId) {
```
norbye, line 373:
```suggestion
export function fetchUserFollowing(eventId: EntityId, currentUserId: EntityId) {
```

norbye (approved):
veri nice

app/actions/EventActions.ts
eikhr: It 100% is. I kind of feel like whoever wrote the funciton didn't even know what the endpoint returned. I didn't change it because I was considering adding an endpoint that did return a boolean, but I guess I can change it as I never got around to doing that.

app/actions/EventActions.ts
```diff
@@ -375,14 +370,15 @@ export function unfollow(followId: EntityId, eventId: EntityId) {
   });
 }
 
-export function isUserFollowing(eventId: EntityId) {
-  return callAPI<boolean>({
-    types: Event.IS_USER_FOLLOWING,
+export function fetchFollowers(eventId: EntityId, currentUserId: EntityId) {
+  return callAPI({
+    types: Event.FETCH_FOLLOWERS,
```
norbye, line 375: Quick comment before the auto-merging hits in for good - I haven't checked what the endpoint returns, but I'm assuming it is just the one object as both the event and user is specified. If that is correct, could you rename this in a way that reflects it's singular nature? If I'm wrong the plural you're currently using is just fine and I have no further comments(:

eikhr: I think it would be good to add an endpoint for checking just one user, though. It is a bit unnecessary to fetch all of them

norbye: Lol aight that's weird.. then I feel like it makes no sense to have to supply the current user (which is weird in either way), but I guess the backend endpoint requires it for some reason 🤷

eikhr: No, actually we supply the current user for frontend reasons. In order to check if the current user is following the event, we need to know who they are. We don't have access to that in the `events` reducer, so we need to attach it to the `fetchFollowers`-action

eikhr: This code is kind of a mess, but I think fixing it should be done in a separate PR

norbye: > I think it would be good to add an endpoint for checking just one user, though. It is a bit unnecessary to fetch all of them

absolutely - as far as I know there is nowhere in the webapp where we allow other users to see who else is following an event - so it should be fine to alter the current implementation I think

eikhr: Yeah, no. Almost all the followers stuff is unused, I think a lot more functionality around it was planned at some point.

PR #4639, opened 2024-04-13
Require event content in editor

ivarnakken:
# Description

Currently only required on the server - which caused issues for some

# Result

image

# Testing

See image above - it now requires it!


itsisak (approved):
Got to love consistency between backend and frontend

PR #4644, opened 2024-04-17
Reduce click-area of mazemap checkbox label

norbye:
# Description

Reduce click-area and margin of mazemap checkbox label as reported in https://github.com/webkom/lego/issues/3567.

Also reduces the click area of all checkboxes that are wrapped in the Field component across the webapp.

# Result

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Reduced size and margin
        </td>
        <td>image</td>
        <td>image
        </td>
    </tr>
</table>

# Testing

---

Resolves https://github.com/webkom/lego/issues/3567

### DIFF
--- a/app/components/Form/CheckBox.css
+++ b/app/components/Form/CheckBox.css
@@ -115,3 +115,7 @@
     transform: scale(1);
   }
 }
+
+.fieldLabel {
+  width: fit-content;
+}
--- a/app/components/Form/CheckBox.tsx
+++ b/app/components/Form/CheckBox.tsx
@@ -64,7 +64,11 @@ const CheckBox = ({
 const RawField = createField(CheckBox, { inlineLabel: true });
 
 const StyledField = ({ fieldClassName, ...props }: FormProps) => (
-  <RawField fieldClassName={fieldClassName} {...props} />
+  <RawField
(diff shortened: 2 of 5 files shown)


ivarnakken:
Is this a general problem with all checkboxes in forms or just this specific one?

norbye:
> Is this a general problem with all checkboxes in forms or just this specific one?

Hmm good point - I figured we could just leave it as is everywhere as I've never seen it to be a problem - but it does seem like the more intuitive way to make it fit-content for all labels across the webapp

ShaileshS1702 (approved):
LGTM!

eikhr (approved):
This happens with all checkboxes in the webapp, but I don't think it's as much of a problem in forms.

norbye:
I agree that it's not the biggest of problems across the webapp - but it isn't really ideal either, soo I changed the PR.

Updated it to fit-content for all labels that are wrapped in the `Field` component. Have tested a few occurrences and they look good (generally this shouldn't influence any styling, as all the label components that are affected is wrapped with a flex component that has not been modified)

eikhr (approved):

PR #4646, opened 2024-04-18
Ensure permissions are present to display list

norbye:
# Description

## Background
In the state, we store all groups in a long list together - more or less like so;
```
groups: {
    entities: [a lot of group objects],
    fetching: false,
    ...
}
```

This is used both for fetching the list of groups that are in the sidepanel of abakus.no/admin/groups and the individual groups. Thus the type of entities is UnknownGroup[] (consisting of DetailedGroup, PublicGroup, +++)

Thus, when you're navigating /admin/groups, the state will contain all the groups, because they are fetched to be in the sidebar. Those groups however are PublicGroup instances (not very detailed, as they're not needed to be for the list.  
That means that they do not have e.g. `permissions` and `parentPermissions`.

When you navigate to another group, it will try to display the group, and as it already exists, it will do so. However, the detailed information has not yet been loaded - so while we're trying to display a DetailedGroup object we only have the PublicGroup so far.

## Current solution

Avoiding casting the group to something that it's not, but rather setting default values for the params that might not be loaded yet.

&nbsp;
**NB** This does make it look as if the group has no permissions when you're navigating from permission to permission, as the `fetching` state is currently not group-specific (and setting it to true for just fetching one group would mess up the places where we display the whole list, like the sidebar) so we have no way of knowing if it's fetching or just empty

# Result

# Testing

---

Resolves ABA-919

### DIFF
--- a/app/routes/admin/groups/components/AddGroupPermission.tsx
+++ b/app/routes/admin/groups/components/AddGroupPermission.tsx
@@ -6,10 +6,10 @@ import { SubmitButton } from 'app/components/Form/SubmitButton';
 import TextInput from 'app/components/Form/TextInput';
 import { useAppDispatch } from 'app/store/hooks';
 import { createValidator, matchesRegex, required } from 'app/utils/validation';
-import type { DetailedGroup } from 'app/store/models/Group';
+import type { DetailedGroup, UnknownGroup } from 'app/store/models/Group';
 
 type Props = {
-  group: DetailedGroup;
+  group: UnknownGroup;
 };
 
 const validate = createValidator({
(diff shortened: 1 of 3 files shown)


eikhr (approved):
This is definitely a problem with the way the redux system works currently. I think this is a good solution for now.

Regarding the fetching state, we could use pagination fetching state for the group list which would allow us to use the `groups.fetching` state for the single group. But i guess that's not really ideal as the groups are not paginated.

norbye:
> Regarding the fetching state, we could use pagination fetching state for the group list which would allow us to use the `groups.fetching` state for the single group. But i guess that's not really ideal as the groups are not paginated.

Yeah, that doesn't sound completely ideal either - as that solution could imply the need of a loading indicator for the pagination stuff in the sidebar - and would get messed up if we wanted to paginate groups at some point.  
Any other fetching state than a specific one for the individual group feels a little wrong imo.

PR #4650, opened 2024-04-20
Finish migration from createEntityReducer to legoAdapter

eikhr:
# Description

The events slice was the last slice still using `createEntityReducer`! That means we can finally get rid of `createEntityReducer`🎉

I have done minimal refactoring of the events code, as I don't want to spend too much time on it right now. I will do some more refactoring to make it more similar to the other slices and fix some type errors later.

# Result

Screenshot 2024-04-29 at 12 39 23

It removes a lot of technical dept from the codebase, and allows us to more easily improve the redux system further as it no longer needs backwards-compatibility with `createEntityReducer`.

# Testing

I have clicked around the event pages, checking that stuff still works

---

Resolves ABA-688


app/reducers/__tests__/events.spec.ts
```diff
@@ -1,241 +1,44 @@
-// Hack because we have circular dependencies
-// (companies -> events -> index -> frontpage -> events)
-// This import resolves dependencies properly..
 import 'app/reducers';
```
ivarnakken, line 1: Is the import still needed though?

ivarnakken (approved):
Awweesome! 😄 Looks good

app/reducers/__tests__/events.spec.ts
eikhr: Nope, I think I must have inteded to remove it along with the comment👍

PR #4655, opened 2024-04-24
Fix incorrect prop name in timeline acitivity

eikhr:
# Description

Timeline page is blank in prod. I think it must have broken during some rebase, it's just a wrong prop name.

# Result

It works now:)

# Testing

Typescript no longer complains, and the page renders in local dev.

---

Resolves https://abakus.sentry.io/issues/5098383302/events/a37e703bacb746e5b9d191ac36f62821/


itsisak (approved):
🙌

falbru (approved):

PR #4657, opened 2024-04-28
Improve typescript types of form fields

eikhr:
# Description

Improves prop types in a bunch of form fields, and fixes some types in the `createField` function which caused basically all fields in react-final-form forms to get type errors.

# Result

Reduced number of type errors from 660 to 607

# Testing

I believe the only non-type change I made was overwriting `value` to undefined in the CheckBox-component (it uses `checked` instead), and I have tested that it still works both in forms and stand-alone.


ivarnakken (approved):
This is beyond nice! I've wished for this for ages 😍😍😍

PR #4679, opened 2024-05-03
Use react-aria for improving Button and adding LinkButton

eikhr:
# Description

I have fallen in love with `react-aria`, it is an unstyled component library that basically just feels like better native HTML-components. It is very unopinionated, and basically just gives us tools to make our own components easier. It also has a huge focus on making accessibility easy, which I think would be good for us as the webapp is not suuper accessible.

In this PR I have used the `react-aria` Button and Link components to slightly improve the buttons and improve buttons that are links a bit more. For both of them there is small improvements to the hover and click effects (darkening the button), that makes them more accurately reflect the behaiviour of the button and be more consistant across platforms. For link buttons we also had some weird behaiviour when using keyboard navigation, that is now fixed.

Some slightly annoying changes, are that I had to rename all `onClick` to `onPress` as react-aria does some processing to clicks/taps to make it more consistent cross-platform.

I also had to change the way we add button margins in groups of buttons, as it didn't work well with vertically stacked buttons (which there are now that LinkButton uses the same css). Now horizontal groups of buttons need to add spacing manually, just like vertical groups do.

# Result

### Before:
https://github.com/webkom/lego-webapp/assets/8343002/27b04ba1-35ae-4f9b-a832-d793ad1c42ca
### After:
https://github.com/webkom/lego-webapp/assets/8343002/7a6e9f94-8652-45db-93e5-72687b273f08

### Before:
https://github.com/webkom/lego-webapp/assets/8343002/688a139d-f5ed-49bc-b0c1-8591d47a0414
### After:
https://github.com/webkom/lego-webapp/assets/8343002/6dc5fbee-dcf7-4ed7-8fd1-772195d7d323

## Button margins
Because of the changes to button margins, I went through most (all?) forms in the webapp to fix the margin, and also made the styling of the buttons a bit more consistent. They should now all be aligned left with `--spacing-md` margin.

# Testing

Buttons still work, and I have looked at a whole bunch of buttons to verify that I didn't break any styling.

I did also change `lego-bricks` to no longer require `react-router`. I have not tested that this works.

### DIFF
--- a/app/components/AnnouncementInLine/index.tsx
+++ b/app/components/AnnouncementInLine/index.tsx
@@ -1,5 +1,5 @@
-import { Button, Icon } from '@webkom/lego-bricks';
-import { Link } from 'react-router-dom';
+import { Icon, LinkButton } from '@webkom/lego-bricks';
+import { useNavigate } from 'react-router-dom';
 import { useAppSelector } from 'app/store/hooks';
 import type { AnnouncementCreateLocationState } from 'app/routes/announcements/components/AnnouncementsCreate';
 import type { UnknownEvent } from 'app/store/models/Event';
@@ -13,6 +13,7 @@ type Props = {
 };
 
 const AnnouncementInLine = ({ event, meeting, group }: Props) => {
+  const navigate = useNavigate();
   const actionGrant = useAppSelector((state) => state.allowed.announcements);
 
   if (!actionGrant) {
@@ -20,21 +21,20 @@ const AnnouncementInLine = ({ event, meeting, group }: Props) => {
   }
 
   return (
-    <Link
-      to="/announcements"
-      state={
-        {
-          event,
-          meeting,
-          group,
-        } as AnnouncementCreateLocationState
-      }
+    <LinkButton
+      onPress={() => {
+        navigate('/announcements', {
+          state: {
+            event,
+            meeting,
+            group,
+          } as AnnouncementCreateLocationState,
+        });
+      }}
     >
-      <Button>
-        <Icon name="send-outline" size={18} />
-        Send kunngjøring
-      </Button>
-    </Link>
+      <Icon name="send-outline" size={18} />
+      Send kunngjøring
+    </LinkButton>
   );
 };
 
--- a/app/components/AnnouncementInLine/index.tsx
+++ b/app/components/AnnouncementInLine/index.tsx
@@ -1,5 +1,4 @@
 import { Icon, LinkButton } from '@webkom/lego-bricks';
-import { useNavigate } from 'react-router-dom';
 import { useAppSelector } from 'app/store/hooks';
 import type { AnnouncementCreateLocationState } from 'app/routes/announcements/components/AnnouncementsCreate';
 import type { UnknownEvent } from 'app/store/models/Event';
@@ -13,7 +12,6 @@ type Props = {
 };
 
 const AnnouncementInLine = ({ event, meeting, group }: Props) => {
-  const navigate = useNavigate();
   const actionGrant = useAppSelector((state) => state.allowed.announcements);
 
   if (!actionGrant) {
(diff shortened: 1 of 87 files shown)


packages/lego-bricks/src/components/Button/Button.module.css
```diff
@@ -62,22 +71,14 @@
   > a {
```
eikhr, line 71: The `+ .button` rule above here causes issues with vertical lists of Buttons/LinkButtons, like the event admin panel

app/components/AnnouncementInLine/index.tsx
```diff
@@ -13,28 +13,28 @@ type Props = {
...
+    <LinkButton
+      onPress={() => {
+        navigate('/announcements', {
+          state: {
+            event,
+            meeting,
+            group,
+          } as AnnouncementCreateLocationState,
+        });
+      }}
```
itsisak, line 33: Just an idea, but could we move this logic to the `LinkButton`-definition so that you could use it like this in this case:

```jsx
<LinkButton href='/announcements' state={ { event, meeting, group} as AnnouncementCreateLocationState }>
```

I think this could be a bit more clean and `react-router-dom` is already a dependency for `lego-bricks`, although the current solution is also okay.

itsisak (commented):
Looks great! Nice to have better accessibility and have accurate hover effects. Also love to have a standardized `<LinkButton>` component 🚀

app/components/AnnouncementInLine/index.tsx
norbye: ➕

norbye:
> Some slightly annoying changes, are that I had to rename all onClick to onPress as react-aria does some processing to clicks/taps to make it more consistent cross-platform. I also had to rename disabled to isDisabled, and I have no idea why they renamed this...

Is there any reason why you didn't keep it as onClick and disabled in the props to the Button component in lego-bricks and then just do
`<AriaButton onPress={onClick} isDisabled={disabled} />`? Just thinking if we prefer the naming we previously used 🤷

app/components/AnnouncementInLine/index.tsx
eikhr: Yes, I didn't realize react-router was a lego-bricks dependency. It that case it might even be possible to get rid of the RouterContext thing

eikhr:
I think the rename to `onPress` is good because it changes the behaviour a bit. It might be confusing that it doesn't have a normal MouseEvent as argument if it's called `onClick`.

As for `isDisabled`, nope, no reason. I'll change it back so it's more consistent with the rest of the webapp:)

app/components/AnnouncementInLine/index.tsx
eikhr: I have changed my mind here. I don't think we should have `react-router-dom` as a dependency for `lego-bricks` as not all our projects use `react-router`, f.ex. kvittering uses `next` instead. I believe this means we cannot use `lego-bricks` for that project. 
But I do agree that this is a bit ugly, perhaps we could add this functionality in a separate `LinkButton`-component in `lego-webapp`

eikhr: Nevermind, I figured out a way to make it an optional peer dependency:)

app/components/Form/SubmitButton.tsx
```diff
@@ -1,32 +1,33 @@
...
 
 export const SubmitButton = ({
   children,
-  onClick,
+  onPress,
   className,
   allowPristine,
   danger = false,
-  disabled = false,
+  isDisabled = false,
```
ivarnakken, line 22: Any reason to keep this

ivarnakken (approved):
I love it! 😍  These bugs have bothered me for so long

app/components/Form/SubmitButton.tsx
eikhr: Nope, I just forgot

PR #4711, opened 2024-05-19
Cache theme in localStorage

norbye:
# Description

Allow the site to set the previously selected theme before it is overridden by the backend

# Result

Werks

> [!CAUTION]

# Testing

Run some reloads and hard reloads in dark mode and it seems to work as expected

---

Resolves ABA-788


itsisak (approved):
This has been very annoying.. Great fix!

PR #4714, opened 2024-05-22
Resolve readme container missing height on public frontpage

norbye:
# Description

As our beloved prod-testers have reported through the proper channels, the readmes do not display properly on the public frontpage.

Turns out that the ref used to get the proper size of the container does not update the component state (which makes sense), and the only reason this works locally and when authenticated is that the component is always re-rendered after the images are fully loaded (either locally from being slow or when logged in by toggling the component).

Also added some long-awaited padding and styling to the public frontpage to match the authenticated one.

Minor changes
* Update fp-2023->2024
* Minor styling of the public frontpage

# Result

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td><b>Public</b><br>
            - Reduce top padding<br>- Add side margins<br>- Make cybdat inline to allow linebreaks
        </td>
        <td>image
        </td>
        <td>image
        </td>
    </tr>
    <tr>
        <td><b>Authenticated:</b><br>Remove padding from title</td>
        <td>image
</td>
        <td>image
</td>
    </tr>
</table>

# Testing

---

Resolves snapchat-complaints

### DIFF
--- a/app/routes/frontpage/components/PublicFrontpage.css
+++ b/app/routes/frontpage/components/PublicFrontpage.css
@@ -2,25 +2,38 @@
 
 .container {
   display: grid;
-  grid-gap: 40px;
+  grid-gap: var(--spacing-xl);
   grid-template-columns: 1fr 1fr 1fr;
   grid-auto-rows: auto;
   grid-template-areas: 'welcome welcome login' 'events events hsp' 'article article readme' 'links links links';
+  padding: 0 var(--lego-default-padding);
 
   @media (--mobile-device) {
-    grid-gap: 20px;
     grid-template-columns: 1fr;
     grid-template-areas: 'welcome' 'login' 'events' 'hsp' 'article' 'readme' 'links';
+    padding: 0 var(--spacing-md);
   }
+
+  h1 {
+    margin-top: 0;
+  }
+
+  h3 {
+    margin-top: 0;
+    margin-left: 0;
+  }
+}
+
+.welcome i {
+  display: inline;
 }
 
 .login {
   height: fit-content;
 }
 
-.welcome,
 .links {
-  padding: 0 10px 10px;
+  padding: 0 var(--lego-default-padding) var(--lego-default-padding);
 }
 
 .hsp {
--- a/app/routes/frontpage/components/PublicFrontpage.tsx
+++ b/app/routes/frontpage/components/PublicFrontpage.tsx
@@ -1,5 +1,6 @@
 import { Button, Card, Container, Flex, LinkButton } from '@webkom/lego-bricks';
 import { usePreparedEffect } from '@webkom/react-prepare';
+import moment from 'moment-timezone';
 import { fetchData, fetchReadmes } from 'app/actions/FrontpageActions';
 import buddyWeekGraphic from 'app/assets/frontpage-graphic-buddyweek.png';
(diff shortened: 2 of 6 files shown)


falbru (commented):
Nice fix! Although I prefer the old layout where the cards under "Nyttige lenker" were tiled 3x2 instead of 2x3 like this:
image

norbye:
> Nice fix! Although I prefer the old layout where the cards under "Nyttige lenker" were tiled 3x2 instead of 2x3 like this: 

Fair - updated that in the last push - they still break down to 2 per row as soon as soon as the screen is a tiny bit too small, but the threshold is pretty much exactly the same as what's currently in prod

app/routes/frontpage/components/PublicFrontpage.css
```diff
@@ -2,25 +2,38 @@
...
 }
 
 .login {
   height: fit-content;
 }
 
-.welcome,
 .links {
-  padding: 0 10px 10px;
+  padding: 0 var(--lego-default-padding) var(--lego-default-padding);
```
ivarnakken, line 36: Do we use this?

app/routes/frontpage/components/PublicFrontpage.css
```diff
@@ -2,25 +2,38 @@
 
 .container {
   display: grid;
-  grid-gap: 40px;
+  grid-gap: var(--spacing-xl);
   grid-template-columns: 1fr 1fr 1fr;
   grid-auto-rows: auto;
   grid-template-areas: 'welcome welcome login' 'events events hsp' 'article article readme' 'links links links';
+  padding: 0 var(--lego-default-padding);
```
ivarnakken, line 9: .

ivarnakken (approved):

falbru (approved):
Looks good! Although I agree with Ivar's comments

app/routes/frontpage/components/PublicFrontpage.tsx
```diff
@@ -105,7 +105,7 @@ const HspInfo = () => (
 
 const usefulLinksConf = [
   {
-    title: 'Fadderperioden 2023',
+    title: 'Fadderperioden 2024',
```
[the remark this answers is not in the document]
norbye: Absolutely(:

app/routes/frontpage/components/PublicFrontpage.css
norbye: That is the variable that's already used on the authenticated frontpage - so I only added it to use it on the public one as well for consistency.

That is however the only other place the variable is used, so I don't have any objections about yeeting it and rather just use `--spacing-xl`, though I don't really mind keeping it either..

app/routes/frontpage/components/PublicFrontpage.tsx
```diff
@@ -1,5 +1,6 @@
 import { Button, Card, Container, Flex, LinkButton } from '@webkom/lego-bricks';
 import { usePreparedEffect } from '@webkom/react-prepare';
+import moment from 'moment';
```
ivarnakken, line 3: we use `moment-timezone`, no?

norbye: Hmm good catch!

PR #4739, opened 2024-06-08
Create accordion component in lego-bricks

norbye:
# Description

We use animated accordion a few places, so it seemed like it was time to make a shared component to take care of that. The component supplies very little styling in order to be flexible to more usecases, but takes care of the functionality needed to animate the transitions.

Migrate Poll and LatestReadme components to use the component

# Result

No significant visual changes

# Testing

---

Resolves ABA-944


falbru (approved):
Very nice! Love the snappy animation for the poll :)

app/components/Poll/index.tsx
```diff
@@ -84,19 +76,14 @@ const Poll = ({
...
 
   const canAnswer = !hasAnswered && isValid;
 
   return (
-    <Flex
-      alignItems="center"
-      className={cx(styles.poll, expanded ? styles.expanded : undefined)}
-      column
-    >
+    <Flex alignItems="center" className={cx(styles.poll)} column>
```
ivarnakken, line 86:
```suggestion
    <Flex alignItems="center" className={styles.poll} column>
```

packages/lego-bricks/src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  open: boolean;
+  rotateClassName: string;
+};
+
+type Props = {
+  disabled?: boolean;
+  defaultOpen?: boolean;
+  animated?: boolean;
+  wrapperClassName?: string;
+  triggerPosition?: 'over' | 'under';
```
ivarnakken, line 17: I feel like it's more common to say "top" and "bottom" when referring to positions, but this is fine

packages/lego-bricks/src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  Trigger: React.ComponentType<TriggerProps>;
```
ivarnakken, line 18: Why the capital T? Should we call this `triggerComponent` to match the Dropdown and Popover component?

ivarnakken (approved):
Awesome!! 😍 I like

The only thing I noticed in the preview was that the spacing between the text and arrow seemed a bit much, but I think it's fine!

Bestem0r (approved):
Looks good!

app/components/Poll/index.tsx
norbye: Good catch!

packages/lego-bricks/src/components/Accordion/index.tsx
norbye: Absolutely agree

packages/lego-bricks/src/components/Accordion/index.tsx
norbye: Whop should've checked how those components named it - updating now

norbye: Realized that I can change the variable name, but it has to be capitalized within the Accordion component as only capitalized variables can be used as JSX elements. (`<triggerComponent />` is not allowed)

norbye:
> The only thing I noticed in the preview was that the spacing between the text and arrow seemed a bit much, but I think it's fine!

Hmm yeah we could change that eventually if we want to - not going to touch it now as it only affects the storybook, and no places where it is used in the webapp

PR #4756, opened 2024-06-22
Prevent endless skeleton for interest groups with < 10 members

ivarnakken:
# Description

im back

See linear issue for context

# Result

It works, but I don't want to show members here. Check out http://localhost:3000/interest-groups/201 if you don't believe me 😝 

# Testing

Hard refresh, and no skeleton. Added a default value of false to prevent any other similar bugs


falbru (approved):
Nice! Works for me

itsisak (approved):
LGTM

PR #4768, opened 2024-06-28
Make icon on latest readme comp "act" as a button

ivarnakken:
# Description

Although you can click on the entire component to expand it, I think it's good for the icon to "act" (or, look) as a button since the user may assume (s)he should click on it.

Also, this lets the user "tab" to and use the component. So I suppose that's a "yay" for universal design

# Result

https://github.com/webkom/lego-webapp/assets/69514187/462ee059-ce03-439b-b6a5-72f6540db8f4

# Testing

See video above

### DIFF
--- a/app/routes/frontpage/components/LatestReadme.tsx
+++ b/app/routes/frontpage/components/LatestReadme.tsx
@@ -35,6 +35,7 @@ const LatestReadme = ({
               {readmeIfy('readme')}
               {!disabled && (
                 <Icon
+                  onClick={onClick}
                   name="chevron-forward-outline"
                   className={rotateClassName}
                 />


itsisak (approved):
I agree, this helps boost accessibility

norbye (approved):

eikhr (approved):

falbru:
I'm a bit confused. For me it looks like the video you've provided exactly how it behaves right now. I can also tab to the icon to expand it, so I'm not noticing any difference.

ivarnakken:
> I'm a bit confused. For me it looks like the video you've provided exactly how it behaves right now. I can also tab to the icon to expand it, so I'm not noticing any difference.

The issue is in `master`, which has not been deployed. I assume you only tested in prod?

falbru:
yup hehe. Then it makes sense :)

PR #4777, opened 2024-07-03
Introduce a `placeholder-color` variable

ivarnakken:
# Description

Cleaner than having to always remember to use the same color in all
these places.

The placeholder color on the search input in the hamburger menu uses a
different color, `color-gray-6`, because its background is too light for
`placeholder-color`.

# Result

No visual changes.

# Testing

Tested that the variable works.


eikhr (approved):
nice!

norbye (approved):

PR #4790, opened 2024-07-09
Update various spacings

ivarnakken:
# Description

There's some buggy behaviour where our webapp inherits the `line-height` from the `lego-editor` of `1.8`. I've tried to update most places that are affected by this if we ever update `lego-editor` with the latest `lego-bricks` variables ..

# Result

Most changes are barely noticeable.

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
			image
        </td>
    </tr>
</table>

# Testing

All changes were tested. That's why there's not that many ...

### DIFF
--- a/app/routes/forum/components/ForumEditor.tsx
+++ b/app/routes/forum/components/ForumEditor.tsx
@@ -84,7 +84,7 @@ const ForumEditor = () => {
         {({ handleSubmit }) => (
           <Form onSubmit={handleSubmit}>
             <Field
-              placeholder="Title"
+              placeholder="Tips til utveksling, hvordan velge veileder eller er indøk virkelig homo?"
               name="title"
               label="Tittel"
               component={TextInput.Field}
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -172,11 +172,9 @@ const EventItem = ({
           style={{ borderColor: colorForEventType(event.eventType) }}
           className={styles.eventItemCompact}
         >
-          <Flex width="100%">
-            <Link to={`/events/${event.slug}`}>
-              <h3 className={styles.eventItemTitle}>{event.title}</h3>
-            </Link>
-          </Flex>
+          <Link to={`/events/${event.slug}`}>
+            <h3 className={styles.eventItemTitle}>{event.title}</h3>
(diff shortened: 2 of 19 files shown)


norbye:
Except for the random unrelated fixes (that I love<3) this feels like a sort of hacky solution..

Don't have my computer handy so I can't check rn, but have you tried to investigate why these variables are inherited? I tried to browse the lego-editor repo and most of the styling seemed to be scoped (except for the styling that is imported from lego-webapp and thus might be outdated at some point - we might want to look into a way to import the variables without any of the css rules)

norbye:
> > this feels like a sort of hacky solution..
> 
> 
> 
> Are you referring to the places where I've updated the `line-height` to the "correct" `1.3`? I know we should fix the cascading bug from the editor package, but when we do, these places will break. There's already a lot of `line-height: 1.3` around in our webapp because of this, and it won't be difficult to just remove all of them when we've resolved the bug.

Ahh, then I might've judged that one a bit prematurely.. sounds good

norbye (approved):

app/routes/forum/components/ForumEditor.tsx
```diff
@@ -84,7 +84,7 @@ const ForumEditor = () => {
         {({ handleSubmit }) => (
           <Form onSubmit={handleSubmit}>
             <Field
-              placeholder="Title"
+              placeholder="Tips til utveksling, hvordan velge veileder eller er indøk virkelig homo?"
```
falbru, line 87: 🤔

falbru (approved):
Looks good

app/routes/forum/components/ForumEditor.tsx
Bestem0r: Yeah, maybe rephrase this?

Bestem0r (approved):
Looks good to me

itsisak (approved):

app/routes/forum/components/ForumEditor.tsx
ivarnakken: oh yea that wasn't _really_ supposed to get added 😆

PR #4796, opened 2024-07-10
Prevent text overflow on error boundary

ivarnakken:
# Description

This bug still confuses me .. I'm still not sure why the overflow actually happens

# Result

<table>
    <tr>
        <td width="13%">Description</td>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            iPhone 12 Pro (390x844)
        </td>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

Tested on larger viewports as well hehe

### DIFF
--- a/app/components/ErrorBoundary/ErrorBoundary.css
+++ b/app/components/ErrorBoundary/ErrorBoundary.css
@@ -3,6 +3,10 @@
 .container {
   width: 40%;
   margin: 0 auto;
+  line-height: 1.3;
+  word-wrap: break-word;
+  white-space: pre-wrap;
+  word-break: break-word;
 
   @media (--mobile-device) {
     width: 60%;
--- a/packages/lego-bricks/src/components/Card/Card.module.css
+++ b/packages/lego-bricks/src/components/Card/Card.module.css
@@ -4,10 +4,19 @@
   border-radius: var(--border-radius-lg);
 }
 
+.withIcon {
+  gap: var(--spacing-md);
+
+  @media (max-width: 992px) {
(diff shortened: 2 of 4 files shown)


app/components/ErrorBoundary/ErrorBoundary.css
```diff
@@ -3,6 +3,10 @@
 .container {
   width: 40%;
   margin: 0 auto;
+  line-height: 1.3;
+  word-wrap: break-word;
+  white-space: pre-wrap;
+  word-break: break-word;
```
ivarnakken, line 9: These did the trick. The rest is just bonus

packages/lego-bricks/src/components/Card/Card.module.css
```diff
@@ -4,10 +4,19 @@
   border-radius: var(--border-radius-lg);
 }
 
+.withIcon {
+  gap: var(--spacing-md);
+
+  @media (max-width: 992px) {
```
ivarnakken, line 10: We should definitely look more into why viewport variables in `lego-bricks` don't work

norbye (approved):

itsisak (approved):

eikhr (approved):
Looks good!
I think it happens because by default it doesn't break long words, so the long reference-id forces the card content to be super wide.

ivarnakken:
> I think it happens because by default it doesn't break long words, so the long reference-id forces the card content to be super wide.

ah yes of course! I didn't think of that!

PR #4797, opened 2024-07-10
Handle more edge cases when sorting companies

ivarnakken:
# Description

Among other things we now check for `null`, which prevents an error when sorting on student contacts.

# Testing

It's now possible to sort on all kinds of things, both ascending and descending.

### DIFF
--- a/app/routes/bdb/SortCompanies.ts
+++ b/app/routes/bdb/SortCompanies.ts
@@ -7,29 +7,45 @@ import type { TransformedAdminCompany } from 'app/reducers/companies';
 import type { ParsedQs } from 'qs';
 
 const sortByAttribute = (attribute) => (ascending) => (a, b) => {
-  if ((!a[attribute] && !b[attribute]) || a[attribute] === b[attribute]) {
+  const aAttribute = a?.[attribute];
+  const bAttribute = b?.[attribute];
+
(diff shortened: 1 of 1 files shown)


app/routes/bdb/SortCompanies.ts
```diff
@@ -7,29 +7,43 @@ import type { TransformedAdminCompany } from 'app/reducers/companies';
 import type { ParsedQs } from 'qs';
 
 const sortByAttribute = (attribute) => (ascending) => (a, b) => {
-  if ((!a[attribute] && !b[attribute]) || a[attribute] === b[attribute]) {
+  const aAttr = a?.[attribute];
+  const bAttr = b?.[attribute];
```
norbye, line 11: Not critical - but it doesn't hurt to avoid abbreviations where possible;
```suggestion
  const aAttribute = a?.[attribute];
  const bAttribute = b?.[attribute];
```

norbye (approved):
stonks

app/routes/bdb/SortCompanies.ts
ivarnakken: Agreed. Changed it now 😄

Bestem0r (approved):
Looks good! Clever use of `==`

PR #4803, opened 2024-07-12
Introduce Lucide icons

ivarnakken:
# Description

I have long wanted to replace our icons from ionicons. They are lacking a bunch of icons, and ones they have are in my opinion not very "consistently" designed. Lucide, on the other hand, is heavily maintained with new icons regularly, with much more consistent, "modern", "stricter" and cleaner designs.

Another (more technical) difference is that we now have to import the icons we wish to use. On one side this acts as a safety, preventing us from misspellings names or using deprecated icons. But, best of all, we won't have to fetch the icons seperatly, meaning the icons will always be visible (despite connectivity issues or issues with ionicons) 😍 Lastly, the fact that it's an npm package makes both setting up and updating the package easier compared to ionicons.

I believe it's better (read; easier) to support both Lucide and ionicons while gradually moving completely over to Lucide.

# Result

You should probably check it out yourself and get a feel for it.

- [x, espacially] Changes look good with slower Internet connections.

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

Build works fine. Tested with ssr.

### DIFF
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -1,4 +1,13 @@
 import { Flex, Icon, Image } from '@webkom/lego-bricks';
+import {
+  AlarmClock,
+  Calendar,
+  CalendarClock,
+  Clock,
+  CircleAlert,
+  CircleCheckBig,
+  Timer,
+} from 'lucide-react';
 import { Link } from 'react-router-dom';
 import Pill from 'app/components/Pill';
 import Tag from 'app/components/Tags/Tag';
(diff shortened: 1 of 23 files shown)


falbru (approved):
Love this 🔥 A lot more consistent, and the technical advantages you mentioned are very nice 👌

Although I must admit that the icons look a bit cartoonish imo, but I still think it's a big improvement!

ivarnakken:
> Although I must admit that the icons look a bit cartoonish imo, but I still think it's a big improvement!

That might be because of the stroke width. It's a bit thicker now, but we can change it if we want to

eikhr (approved):
I like it!

PR #4814, opened 2024-07-21
Refactor bdb page

ivarnakken:
# Description

Most notable is the use of the Table component, which has also been slightly updated to support search filtering with intersection of results instead of union.

# Result

While this does look and function "better", I believe the best part of this is the reduced technical debt

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

Filtering and etc. still works fine in other tables.


falbru:
Hmm something weird is happening on my machine. The colored background appears only as a straight line:
image

Using Firefox on Linux

ivarnakken:
> Hmm something weird is happening on my machine. The colored background appears only as a straight line:

ah yes typical Firefox ...

I've pretty much spent the entire day trying to fix this, and I've come up with this ..

<table>
    <tr>
        <td>Firefox</td>
        <td>Chrome</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>
 
On Firefox there's some extra "margin" around the buttons. There's no way of removing it, and there's no reason for it to be there. Good job Firefox

falbru (approved):
Awesome 🔥 It now works on the best browser 😛

PR #4828, opened 2024-07-28
Remove duplicate global css file in lego-bricks

ivarnakken:
# Description

Similar to what we did for the variables file: https://github.com/webkom/lego-webapp/pull/4558/files

Not sure why the file is called global in lego-bricks and global_s_ in lego-webapp - I'll let it be 🙃 

# Testing

Seems to work fine. Haven't tested lego-bricks in other projects


norbye (approved):

falbru (approved):

ivarnakken:
When I come to think of it - do we even need the global styling in lego-bricks? You wouldn't necessarily want that in other projects. Currently, it causes an issue where the global styling from lego-editor (which uses lego-bricks) overrides the global styling from lego-webapp. Thoughts? I think I will remove it after all

Bestem0r (approved):

ivarnakken:
Ended up removing the import after all

PR #4846, opened 2024-08-02
Use a cache key to prevent lock file issues in drone pipeline

ivarnakken:
# Description

Tired of the cache always being wrong and causing problems, especially lately for basically all PRs that have dependency updates

# Testing

I think it works

---

Resolves ABA-980

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -140,12 +140,6 @@ steps:
       event: [push]
       branch:
         exclude: [build]
-    depends_on:
-      - postgres
-      - minio
-      - thumbor
-      - redis
-      - api
     volumes:
       - name: yarn_cache
         path: /tmp/cache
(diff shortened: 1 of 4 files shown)


eikhr (approved):
Will the caches ever be deleted? We should probably pay some attention to how much storage is used, but it seems like an ok solution.

ivarnakken:
> Will the caches ever be deleted? We should probably pay some attention to how much storage is used, but it seems like an ok solution.

Good point! I don't think they'll be automagically deleted.

A possible solution would be to add some time-based expiration (e.g. every 30 days):

```yml
- name: cleanup_caches
  image: alpine:latest
  when:
    event: [push]
    branch:
      exclude: [build]
  volumes:
    - name: yarn_cache
      path: /tmp/cache
  commands:
    - find /tmp/cache -type d -mtime +30 -exec rm -rf {} +
```

falbru (approved):

PR #4848, opened 2024-08-03
Prevent picture edit modal from closing when tagging

ivarnakken:
# Description

The modal would close when tagging users, causing changes to not be saved.

# Testing

Tagging now works. Closing the modal like normal still works fine.

### DIFF
--- a/app/routes/photos/components/GalleryPictureEditModal.tsx
+++ b/app/routes/photos/components/GalleryPictureEditModal.tsx
@@ -2,11 +2,13 @@ import {
   Button,
   ButtonGroup,
   Flex,
-  Modal,
+  Icon,
   Image,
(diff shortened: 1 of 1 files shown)


app/routes/photos/components/GalleryPictureEditModal.tsx
```diff
@@ -101,6 +103,7 @@ const GalleryPictureEditModal = () => {
     <Modal
       onOpenChange={(open) => !open && navigate(`/photos/${gallery.id}`)}
       isOpen
+      isDismissable={false}
```
ivarnakken, line 106: This is the fix - the rest is just bonus

falbru (approved):
Looks good

PR #4895, opened 2024-08-16
Fix bug in table filtering breaking group member list

eikhr:
# Description

The problem was that the Table-component was filtering on all the things that were passed in through the `filters`-prop. This may sound correct, but I think we only want it to filter on the ones that are also defined as columns in the table. These are the ones with filters in the table UI.

I think this is the best way to do it, as we often pass all query parameters to the table filters even though not all of them should directly filter the table contents.

# Result

Group members now appear.

# Testing

The filters on group member list page now works as intended, and I also tested some filtering in BDB to ensure I didn't break that.

---

Resolves ABA-1008

### DIFF
--- a/app/components/Table/index.tsx
+++ b/app/components/Table/index.tsx
@@ -150,7 +150,7 @@ const Table = <T extends { id: EntityId }>({
         filterIndex,
       } = columns.find(
         (column) => column.filterIndex === key || column.dataIndex === key,
-      ) || {};
+      ) || { inlineFiltering: false };
 
       if (!inlineFiltering) return true;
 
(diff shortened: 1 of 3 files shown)


app/components/Table/index.tsx
```diff
@@ -150,7 +150,7 @@ const Table = <T extends { id: EntityId }>({
         filterIndex,
       } = columns.find(
         (column) => column.filterIndex === key || column.dataIndex === key,
-      ) || {};
```
eikhr, line 153: This is the fix, the rest is just types

app/components/Table/index.tsx
```diff
@@ -150,7 +150,7 @@ const Table = <T extends { id: EntityId }>({
...
+      ) || { inlineFiltering: false };
```
eikhr, line 153: This is the fix, the rest is just types

ivarnakken (approved):

PR #4904, opened 2024-08-20
Bump lego-bricks version

eikhr:
# Description

npm version was far behind

PROBLEMS:

It doesn't work without `react-router-dom`. Seems a bit complicated to fix...

### DIFF
--- a/packages/lego-bricks/package.json
+++ b/packages/lego-bricks/package.json
@@ -19,6 +19,7 @@
   ],
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "import": "./dist/lego-bricks.js",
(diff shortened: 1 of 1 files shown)


ivarnakken:
> It doesn't work without react-router-dom. Seems a bit complicated to fix...

Is it better [to just enforce it](https://github.com/webkom/lego-webapp/pull/4904/files#diff-cd894cd1a52f5bb3bbfa8f63d70415d792de00b5e5f2cea0083068bafa2aec21L54-L56)?

---

Also, are we sure it's a peer dependency and not just a regular dependency? Or both? This is confusing indeed

ivarnakken:
Superseded by https://github.com/webkom/lego-webapp/pull/4968

PR #4946, opened 2024-09-08
Improve flaky cypress test

eikhr:
Wait for errors to disappear before pressing submit

# Description

This is the one that was constantly failing after moving to server-room mando.

# Testing

Have run the test on both mando and hetzner agents.


ivarnakken (approved):
another week, another flaky e2e test 😋

PR #4964, opened 2024-09-16
Display correct registrationDeadline in event editor

norbye:
# Description

The little text in the editor was using a variable that was not set - and therefore always showed the current time.

Now it uses the same function as is used when it is displayed, and works better.

# Result

image

# Testing

---


ivarnakken (approved):
:shipit:

PR #4988, opened 2024-09-24
Support anonymized survey submissions

eikhr:
# Description

Currently we send the user object with every survey submission, which is not ideal because we tell users they are anonymous. This makes changes to frontend to support the changes to the API endpoint removing the user field from the submissions.

Backend PR: https://github.com/webkom/lego/pull/3529

# Result

No visual or functional changes.

# Testing

Needs more testing with non-admin users

---

Resolves ABA-273


norbye (approved):

ivarnakken (approved):
Nice!🤩

PR #4975, opened 2024-09-22
Fix button styling in bricks

Bestem0r:
# Description

The latest version of `lego-bricks` incorrectly styles buttons, adding a gray background and border. This is likely caused by the recent removal of the global css import from `lego-webapp`. This PR fixes the styling.

# Result

- [x] Changes look good on both light and dark theme.
- [x] Changes look good with different viewports (mobile, tablet, etc.).
- [x] Changes look good with slower Internet connections.

| Before | After |
| -- | -- |
| Screenshot from 2024-09-22 17-17-49 | Screenshot from 2024-09-22 17-17-58 |
| Screenshot from 2024-09-22 17-27-25 | Screenshot from 2024-09-22 17-27-33 |

# Testing

- [x] I have thoroughly tested my changes.

Navigate through the application and ensure all buttons, icon buttons and link buttons have the correct styling.

PR #4997, opened 2024-09-25
Bump lego-editor

Bestem0r:
# Description

Bumps `lego-editor` in order to remove `width: 140px` from `globals.css`. This also caused all buttons not using the `lego-bricks` `<Button>` to have default button styling. To handle this,  global button css has been added.

# Testing

Verify that all tables look and behave correctly.

### DIFF
--- a/app/components/CommentForm/CommentForm.css
+++ b/app/components/CommentForm/CommentForm.css
@@ -7,6 +7,5 @@
 }
 
 .submitButton:disabled {
-  opacity: 0;
-  transition: opacity var(--easing-medium);
+  display: none;
 }
--- a/app/styles/globals.css
+++ b/app/styles/globals.css
@@ -98,7 +98,6 @@ thead {
 th,
 td {
   padding: var(--spacing-sm);
-  width: 140px;
   display: table-cell;
 }
 
--- a/app/styles/globals.css
+++ b/app/styles/globals.css
@@ -14,6 +14,12 @@ html {
   font-size: var(--font-size-md);
 }
 
+button {
+  cursor: pointer;
+  border: none;
+  background-color: transparent;
+}
+
 body {
   line-height: 1.3;
   font-family: Inter, sans-serif;
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@sentry/node": "^7.101.1",
     "@stripe/react-stripe-js": "^1.14.2",
     "@stripe/stripe-js": "^1.46.0",
-    "@webkom/lego-editor": "^2.5.0",
+    "@webkom/lego-editor": "^2.5.1",
     "@webkom/react-meter-bar": "^2.0.0",
     "@webkom/react-prepare": "^1.0.1",
     "animate.css": "^4.1.1",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@sentry/node": "^7.101.1",
     "@stripe/react-stripe-js": "^1.14.2",
     "@stripe/stripe-js": "^1.46.0",
-    "@webkom/lego-editor": "^2.5.2",
+    "@webkom/lego-editor": "^2.6.1",
     "@webkom/react-meter-bar": "^2.0.0",
     "@webkom/react-prepare": "^1.0.1",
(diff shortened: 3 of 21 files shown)


ivarnakken (approved):

ivarnakken:
I'm not a fan of [f680726](https://github.com/webkom/lego-webapp/pull/4997/commits/f68072662e9585d0a926ed23e109f9d499738ce5). These buttons use the `button` tag for a reason, and they don't look good as a `Button`. In e.g. the dropdown it looks better to fill the background like for the other items. I vote to revert this commit.

Bestem0r:
> I'm not a fan of [f680726](https://github.com/webkom/lego-webapp/pull/4997/commits/f68072662e9585d0a926ed23e109f9d499738ce5). These buttons use the `button` tag for a reason, and they don't look good as a `Button`. In e.g. the dropdown it looks better to fill the background like for the other items. I vote to revert this commit.

Hmm, imo it looks slightly more consistent compared to the rest of the website, but I don't have a strong opinion on it. Refactoring all buttons is definitely out of scope for this PR anyways. Will revet it and add some global styling

ivarnakken:
> Hmm, imo it looks slightly more consistent compared to the rest of the website, but I don't have a strong opinion on it. Refactoring all buttons is definitely out of scope for this PR anyways. Will revet it and add some global styling

In some special cases it makes sense to just use a `button`, like on the unauthenticated login component. I also think it looks better to "fill" the entire dropdown item background. I agree that it's good to be consistent though, but that's what we have the `Button` component for.

Bestem0r:
Good now? @ivarnakken

app/components/CommentForm/CommentForm.css
```diff
@@ -7,6 +7,5 @@
 }
 
 .submitButton:disabled {
-  opacity: 0;
-  transition: opacity var(--easing-medium);
+  display: none;
```
ivarnakken, line 10: I like the fading though 😢

app/styles/globals.css
```diff
@@ -14,6 +14,12 @@ html {
   font-size: var(--font-size-md);
 }
 
+button {
+  cursor: pointer;
+  border: none;
+  background-color: transparent;
+}
```
ivarnakken, line 21: With this, do we still need https://github.com/webkom/lego-webapp/pull/4975/files? Also, very nitpick but can we move this to after the `body` selector so that there's some sort of hierarchy here

package.json
```diff
@@ -46,7 +46,7 @@
     "@sentry/node": "^7.101.1",
     "@stripe/react-stripe-js": "^1.14.2",
     "@stripe/stripe-js": "^1.46.0",
-    "@webkom/lego-editor": "^2.5.0",
+    "@webkom/lego-editor": "^2.5.2",
```
ivarnakken, line 49: We're at v2.6.1 now

app/components/CommentForm/CommentForm.css
eikhr: Yeah, but I'm not a huge fan of having the button still be there, only invisible.

app/styles/globals.css
eikhr: Is this global.css file included in lego-bricks? The button styling should work for other projects using lego-bricks as well.

ivarnakken: Should it though? Is it not just lego-webapp specific to use regular `button`s like we do? Not a fan of component libraries that add global styling

app/components/CommentForm/CommentForm.css
ivarnakken: hehe I guess so

app/styles/globals.css
Bestem0r: Not really, but that would break the styling of e.g. the `lego-editor` demo @ivarnakken

eikhr: Ok, I agree `lego-bricks` shouldn't add any global styling. But do the buttons in `lego-bricks` look good even without this? If they do, and this is just for the non-`lego-bricks` buttons, I think this is good:)

PR #4998, opened 2024-09-25
Add hover animation to eventList

magnusbrecke:
# Description

Added a hover effect over the event items in eventlist

# Result

Before: 

https://github.com/user-attachments/assets/dea17890-035e-469b-8278-4ba3e01b87b3

After: 

https://github.com/user-attachments/assets/68c4b9c8-66d1-4940-9c43-5894a83c1d1e

- [ x ] Changes look good on both light and dark theme.
- [ x ] Changes look good with different viewports (mobile, tablet, etc.).
- [ x ] Changes look good with slower Internet connections.

# Testing

- [ x ] I have thoroughly tested my changes.

---

Resolves ABA-1084


app/components/EventItem/styles.css
```diff
@@ -6,6 +6,13 @@
   display: flex;
   justify-content: space-between;
   line-height: 1.3;
+  transition: var(--easing-fast);
+}
+
+.eventItemHover {
```
falbru, line 12:
```suggestion
.eventItem:hover {
```

You can do this instead of using TypeScript!

app/components/EventItem/styles.css
```diff
@@ -48,6 +55,12 @@
   max-height: 80px;
   object-fit: contain;
   background: white;
+  margin: 5px 10px 5px 0;
+  transition: var(--easing-fast);
+}
+
+.companyLogoHover img {
```
falbru, line 62: Same here
```suggestion
.companyLogo:hover img {
```

falbru (commented):
Very nice work! Looks great 🔥

I forgot to mention it, but you don't actually need to use TypeScript in order to track if an element is hovered or not. You can use the CSS [selector](https://www.w3schools.com/cssref/css_selectors.php) `:hover` instead!

sudhanshuku01:
Can I work on this ?

app/components/EventItem/styles.css
eikhr: ➕

app/components/EventItem/styles.css
```diff
@@ -6,6 +6,13 @@
   display: flex;
   justify-content: space-between;
   line-height: 1.3;
+  transition: var(--easing-fast);
+}
+
+.eventItemHover {
+  background-color: var(--additive-background);
+  border-top-right-radius: 10px;
+  border-bottom-right-radius: 10px;
```
ivarnakken, line 15: Do we only want border on the right side? This feels kinda off from the rest of the webapp. Besides, please use border radius variables like `var(--border-radius-sm)` for this instead of `10px`😅

app/components/EventItem/styles.css
```diff
@@ -48,6 +55,12 @@
   max-height: 80px;
   object-fit: contain;
   background: white;
+  margin: 5px 10px 5px 0;
+  transition: var(--easing-fast);
+}
+
+.companyLogoHover img {
+  border-radius: 10px;
```
ivarnakken, line 63: See other comment. Please use `var(--border-radius-sm)`

app/components/EventItem/styles.css
```diff
@@ -48,6 +55,12 @@
   max-height: 80px;
   object-fit: contain;
   background: white;
+  margin: 5px 10px 5px 0;
```
ivarnakken, line 58: These should be spacing variables. It's important that all the spacings in the webapp are similar so that things don't feel off😁

ivarnakken (commented):
I have wanted this to be a link for so long!😍

app/components/EventItem/styles.css
magnusbrecke: I will fix border radius with the variable instead of 10px. But about if we should have the border radius or not, if you look at the job listing page, there are border radiuses on those elements. So I am not entirely sure what is most consistent, no border radius or border radius.

ivarnakken: The items on the joblisting page also has border radiuses. I think we should do the same here, but I was just contemplating wether we only wanted it on the right side. The colored border on the left side is kind off atm

magnusbrecke:
New Changes:

https://github.com/user-attachments/assets/163017f6-4be8-402d-becd-a800127bd9e3

falbru (approved):
Awesome!

eikhr (approved):
Looks very good!

norbye (approved):

app/components/EventItem/styles.css
```diff
@@ -48,6 +55,12 @@
   max-height: 80px;
   object-fit: contain;
   background: white;
+  margin: var(--spacing-sm) var(--spacing-md) var(--spacing-sm) 0;
+  transition: var(--easing-fast);
+}
+
+.eventItem:hover .companyLogo img {
+  border-radius: var(--border-radius-sm);
 }
```
ivarnakken, line 64: This feels kind of off - why are we adding border-radius to the images as well

app/components/EventItem/index.tsx
ivarnakken, line 1: You don't have to fix this now in this PR, but it would be nice if you did the same to the other event item types 😍

ivarnakken (commented):
Looking better!

app/components/EventItem/index.tsx
magnusbrecke: I agree I have already started working on it!

magnusbrecke:
New additions to the page where the hover effect has been implemented:

Profile page:

https://github.com/user-attachments/assets/2278e60c-803d-4582-b01f-3c967b206dfe

Authenticated front-page:

https://github.com/user-attachments/assets/6b59752a-65a0-4a6c-9167-945ce6df49c0

app/components/EventItem/styles.css
magnusbrecke: I tried both with and without, but personally felt with felt more natural. No other reason than that

itsisak (approved):
LGTM.

In the future it could be nice with a standardized component for a link with this effect. But this is good for now!

app/components/EventItem/index.tsx
```diff
@@ -155,17 +156,18 @@ const EventItem = ({
...
+          to={`/events/${event.slug}`}
           style={{ borderColor: colorForEventType(event.eventType) }}
           className={styles.eventItemCompact}
         >
-          <Link to={`/events/${event.slug}`}>
-            <h3 className={styles.eventItemTitle}>{event.title}</h3>
-          </Link>
+          {/* <Link to={`/events/${event.slug}`}> */}
+          <h3 className={styles.eventItemTitle}>{event.title}</h3>
+          {/* </Link> */}
```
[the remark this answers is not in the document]
magnusbrecke: yeah thx fixed now

app/components/EventItem/styles.css
```diff
@@ -1,11 +1,18 @@
...
   display: flex;
   justify-content: space-between;
   line-height: 1.3;
+  transition: var(--easing-fast);
+}
+
+.eventItem:hover {
+  background-color: var(--additive-background);
+  border-radius: var(--border-radius-sm);
 }
```
eikhr, line 16:
```suggestion

  &:hover {
    background-color: var(--additive-background);
    border-radius: var(--border-radius-sm);
  }
}
```
Use `&:hover` nested inside the class to add hover styling to the same classname:)

app/components/EventItem/styles.css
```diff
@@ -1,11 +1,18 @@
...
   margin-bottom: var(--spacing-md);
   display: flex;
   justify-content: space-between;
   line-height: 1.3;
+  transition: var(--easing-fast);
+}
+
+.eventItem:hover {
+  background-color: var(--additive-background);
+  border-radius: var(--border-radius-sm);
```
eikhr, line 15: Joblistings use `--border-radius-lg`. I would love for this to be consistant with that

eikhr (approved):
Huge improvement!

app/components/EventItem/styles.css
magnusbrecke: I also removed the radius from the left side
It is necessary when the bordrer radius is that big

falbru (approved):
Awesome

ivarnakken:
Remember to squash merge all the commits

PR #5001, opened 2024-09-25
Remove adminComment/note field from companies

juniwbjerde:
# Description
Removed the field `adminComment` from company. It was only used in the bdb table, and just took up unnecessary space and was never really used for anything... 

# Testing

---

Corresponding backend changes: https://github.com/webkom/lego/pull/3643

Resolves ABA-1052


ivarnakken:
See comment on backend PR. I thought we only wanted to remove it from the table? I know it's somewhat kinda used sometimes😆😅

juniwbjerde:
@ivarnakken I've answered you in the backed PR as well, but what does "somewhat kinda somethimes" mean? Falk got the impression that Bedkom never really use it as far as i understood

ivarnakken:
> @ivarnakken I've answered you in the backed PR as well, but what does "somewhat kinda somethimes" mean? Falk got the impression that Bedkom never really use it as far as i understood

I've seen it be used, but it's never been used very thoroughly. I thought we discussed that we wanted to improve the comments and make something similar to itDAGENE's system. I agree that it should be removed from the table though. Not sure if we have to remove it as a feature all together, because ideally we would want them to use it more. I guess our job is to make it more intuitive and useful. Perhaps Falk or Isak can show itDAGENE's admin page

ivarnakken:
Oh I read your comment on lego now! I totally forgot we had "normal" comments as well, so in that case I'm totally in for completely yeeting it😁sorry hehe

app/routes/bdb/components/BdbPage.tsx
```diff
@@ -222,15 +222,12 @@ const BdbPage = () => {
       },
     },
     {
-      title: 'Notat',
-      dataIndex: 'comment',
+      // Using an empty column for filtering
+      title: '',
+      dataIndex: 'Filter',
```
ivarnakken, line 227:
```suggestion
      dataIndex: 'active',
```

app/routes/bdb/components/BdbPage.tsx
```diff
@@ -222,15 +222,12 @@ const BdbPage = () => {
...
       centered: false,
```
ivarnakken, line 228: Is this needed now?

eikhr (approved):

Bestem0r (approved):
Nice!

PR #4974, opened 2024-09-22
Simplify event detail page by moving stuff into components

eikhr:
# Description

Moved stuff out of event detail page, into separate components. Fixes a bunch of typescript errors too

# Result

No visual changes

# Testing

- [x] I have thoroughly tested my changes.

Please describe what and how the changes have been tested, and provide instructions to reproduce if necessary.

PR #5002, opened 2024-09-26
Hide registration countdown button when already admin-registered

eikhr:
# Description

Fixes a bug and some small mistakes in the code from #4977 and #4974 

# Result

This no longer happens:
Screenshot 2024-09-26 at 10 12 28

# Testing

Tested that the bug is no longer present, and that it doesn't break regular registrations.


juniwbjerde (approved):
lgtm!

PR #5019, opened 2024-10-03
Fix attendance modal overflowing when searching on mobile

norbye:
# Description

The previously used `100vh` became bigger than the `var(--visual-viewport-height)` that was introduced to the overlay 5 months ago, which caused the page to overflow when visual-viewport-height shrinked when users opened their keyboards to search (on mobile).

# Result

This is now fixed, so the div does not move when you search for attendants.

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            ...
        </td>
        <td>
            ...
        </td>
        <td>
            ...
        </td>
    </tr>
</table>

# Testing

---


ivarnakken (approved):
Awesome!

PR #5041, opened 2024-10-09
Improve event items on frontpage

ivarnakken:
# Description

Better spacing, look and the entire thing is now clickable!

# Result

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
			<video src="https://github.com/user-attachments/assets/b7cd367a-9ab4-4623-bce6-fb8ea5e09522" />
        </td>
        <td>
			<video src="https://github.com/user-attachments/assets/5a7a8de6-39a4-4ec7-8814-aab25adb32f1" />
        </td>
    </tr>
</table>

# Testing

Checked different viewports and themes (and with skeleton components on slower connections).

### DIFF
--- a/app/routes/frontpage/components/CompactEvents.css
+++ b/app/routes/frontpage/components/CompactEvents.css
@@ -8,7 +8,7 @@
   text-align: left;
   gap: var(--spacing-xl);
 
-  @media (--small-viewport) {
+  @media (--medium-viewport) {
     flex-direction: column;
   }
 }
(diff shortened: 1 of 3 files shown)


app/routes/frontpage/components/CompactEvents.css
```diff
@@ -24,29 +24,20 @@
...
-  display: flex;
-  align-items: center;
-}
-
 .eventItem {
-  height: 26px;
-}
-
-.compactEvents a {
+  height: 28.8px; /* Used for the skeleton component */
```
falbru, line 28: What does this mean?

falbru (approved):

app/routes/frontpage/components/CompactEvents.css
ivarnakken: The height of an event item. This is never dynamic, and it's therefore "safe" to "hard code". The skeleton component needs to know the height of the component since it does not have any children content

itsisak (approved):
This has been so annoying, great fix!

PR #5051, opened 2024-10-12
Globally remove default styling from regular buttons

ivarnakken:
# Description

At least the stylings we don't want - I've e.g. kept outline and font-weights.

This looks much worse on mobile, where a blue color is added as well. It's not "that" noticeable on laptop

# Result

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
			image
        </td>
        <td>
           image
        </td>
    </tr>
    <tr>
        <td>
			image
        </td>
        <td>
           image
        </td>
    </tr>
</table>

# Testing

Looks good now, both with and without ssr.

---

Resolves ABA-1118

### DIFF
--- a/app/styles/globals.css
+++ b/app/styles/globals.css
@@ -18,6 +18,8 @@ button {
   cursor: pointer;
   border: none;
   background-color: transparent;
+  font-size: inherit;
+  color: inherit;
 }
 
 body {


Bestem0r (approved):
Great!

eikhr (approved):

PR #5043, opened 2024-10-10
Clean up code and improve consistency in profile page

eikhr:
# Description

Fix a bunch of type errors, split up large components, reuse components more (also improves styling consistency).

Small styling changes.

# Result

Type errors: 358 -> 333

- [x] Changes look good on both light and dark theme.
- [x] Changes look good with different viewports (mobile, tablet, etc.).
- [x] Changes look good with slower Internet connections.

<table>
    <tr>
        <td width="25%">Description</td>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Profile info on mobile
        </td>
        <td>

Screenshot 2024-10-10 at 18 30 36
Screenshot 2024-10-10 at 18 32 07
        </td>
        <td>

Screenshot 2024-10-10 at 18 29 53
Screenshot 2024-10-10 at 18 31 32
        </td>
    </tr>
    <tr>
        <td>
            E-mail lists and permissions
        </td>
        <td>

Screenshot 2024-10-10 at 18 33 45
        </td>
        <td>

Screenshot 2024-10-10 at 18 33 12
        </td>
    </tr>
</table>

# Testing

- [x] I have thoroughly tested my changes.

'--

Resolves ABA-1097

PR #5055, opened 2024-10-14
Fix crash on user profile page

eikhr:
# Description

I made an oopsie in #5043
It crashes when clicking a user profile on the event detail page because of invomplete data in the redux store.

# Result

no more crash:)

# Testing

---

Resolves [LEGO-WEBAPP-25Y](https://abakus.sentry.io/issues/5986651245/)


jonasdeluna (approved):
Great work!

PR #5068, opened 2024-10-16
Make semester statuses tags

Bestem0r:
# Description

Replaces the filled cell semester status with tags representing the semester statuses of the company.

# Result

| Before | After | 
| -- | -- |
| image | image |
| image | image |

# Testing

Open bdb and verify the table has tags.

Resolves [ABA-1120](https://linear.app/abakus-webkom/issue/ABA-1120/turn-semester-statuses-into-pills-on-bdb)

### DIFF
--- a/app/routes/bdb/utils.tsx
+++ b/app/routes/bdb/utils.tsx
@@ -29,7 +29,7 @@ export const NonEventContactStatusConfig: Record<
   [NonEventContactStatus.NOT_INTERESTED]: {
     displayName: 'Ikke interessert',
     color: 'var(--danger-color)',
-    textColor: '#ff0000',
+    textColor: '#fff',
   },
   [NonEventContactStatus.CONTACTED]: {
     displayName: 'Kontaktet',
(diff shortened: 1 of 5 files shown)


app/routes/bdb/utils.tsx
```diff
@@ -39,7 +39,7 @@ export const NonEventContactStatusConfig: Record<
   [NonEventContactStatus.NOT_CONTACTED]: {
     displayName: 'Ikke kontaktet',
     color: 'var(--additive-background)',
-    textColor: '#000',
+    textColor: 'var(--additive-text)',
```
ivarnakken, line 42: I don't think `--additive-text` is a thing?

app/routes/bdb/utils.tsx
```diff
@@ -29,7 +29,7 @@ export const NonEventContactStatusConfig: Record<
   [NonEventContactStatus.NOT_INTERESTED]: {
     displayName: 'Ikke interessert',
     color: 'var(--danger-color)',
-    textColor: '#ff0000',
+    textColor: '#fff',
```
ivarnakken, line 32: Could we maybe do the `--color-absolute-white`? I know this is the same, but I prefer to use our variable in case we want to change them

ivarnakken (approved):
Looks very good! 😍 

I noticed that the text colors in the dropdown does not match the ones in the tags - is this an easy fix?
image

falbru (approved):
Awesome! 🥰

PR #5091, opened 2024-10-23
Add lending system MVP

eikhr:
# Description

Adds a simple MVP for the lending system. Currently you can only set description, location and image on the lendable objects. It also supports setting view-permissions per object for specific groups.

# Result

Added to the navigation menu
Screenshot 2024-10-23 at 21 16 11

Empty state:
Screenshot 2024-10-23 at 21 16 26

List:
Screenshot 2024-10-23 at 21 19 02

Detail:
Screenshot 2024-10-23 at 21 19 24

Detail as admin without view-rights (lending-rights) to this object (admins can see everything):
Screenshot 2024-10-23 at 21 19 40

Create/edit form
Screenshot 2024-10-23 at 21 21 05

Detail page on mobile:
Screenshot 2024-10-23 at 21 22 22

List page on mobile:
Screenshot 2024-10-23 at 21 22 43

# Testing

It works:)

---

Resolves ABA-1046

### DIFF
--- a/app/actions/ActionTypes.ts
+++ b/app/actions/ActionTypes.ts
@@ -111,6 +111,13 @@ export const Joblistings = {
   DELETE: generateStatuses('Joblistings.DELETE') as AAT,
 };
 
+export const LendableObjects = {
+  FETCH: generateStatuses('LendableObject.FETCH_ALL') as AAT,
+  CREATE: generateStatuses('LendableObject.CREATE') as AAT,
+  EDIT: generateStatuses('LendableObject.EDIT') as AAT,
+  DELETE: generateStatuses('LendableObject.DELETE') as AAT,
+};
+
 /**
  *
  */
--- a/app/routes/lending/components/LendableObjectDetail/index.tsx
+++ b/app/routes/lending/components/LendableObjectDetail/index.tsx
@@ -1,0 +1,75 @@
+import { Page, LinkButton, PageCover, Card } from '@webkom/lego-bricks';
+import { usePreparedEffect } from '@webkom/react-prepare';
+import { Warehouse } from 'lucide-react';
+import { Helmet } from 'react-helmet-async';
+import { useParams } from 'react-router-dom';
+import { fetchLendableObjectById } from 'app/actions/LendableObjectActions';
+import {
+  ContentMain,
+  ContentSection,
+  ContentSidebar,
+} from 'app/components/Content';
+import DisplayContent from 'app/components/DisplayContent';
+import TextWithIcon from 'app/components/TextWithIcon';
+import { selectLendableObjectById } from 'app/reducers/lendableObjects';
+import { useAppDispatch, useAppSelector } from 'app/store/hooks';
+
+export const LendableObjectList = () => {
+  const { lendableObjectId } = useParams<'lendableObjectId'>();
+  const dispatch = useAppDispatch();
+  const fetching = useAppSelector((state) => state.lendableObjects.fetching);
+
+  usePreparedEffect(
+    'fetchLendableObjectById',
+    () => dispatch(fetchLendableObjectById(lendableObjectId!)),
+    [],
+  );
+
+  const lendableObject = useAppSelector((state) =>
+    selectLendableObjectById(state, lendableObjectId!),
+  );
+
+  const title = lendableObject ? `Utlån: ${lendableObject.title}` : undefined;
+
+  return (
+    <Page
+      title={title}
+      cover={<PageCover image={lendableObject?.image} skeleton={fetching} />}
+      actionButtons={
+        lendableObject &&
+        'actionGrant' in lendableObject &&
+        lendableObject.actionGrant.includes('edit') ? (
+          <LinkButton href={`/lending/${lendableObjectId}/edit`}>
+            Rediger
+          </LinkButton>
+        ) : undefined
+      }
+      back={{ href: '/lending' }}
+      skeleton={fetching}
+    >
+      <Helmet title={title || 'Utlån'} />
+      {lendableObject && (
+        <ContentSection>
+          <ContentMain>
+            {lendableObject && !lendableObject.canLend && (
+              <Card severity="warning">
+                Du har ikke tilgang til å låne dette objektet, men kan se det
+                pga. admin-rettigheter.
+              </Card>
+            )}
+            <DisplayContent content={lendableObject.description} />
+          </ContentMain>
+          <ContentSidebar>
+            <TextWithIcon
+              iconNode={<Warehouse />}
+              size={20}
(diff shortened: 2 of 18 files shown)


ShaileshS1702 (commented):
Very interesting!

app/actions/ActionTypes.ts
```diff
@@ -111,6 +111,16 @@ export const Joblistings = {
   DELETE: generateStatuses('Joblistings.DELETE') as AAT,
 };
 
+/**
+ *
+ */
```
ivarnakken, line 116: ?

app/routes/lending/components/LendableObjectDetail/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <Helmet title={title || 'Utlån'} />
+      {lendableObject && (
+        <ContentSection>
+          <ContentMain>
+            {lendableObject && !lendableObject.canLend ? (
+              <Card severity="warning">
+                Du har ikke tilgang til å låne dette objektet, men kan se det
+                pga. admin-rettigheter.
+              </Card>
+            ) : null}
```
ivarnakken, line 45:
```suggestion
            {lendableObject && !lendableObject.canLend && (
              <Card severity="warning">
                Du har ikke tilgang til å låne dette objektet, men kan se det
                pga. admin-rettigheter.
              </Card>
            )}
```
any reason to not just do this?

app/routes/lending/components/LendableObjectEdit/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Page
+      title={title}
+      back={{ href: `/lending/${lendableObjectId}` }}
+      skeleton={fetching}
+    >
+      <Helmet title={title} />
+      {fetching ? (
+        <LoadingIndicator loading />
+      ) : (
```
ivarnakken, line 25: Is this not handled by the `Page` component?

app/routes/lending/components/LendableObjectList/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <Link to={`/lending/${lendableObject.id}`}>
+      <Card isHoverable hideOverflow className={styles.lendableObjectCard}>
+        <Image
+          className={styles.lendableObjectImage}
+          src={lendableObject.image || abakus_icon}
+          alt={`${lendableObject.title}`}
+        />
+        <div className={styles.lendableObjectFooter}>
+          <h3>{lendableObject.title}</h3>
```
ivarnakken, line 38: I think `h3` is a bit large, and I'm not sure if I like that it's centered. It looks quite different from other list items on the webapp, e.g. events, articles, joblistings etc.

app/routes/lending/components/LendableObjectList/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+            <LinkButton href="/lending/new">Nytt utlånsobjekt</LinkButton>
+          )}
+        </>
+      }
+      sidebar={filterSidebar({
+        children: (
+          <FilterSection title="Søk">
+            <TextInput
+              prefix="search"
+              placeholder="Søk"
```
ivarnakken, line 84: I don't like when the placeholder is the same as the label, it's redundant

app/routes/lending/components/LendableObjectList/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+              onChange={(e) => setQueryValue('search')(e.target.value)}
+            />
+          </FilterSection>
+        ),
+      })}
+    >
+      <Helmet title={title} />
+
+      <LoadingIndicator loading={fetching}>
```
ivarnakken, line 93: Maybe we could do skeleton components instead of just loading indicators? 😎

app/routes/lending/useFetchedLendableObject.ts
```diff
@@ -0,0 +1,38 @@
...
+  const dispatch = useAppDispatch();
+  const [fetching, setFetching] = useState(true);
+
+  usePreparedEffect(
+    'fetchLendableObjectById',
+    () =>
+      dispatch(fetchLendableObjectById(lendableObjectId)).then(() =>
+        setFetching(false),
+      ),
+    [],
```
ivarnakken, line 22:
```suggestion
    [lendableObjectId],
```
Shouldn't the id be in the dependency array?

app/routes/lending/useFetchedLendableObject.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return { lendableObject, fetching } as
+    | {
+        lendableObject?: UnknownLendableObject;
+        fetching: true;
+      }
+    | {
+        lendableObject: DetailLendableObject;
+        fetching: false;
+      };
```
ivarnakken, line 37: Can't this be handled by the reducer?

ivarnakken (approved):
Looking really good!

I commented on everything I could spot, and I'll let you decide if you want to iterate on it later

app/actions/ActionTypes.ts
eikhr: hahaha, all the others had it so I figured I'd follow convention

app/routes/lending/components/LendableObjectEdit/index.tsx
eikhr: No, but I suppose it could be...

app/routes/lending/components/LendableObjectList/index.tsx
eikhr: Yes, but not yet. I think I want to change this view quite a bit

app/routes/lending/useFetchedLendableObject.ts
eikhr: Nice catch! We should get the `useEffect` dependency linter to lint these too

app/routes/lending/useFetchedLendableObject.ts
eikhr: Kind of, but the types will be wrong. The reducer simply returns:
```typescript
lendableObject?: UnknownLendableObject;
fetching: boolean;
```
which turns into:
```typescript
lendableObject?: DetailLendableObject;
fetching: boolean;
```
if you give the selector a type argument `<DetailLendableObject>`. Because of this it will assume that we have a `DetailLendableObject` whenever `lendableObject !== undefined`. However, we might still be fetching the detailed object and only have a `ListLendableObject` in the redux state. This could cause errors if we access properties that only the detailed serializer includes.

These types worked out very well when used on the page, so I am trying ot figure out a good way of generalizing this to all selectors. Hopefully you will see a PR on that soon:)

PR #5097, opened 2024-10-24
Import from lodash-es instead of lodash

eikhr:
Vite is not a huge fan of commonjs modules, es-modules also allow for better tree-shaking of the lodash package. (meaning we don't include code we don't use in lego-webapp)


ivarnakken (approved):
Nice

You might as well remove this, no? https://github.com/webkom/lego-webapp/blob/lodash-es/config/webpack.client.js#L130-L134

norbye (approved):

eikhr:
lol, this actually doesn't work with webpack. I guess it will have to be included in a larger vite PR...

PR #5098, opened 2024-10-24
Fix errors when loading members on pages when unauthorized

itsisak:
When not logged in and trying to view pages that normally contains memberships, a 401 error would occur.

The previous approach using `toString()` does not work with SSR as it minimizes / bundles the code so the actual function name becomes something like `d(e)` not `fetchAllMemberships`. 

Resolves ABA-1129


eikhr (approved):
Hahahah, great code🙃

norbye (approved):

ivarnakken (approved):
:shipit: lgtm

PR #5123, opened 2024-10-30
Remove author field from meeting detail

ivarnakken:
# Description

I'd like to make this resemble the event details more, but I have a trouble finding an icon for both "author" and "referent", so my plan is to iterate on this by first removing the author field and get people used to that :)

The first report changelog will say it was "opprettet" so that should be sufficient

image

# Testing

It is now hidden if there are any report changelogs

### DIFF
--- a/app/routes/meetings/components/MeetingDetail.tsx
+++ b/app/routes/meetings/components/MeetingDetail.tsx
@@ -211,6 +211,7 @@ const MeetingDetails = () => {
                     iconNode={<ListRestart />}
                     size={20}
                     className="secondaryFontColor"
+                    data-test-id="changelog-dropdown-trigger"
                   />
                 }
               >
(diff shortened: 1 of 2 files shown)


eikhr:
I don't like this. It is a bit cumbersome to have to check the report logs to find the meeting author. For the Webkom meetings it is of course obvious who is arranging the meeting, but this isn't necessarily the case for all meetings.

ivarnakken:
> I don't like this. It is a bit cumbersome to have to check the report logs to find the meeting author. For the Webkom meetings it is of course obvious who is arranging the meeting, but this isn't necessarily the case for all meetings.

hmm I suppose so - I'll have to think about a solution hehe

danielyanghansen (approved):
I like the change!

PR #5124, opened 2024-10-30
Repace usage of deprecated event.which with event.key

eikhr:
# Description

`event.which` is deprecated and may be removed from future browser versions. https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which
It can be replaced with `event.key`.

# Result

No functional or visual changes

# Testing

---

Resolves ABA-1158


ivarnakken (approved):
I like

norbye (approved):
LGTM

PR #5141, opened 2024-11-04
Update waiting list placement calculation for more performant backend

eikhr:
# Description

Appearently my code in https://github.com/webkom/lego/pull/3642 caused "significant performance issues" and "increased loading times" on the event detail page. This is true, my bad.

I believe the issue getting the fairly expensive inherited groups for each person in the waiting list. Instead we can just get child-groups for all the applicable groups in the pools. This should be cheaper because there are fewer pools.

# Result

No visual changes

# Testing

---


ivarnakken (approved):
😄

jonasdeluna (approved):
Awesome! 🚀

PR #5162, opened 2024-11-06
Update event type model

ivarnakken:
# Description

These fields are now possibly null if the user is unauthenticated


Bestem0r (approved):
Great! 🤩

PR #5165, opened 2024-11-06
Add semester auto select to interest page

Bestem0r:
# Description

Makes the selected semester under company interest  the current semester by default.

# Result

https://github.com/user-attachments/assets/7ef7bc5f-8e3a-449a-a438-6d2a2b85c55c

# Testing

Resolves [ABA-1048](https://linear.app/abakus-webkom/issue/ABA-1103/the-selected-semester-under-company-interest-should-be-the-current)

### DIFF
--- a/app/components/Form/SelectInput.tsx
+++ b/app/components/Form/SelectInput.tsx
@@ -80,7 +80,8 @@ const NO_OPTIONS_MESSAGE = 'Ingen treff';
 const LOADING_MESSAGE = 'Laster inn ...';
 
 const SelectInput = <
-  Option extends { label: string; value: number },
+  T,
+  Option extends { label: string; value: T },
   IsMulti extends boolean = false,
 >({
   name,
--- a/app/routes/companyInterest/components/CompanyInterestList.tsx
+++ b/app/routes/companyInterest/components/CompanyInterestList.tsx
@@ -241,8 +241,8 @@ const CompanyInterestList = () => {
                       language: 'norwegian',
                     })
                   : 'Vis alle semestre',
-                value: Number(currentCompanySemester?.id) ?? 9999,
-                year: currentCompanySemester?.year ?? 9999,
+                value: Number(currentCompanySemester?.id),
+                year: Number(currentCompanySemester?.year),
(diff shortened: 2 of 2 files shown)


app/routes/companyInterest/components/CompanyInterestList.tsx
```diff
@@ -226,9 +233,25 @@ const CompanyInterestList = () => {
...
+              value={{
+                label: currentCompanySemester
+                  ? semesterToText({
+                      semester: currentCompanySemester.semester,
+                      year: currentCompanySemester.year,
+                      language: 'norwegian',
+                    })
+                  : 'Vis alle semestre',
+                value: Number(currentCompanySemester?.id) ?? 9999,
+                year: currentCompanySemester?.year ?? 9999,
```
ivarnakken, line 245: 💀 why not just the current year?

app/components/Form/SelectInput.tsx
```diff
@@ -80,7 +80,8 @@ const NO_OPTIONS_MESSAGE = 'Ingen treff';
 const LOADING_MESSAGE = 'Laster inn ...';
 
 const SelectInput = <
-  Option extends { label: string; value: number },
+  T,
+  Option extends { label: string; value: T },
```
ivarnakken, line 84: is this used?

ivarnakken (approved):
Looks mostly good! 😄 Nicely done

app/components/Form/SelectInput.tsx
Bestem0r: It's maybe not the best solution, but it fixes typescript errors when `value` is not a `number` (the most common case afaik). I assume `value: any` would not be better?

app/routes/companyInterest/components/CompanyInterestList.tsx
Bestem0r: If `currentCompanySemester` is `undefined` (which it really shouldn't since the default option is the current semester), the input should default to `"Vis alle semestre"`. Looking at the old implementation, the value defaulted to `''`, but `year` is now a `number`. Would `Number(currentCompanySemester?.year)` be better?

PR #5171, opened 2024-11-10
Create Label component and improve radio/checkbox labels

eikhr:
# Description

Refactored form labels to be a reusable component. This allows for less duplicated code and more consistency when using stand-alone labels outside of the `Field`-component.

Also added lots of missing required-stars (*) to various forms.

# Result

- Minor changes to label styling and spacing
- Added missing required stars
- Replaced info-tooltip icon
----

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Company interest form
        </td>
        <td>

Screenshot 2024-11-10 at 23 39 24
        </td>
        <td>

Screenshot 2024-11-10 at 23 39 38
        </td>
    </tr>
</table>

# Testing

I have tested a lot of forms and different field types. Normal fields still work and focus as expected when clicking the label. Checkboxes and radio buttons select option when label is clicked, but nothing when the group-label is clicked.

---

Resolves ABA-1179

### DIFF
--- a/app/components/Form/Label.tsx
+++ b/app/components/Form/Label.tsx
@@ -1,0 +1,77 @@
+import { Flex, Icon } from '@webkom/lego-bricks';
+import cx from 'classnames';
+import { HelpCircle } from 'lucide-react';
+import Tooltip from 'app/components/Tooltip';
+import styles from './Label.module.css';
+import type { HTMLProps, ReactNode } from 'react';
+
+type LabelTextProps = {
+  label: ReactNode;
+  description?: string;
+  required?: boolean;
+  className?: string;
+};
+
+export const LabelText = ({
+  label,
+  description,
+  required,
+  className,
+}: LabelTextProps) => (
+  <div className={cx(styles.label, className)}>
+    {label}
+    {required && <span className={styles.required}>*</span>}
+    {description && (
+      <Tooltip content={description} className={styles.description}>
+        <Icon size={20} iconNode={<HelpCircle />} />
+      </Tooltip>
+    )}
+  </div>
+);
+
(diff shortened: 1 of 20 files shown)


danielyanghansen (changes requested):
Great change!
Just one thing I'm unsure about:
I can't see the queries reducer slice being used anywhere, so it seems better to split this into two PRs' if we're not using the reducer now

app/components/Form/Label.tsx
```diff
@@ -0,0 +1,96 @@
...
+  <fieldset
+    className={cx(styles.fieldSet, fieldSetProps.className)}
+    {...fieldSetProps}
+  >
+    <legend className={styles.fieldSetLegend}>
+      <LabelText label={legend} description={description} required={required} />
+    </legend>
+    {children}
+  </fieldset>
+);
```
ivarnakken, line 96: I like! But why is it in the `Label.tsx` file?

ivarnakken (approved):
Very clean! Had some scepticism to some of the styling removals, but I tested with ssr and looks perfect! 😄

eikhr:
> Great change! Just one thing I'm unsure about: I can't see the queries reducer slice being used anywhere, so it seems better to split this into two PRs' if we're not using the reducer now

Whoops, you are completely right. This was left in by accident, it is something I'm testing out for a different PR:)

app/components/Form/Label.tsx
eikhr: It looks exactly the same, and is typically used as a label for a group of checkboxes/radiobuttons. But I suppose it makes sense to have a separate file

danielyanghansen (approved):
:shipit:

PR #5218, opened 2024-12-10
Support previous events on event list (and a toggle switch!)

ivarnakken:
# Description

Specialisation project has finally been delivered! 🥳

This feature has been highly requested by various people. 

In addition, introduce a `ToggleSwitch` component for binary values (yes/no).

# Result

<p>
	<video src="https://github.com/user-attachments/assets/db50da38-ce08-4c6d-81db-b097b63a9f73" />
</p>

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
			image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

Previous events are still paginated properly.

The toggle switch has bene tested as both a field and a "normal" component.

---

Resolves ABA-492

### DIFF
--- a/app/components/Form/ToggleSwitch.tsx
+++ b/app/components/Form/ToggleSwitch.tsx
@@ -1,0 +1,85 @@
+import { Flex } from '@webkom/lego-bricks';
+import cx from 'classnames';
+import { Keyboard } from 'app/utils/constants';
+import { createField } from './Field';
+import styles from './ToggleSwitch.module.css';
+import type {
+  ComponentProps,
+  InputHTMLAttributes,
+  KeyboardEvent,
+  MouseEvent,
+} from 'react';
+import type { Overwrite } from 'utility-types';
+
+type Props = {
+  className?: string;
+} & Overwrite<
+  InputHTMLAttributes<HTMLInputElement>,
+  { onChange?: (checked: boolean) => void }
+>;
+
+const ToggleSwitch = ({
+  id,
+  checked,
+  value,
+  className,
+  disabled,
+  ...props
+}: Props) => {
+  const handleKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === Keyboard.ENTER) {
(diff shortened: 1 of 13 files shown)


app/components/Form/ToggleSwitch.tsx
```diff
@@ -0,0 +1,84 @@
...
+      ? checked === 'true'
+      : !!checked;
+
+  return (
+    <Flex wrap alignItems="center" gap="var(--spacing-sm)">
+      <div className={cx(styles.toggleSwitch, className)} onClick={handleClick}>
+        <input
+          {...props}
+          type="checkbox"
+          role="switch"
```
eikhr, line 60:
```suggestion
          role="switch"
          aria-checked={isChecked}
```
The `aria-checked` attribute is required when using the "switch" role. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role#associated_aria_roles_states_and_properties

app/components/Form/ToggleSwitch.tsx
```diff
@@ -0,0 +1,84 @@
...
+>;
+
+const ToggleSwitch = ({
+  id,
+  checked,
+  value,
+  className,
+  disabled,
+  ...props
+}: Props) => {
```
eikhr, line 28: It may be worth looking at the `ToggleButton` from `react-aria-components`. It may simplify this component a bit:)
https://react-spectrum.adobe.com/react-aria/ToggleButton.html

app/components/Form/ToggleSwitch.tsx
ivarnakken: Added!

app/components/Form/ToggleSwitch.tsx
ivarnakken: Clever! I refactored it now - ended up much cleaner and "safer"

PR #5230, opened 2024-12-12
Include registration statuses to event items

ivarnakken:
# Description

Makes it much easier to differentiate between the list of dates in the items.
There's no need to show the activation time for registrations if the user has
already registered.

Besides, people prefer to use the event list on /events rather than the
one on the user profile to view upcoming events.

# Result

Look at _Deloitte AS_ and _Fadderpåmelding Kommunikasjonsteknologi_ to see event items with 2 dates ... difficult to differentiate between the two:

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

Tested all event types, and the three different registration statuses (unregistered, waiting list and registered/admitted)

### DIFF
--- a/app/components/EventItem/index.tsx
+++ b/app/components/EventItem/index.tsx
@@ -1,6 +1,5 @@
 import { Flex, Icon, Image } from '@webkom/lego-bricks';
 import {
-  AlarmClock,
   Calendar,
   CalendarClock,
   CircleAlert,
@@ -8,6 +7,7 @@ import {
   Clock,
   Timer,
 } from 'lucide-react';
+import moment from 'moment-timezone';
 import { Link } from 'react-router-dom';
 import Pill from 'app/components/Pill';
 import Tag from 'app/components/Tags/Tag';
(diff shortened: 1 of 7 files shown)


app/components/EventItem/index.tsx
```diff
@@ -205,12 +201,20 @@ const EventItem = ({
...
         >
-          <Flex column gap="var(--spacing-xs)" className="secondaryFontColor">
-            <div>
-              <h3 className={styles.eventItemTitle}>{event.title}</h3>
-              {event.totalCapacity > 0 && <Attendance event={event} />}
-            </div>
-            <TimeStartAndRegistration event={event} />
+          <Flex column gap="var(--spacing-sm)" className="secondaryFontColor">
+            <h3 className={styles.eventItemTitle}>{event.title}</h3>
+            {event.totalCapacity != null && event.totalCapacity > 0 && (
```
norbye, line 206:
```suggestion
            {event.totalCapacity !== null && event.totalCapacity > 0 && (
```
I like ===

norbye (approved):
Love it!! Have been missing a nice overview of events I've registered to for so long

app/components/EventItem/index.tsx
ivarnakken: This is actually intentional so to include `undefined` 😄

norbye: Ah, I thought `!=` would cover more than just `undefined` and `null`, but it seems I was mistaken

PR #5232, opened 2024-12-13
Extend DatePicker to support ranges

ivarnakken:
# Description

Better than having to use two separate date pickers to simulate it. The plan is to use this to filter on events on /events.

# Result

Notice how the regular dropdown does not update which month is shown as default according to the value it is given (unlike the new range version 😎):

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

- Tested when start date and end date are the same days (but different times of the day)
- Tested as both fields and regular components (see images above)
- Tested that changes to global table styles does not affect the Table component
- Tested that it looks good on smaller viewports:
	image

- Tested that the regular date picker still works :)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -281,6 +281,7 @@ steps:
       - install_cypress
       - ssr
     environment:
+      TZ: Europe/Oslo
       CYPRESS_API_BASE_URL: http://api:8000
       CYPRESS_RESET_DB_API: http://legocypresshelper:3030
       CYPRESS_BASE_URL: http://ssr:3000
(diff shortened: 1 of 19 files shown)


ivarnakken:
> Came onto some situations while testing where the choice of date felt a little buggy (regarding which of the two dates should be moved to the new location) - but didn't get any clear cases to reproduce just yet so might be simpler to try it out and change it after a little bit.

I made some changes to this like 30 min ago - did you use the up-to-date code? 

> One thing that I think should be changed before merging this is closing the dialog after selecting your second date. As times are below dates I would naturally select a date first and that makes it annoying that it decides to close on its own before I've gotten to do that

Very good point! I'll fix that asap

norbye:
> I made some changes to this like 30 min ago - did you use the up-to-date code?

At first I didn't, then I did - might be why I struggled to reproduce it properly. Looked smoother now

PR #5233, opened 2024-12-14
Improve and clean up search menu

ivarnakken:
# Description

[Refactor and improve the search menu](https://github.com/webkom/lego-webapp/commit/b0afb9671f1f5f45a7b518e9b51d0207d7aa7fa0), and [remove profile and tags links](https://github.com/webkom/lego-webapp/commit/02f468b5d67158d652911596bf7d0ca57ab8b77e) (the user's have spoken; they don't think the links are needed as they just make the menu more cluttered).

# Result

Test it yourself to get a feel for the animations :))

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
</table>

# Testing

- Quite extensively tested all functionality, responsiveness etc.
- Tested on my iPhone (not just in the devtools).


app/components/Search/index.tsx
```diff
@@ -6,6 +6,7 @@ import { useIsLoggedIn } from 'app/reducers/auth';
 import { selectAutocompleteRedux } from 'app/reducers/search';
 import { useAppDispatch, useAppSelector } from 'app/store/hooks';
 import { Keyboard } from 'app/utils/constants';
+import FancyNodesCanvas from '../Header/FancyNodesCanvas';
```
norbye, line 9: Not a fan of this, feels weird when it's not there before and then it's there when you open the search bar.

Not as noticeable on dark mode, but on light mode it feels weird weird

Thinking about it it might also be because the gray color of the FancyNodes is slightly different from the gray color that is now set as the background - so the entire background changes when the search bar is opened

app/components/Search/Search.module.css
```diff
@@ -1,22 +1,53 @@
 @import url('~app/styles/variables.css');
 
+.wrapper {
+  position: fixed;
+  inset: var(--development-banner-height) 0 0;
```
norbye, line 5: What's the advantages of this approach over what we did previously?

norbye (approved):
LGTM

Not a critical ask, but when the screen resizes and some menu items have to be scrolled to be reached it would be sick to have the same shadow as we do in the attendees modal

app/components/Search/index.tsx
norbye: Anyways the fancy nodes feel even more noticeably old when I'm getting used to seeing the rest of the website without them - so opening the menu feels like an ugly blast from the past

app/components/Search/Search.module.css
ivarnakken: The search bar is now aligned with the "close" button, and on development mode it gets pushed further down because of the development banner that says "This is a development version" or something. This is just so that it's always aligned, even on development

app/components/Search/index.tsx
ivarnakken: I agree - I vote to remove entirely

ivarnakken:
> Not a critical ask, but when the screen resizes and some menu items have to be scrolled to be reached it would be sick to have the same shadow as we do in the attendees modal

That would be nice! I considered it, but must have forgotten

app/components/Search/Search.module.css
norbye: Ahahh, aight

ivarnakken: Just to clarify further; the development banner was not visible in the search menu before (see image above). I though it was nice to include :)

PR #5235, opened 2024-12-15
Fix background color on flat buttons on hover

ivarnakken:
# Description

Fetching data from the GitHub uptime repo hehe - couldn't find a better api

# Result

Notice the pulse!

https://github.com/user-attachments/assets/81d077f9-1d26-444d-9988-848095d4568d

# Testing

It works? Apparently one service has degraded performance 😩

### DIFF
--- a/app/actions/StatusActions.ts
+++ b/app/actions/StatusActions.ts
@@ -1,0 +1,50 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+export type SystemStatus = {
+  status: 'operational' | 'degraded' | 'major';
+  message: string;
+};
+
+type UptimeService = {
+  name: string;
+  status: 'up' | 'down' | 'degraded';
+  uptime: string;
+};
+
+export const fetchSystemStatus = createAsyncThunk(
+  'status/fetch',
+  async (): Promise<SystemStatus> => {
+    const response = await fetch(
+      'https://raw.githubusercontent.com/webkom/uptime/master/history/summary.json',
+    );
+
+    if (!response.ok) {
+      throw new Error('Failed to fetch system status');
+    }
+
+    const data = (await response.json()) as UptimeService[];
+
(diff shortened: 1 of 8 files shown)


norbye (approved):
cool!

app/actions/StatusActions.ts
```diff
@@ -0,0 +1,50 @@
...
+type UptimeService = {
+  name: string;
+  status: 'up' | 'down' | 'degraded';
+  uptime: string;
+};
+
+export const fetchSystemStatus = createAsyncThunk(
+  'status/fetch',
+  async (): Promise<SystemStatus> => {
+    const response = await fetch(
```
eikhr, line 17: This is a very clean way of fetching, but it doesn't follow convention. Also, I'm pretty sure the `SystemStatus.FETCH` won't work now, meaning loading state doesn't get set.

I suppose it's fine if we don't need loading state, but in that case just remove the unused `SystemStatus.FETCH` action types.

ivarnakken: > This is a very clean way of creating a fetching thunk, but it doesn't follow lego-webapp convention.

It's sort of a mixture between "the convetion" and the readme fetching way. You're probably right about the loading states. Not sure what's the best way to do this is - do you have any input? I feel like you know our redux implementation _much_ better than I do

Bestem0r (approved):
Looks good!

eikhr (approved):

falbru:
@ivarnakken Merge this s00n?

ivarnakken:
> @ivarnakken Merge this s00n?

Sure, but I think I was waiting for the backend to not be in a degraded state

itsisak:
> > @ivarnakken Merge this s00n?
> 
> 
> 
> Sure, but I think I was waiting for the backend to not be in a degraded state

Well that's fixed now ;)

PR #5250, opened 2024-12-23
Only validate payment deadline if event is priced

norbye:
# Description

The date of the payment deadline was validated even though the event was not priced. As such, the warning that the date had to be changed was not shown, even though the event was not correctly validated if the unregistration deadline was later than a week after the creation date of the event (which is the default value for the payment deadline it seems).

Now the validation is only applied if isPriced is true.

# Result

Werks as expected.

# Testing

---

Resolves ABA-1211


ivarnakken (approved):
Makes sense! Nice

eikhr (approved):

PR #5252, opened 2024-12-25
Improve and fix navigation tabs on smaller viewports

ivarnakken:
# Description

The tabs now overflow nicely, with both a fade and arrow to indicate the scroll behaviour.

The tabs are now also scrolled into view if clicked on.

# Result

- [yes] Changes look good with different viewports (mobile, tablet, etc.).

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Same as before when all tabs fit into view
        </td>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
    <tr>
        <td>
            At the beginning
        </td>
        <td>
            image
        </td>
        <td>
            image
        </td>
    </tr>
    <tr>
        <td>
            At the end
        </td>
        <td>
            image
        </td>
        <td>
        	image
        </td>
    </tr>
</table>

https://github.com/user-attachments/assets/18876979-2987-4056-894c-0bfb338a5c7c

# Testing

Tested quite extensively on smaller views. A lot of edge cases have been tested, e.g. that arrows and shadows disappear when they should etc.

### DIFF
--- a/packages/lego-bricks/src/components/Tabs/TabContainer.tsx
+++ b/packages/lego-bricks/src/components/Tabs/TabContainer.tsx
@@ -1,5 +1,6 @@
 import cx from 'classnames';
-import { useEffect, useState } from 'react';
+import { ChevronLeft, ChevronRight } from 'lucide-react';
+import { useEffect, useState, useRef } from 'react';
 import { useLocation } from 'react-router-dom';
 import styles from './TabContainer.module.css';
 import type { ReactNode } from 'react';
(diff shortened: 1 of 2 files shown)


Bestem0r (approved):
Nice! Looks a lot better

packages/lego-bricks/src/components/Tabs/TabContainer.tsx
```diff
@@ -23,15 +49,62 @@ export const TabContainer = ({ className, lineColor, children }: Props) => {
...
+      element.addEventListener('scroll', checkScroll);
+      window.addEventListener('resize', checkScroll);
+      checkScroll();
+    }
+
+    return () => {
+      element?.removeEventListener('scroll', checkScroll);
+      window.removeEventListener('resize', checkScroll);
+    };
+  }, []);
```
eikhr, line 75:
```suggestion
      checkScroll();

      return () => {
        element.removeEventListener('scroll', checkScroll);
        window.removeEventListener('resize', checkScroll);
      };
    }
  }, []);
```
Move the event listener removal into the if statement to make sure to only remove listeners we actually added.

eikhr (approved):
Looks great! I love this solution

packages/lego-bricks/src/components/Tabs/TabContainer.tsx
ivarnakken: Wouldn't it be safer to run the cleanup function regardless of whether the element exists or not during unmount? I suppose there's always a chance of `element` becoming falsy during the lifecycle

PR #5259, opened 2024-12-29
Support date range filtering on events

ivarnakken:
# Description

Highly requested and very useful!

Had to make some changes to the date picker and query hook.

# Result

image

# Testing

Tested on hard refreshes, many different date ranges, together with the `query.showPrevious` and the "Load more" button.

### DIFF
--- a/app/components/Form/DatePicker.tsx
+++ b/app/components/Form/DatePicker.tsx
@@ -73,7 +73,6 @@ const DatePicker = ({
     }
     return moment().add(1, 'month');
   });
-  const [selectingEnd, setSelectingEnd] = useState(false);
 
   const parsedValue = useMemo(() => {
     if (range && Array.isArray(value)) {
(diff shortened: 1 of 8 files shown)


falbru (changes requested):
Awesome feature!

Found a bug where I could pick a start date that was after the end date:

image

I first selected 1, then 16 and then 23.

ivarnakken:
> Found a bug where I could pick a start date that was after the end date:

Nice catch :) I noticed there was some buggy behaviour but it was difficult to reproduce; thanks for the example. I'll try to make it less prone to bugs but there are a ton of edge cases to this .. 😩

ivarnakken:
I made several improvements to the date picker now - try it out! It should now cover all edge cases pretty nicely. I tried to follow a similar logic to [shadcn](https://ui.shadcn.com/docs/components/date-picker) and [NextUI](https://nextui.org/docs/components/date-range-picker)

falbru (approved):
Awesome! Works for me now!

PR #5261, opened 2024-12-30
Add semester arrows to BDB

Bestem0r:
# Description

Adds arrows to each side of the semester dropdown in BDB to easily go to the next/previous semester.

# Result

https://github.com/user-attachments/assets/31cc1518-41dd-4d8b-b058-72793c9f9540

# Testing

Open BDB and verify that the semester arrows work correctly.

---

Resolves [ABA-1048](https://linear.app/abakus-webkom/issue/ABA-1216/add-semester-navigation-buttons)

### DIFF
--- a/app/routes/bdb/components/BdbPage.tsx
+++ b/app/routes/bdb/components/BdbPage.tsx
@@ -176,20 +176,19 @@ const BdbPage = () => {
       </Card>
 
       <Flex width="fit-content">
-        <button
-          onClick={() =>
+        <Icon
+          iconNode={<ChevronLeft />}
+          onPress={() =>
             setQuery({
               ...query,
               semester: getSemesterSlugOffset(
                 currentCompanySemester!.id,
                 companySemesters,
-                -1,
+                'previous',
               ),
             })
           }
-        >
-          <Icon iconNode={<ChevronLeft />} />
-        </button>
+        />
         <SelectInput
           name="semester"
           options={companySemesters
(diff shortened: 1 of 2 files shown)


app/routes/bdb/utils.tsx
```diff
@@ -158,6 +158,18 @@ export const getSemesterSlugById = (
   return `${semester.year}${semester.semester === 'autumn' ? 'h' : 'v'}`;
 };
 
+export const getSemesterSlugOffset = (
```
ivarnakken, line 161: Is this not kind of over engineered? Do we ever need an offset greater than 1? I feel like it would be more practical to just get a direction as an arg; `previous`/`next` or `left`/`right`

app/routes/bdb/components/BdbPage.tsx
```diff
@@ -174,6 +176,20 @@ const BdbPage = () => {
...
+              semester: getSemesterSlugOffset(
+                currentCompanySemester!.id,
+                companySemesters,
+                -1,
+              ),
+            })
+          }
+        >
+          <Icon iconNode={<ChevronLeft />} />
+        </button>
```
ivarnakken, line 192:
```suggestion
        <Icon iconNode={<ChevronLeft />} onPress={() =>
            setQuery({
              ...query,
              semester: getSemesterSlugOffset(
                currentCompanySemester!.id,
                companySemesters,
                -1,
              ),
            })
          }
         />
        </button>
```

Just add the onPress directly to the icon

ivarnakken (changes requested):
Nice! I find this to be very useful

app/routes/bdb/utils.tsx
Bestem0r: Yeah you are probably right, will change it 👍

ivarnakken (approved):
Not sure about the `offset` argument name but who cares! Very nice!

PR #5262, opened 2024-12-30
Introduce a collapsableAfter prop to the table component

ivarnakken:
# Description

For tables that are not paginated, it's good to be able to collapse them
to prevent long render times

# Result

image

# Testing


falbru (approved):
Very clean! We should use collapsable more often! It's very annoying to scroll down long tables/lists

norbye (approved):

PR #5271, opened 2025-01-06
Sorting and filtering by presence in attendee list

emillub:
# Description
Add functionality for sorting by presence in attendee list as requested by Fagkom.
Also add functionality for filtering.

# Result

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>

image
</td>
        <td>

image
image
        </td>
    </tr>
</table>

# Testing

---

Resolves ABA-1181

### DIFF
--- a/app/components/Table/index.tsx
+++ b/app/components/Table/index.tsx
@@ -138,11 +138,7 @@ const Table = <T extends { id: EntityId }>({
   }, [sort, data]);
 
   const filter: (item: T) => boolean = (item) => {
-    const activeFilters = Object.entries(filters).filter(
-      ([_, value]) => value?.length,
-    );
-
-    if (activeFilters.length === 0) {
+    if (Object.values(filters).every(isEmpty)) {
       return true;
     }
 
@@ -167,6 +163,7 @@ const Table = <T extends { id: EntityId }>({
       if (typeof filters[key] === 'boolean') {
         return filters[key] === get(item, index);
       }
+
       return filters[key]?.some((arrayFilter) =>
         String(filterMapping(get(item, index) ?? ''))
           .toLowerCase()
--- a/app/components/Table/index.tsx
+++ b/app/components/Table/index.tsx
@@ -163,11 +163,7 @@ const Table = <T extends { id: EntityId }>({
       if (typeof filters[key] === 'boolean') {
         return filters[key] === get(item, index);
       }
-      console.log(
-        filters[key],
-        get(item, index),
-        filterMapping(get(item, index) ?? ''),
-      );
+
       return filters[key]?.some((arrayFilter) =>
         String(filterMapping(get(item, index) ?? ''))
           .toLowerCase()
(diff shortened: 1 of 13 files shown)


emillub:
I made a small alteration to the filtering function in Table/Index as it gave zero results when manually removing all filters. I am not completely sure that this doesnt mess anything up elsewhere. Could someone with more experience take a look? Or I could revert back to only sorting?

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
ivarnakken, line 1: We should use the `Presence` enum defined in `app/store/models/Registration.ts`

ivarnakken (approved):
This looks good (except for the nitpick about using the enum)! 😍 Very nice feature!

Not sure about the changes to the Table component filtering, but it seems alright. @norbye knows more than me here

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
```diff
@@ -271,6 +282,20 @@ export const RegisteredTable = ({
...
+          'LATE',
+          'UNKNOWN',
+          'PRESENT',
+        ];
+        if (
+          presenceSortingOrder.indexOf(a['presence']) >
+          presenceSortingOrder.indexOf(b['presence'])
+        )
+          return 1;
+        else return -1;
```
norbye, line 297: If we are going to shorten it down by skipping curly brackets, might we also throw in a ternary?
```suggestion
        return presenceSortingOrder.indexOf(a['presence']) > presenceSortingOrder.indexOf(b['presence']) ? 1 : -1;
```
or even neater
```suggestion
        return (presenceSortingOrder.indexOf(a['presence']) - presenceSortingOrder.indexOf(b['presence']));
```

norbye (approved):
Love it!

I second @ivarnakken in that you should use the enum, and added some nitpick of my own - although that's more minor stuff which in my opinion makes it more compact and readable.

app/components/Table/index.tsx
```diff
@@ -163,7 +167,6 @@ const Table = <T extends { id: EntityId }>({
       if (typeof filters[key] === 'boolean') {
         return filters[key] === get(item, index);
       }
-
```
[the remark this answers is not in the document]
emillub: Haha I didnt notice i deleted it. I agree there should be a blank line there

emillub:
Did I use the enum correctly? @ivarnakken
Also; Since the item being filtered is made lowercase I assume the filter is supposed to be case insensitive. I modified the filtering so that the filter value is also made lowercase to avoid my previous workaround. Does it look okay? @norbye

app/routes/events/components/EventAdministrate/RegistrationTables.tsx
emillub: Done! I think

norbye:
> Also; Since the item being filtered is made lowercase I assume the filter is supposed to be case insensitive. I modified the filtering so that the filter value is also made lowercase to avoid my previous workaround. Does it look okay? @norbye

Awesome! You made it case insensitive the same way I would've, and it seems to work just fine:) also fixes case insensitive searching so thats perfect

Remember to squash merge (or clean up your commits) when you're finishing up

PR #5281, opened 2025-01-15
bug-fix css for eventType name

tindevw:
# Description
Event Type text and buttons were positioned wrong. Aligned it correctly by changing JustifyComponent to flex-end in the ButtonGroup. 

# Result

- [x]  Changes look good on both light and dark theme.

before:
image
image

after:  

image
image
          
 

# Testing

---

Resolves ABA-1230

### DIFF
--- a/packages/lego-bricks/src/components/Layout/Page/Page.module.css
+++ b/packages/lego-bricks/src/components/Layout/Page/Page.module.css
@@ -39,8 +39,9 @@
 }
 
 .buttonGroup {
-  flex-grow: 1;
+  flex-grow: 0;
   justify-content: flex-end;
+  margin-left: auto;
 }
 
 .description {


Viljen789 (commented):
Looks good, maybe a small mistake with the extra newline, but the fix is amazing!

app/routes/events/components/EventDetail/index.tsx
```diff
@@ -173,7 +173,14 @@ const EventDetail = () => {
...
         event && (
-          <div className={styles.eventType}>
+          <div
+            className={`${styles.eventType} justifyContent__flex-end`}
+            style={{
+              display: 'flex',
+              justifyContent: 'flex-end',
+              width: '100%',
+            }}
+          >
```
ivarnakken, line 183: No inline css please🙏😅 nice to be consistent.

Besides, the bug is actually in the Page component and not here specifically, meaning all action buttons on all pages are affected.

ivarnakken (changes requested):

app/routes/events/components/EventDetail/index.tsx
tindevw: Thanks for pointing it out! I was not entirely sure how to address it. But I can take a closer look at it later maybe with some help, as I'm still figuring out how things works. If it's urgent, though, perhaps someone else can take it on. :)

tindevw: I made changes to the ButtonGroup now, and it seems to have improved the alignment as the buttons have now shifted to the right position. 
image
image

packages/lego-bricks/src/components/Button/ButtonGroup.tsx
```diff
@@ -12,7 +12,7 @@ type Props = {
 export const ButtonGroup = ({
   vertical = false,
   centered = false,
-  justifyContent,
+  justifyContent = 'flex-end',
```
ivarnakken, line 15: This will effect all button groups everywhere and might not be what we want. Look at the action button styling for the page component

tindevw: I will. Thank you.

eikhr (approved):
Looks great! :shipit:

falbru (approved):
Awesome!

ShaileshS1702 (approved):
Nice work!

ivarnakken (approved):
I think this is good! 😍 Awesome job!

PR #5297, opened 2025-01-24
Define link for groups with the board type

norbye:
# Description

Defined link path for board groups to make badges clickable

# Result

Is clickable

# Testing

I clicked

---

Resolves ABA-1243


eikhr (commented):
👍

eikhr (approved):

PR #5300, opened 2025-01-25
Fix cypress stripe tests

norbye:
# Description

Got the websockets to work out locally, with stripe, so I got the rest of the tests to pass as well.

Had to update some of the intercepts as stripe has evolved.

# Result

No visual changes, but it all works out fine when I'm testing it locally

# Testing

---

Resolves ABA-1247


eikhr (approved):
Looks good! Will be really great to get Stripe updated:)

ivarnakken (approved):
Love it

PR #5319, opened 2025-02-01
Add skeletons and accordions to survey result and submissions

itsisak:
# Description

For larger events, the survey page is suffers from performance issues since there is a large amount of submissions. This PR aims to improve user feedback. 

Also make each submission into accordions for better usability but also remove the content (the answers) from the virtual dom when they are not expanded to improve performance. This requires disabling animations for the accordions, but I think that is fine. 

There was an idea to paginate the page with individual answers, however since the result/summary needs all submissions anyways that solution does not make  much sense.

# Result

Images of skeleton components are not included, but they are added to the following places:
 - /surveys
 - PieChart, BarChart and TextAnswers (used on /surveys/<id>/submissions/summary and on /events/<id>/administrate/statistics but does not seem to apply there)
 - /surveys/<id>/submissions/individual

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Accordions on `/surveys/<id>/submissions/individual`
        </td>
        <td> 

Screenshot 2025-02-01 at 16 48 37
        </td>
        <td> 

Screenshot 2025-02-01 at 16 38 27
        </td>
    </tr>
</table>

# Testing

Seems to work :)

---

Resolves ABA-1259

### DIFF
--- a/app/components/Chart/BarChart.tsx
+++ b/app/components/Chart/BarChart.tsx
@@ -1,15 +1,24 @@
+import { Skeleton } from '@webkom/lego-bricks';
 import { BarChart, Bar, Cell, XAxis, YAxis } from 'recharts';
 import { CHART_COLORS } from 'app/components/Chart/utils';
+import styles from './Chart.module.css';
 import type { DistributionDataPoint } from 'app/components/Chart/utils';
 
 type Props = {
   distributionData: DistributionDataPoint[];
   chartColors?: string[];
+  fetching?: boolean;
 };
 const DistributionBarChart = ({
   chartColors = CHART_COLORS,
   distributionData,
+  fetching = false,
 }: Props) => {
+
+  if (fetching) {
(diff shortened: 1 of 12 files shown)


ivarnakken (approved):
Let's gooo 🚀 😍

norbye (approved):
Sweet!

> This requires disabling animations for the accordions, but I think that is fine.

You could also look into the accordion component if you want to to fix this behaviour at some point(: it cannot be removed from the DOM while it's animating, but there might be a neat way to remove it after the animation has completed, idk

itsisak:
> You could also look into the accordion component if you want to to fix this behaviour at some point(: it cannot be removed from the DOM while it's animating, but there might be a neat way to remove it after the animation has completed, idk

I agree, that would be nice. Currently I believe the opening animation also requires the content to be in the DOM before the animation. Might look into it

PR #5323, opened 2025-02-03
Add semester status filtering in BDB

Bestem0r:
# Description

Adds inline filtering to semester statuses in BDB Pro. Choosing multiple filters works as `or` on the semester statuses.

[Backend PR](https://github.com/webkom/lego/pull/3725)

# Result

https://github.com/user-attachments/assets/c74165cf-549f-4c35-af31-7829752502ce

# Testing

Resolves [ABA-1048](https://linear.app/abakus-webkom/issue/ABA-1215/add-filtering-and-sorting-options-to-bdb)

### DIFF
--- a/app/components/Search/utils.tsx
+++ b/app/components/Search/utils.tsx
@@ -161,7 +161,7 @@ const LINKS: Array<Link> = [
     key: 'bdb',
     title: 'Bedriftsdatabase',
     icon: <Database />,
-    url: '/bdb/bdb-pro',
+    url: '/bdb',
   },
   {
     admin: true,
--- a/app/components/Search/utils.tsx
+++ b/app/components/Search/utils.tsx
@@ -161,7 +161,7 @@ const LINKS: Array<Link> = [
     key: 'bdb',
     title: 'Bedriftsdatabase',
     icon: <Database />,
-    url: '/bdb',
+    url: '/bdb/bdb-pro',
   },
   {
     admin: true,
--- a/app/routes/bdb/components/BdbDetail.tsx
+++ b/app/routes/bdb/components/BdbDetail.tsx
@@ -452,7 +452,7 @@ const BdbDetail = () => {
       }
       title={title}
       back={{
-        href: '/bdb',
+        href: '/bdb/bdb-pro',
       }}
       actionButtons={
         <LinkButton key="edit" href={`/bdb/${companyId}/edit`}>
--- a/app/routes/bdb/components/BdbDetail.tsx
+++ b/app/routes/bdb/components/BdbDetail.tsx
@@ -452,7 +452,7 @@ const BdbDetail = () => {
       }
       title={title}
       back={{
-        href: '/bdb/bdb-pro',
+        href: '/bdb',
       }}
       actionButtons={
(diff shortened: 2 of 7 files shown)


ch0rizo (commented):
Very cool, very nice

app/components/Search/utils.tsx
```diff
@@ -161,7 +161,7 @@ const LINKS: Array<Link> = [
     key: 'bdb',
     title: 'Bedriftsdatabase',
     icon: <Database />,
-    url: '/bdb',
+    url: '/bdb/bdb-pro',
```
norbye, line 164: Just want to pick your brains on this one..

Why change the URL at all? Isn't it sufficient to add some pro badges on the titles and navbar? Considering that we're not keeping the "not-pro" version.

Also, if you're changing the URL, have you considered just making it `bdb-pro` or `bdb/pro` instead of `bdb/bdb-pro`?

And if you're changing it, ensure backwards compatibility with the URLs - would be a shame if any of the old paths (such as `/bdb` or `/bdb/*` stops working)

norbye (changes requested):
Slapping on a small changes requested - just cause I want to hear about the path stuff to ensure that we're not leaving unnecessary invalid paths in our wake.

Otherwise it looks very nice

app/components/Search/utils.tsx
ivarnakken: I don't want this

Especially not a duplicate "/bdb/bdb"

Bestem0r: It was changed in order to differentiate between pages when I used the sidebar filter component. Now that the filtering is inline, it is no longer needed. I agree that `/bdb/bdb-pro` doesn't look very good, so I'll revert the change 👍

app/routes/bdb/components/BdbDetail.tsx
```diff
@@ -452,7 +452,7 @@ const BdbDetail = () => {
       }
       title={title}
       back={{
-        href: '/bdb',
+        href: '/bdb/bdb-pro',
```
norbye, line 455: 👀

app/routes/bdb/components/CompanyEditor.tsx
```diff
@@ -157,7 +157,7 @@ const CompanyEditor = () => {
             message="Er du sikker på at du vil slette denne bedriften?"
             onConfirm={() =>
               dispatch(deleteCompany(companyId)).then(() => {
-                navigate('/bdb');
+                navigate('/bdb/bdb-pro');
```
norbye, line 160: 👀

app/routes/company/components/CompaniesPage.tsx
```diff
@@ -163,7 +163,7 @@ const CompaniesPage = () => {
       })}
       actionButtons={
         (actionGrant.includes('create') || actionGrant.includes('edit')) && (
-          <LinkButton href="/bdb">Bedriftsdatabasen</LinkButton>
+          <LinkButton href="/bdb/bdb-pro">Bedriftsdatabasen</LinkButton>
```
norbye, line 166: 👀

norbye (changes requested):

norbye (approved):
Nice!

PR #5329, opened 2025-02-08
Add pro badge to BDB

Bestem0r:
# Description

Rebasing messed the other PR up, so here's a new one... 

# Result

| Before | After |
| -- | -- |
| image | image |
| image | image |

# Testing

Resolves [ABA-1038](https://linear.app/abakus-webkom/issue/ABA-1083/add-pro-badge-to-bdb)


app/routes/bdb/components/bdb.module.css
norbye, line 1: It doesn't seem like any of these classes are actually used. Any reason to keep them?

norbye (commented):
Gotta love a proper rebase<3

Kinda bugging me that BDB ends up being sorted at the back just because it is now rendered as a JSX element and not text. Thoughts on adding a `sortTitle` property or something that can be used to sort instead of `title` when it is not a string?

ch0rizo:
Pro😎

Bestem0r:
> Gotta love a proper rebase<3
> 
> Kinda bugging me that BDB ends up being sorted at the back just because it is now rendered as a JSX element and not text. Thoughts on adding a `sortTitle` property or something that can be used to sort instead of `title` when it is not a string?

Yeah, I agree. I'll see what I can do

app/routes/bdb/components/bdb.module.css
Bestem0r: Seems like `.deleteFile` and `.surveyContainer` are the only ones used. Removing the rest.

norbye (approved):
Nice!

PR #5335, opened 2025-02-11
Fix sidebar bug in achievements and move link

norbye:
# Description

Fixed some bugs in the achievements
* The location check used to toggle the sidebar was above the component, which caused it to only be reloaded when you refresh the page, not when navigating
* The location check is sorta suboptimal, but at least now it covers both `/achievements` and `/achievements/`
* The overview page was accessible from both `/achievements` and `/achievements/overview`. The user page used the former, and the internal navigation links used the latter, and the nav-link active state only listened to the latter. Now they all use `/achievements`
* The "Topplister" button used to be placed next to the badges, now it is displayed under the trophies - which makes more sense imo

# Result

Resolved the bugs mentioned above.

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>image</td>
        <td>image</td>
    </tr>
</table>

# Testing

---

### DIFF
--- a/app/routes/achievements/components/index.tsx
+++ b/app/routes/achievements/components/index.tsx
@@ -4,9 +4,11 @@ import { RadioButton } from 'app/components/Form';
 import useQuery from 'app/utils/useQuery';
 import { AchievementTabs } from './utils';
 
-const showFilters = location.pathname === '/achievements/overview';
-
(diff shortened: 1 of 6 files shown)


jonasdeluna (approved):
Awesome! 🔥

PR #5341, opened 2025-02-12
Improve `isMobileViewport` hook

ivarnakken:
🤷🏼‍♂️ thought it was kinda cool to support blackberry and iemobile

### DIFF
--- a/app/utils/isMobileViewport.ts
+++ b/app/utils/isMobileViewport.ts
@@ -5,7 +5,20 @@ const isMobileViewport = () => {
     return false;
   }
   const width = window?.innerWidth;
-  return width < 768;
+
+  // Check user agent string
+  const userAgent = navigator.userAgent.toLowerCase();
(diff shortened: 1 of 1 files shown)


app/utils/isMobileViewport.ts
```diff
@@ -1,8 +1,24 @@
...
+    return false;
+  }
+  const width = window?.innerWidth;
+
+  // Check user agent string
+  const userAgent = navigator.userAgent.toLowerCase();
+  const isMobileUA =
+    /mobile|android|iphone|ipad|ipod|blackberry|iemobile|opera mini/i.test(
+      userAgent,
+    );
```
eikhr, line 14: Can we also check the User-Agent header when SSR?

ivarnakken: seems to work fine? 🤷🏼‍♂️

PR #5358, opened 2025-02-14
Fix Sentry crash with node v20

eikhr:
# Description

Staging is down because our old Sentry version crashes when running node 20. Updating Sentry should fix it 🤞

# Testing

- [ish] I have thoroughly tested my changes.

It seems to work locally, but to be sure we need to test that errors are reported correctly in staging.

---

Resolves ABA-1286


jonasdeluna (approved):
📈📈📈📈

PR #5359, opened 2025-02-14
Start migration to react router route modules

eikhr:
# Description

I think the best way to move away from Webpack is to start using a framework. Imo. the best way to do this is by migrating to the "framework mode" in react-router v7, which is basically what was previously called Remix. It is similar to Next.js, but provides an easier migraiton path because we are already using react-router.

In this PR, I have started migrating some routes to the format that is used by the file system based router in react-router v7. It is currently not using any framework features, but instead converting them back to the library-mode react-router routes we are already using.

My plan is to keep migrating all the routes, and once everything is done, attempt to start using the framework features in vite mode:)

The migrated routes look like this:
Screenshot 2025-02-15 at 14 12 50

FEEDBACK VERY WELCOME!!

Links:
- Docs on migration: https://reactrouter.com/upgrading/router-provider#framework-adoption-from-routerprovider
- Docs on file system routing: https://reactrouter.com/how-to/file-route-conventions

# Testing

Stuff works the same as previously, as the changes are mostly moving files around.

---


app/routes/admin/email/index.tsx
```diff
@@ -1,61 +1,67 @@
...
+        lazy: () =>
+          import('app/routes/admin_.email.lists.$emailListId/route').then(
+            convert,
+          ),
+      },
+      {
+        path: 'lists/:emailListId',
+        lazy: () =>
+          import('app/routes/admin_.email.lists.$emailListId/route').then(
+            convert,
```
jonasdeluna, line 25: is this really the intended way to use this routing system?

jonasdeluna:
This is nice, do we need to migrate everything at once before moving over?

app/routes/admin/email/index.tsx
eikhr: No no, this is just temporary to make it work with our current routing until all the routes are migrated. Once everything is migrated we can start using the file-system routing, meaning that react-router will automatically create the routes based on the folder names (f.ex. `admin_.email.lists.$emailListId` which corresponds to a url like `/admin/email/lists/10`)

eikhr:
> This is nice, do we need to migrate everything at once before moving over?

Nope, if you look at `app/utils/convertRoute.ts` I added a temporary conversion function, that makes the migrated routes still work with the current old router:)
The plan is to do it gradually over multiple PRs to avoid merge conflicts as much as possible

ivarnakken (approved):
Probably no issues! :shipit: ™️

jonasdeluna (approved):
💎👐

eikhr:
After some investigation into Remix/react-router framework mode, I think it will be too much work to migrate. It seems like there is no good way to set up redux with SSR, which is hard to replace at this point. Sooo, I'm just going to close this:)

Currently looking into https://vike.dev/ instead, it seems a lot less opinionated so should be easier for us to migrate too!

PR #5380, opened 2025-02-26
Migrate from Webpack to Vite/Vike

eikhr:
# Description

https://vike.dev/ is a very customizable framwork, which allows us to make most of the existing code work and migrate gradually to the recommended way of doing things.

TODO (Before merging):
- [ ] Fix cypress tests
- [x] Update Dockerfile for production builds

TODO (after merging):
- [ ] Move routing from react-router to built-in Vike router

# Result

Better??

# Testing

- We will have to test in staging and fix issues as they come up

Most of the important features still work:)

---

Resolves ABA-1141


ivarnakken (approved):
lgtm

jonasdeluna (approved):
This is probably good but it was too long to read unfortunately.

falbru (approved):
lgtm

PR #5381, opened 2025-02-26
Add complete profile achievement

Arashfa0301:
Add complete profile achievement

### DIFF
--- a/app/utils/achievementConstants.ts
+++ b/app/utils/achievementConstants.ts
@@ -124,6 +124,14 @@ const AchievementsInfo: Record<string, AchievementData[]> = {
       hidden: true,
     },
   ],
+  complete_profile: [
+    {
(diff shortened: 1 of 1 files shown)


app/utils/achievementConstants.ts
```diff
@@ -124,6 +124,14 @@ const AchievementsInfo: Record<string, AchievementData[]> = {
       hidden: true,
     },
   ],
+  complete_profile: [
+    {
+      name: 'Komplet spiller',
```
jonasdeluna, line 129: *Komplett
is there any other better name for this in general though?

Arashfa0301: Do you have any suggestions? Otherwise, i kinda like it

jonasdeluna (approved):

PR #5383, opened 2025-02-26
Fix search in table

Bestem0r:
# Description

A fix for an edge case in BDB filtering accidentally broke inline search. This has been resolved while keeping the original fix intact.

# Result

### Search works

https://github.com/user-attachments/assets/0ce38b6b-7891-4797-b22f-6731a529c221

### BDB filtering works

Previously, changing semester and then editing the filter would cause the semester to reset due to synchronization issues between the internal table component filter and the external url filter.

https://github.com/user-attachments/assets/866eaf12-ec7a-47e7-b7fd-dcd123ad671f

# Testing

Verify that search and BDB filtering works.

---

Resolves [ABA-1307](https://linear.app/abakus-webkom/issue/ABA-1307/fix-searching-in-tables)

### DIFF
--- a/app/components/Table/index.tsx
+++ b/app/components/Table/index.tsx
@@ -2,7 +2,7 @@ import { Button, Flex, Icon } from '@webkom/lego-bricks';
 import cx from 'classnames';
 import { debounce, isEmpty, get, isEqual } from 'lodash';
 import { ChevronDown, ChevronUp } from 'lucide-react';
-import { useEffect, useMemo, useState, useRef, useCallback } from 'react';
+import { useEffect, useMemo, useState, useRef } from 'react';
 import InfiniteScroll from 'react-infinite-scroller';
 import BodyCell from './BodyCell';
 import HeadCell from './HeadCell';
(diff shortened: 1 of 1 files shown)


app/components/Table/index.tsx
```diff
@@ -106,11 +106,37 @@ const Table = <T extends { id: EntityId }>({
...
   useEffect(() => {
-    setFilters(queryFiltersToFilters(props.filters));
+    if (isUpdatingFromInternalChange.current) {
+      isUpdatingFromInternalChange.current = false;
+      return;
+    }
+
+    if (!isEqual(props.filters, prevPropsFilters.current)) {
+      prevPropsFilters.current = props.filters;
+      setFilters(queryFiltersToFilters(props.filters));
```
norbye, line 125: Just curious, have you tried with just this change? I'd imagine that even if the changes are internal or external the isEqual would avoid the endless looping (if that was the issue, that is)

norbye (approved):
Nice stuff!

Out of curiosity, did this bug present locally or just using ssr? Didn't think we found it when we made the change, or did we just not test it that thoroughly?

app/components/Table/index.tsx
Bestem0r: Good catch! Yes, it seems like the comparison is the only thing needed

app/components/Table/index.tsx
```diff
@@ -106,11 +106,26 @@ const Table = <T extends { id: EntityId }>({
   const [filters, setFilters] = useState<Filters>(
     queryFiltersToFilters(props.filters),
   );
+  const prevPropsFilters = useRef(props.filters);
 
   useEffect(() => {
-    setFilters(queryFiltersToFilters(props.filters));
+    if (!isEqual(props.filters, prevPropsFilters.current)) {
```
norbye, line 112: Have you tried to compare `props.filters` and `filters` instead of storing it in the useRef?

Bestem0r: Yeah that causes a whole bunch of "too many calls" errors and crashing the page with a "the operation is insecure" error message 🤷‍♂️

norbye: Hmm yeah that sounds annoying

PR #5387, opened 2025-02-27
Add Accordion option for removing children when closed

itsisak:
# Description

Adding an option to remove the Accordion children from the DOM when it is closed.

This is an option that may improve performance, but it may impact the smoothness of the animation if they are combined.

Using the approach with `useCallback` and `ResizeObserver` also fixes issues where the content of an accordion in other ways may be changing, such as loading images.

# Testing

---

Resolves ABA-1166 and ABA-1264

### DIFF
--- a/packages/lego-bricks/src/components/Accordion/index.tsx
+++ b/packages/lego-bricks/src/components/Accordion/index.tsx
@@ -1,5 +1,5 @@
 import cx from 'classnames';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
 import styles from './Accordion.module.css';
 
 type TriggerProps = {
@@ -13,7 +13,7 @@ type Props = {
   disabled?: boolean;
   defaultOpen?: boolean;
   animated?: boolean;
-  removeChildrenOnClose?: boolean;
+  persistChildren?: boolean;
   wrapperClassName?: string;
   triggerPosition?: 'top' | 'bottom';
   triggerComponent: React.ComponentType<TriggerProps>;
@@ -27,14 +27,13 @@ export const Accordion = ({
   disabled = false,
   defaultOpen = false,
   animated = true,
-  removeChildrenOnClose = false,
+  persistChildren = false,
   triggerPosition = 'top',
   wrapperClassName,
   triggerComponent: TriggerComponent,
   children,
 }: Props) => {
   const [open, setOpen] = useState(defaultOpen);
-  const [initialRender, setInitialRender] = useState(true);
   const [containerHeight, setContainerHeight] = useState(0);
   const [isTransitioning, setTransitioning] = useState(false);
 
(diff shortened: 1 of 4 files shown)


ch0rizo (approved):
Very good, very nice

packages/lego-bricks/src/components/Accordion/index.tsx
```diff
@@ -26,26 +27,35 @@ export const Accordion = ({
   disabled = false,
   defaultOpen = false,
   animated = true,
+  removeChildrenOnClose = false,
```
norbye, line 30: Any reason not to leave this as default true? Dunno if there are any places where it is particularly advantageous to keep it rendered when hidden?

packages/lego-bricks/src/components/Accordion/index.tsx
```diff
@@ -26,26 +27,35 @@ export const Accordion = ({
...
     if (initialRender && animated) {
       // setOpen is triggered through useEffect to ensure proper order of actions
       setInitialRender(false);
     } else {
       setOpen(!open);
     }
+    setTransitioning(true);
   };
 
   // Ensures that the open flag is triggered after the initialRender flag is changed
```
norbye, line 61: With this beautiful ResizeRender thingy, have you checked to see if the following useEffect and the initialRender thingys might not be needed any more?

packages/lego-bricks/src/components/Accordion/index.tsx
```diff
@@ -84,8 +99,9 @@ export const Accordion = ({
         style={{
           height: open ? openHeight : 0,
         }}
+        onTransitionEnd={() => setTransitioning(false)}
       >
-        <div ref={childrenWrapperRef}>{children}</div>
+        <div ref={childrenWrapperRef}>{renderChildren && children}</div>
```
norbye, line 104: Not sure if it makes much of a difference - but would it be neater to do the renderChildren check outside this div? Currently the unrendering would trigger a size change of the div, but if you hide the div in its entirety that might not be the case?

This is really minor though and just a random thought that just hit me

norbye (approved):
Awesome!

ivarnakken (approved):
Looks good!

packages/lego-bricks/src/components/Accordion/index.tsx
itsisak: I was a bit split on this. I tested it mostly with the readme container on the frontpage which contain some images. Whit `removeChildrenOnClose` and animation the opening animation was a bit wonky as the height of children changes.

Still I dont think its that bad, but I havent tested it much other places. I will look into it tomorrow probably.

If it seems to be generally fine I will change it to default to true and maybe just disable it for the readme container in particular.

packages/lego-bricks/src/components/Accordion/index.tsx
itsisak: I dont not think to check that to be honest. I just figured to not mess to much with the awesome code that was there before. 😉  I will see if I can clean up this component further.

packages/lego-bricks/src/components/Accordion/index.tsx
itsisak: I will test it and make the check outside the div if it doesn't make much difference. Adding the check there was simply the quickest and easiest when I made it. 😆

norbye (approved):
Sikkert snill
> If it works for you it works for me

PR #5393, opened 2025-03-03
Feat add filtering to achievement leaderboard

jonasdeluna:
# Description

Adds filtering to the leaderboard

This will be tested extensively in staging but local testing shows it works well.

# Result

image

> [!CAUTION]

# Testing

Tested locally.

---

Resolves ABA-1314

### DIFF
--- a/app/utils/achievementConstants.ts
+++ b/app/utils/achievementConstants.ts
@@ -198,7 +198,7 @@ const AchievementsInfo: Record<string, AchievementData[]> = {
       hidden: false,
     },
     {
-      name: 'Studentenes forkjemper',
+      name: 'Studentenes\nforkjemper',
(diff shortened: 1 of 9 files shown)


app/utils/achievementConstants.ts
```diff
@@ -198,7 +198,7 @@ const AchievementsInfo: Record<string, AchievementData[]> = {
       hidden: false,
     },
     {
-      name: 'Studentenes forkjemper',
+      name: 'Studentenes\nforkjemper',
```
itsisak, line 201: I would think this is better to achieve by restricting the size of the box with css instead.

itsisak (approved):
Haven't tested, but it looks good

PR #5401, opened 2025-03-05
Fix company creation form

ivarnakken:
# Description

The issue was student contacts that expect a company relation, but this does not exists while creating the company duh. Student contacts are now added in a different form, for some reason? This is still a fix ™️ 

# Testing

It works 👍🏼 and I'm able to add student contacts via the other form


jonasdeluna (approved):
:chart_with_upwards_trend:

PR #5402, opened 2025-03-05
Add LendingRequest redux boilerplate

eikhr:
# Description

To facilitate for easier parallel development on the lending system, I have added a bunch of redux boilerplate that is needed for some of the other tasks.

It is kinda broken because I have added features that are not yet added to the backend, but it doesn't matter as long as they are unused

# Testing

- [-] I have thoroughly tested my changes.

If something is broken, it can be fixed in the PR using the function.

---

Resolves ABA-1324


itsisak (approved):

ch0rizo (approved):
🤲

PR #5408, opened 2025-03-05
Fix Dockerfile not working without locally installed node_modules

eikhr:
# Description

I tested the Dockerfile with node_modules installed locally, making it work locally even though it copied the node_modules from the `setup`-step into the wrong directory.


Arashfa0301 (approved):
very nice

PR #5415, opened 2025-03-06
Fix docker healthcheck

eikhr:
# Description

Healthcheck was failing, making it impossible to deploy... Appearently the docker image needs curl in order to do the check

Also I accidentally broke the /healthz endpoint a while ago

I'm not 100% sure it works, so I will try deploying to staging before merging

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ COPY package.json yarn.lock ./
 COPY lego-webapp/package.json lego-webapp/package.json
 COPY packages packages
 
-RUN apk add curl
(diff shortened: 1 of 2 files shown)


jonasdeluna (approved):
:chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend: :chart_with_upwards_trend:

PR #5417, opened 2025-03-06
Migrate all routes from react-router to Vike routing

eikhr:
# Description

Time for another large refactoring PR! This one replaces `react-router` with the built-in file-system routing of Vike.

https://vike.dev/routing

It does have some downsides, but I think using the framework features will simplify our config and make maintanence easier. Downsides include that all pages are now called "+Page.tsx", making them hard to search for, and that optional url parameters require workarounds becuase Vike for some reason doesn't support them. Also Vike ignores "pages" directories when doing fs-routing, meaning our "pages"-url required some manual routing config.

Buuut, on the plus-side, we can now start using the other framework features of Vike like the +loader.ts files which can replace react-prepare! 
Oh! Another thing is that this improves the way we handle navigation in `lego-bricks` a ton. Instead of dynamically importing `react-router` it now supports using any routing framework with **any** navigation options, configured in `LegoBricksProvider`.

# TO ME LATER: Remeber to squash this one!

# Testing

---

Resolves ABA-1333


eikhr:
LinkButton is still good! Instead of Link we can now just use normal <a> tags:)

I also added a new Link-component that knows if it is currently active, like the NavLink from react-router

PR #5423, opened 2025-03-07
Fix publicly served abakus logo used in emails

eikhr:
# Description

The Abakus-logo image used in emails is fetched from https://abakus.no/185f9aa436cf7f5da598.png [(hardcoded for some reason)](https://github.com/webkom/lego/blob/9a862350574616afd9a1e4a26c8dc5a532ba428b/lego/templates/email/base.html#L228). I assume this was some hash generated by webpack. I added the logo in dark and light to the public assets, and a redirect from the old path.

---

Resolves ABA-1344


norbye (approved):
Looks good, though trying the vike approach might feel cleaner

norbye (approved):

PR #5425, opened 2025-03-08
Feat add feature flags

jonasdeluna:
# Description

This adds feature flags and i added one to lending.
Reloads is only required for the allowed part of this. The hook only part will still work. This is only due to fetchMeta not being called after editing so is really only a thing for admins. For regular users who did not refresh within the span of the activation, there is no way to force them to refetch meta either. The hook part will again still work. This is a minor edgecase.

Also improve design of sudo page

# Result

[Screencast_20250308_141059.webm](https://github.com/user-attachments/assets/2114fb1b-1e45-4240-b5e8-7d18a823d8d6)

image

> [!CAUTION]

# Testing

---

Resolves Resolves ABA-1345, ABA-1341

### DIFF
--- a/lego-webapp/components/Form/ObjectPermissions.tsx
+++ b/lego-webapp/components/Form/ObjectPermissions.tsx
@@ -71,7 +71,7 @@ const ObjectPermissions = ({
   ].filter(Boolean);
 };
 
-const toIds = (mapping) => mapping.value;
+export const toIds = (mapping) => mapping.value;
 
 export const normalizeObjectPermissions = ({
   requireAuth,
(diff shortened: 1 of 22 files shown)


ch0rizo:
faq-air-tsa-security-v2020-header-scaled (1)

ShaileshS1702 (approved):
LGTM!

norbye:
Dunno if this is the right place to request it - but have you considered adding some sort of visual indication to the navbar/wherever if an item is feature flagged? I feel like it is bound to happen that we end up feature flagging a part of the website and then forget that no one other than us can use it

jonasdeluna:
> Dunno if this is the right place to request it - but have you considered adding some sort of visual indication to the navbar/wherever if an item is feature flagged? I feel like it is bound to happen that we end up feature flagging a part of the website and then forget that no one other than us can use it

Hiding the navbar right now is done through the field "hide routes" which just sets the appropriate allowed value in the meta-info action in the backend. There is no info to the client that it is hidden due to a feature flag. Using the feature flags in the navbar itself turned out tricky, as it only let me do it if i checked every navbar item for a feature flag (conditional hooks).

jonasdeluna:
There is no special cases for admins as previously mentioned at the moment in the backend PR

PR #5427, opened 2025-03-09
Remove unnecessary variables.css imports

eikhr:
# Description

All the variables are imported in `global.css`, so they are available everywhere. The imports simply duplicated css into all the output css files. 

# Result

Smaller css bundles

# Testing

Things look the same, tested in dev and build.


itsisak (approved):
Very nice, very clean

PR #5431, opened 2025-03-10
Add recurring meetings

jonasdeluna:
# Description

Add weekly recurring meetings

# Result
Disclaimer:
The parent item will only be stuck at the top. This leads to the "duplicate" effect ONLY for the author of the recurring meetings so it does not get lost in pagination. I think this is sensible because it is both a meeting and a parent which will instantiate more meetings later on.

image

> [!CAUTION]
# Testing

---

Resolves Resolves ABA-385


lego-webapp/pages/meetings/MeetingEditor.tsx
```diff
@@ -81,6 +89,8 @@ export type MeetingFormValues = {
   description?: string;
   date?: [Dateish, Dateish];
   useMazemap: boolean;
+  isRecurring: boolean;
+  isTemplate?: boolean;
```
falbru, line 93:
```suggestion
  isTemplate: boolean;
```

falbru (approved):
LGTM

magnusbrecke (approved):
Productive feature! Nice!

PR #5436, opened 2025-03-12
Truncate compact elements dynamically

norbye:
# Description

Truncating the compact event titles by a fixed amount of characters makes it horribly static.

There are probably better ways to do this that depend less on hardcoded max-widths, but I believe this is an aight improvement for now.

The `6.5rem` are an approximated width of the colored circle and the date, and thus might have to be adjusted at some point if I'm incorrect - although it seems to work for march, which is one of the longer months, as it's not shortened to three letters.

`text-overflow: ellipsis;` appears to be well supported (https://caniuse.com/?search=text-overflow%3A%20ellipsis). In the case that you use a browser that does not support it, `max-width` and `overflow: hidden` should still be supported so you just won't get the "..." at the end.

# Result

- [x] **Changes look good with different viewports (mobile, tablet, etc.).**

> [!CAUTION]

## Before

https://github.com/user-attachments/assets/11fd9032-315a-4b76-83a5-077aa8292147

* Pay attention when the red dev-mode bar at the top shrinks. That means that the compact events are overflowing (their width is wider than the width of the screen)
* Also notice when the right column shrinks in size, which is due to the compact events taking up more space on the left column

## After

https://github.com/user-attachments/assets/c152f8bc-cd68-45cf-9431-7714d62ff3e2

* Notice how the titles are not truncated when there is enough space so you can see the whole thing

# Testing

---

### DIFF
--- a/lego-webapp/pages/index/CompactEvents.module.css
+++ b/lego-webapp/pages/index/CompactEvents.module.css
@@ -41,3 +41,32 @@ div.compactEvents {
   transform: rotate(-20deg);
   color: var(--lego-link-color);
 }
+
+/**
+ * Calculate max-width of the title based on the break-points for x columns
+*/
+
+.eventItemTitle {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  max-width: calc((var(--lego-max-width) / 3) - 4 * var(--spacing-md) - 6.5rem);
+}
+
+@media (--lego-max-width) {
+  .eventItemTitle {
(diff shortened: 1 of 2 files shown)


jonasdeluna:
Are the before and after swapped or something? I feel like they are not truncated in the before

jonasdeluna (approved):
Nice! 🔥📈💎👐🚀

Viljen789:
Cool functionality! Really like it!

ivarnakken (approved):
I like!

itsisak (approved):
Very nice!

PR #5442, opened 2025-03-13
Render MazemapEmbed once per page

ollfkaih:
# Description

Now the mazemap should load a maximum of once per page.
I thought of complicating this, then I saw persistchildren existed, rendering children on page load.
@eikhr already made mazemap.min.js @ 3,51MB preload, so downloading the rest at 0,5-1MB to have instant load of mazemap is worth it IMO. 
Otherwise we would need to refactor and wait to load the script until map is opened to have any real data saving effect anyways. Also, that big main js file would be cached in browser after loading any map so we'd be saving less than 1MB most of the time 🤷

# Result

> [!CAUTION]

<table>
    <tr>
        <td >Description</td>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Loads before open, and stays loaded when re-opening (if people do that)
        </td>
        <td>         <video width='20px' height='400px' src=https://github.com/user-attachments/assets/ff21ff01-4b28-4622-8bc0-93350dca55cc />        </td>
        <td>
<video height='400px' src=https://github.com/user-attachments/assets/3f61d0bb-363c-49eb-99ba-ffbc9d33d092 />
        </td>
    </tr>
</table>

# Testing

### DIFF
--- a/lego-webapp/components/MazemapEmbed/MazemapAccordion.tsx
+++ b/lego-webapp/components/MazemapEmbed/MazemapAccordion.tsx
@@ -1,8 +1,9 @@
 import { Accordion, Flex, Icon } from '@webkom/lego-bricks';
 import { ChevronRight } from 'lucide-react';
+import { Helmet } from 'react-helmet-async';
 import mazemapLogo from '~/assets/mazemap.svg';
 import MazemapLink from '~/components/MazemapEmbed/MazemapLink';
-import { MazemapEmbed } from '~/components/MazemapEmbed/index';
+import { MazemapEmbed, mazemapScript } from '~/components/MazemapEmbed/index';
 import styles from './MazemapAccordion.module.css';
 import type { ComponentProps } from 'react';
 
@@ -14,6 +15,7 @@ export const MazemapAccordion = ({ defaultOpen = false, ...props }: Props) => {
   return (
     <>
       <Flex column className={styles.mazemapAccordion}>
+        <Helmet title="Mazemap">{mazemapScript}</Helmet>
         <Accordion
           persistChildren
           defaultOpen={defaultOpen}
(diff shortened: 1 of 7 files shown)


jonasdeluna:
I might be misunderstanding but is this how it worked before the accordion changes? I thought it was changed to load as the accordion is expanded? I think making it load only once per page is nice though, how about making it load only when expanded but also once if possible?

jonasdeluna (approved):

itsisak:
> I might be misunderstanding but is this how it worked before the accordion changes? I thought it was changed to load as the accordion is expanded? I think making it load only once per page is nice though, how about making it load only when expanded but also once if possible?

Yes, this was the default behavior before, I thought I added `persistChildren` to all the Accordions that would benefit from it, but forgot this one 😅 I don't think making it only load when expanded is necessary or better. Using the `persistChildren` also has the effect of ensuring a smooth opening animation which is nice.

itsisak (approved):
Very nice!

lego-webapp/components/MazemapEmbed/MazemapAccordion.tsx
```diff
@@ -14,7 +15,9 @@ export const MazemapAccordion = ({ defaultOpen = false, ...props }: Props) => {
   return (
     <>
       <Flex column className={styles.mazemapAccordion}>
+        <Helmet title="Mazemap">{mazemapScript}</Helmet>
```
eikhr, line 18:
```suggestion
        <Helmet>{mazemapScript}</Helmet>
```
This sets the title of all meetings and events using mazemap to "Mazemap"
Screenshot 2025-03-14 at 19 35 17

eikhr (changes requested):

lego-webapp/components/MazemapEmbed/MazemapAccordion.tsx
eikhr: Same for "Mazemap Search"

PR #5447, opened 2025-03-17
Add gala achievement

aleksandervestlund:
* Primarily developed by @karolipb
* Add frontend support for gala achievement.
* The different colours are probably not properly configured.
* See backend PR webkom/lego#3789.


ivarnakken (approved):
Very cool feature! 🤩 However, I'm not a fan of calling it "gala" with 1 L when it's normally written with 2

@jonasdeluna can tell if the colors/rarity is wrong 🤷🏼‍♂️

PR #5453, opened 2025-03-18
Throttle theme changes instead of debouncing

itsisak:
# Description

Currently the implementation of theme changing uses debounce to prevent spamming the server. This results in slow feedback (toast) for the user. Using throttling instead it sends a request instantly but does not allow further updates until after a delay. 

This way it is still kind of possible to spam the server (only 2s between each request), if we want to eliminate that too I did try out a custom function which combines the functionality of throttling and debounce (instant feedback on first click, then debounce rest of calls), its okay, but i might prefer the user feedback with just throttle.

Also moved more logic to themeUtils, both to make this PR work and its more clean imo. 

The custom event allows updates to the theme to not send a API request to the backend and only be used for syncing local state.

Also anyone know if there is a reason for why there is no `Theme` `enum` with `dark` and `light`? A lot of "magic" strings. If not I could make that. 

# Result

*Before (with debounce):*

https://github.com/user-attachments/assets/a3d775dd-8e7f-4447-a8a0-8c65d5564642

*After (with throttlling):*

https://github.com/user-attachments/assets/ae897ee5-3c74-45ce-9934-24f52a36b859

# Testing

- [ I have thoroughly spammed the theme change button probably 100000000 times] I have thoroughly tested my changes.

---

Resolves ABA-1362

### DIFF
--- a/lego-webapp/pages/lending/@lendableObjectId/request/@lendingRequestId/+Page.tsx
+++ b/lego-webapp/pages/lending/@lendableObjectId/request/@lendingRequestId/+Page.tsx
@@ -47,7 +47,7 @@ type Params = {
 };
 
 const LendingRequest = () => {
-  const { query, setQueryValue } = useQuery({
+  const { query } = useQuery({
     fromAdmin: '',
   });
 
--- a/lego-webapp/components/Header/ToggleTheme.tsx
+++ b/lego-webapp/components/Header/ToggleTheme.tsx
@@ -1,11 +1,6 @@
 import { Icon } from '@webkom/lego-bricks';
 import cx from 'classnames';
-import { debounce } from 'lodash';
 import { MoonStar, Sun } from 'lucide-react';
-import { useCallback } from 'react';
-import { updateUserTheme } from '~/redux/actions/UserActions';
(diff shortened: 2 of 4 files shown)


ivarnakken (approved):
Very nice use of throttle! Good job!

lego-webapp/pages/lending/@lendableObjectId/request/@lendingRequestId/+Page.tsx
```diff
@@ -47,7 +47,7 @@ type Params = {
 };
 
 const LendingRequest = () => {
-  const { query, setQueryValue } = useQuery({
+  const { query } = useQuery({
```
Bestem0r, line 50: Great! Awesome work on this file 🤩

Bestem0r (approved):

lego-webapp/pages/lending/@lendableObjectId/request/@lendingRequestId/+Page.tsx
itsisak: Thank you so much! I spent 2 hours working on it!

PR #5457, opened 2025-03-19
Replace yarn with pnpm

eikhr:
# Description

It has very nice monorepo features, and is super fast!

# Result

All yarn commands are now replaced with pnpm (f.ex. `pnpm i` instead of `yarn install` and `pnpm dev:staging` instead of `yarn dev:staging`). The internal packages (lego-bricks) are now directly linked to lego-webapp, so any changes are reflected instantly and cause hot-reload.

# Testing

- I have verified that changes in `lego-bricks` are hot-realoaded in `lego-webapp`
- I'm like 90% sure the Dockerfile still works!

---

Resolves ABA-1192


norbye:
Please please please test this for a while in staging

PR #5458, opened 2025-03-19
Fix modal height

Bestem0r:
# Description

This has annoyed me for quite some time...
Previously, when zooming in using two finger pinch, the modal height would shrink. This has now been fixed in `lego-bricks`.

# Result

| Before | After |
| -- | -- |
| image | image |

# Testing

Resolves [ABA-1365](https://linear.app/abakus-webkom/issue/ABA-1365/incorrect-modal-height-on-zoom)

### DIFF
--- a/packages/lego-bricks/package.json
+++ b/packages/lego-bricks/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@webkom/lego-bricks",
-  "version": "1.3.6",
+  "version": "1.3.8",
   "description": "Component library for lego and other Abakus projects",
   "author": "webkom",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


eikhr (approved):
Looks very good!

norbye:
Have you tested how this works with mobile devices? Some times 100vh gets jiffy because it counts the areas that are under the navbar that dissapears when you scroll. Might be a usecase for 100svh or the other unit if it's an issue. At least I think that's why I originally switched from 100vh to that other unit

norbye:
And by testing on mobile devices I mean testing opening it on an actual phone, not the viewport in a desktop browser

Bestem0r:
> Have you tested how this works with mobile devices? Some times 100vh gets jiffy because it counts the areas that are under the navbar that dissapears when you scroll. Might be a usecase for 100svh or the other unit if it's an issue. At least I think that's why I originally switched from 100vh to that other unit

Seems to be working correctly on Chrome Android at least.

ch0rizo (approved):
🪨

PR #5460, opened 2025-03-19
Remove pnpm cache to fix docker build

eikhr:
# Description

Docker build worked on my machine, but not appearently not in ci... This fixes it


danielyanghansen (approved):
*Christoffer Voice* Looks good to me. Jeg bare godkjenner jeg

norbye (approved):
why not

PR #5470, opened 2025-03-20
Fix events page scrolling to top on fetch more

eikhr:
# Description

Whenever you pressed the "fetch more" button on the events page, it scrolled up to the navigation tabs. I fixed it:)

---

Resolves ABA-1351


norbye (approved):
LGTM

jonasdeluna (approved):
nice! This appears to also be an issue on the membership admin list but maybe this will also fix that?

ntnuitennis (approved):
Great job!
- ivar hehe

eikhr:
> nice! This appears to also be an issue on the membership admin list but maybe this will also fix that? *Nvm it happens there only occasionally idk why so it might not be a bug there after all

Yeah, I think this will fix the issue on all pages using the Tabs-components

PR #5476, opened 2025-03-21
Move react-meter-bar component into lego-bricks

eikhr:
# Description

For some reason have a separate repo for [react-meter-bar](https://github.com/webkom/react-meter-bar) which is just a single component used for password strength in lego-webapp. 

Screenshot 2025-03-21 at 12 47 12

I figured it would be simpler to just have it in `lego-bricks`.

# Result

It's now in lego-bricks.


Viljen789 (approved):
Awesome Pawesome

itsisak (approved):
very cool very good

PR #5487, opened 2025-03-25
Add event count to leaderboard

jonasdeluna:
# Description

Sorting is frontend only. But usually high event count => high achievement score so its probably fine. I think its okayish fast due to pagination and indices now. I dont think anyone outside the top 50 in achievement score are in the top 50 of event count.
Also add the option for achievementGroup to not be leveled to prepare for seasonal trophies like easter egg hunt or other ones that may come up later. This allows us to display only the owned trophy in color and the other ones (including lower levels) are still greyed out as you dont own them.

image

# Result

> [!CAUTION]

---

Resolves ABA-1385


jonasdeluna:
It's sort of unrelated but I feel like its something people might want to see and compare themselves to other people in. Do you think it should not be a feature at all or have its own leaderboard?

itsisak:
I have to agree with Ivar, this feels a bit out of place. I am also not sure I understood `isLeveled` properly. We currently handle trophies with only 1 item, such as the hidden trophies, did you want something different than that?

ivarnakken:
> Do you think it should not be a feature at all or have its own leaderboard? Grinding events only "pays off" in chunks of 25/50 so it may take some time to show results. I think having this could be like an incentive to keep grinding.

Users can see the count on their user profile, among the event history, where it belongs imo

jonasdeluna:
isLeveled is for trophies that follow the following:
isLeveled: boolean; // Does requirements(level2) => requirements(level1)

Do you think the event count should have its own leaderboard or not be visible at all?

jonasdeluna:
> > Do you think it should not be a feature at all or have its own leaderboard? Grinding events only "pays off" in chunks of 25/50 so it may take some time to show results. I think having this could be like an incentive to keep grinding.
> 
> Users can see the count on their user profile, among the event history, where it belongs imo

Yes but you cant see other peoples count, I think making it a competition can incentivize going to that extra event to climb another spot on the leaderboard. Like going from 125 events to 127 events which does not result in any trophies but might rank up on the leaderboard

ivarnakken:
> Yes but you cant see other peoples count, I think making it a competition can incentivize going to that extra event to climb another spot on the leaderboard. Like going from 125 events to 127 events which does not result in any trophies but might rank up on the leaderboard

I see. What about implementing some sort of "score"/"xp"? 🤩 Users gain xp (a higher score) by attending events, getting trophies, posting overhørts etc. etc. Perhaps that gives us a reason to remove the "amount of events"-trophy? Would then be cool with a trophy (different ranks/levels) for the amount of xp a user has

jonasdeluna:
I think people would be upset if we remove the event count trophy as its probably the most grinded after right now but having a score sounds like a good idea where we can give more customizations as rewards.

ivarnakken:
> I think people would be upset if we remove the event count trophy as its probably the most grinded after right now but having a score sounds like a good idea where we can give more customizations as rewards.

I don't think it's "the most grinded" one atm, but I believe some sort of score system would be able to replace it as something you gain xp from. Perhaps genfors and gala can give extra xp? And in that way get rid of that as well? At least if we had concerns about "too many" achievements being added.

PR #5493, opened 2025-03-26
Make the blog more visible

Bestem0r:
# Description

As discussed in a meeting, this PR aims to improve the visibility of our blog. This has been done by including a link on the authenticated front page as well as in the footer. 

# Result
image

# Testing

Resolves [ABA-1280](https://linear.app/abakus-webkom/issue/ABA-1280/make-utviklarbloggen-more-visible-at-abakusno)

### DIFF
--- a/lego-webapp/components/Footer/index.tsx
+++ b/lego-webapp/components/Footer/index.tsx
@@ -52,6 +52,17 @@ const Footer = () => {
             >
               Backend
             </a>
+            <Circle
+              size="var(--spacing-xs)"
+              color="var(--color-absolute-white)"
+            />
+            <a
+              href="https://webkom.dev"
+              rel="noopener noreferrer"
+              target="_blank"
+            >
+              Utviklerbloggen
+            </a>
           </Flex>
         </div>
 
--- a/lego-webapp/components/Search/utils.tsx
+++ b/lego-webapp/components/Search/utils.tsx
@@ -30,6 +30,7 @@ import {
   ShoppingCart,
   Accessibility,
   Trophy,
+  MessageSquareQuote,
 } from 'lucide-react';
 import ReadmeLogo from '~/components/ReadmeLogo';
 import { Tag } from '~/components/Tags';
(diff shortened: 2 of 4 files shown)


Viljen789 (approved):
Looks good 👍

ivarnakken (commented):
It looks good, but I'm not sure about cluttering up the frontpage with more stuff. This is not really an important link, and far less important than links which we have hidden away in the menu. Imo, this should only be in the menu.

falbru:
I agree with @ivarnakken. Perhaps it could be moved to the footer?

eikhr:
I agree it would make more sense in the footer, perhaps together with the github links. Although that decreases the visibility a lot

eikhr:
Nevermind, I just noticed you did add it to the footer too! Looks good there:)

jonasdeluna:
I feel like we should change it to something else than "endringslogg", as its a dev blog and we can write more stuff there than just changelogs.

jonasdeluna (approved):
Nice but please change endringslogg to utviklerblogg or devblog

ivarnakken:
> Nice but please change endringslogg to utviklerblogg or devblog

Heavy ➕

The nynorsk is kinda cringe now

Bestem0r:
> It looks good, but I'm not sure about cluttering up the frontpage with more stuff. This is not really an important link, and far less important than links which we have hidden away in the menu. Imo, this should only be in the menu.

Yeah that's a good point. I'll move the link to the menu

Bestem0r:
> I feel like we should change it to something else than "endringslogg", as its a dev blog and we can write more stuff there than just changelogs. "Utviklerblogg" is more descriptive.

Yeah true. I was thinking that "endringslogg" made more sense in the footer as it is placed next to the repo links. I'll change it though.

falbru (changes requested):
The footer looks good, but I don't like it on the frontpage. Could we discuss this further?

falbru (approved):
Awesome 😍

PR #5500, opened 2025-03-26
Prevent ToggleButton from submitting form

eikhr:
# Description

IDK why, but it somehow the ToggleButtom becomes a type="submit" button when used in forms. Overriding it fixes the behaviour.

# Result

It no longer submits forms when toggling the button.

# Testing

I have tried it in the Joblisting form which used to have the bug.

---

Resolves ABA-1331


jonasdeluna (approved):
Nice

ch0rizo (approved):
Huge 🪨

PR #5505, opened 2025-03-26
Use lodash-es instead of lodash

eikhr:
# Description

This does the same as #5097 just that this time it actually works because we use vite. It fixes typescript types on all lodash imports, and reduces dependence on the cjs compatibility plugin for vite.

# Result

Cleaner config and more types. ~100 typescript errors are resolved:)

# Testing


jonasdeluna (approved):
Very nice!

ivarnakken (approved):

packages/lego-bricks/package.json
```diff
@@ -72,7 +72,7 @@
     "@storybook/react": "^8.6.4",
     "@storybook/react-vite": "^8.6.4",
     "@storybook/test": "^8.6.4",
-    "@types/express": "^5.0.0",
+    "@types/express": "^4.17.21",
```
itsisak, line 75: Why does this need to be a lower version?

itsisak (approved):
Looks good!

Bestem0r (approved):

packages/lego-bricks/package.json
eikhr: It was wrong, we use express v4. Doesn't seem like it mattered much, but best to use correct version of types:)

PR #5522, opened 2025-03-31
Fix issue with image file-key when editing with lego-editor

eikhr:
# Description

Editing editor-content with images didn't work because of a mistake with parsing the `data-file-key` needed by the backend.
This fixes the parsing, so that it is preserved when editing content containing images.

# Testing

Tested editing events in dev mode, and verified that the `data-file-key` is correct after editing the event.


jonasdeluna (approved):
Nice

PR #5525, opened 2025-04-01
Improve image support in lego-editor

eikhr:
# Description

Since the new editor was released, we have received a few complains about the missing caption-feature on images. This PR re-adds it as an optional toggle on images, via a bubble-menu (see below).

It also centers images on the page, which generally looks better:)

# Result

Image when selected in lego-editor:
Screenshot 2025-04-01 at 15 03 27

After clicking the caption button:
Screenshot 2025-04-01 at 15 03 58

- [x] Changes look good on both light and dark theme.
- [x] Changes look good with different viewports (mobile, tablet, etc.).
- [x] Changes look good with slower Internet connections.

> [!CAUTION]
> Make sure your images do not contain any real user information.

<table>
    <tr>
        <td width="25%">Description</td>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Article with images
        </td>
        <td>
            
Screenshot 2025-04-01 at 14 46 33
        </td>
        <td>

Screenshot 2025-04-01 at 14 46 03
        </td>
    </tr>
</table>

# Testing

- [x] I have thoroughly tested my changes.

I have tested both in lego-editor dev and in lego-webapp on staging. So I'm pretty sure it works:)

---

Resolves ABA-1407

PR #5526, opened 2025-04-01
Add tooltips to lego-editor toolbar buttons

eikhr:
# NB! Builds on #5525 

# Description

Some of the toolbar buttons are not entirely self-explanatory based on the icons used. This adds tooltips to explain them further.
I also moved `<Tooltip>` into `lego-bricks` in order to be able to use it in `lego-editor`. This also required adding support for keyboard interaction with the Tooltip component, in the case that it uses `onClick` because of stricter linting in `lego-bricks`:)

# Result

- [ ] Changes look good with different viewports (mobile, tablet, etc.).
  - Our tooltips don't work great on mobile, but that's not directly related to this PR

Looks like this in toolbar:
Screenshot 2025-04-01 at 16 31 02

And like this in image menu:
Screenshot 2025-04-01 at 16 33 15

# Testing

---

Resolves ABA-1408


ivarnakken (approved):
Looks good!

PR #5527, opened 2025-04-02
Fix redirect issue by bumping vike-server

eikhr:
# Description

See title. There was a bug where redirects didn't work.


norbye (approved):
sweet

PR #5532, opened 2025-04-02
Use mazemap from CDN and persist accordion content

eikhr:
# Description

Removes mazemap package, and uses the script and stylesheet tags from mazemap CDN directly as they suggest in the [docs](https://api.mazemap.com/js/v2.2.1/docs/#intro).

Also makes the accordion mazemap persistent like in #5442 

# Result

- Simpler setup as we don't need `mazemap` package:)
- Smoother opening of accordion (videos from #5442, but should look the same)
<table>
    <tr>
        <td >Description</td>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
            Loads before open, and stays loaded when re-opening (if people do that)
        </td>
        <td>         <video width='20px' height='400px' src=https://github.com/user-attachments/assets/ff21ff01-4b28-4622-8bc0-93350dca55cc />        </td>
        <td>
<video height='400px' src=https://github.com/user-attachments/assets/3f61d0bb-363c-49eb-99ba-ffbc9d33d092 />
        </td>
    </tr>
</table>

# Testing

I have tested all mazemap usages in dev. Meetings and events both detail page and editors.

Closes #5442


Viljen789:
Sigma!

Viljen789 (approved):

lego-webapp/components/MazemapEmbed/index.tsx
```diff
@@ -17,7 +16,7 @@ type Props = {
 /** A component that shows a mazemap map of a given poi (e.g. room),
  * largely based on https://api.mazemap.com/js/v2.0.63/docs/#ex-data-poi
  *
- * REQUIRED: {mazemapScript} must be included in the pages <Helmet> component
+ * REQUIRED: {mazemapDeps} must be included in the pages <Helmet> component
```
ShaileshS1702, line 19: I am not sure, but is this comment needed when the helmet component have been removed?

eikhr: Good question. The `<Helmet>` component right under here actually didn't work (I think there is max 1 Helmet per page), but we still need the script that was in the `<Helmet>` to make mazemap work. 
Instead we need to add the script to the page-component's existing `<Helmet>` (so that we only use 1 helmet component). The comment here is meant to explain that we need to remember to add `mazemapDeps`, which includes the script and a stylesheet, to the `<Helmet>` component of any page that uses mazemap.

jonasdeluna (approved):
Nice, is this their own cdn or something? Did not seem familiar

eikhr:
Maybe CDN is the wrong term, this gets it from Mazemap's own servers

PR #5538, opened 2025-04-03
Fix dependency installation command in README

eikhr:
# Description

In the previous version of the README (ever since swapping out yarn for pnpm) there was a mistake in the command for installing dependencies. This command was just `pnpm` instead of `pnpm i` or `pnpm install`. This is likely because with yarn you could installing using simply `yarn`.

# Result

In this PR I have updated the command in `README.md` to reflect the correct usage of the pnpm installation command `pnpm i`.

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
           Quick start section of README.md
        </td>
        <td>
Screenshot 2025-04-03 at 13 44 49
        </td>
        <td>
Screenshot 2025-04-03 at 13 45 13
        </td>
    </tr>
</table>

# Testing

Tried running `pnpm`, noted how it didn't install depenencies. After updating the README, I tried running the new command `pnpm i`, and verified that it now runs the installation of dependencies using pnpm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 > For all the alternative ways to run the project, you will need to run `pnpm build` to compile [lego-bricks](#lego-bricks) the first time you run the project. After this, it will build automatically when changed.
 
 ```bash
-$ pnpm # Install dependencies
+$ pnpm i # Install dependencies
 $ pnpm build # Compile LEGO-BRICKS - only required the first time you run the project
 $ pnpm dev:staging # Start webserver with development backend
 ```


Bestem0r (approved):
Great!

norbye (approved):
Appreciate that you (reportedly) tested this in dark mode and multiple screen sizes<3

itsisak (changes requested):
i think this is easier to understand

itsisak (approved):
Hmm, okay then, even though it might be more confusing as to what `i` means.. 🤔

ch0rizo (approved):
Huge change! 🪨

PR #5542, opened 2025-04-04
Add support for new notification websocket message

eikhr:
# Description

Related backend pr: https://github.com/webkom/lego/pull/3825

Updates the redux state with the newly received notification when receiving the new websocket message implemented in the related backend PR. This makes the notification thing in the header update live when receiving notifications.

# Result

https://github.com/user-attachments/assets/7e149374-53ba-44bc-bb20-945519c901ef

# Testing

As shown in the video, I have tested that the notifications appear correctly when the websocket message is received.

---

Resolves ABA-1420


jonasdeluna (approved):
Nice but i dont understand how can we directly upsert the data i thought it was a plain text string or is there some magic happening in lego adapter or somewhere else

Viljen789 (approved):
Sigma

PR #5549, opened 2025-04-09
Fix custom media query variables in lego-bricks

eikhr:
# Description

Custom media query variables allow us to write `@media (--medium-viewport)` instead of `@media (max-width: 45em)`, this fixes them in `lego-bricks`.

It is kind of hacky, but I import the media query definition from `lego-webapp` into `lego-bricks` using the file path `'../../../lego-webapp/styles/custom-media.css'`. This is the same way we import global variables so I think it's ok.

# Result

No visual changes, just defines all media query values in one place.

# Testing

They all still work after building.

---

Resolves ABA-965

### DIFF
--- a/packages/lego-bricks/src/components/Card/Card.module.css
+++ b/packages/lego-bricks/src/components/Card/Card.module.css
@@ -1,3 +1,5 @@
+@import '../../custom-media.css';
+
 .baseCard {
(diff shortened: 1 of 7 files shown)


ivarnakken (approved):
Awesome

jonasdeluna (approved):

PR #5575, opened 2025-05-07
FIX: font difference mobile

ch0rizo:
# Description

Countdown banner time font is different on mobile compared to laptops

Resolves ABA-1460


jonasdeluna (approved):
nice

PR #5681, opened 2025-08-27
Fix incrementing metrics bug

itsisak:
# Description

If you went in to the statistics page multiple times without refreshing it would double the metrics values. This was due to the initial value being updated. 

# Testing

I went in and out of the statistics page multiple times and it now does not increment it like it used to :)

---

Resolves ABA-1485


falbru (approved):
Awesome!

PR #5698, opened 2025-09-09
Improve trophies on profile

falbru:
# Description

- Remove Topplister button (I think we should just have it in the menu)
- Make the trophies area collapsible

# Result

**before**

image

**after**

[Screencast From 2025-09-09 16-56-21.webm](https://github.com/user-attachments/assets/c769011f-9d9d-4f12-83c0-1f1b245ceae5)

# Testing

---

Resolves ABA-1516


lego-webapp/components/Collapsible/Collapsible.tsx
itsisak, line 1: Why not using `Accordion` from `lego-bricks`?

itsisak (approved):
This looks way cleaner!!

lego-webapp/components/Collapsible/Collapsible.tsx
itsisak: The accordion is unstyled. It only takes in a `triggerComponent` meaning that could be pretty much anything. You can also set the position of said component at the bottom so it. 

In this exact scenario the bottom link is a part of the accordion content.

The code is pretty close to your `Collapsible` component code aswell, just a slight different approach so that the `Accordion` also removes the content from the DOM when it is not expanded by default. Imo that is better for performance (not that it may matter much in this scenario so if it is too buggy just keep it always rendered)

falbru: Idk I like the design where the trophies are truncated and only some are visible, and you can expand it to see all (like it was with the previous design, but better). With the Accordion, you'd only see the trigger component when it's closed and I'm not sure what the trigger component would be so that it'd look nice

PR #5703, opened 2025-09-11
Remove anonymous option for contact forms

itsisak:
# Description

Removes the option for sending anonymous contact forms.

This could be an issue if this is being linked outside of abakus.no intended for externals to use to contact abakus as it is now not possible to submit the contact form without being logged in. The path `abakus.no/contact` seems like a typical entry point for externals to contact abakus on if they dont know of other ways. 

If this contact form is only intented for abakus users anyways it should be okay. 

# Testing

---

See related backend pr: webkom/lego#3842

Resolves ABA-1496


itsisak:
Page when not logged in:

Screenshot 2025-09-15 at 15 38 45

Viljen789 (approved):
Looks good! Nice change

PR #5707, opened 2025-09-13
fix: Refactor Reactions to use react-aria

itsisak:
# Description

Refactored the `Reactions` component to use components from `react-aria-components` instead of the custom logic we had. This results in a more robust and proper solution for handling popovers.

Also added a state in redux indicating wheter emojis have previously been fetched. If they have it will not try to fetch them again. This prevents multiple instances of the reaction picker from fetching emojis again and displaying the loading indicator. This way the picker rarely loads. However the cost of this is that the emojis state might become stale at least until redux is refreshed.

# Result

- It prioritizes opening above the trigger component, but if there is not room it will open where there is
- It will never be cropped due to overflow as the popover now works like a `Dialog` existing entirely outside of the root
- It is not possible to have multiple pickers open 
- It is not possible to interact with other elements while the picker is open
- Uses built in functionality for detecting hover

**Before**

https://github.com/user-attachments/assets/d0fe01d3-754e-42ac-a5a7-c717bc045216

**After**

https://github.com/user-attachments/assets/03fe6217-c9a7-40b7-b334-7316f61a0f47

# Testing

---

Resolves ABA-1514

### DIFF
--- a/lego-webapp/components/Reactions/index.tsx
+++ b/lego-webapp/components/Reactions/index.tsx
@@ -1,14 +1,15 @@
-import { Flex, Icon } from '@webkom/lego-bricks';
+import { Icon } from '@webkom/lego-bricks';
 import cx from 'classnames';
 import { SmilePlus } from 'lucide-react';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useState } from 'react';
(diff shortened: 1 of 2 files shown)


ch0rizo (approved):
jævla nice da!

PR #5729, opened 2025-09-30
fix: change gala event color to pink

sondrejk:
# Description
Just a small color change made for an email request from Arrkom.


ch0rizo (approved):
lgtm, but can't we standarise the colors? Like with variables?

viljeno (approved):
LGTM!

PR #5731, opened 2025-09-30
Prevent updating pools after registrations

tindevw:
# Description

Ensures pool cannot be updated to include additional registration groups once registrations start, instead a new pool must be created. Prevents unintended complications caused by modifying pools mid-registration. 

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
           Disables field
        </td>
        <td>
Screenshot 2025-09-30 at 22 08 29
        </td>
        <td>
Screenshot 2025-09-30 at 22 08 14
        </td>
    </tr>
</table>

# Testing

Resolves ABA-1541


falbru:
It'd also be nice with e.g. a tooltip to inform the user why the fields are disabled, and that the user should create a new pool instead

lego-webapp/pages/events/src/EventEditor/renderPools.tsx
```diff
@@ -64,6 +64,7 @@ const renderPools = ({ fields, startTime, eventStatusType }: poolProps) => (
           component={DatePicker.Field}
         />
         <Field
+          disabled={fields.value[index].registrations?.length > 0}
```
itsisak, line 67: What about if they want to update a pool which has opened but gotten zero registrations?

itsisak: I agree that that is probably not an issue, my only concern about it is that it might seem like a confusing requirement and that it might be easier to enforce it based on whether or not the pool has opened.

tindevw:
> Seems like your initial solution was a simpler fix... 😅 I see now also that there is a slight benefit to be able to set the activation time in the past which is that if you want the pool to already be activated when creating the event, so you select current time but then when clicking create event, time as passed (big surprise) and it will fail..

Yes, I see the issue. I'll look into a better fix for it.

PR #5750, opened 2025-10-15
fix wierd hover effect in footer

celiacisternas:
# Description

fixed hover over "av Webkom", so it doesn't show hovering when hovering over LEGO.

---

Resolves ABA-1575


Viljen789 (approved):
Nice change!

jonasdeluna (approved):
Awesome! ✈️

PR #5795, opened 2025-11-09
Prevent admin functionality on own lendingRequest

itsisak:
# Description

A user should not have any admin functionality on their own requests. This should be fixed in backend too as currently only "approve" is not allowed on own requests.

Also just centering content of statuspill.

# Result

> [!CAUTION]

Screenshot 2025-11-09 at 14 39 30


itsisak:
> Design question: Should it have some background-color of some red/orange color to indicate "you are not allowed"?

Do you mean to color the text? Or like use a red info box, sort of like on events if you have unanswered surveys?

Imo, i think this looks the most clean and less "intrusive", I even considered just removing the whole admin section in this case, but figured people might be a bit confused as to why it might suddenly dissapear

PR #5798, opened 2025-11-13
Background snowflakes

AstridMS5:
# Description

Snowflakes in the background on the front page

# Result

> [!CAUTION]

[Skjermopptak fra 2025-11-13 16-41-13.webm](https://github.com/user-attachments/assets/ca920958-09ae-4ec3-a857-81e8316aed57)

# Testing

---

Resolves ABA-1594


TonyWorep:
> Very nice! How does it look on light mode with white background?

@ch0rizo You don't see the snowflakes at all. We have had the "sucks to suck" mentality for those who use light mode 😝. On the serious note, we are not sure if there are colours other than white that make them snowflakes, if that makes sense. If we find a good colour in light mode then we can implement it so that it will be visible in light mode as well.

PR #5817, opened 2025-12-18
Increase heap space for build

ch0rizo:
# Description

Explicitly increase node heap space and hide sourcemapping (sentry should still have sourcemap)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -161,6 +161,7 @@ steps:
   - name: build
     image: node:20-alpine
     environment:
+      NODE_OPTIONS: --max-old-space-size=4096
     when:
       event: [push]
       branch:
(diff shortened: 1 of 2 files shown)


.drone.yml
itsisak: I think you need to create a new signature for this file whenever you make changes to it so that you wont have to explicitly accept each job.

https://docs.drone.io/signature/

.drone.yml
```diff
@@ -271,7 +273,7 @@ steps:
       - npm i -g corepack
       - corepack enable
       - pnpm --filter lego-webapp exec cypress run --e2e --record
-  
+
```
Viljen789: Diff farming 🔥

PR #5863, opened 2026-02-10
add lending to event creating/editing

### LINKED ISSUE #5862: Add lending to events
Pr wanted the possibility of reserving a lending object from the event creation page

ShaileshS1702:
# Description

Add the possibility of lending an object directly from the event edit form

# Result

> [!CAUTION]

<table>
    <tr>
        <td>Before</td>
        <td>After</td>
    </tr>
    <tr>
        <td>
Dark mode
        </td>
        <td>
            ...
        </td>
        <td>
image        </td>
    </tr>
    <tr>
        <td>
Dark mode with lending objects chosen
        </td>
        <td>
            ...
        </td>
        <td>
image

     </td>
    </tr>
</table>

# Testing

---

Resolves #5862


ch0rizo (approved):
Really cool feature!

lego-webapp/pages/events/src/EventEditor/index.tsx
```diff
@@ -369,19 +387,18 @@ const EventEditor = () => {
...
-
             <EditorSection title="Detaljer" initiallyExpanded={!isEditPage}>
               <Details values={values} />
             </EditorSection>
-
             <EditorSection title="Påmelding" initiallyExpanded={!isEditPage}>
               <Registration values={values} />
             </EditorSection>
-
+            <EditorSection title="Utstyrslån" initiallyExpanded={false}>
```
ch0rizo: Should this maybe be set to `true`?

ch0rizo (commented):
M

jonasdeluna (approved):
This is very cool! Could we have some relation between the event and requests, so that we can see/filter on the event page or in the ending page based on events? The event could for example store its own lending requests?
Could be a separate more extensive feature though

lego-webapp/pages/events/src/EventEditor/index.tsx
```diff
@@ -206,21 +209,40 @@ const EventEditor = () => {
...
+        if (values.saveToImageGallery) {
+          void dispatch(setSaveForUse(key, token, true));
+        }
+
+        // Create lending request if lendingObjects exist and we're creating a new event
+        if (values.lendingObjects && values.lendingObjects.length > 0) {
+          for (let i = 0; i < values.lendingObjects.length; i++) {
+            const lendingRequestData: CreateLendingRequest = {
+              lendableObject: values.lendingObjects[i].value,
+              comment: values.lendingDescription[i] || '',
```
ch0rizo, line 229:
```suggestion
              comment: values.lendingDescription?[i] || '',
```

Crash when an object is selected but no comment typed. Here you fix it with a default empty string. However, `lendingDescription` is not in the `initialValues`. If the user selects a lending object, but never types into any comment field, `values.lendingDescription` is `undefined -> underfined[i]` throws on submit. Basically since `values.lendingDescription` is an array it will fail before hitting the default `''`

lego-webapp/pages/events/src/EventEditor/index.tsx
```diff
@@ -206,21 +209,40 @@ const EventEditor = () => {
...
+
+    eventAction
+      .then((res) => {
+        const key: string = values.cover.split(':')[0];
+        const token: string = values.cover.split(':')[1];
+        if (values.saveToImageGallery) {
+          void dispatch(setSaveForUse(key, token, true));
+        }
+
+        // Create lending request if lendingObjects exist and we're creating a new event
```
ch0rizo, line 224: I am worried that this will create a lending request on editing the event aswell. Because there are no difference between edit and create form. There should gated with isEditPage.

lego-webapp/redux/actions/LendableObjectActions.ts
```diff
@@ -23,6 +24,20 @@ export const fetchAllLendableObjects = () =>
...
+) =>
+  callAPI<EntityId[]>({
+    types: LendableObjects.FETCH_AVAILABLE,
+    endpoint: '/lending/objects/available/',
+    query: { start_date, end_date },
+    meta: {
+      errorMessage: 'Henting av tilgjengelige utlånsobjekter feilet',
+    },
+    propagateError: true,
+  });
```
ch0rizo, line 39: Availability fetch keys off startTime/endTime which is never change in the form (LendingSection.tsx). The form's date field is `name="date"` (a range) -> values.date[0]/values.date[1]. There is no date-driven startTime/endTime on form values. This means:
- On create: startTime/endtime are undefined -> always hits the clear branch, availability is never fetched.
- On edit: they hold the stale initial values -> fetched once with the original dates, never updates when the user picks a new range. 

However `onSubmit` correctly uses  `values.date[0]/values.date[1]` for the request - so the two halves of the feature disagree. A possible solution is to switch the availability effect to `values.date?.[0] / values.date?.[1]`

lego-webapp/pages/events/src/EventEditor/EditorSection/LendingSection.tsx
```diff
@@ -0,0 +1,191 @@
...
+    selectLendableObjectsUnavailable,
+  );
+
+  usePreparedEffect(
+    'fetchAllLendableObjects',
+    () =>
+      dispatch(fetchAllLendableObjects()).catch((err) => {
+        // Log to aid debugging; slice will still mark fetch failure.
+        // eslint-disable-next-line no-console
+        console.error('fetchAllLendableObjects failed', err);
```
ch0rizo, line 46: I would not leave this in prod. If errors is to be announced it should be through user friendly toasts

lego-webapp/pages/events/src/EventEditor/EditorSection/LendingSection.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+      timeout = setTimeout(() => {
+        dispatch(
+          fetchAvailableLendableObjectIdsByDate(
+            values.startTime,
+            values.endTime,
+          ),
+        ).catch((err) => {
+          // eslint-disable-next-line no-console
+          console.error('fetchAvailableLendableObjectIdsByDate failed', err);
```
ch0rizo, line 71: Here as well :)

lego-webapp/redux/actions/callAPI.ts
```diff
@@ -269,6 +269,9 @@ export default function callAPI<
       requestOptions,
     );
 
+    // Attach a requestId to meta for request identity (used to ignore stale responses)
+    const requestId = `${Date.now()}-${Math.random().toString(36).slice(2, 9)}`;
```
ch0rizo, line 273: This is a global change. callAPI.ts is a the foundation of every redux API action in the app. This will add requestId to every action's meta.

Maybe reconsider making a global change for a local need. A proposed solution would be to solve the stale-response handling locally in LendingSection, without touching callAPI at all. The thunk from `dispatch(fetchAvailable…)` already returns a promise - you can await it and compare against the latest dispatched date range (or keep the "latest request" marker in a useRef (probably not a ideal🤷‍♂️)).

lego-webapp/redux/actionTypes.ts
```diff
@@ -88,6 +88,7 @@ export const Joblistings = {
 
 export const LendableObjects = {
   FETCH: generateStatuses('LendableObject.FETCH_ALL'),
+  FETCH_AVAILABLE: generateStatuses('LendableObject.FETCH_AVAILABLE'),
   FETCH_AVAILABILITY: generateStatuses('LendableObject.FETCH_AVAILABILITY'),
```
ch0rizo, line 92: What is the difference? Should there be a single source of truth, if they have the same purpose

lego-webapp/redux/slices/lendableObjects.ts
```diff
@@ -11,27 +11,82 @@ const legoAdapter = createLegoAdapter(EntityType.LendableObjects);
...
+            return;
+          }
+          state.availabilityFetchFailed = false;
+          state.availableIds = action.payload;
+        },
+      );
+      addCase(
+        LendableObjects.FETCH_AVAILABLE.FAILURE,
+        (state, action: AnyAction) => {
+          const requestId = action.meta?.requestId ?? null;
```
ch0rizo, line 63: Consider rewriting this if you didn't change `callAPI.ts`

packages/lego-bricks/src/components/ImageUpload/index.tsx
ch0rizo, line 1: Just curious what is this for?

lego-webapp/redux/actionTypes.ts
ch0rizo: Its a bit confusing with the naming convention. Seems one is for bulk availability and the other one for a specific one

ch0rizo (changes requested):
I think this a nice and useful feature!! There are some bugs and fixes I would take a closer look.

PR #5937, opened 2026-04-13
Add lending request archiving

Bestem0r:
# Description

Adds archiving of lending requests to clear up inbox.

[Backend PR](https://github.com/webkom/lego/pull/4114).

# Result

https://github.com/user-attachments/assets/e8ab0b85-238a-4ec5-8d51-7d3714d3ee46

# Testing

### DIFF
--- a/lego-webapp/pages/lending/+Page.tsx
+++ b/lego-webapp/pages/lending/+Page.tsx
@@ -5,7 +5,10 @@ import { HeartHandshake } from 'lucide-react';
 import { useEffect, useState, useRef } from 'react';
 import { Helmet } from 'react-helmet-async';
 import { fetchAllLendableObjects } from '~/redux/actions/LendableObjectActions';
-import { fetchLendingRequests } from '~/redux/actions/LendingRequestActions';
+import {
+  editLendingRequest,
+  fetchLendingRequests,
+} from '~/redux/actions/LendingRequestActions';
 import { useAppDispatch, useAppSelector } from '~/redux/hooks';
 import { EntityType } from '~/redux/models/entities';
 import { selectLendableObjectsForIndex } from '~/redux/slices/lendableObjects';
@@ -16,26 +19,35 @@ import useQuery from '~/utils/useQuery';
 import FilterSearch from './FilterSearch';
 import ItemIndex from './ItemIndex';
 import styles from './LendingPage.module.css';
-import RequestInbox, { type LendingRequestOrdering } from './RequestInbox';
+import RequestInbox, {
+  type LendingRequestArchivedFilter,
+} from './RequestInbox';
 import {
   REQUEST_INBOX_PAGE_SIZE,
   getNextVisibleCount,
   getVisibleRequestCount,
   shouldFetchMoreRequests,
 } from './requestInboxPagination';
+import type { TransformedLendingRequest } from '~/redux/models/LendingRequest';
 
 const defaultLendingQuery = {
   search: '',
   lendingCategories: [] as FilterLendingCategory[],
-  ordering: '-created_at' as LendingRequestOrdering,
+  archived: 'false' as LendingRequestArchivedFilter,
 };
 
 const LendableObjectList = () => {
   const { query, setQueryValue } = useQuery(defaultLendingQuery);
-  const requestOrdering: LendingRequestOrdering =
-    query.ordering === 'created_at' ? 'created_at' : '-created_at';
+  const requestArchived: LendingRequestArchivedFilter =
+    query.archived === 'true' ? 'true' : 'false';
   const requestQuery = {
-    ordering: requestOrdering,
+    archived: requestArchived,
+  };
+  const activeRequestQuery = {
+    archived: 'false' as const,
+  };
+  const archivedRequestQuery = {
+    archived: 'true' as const,
   };
 
   const dispatch = useAppDispatch();
@@ -56,7 +68,7 @@ const LendableObjectList = () => {
           query: requestQuery,
         }),
       ),
-    [requestOrdering],
+    [requestArchived],
   );
 
   const { pagination: requestsPagination } = useAppSelector((state) =>
@@ -81,11 +93,11 @@ const LendableObjectList = () => {
     selectTransformedLendingRequests(state, { pagination: requestsPagination }),
   );
   const [visibleCount, setVisibleCount] = useState(REQUEST_INBOX_PAGE_SIZE);
-  const previousRequestOrderingRef = useRef(requestOrdering);
+  const previousRequestArchivedRef = useRef(requestArchived);
   const visibleRequestCount = getVisibleRequestCount({
     visibleCount,
-    currentOrdering: requestOrdering,
-    previousOrdering: previousRequestOrderingRef.current,
+    currentArchived: requestArchived,
+    previousArchived: previousRequestArchivedRef.current,
   });
 
   const lendingRequests = originalLendingRequests.slice(0, visibleRequestCount);
@@ -104,6 +116,17 @@ const LendableObjectList = () => {
       fetchMoreLendingRequests();
     }
     setVisibleCount(nextVisibleCount);
+  };
+
+  const handleArchiveRequest = async (
+    requestId: TransformedLendingRequest['id'],
+    archived: boolean,
+  ) => {
+    await dispatch(editLendingRequest({ id: requestId, archived }));
+    await Promise.all([
+      dispatch(fetchLendingRequests({ query: activeRequestQuery })),
+      dispatch(fetchLendingRequests({ query: archivedRequestQuery })),
+    ]);
   };
 
   const objectsActionGrant = useAppSelector(
@@ -137,9 +160,9 @@ const LendableObjectList = () => {
   };
 
   useEffect(() => {
-    previousRequestOrderingRef.current = requestOrdering;
+    previousRequestArchivedRef.current = requestArchived;
     setVisibleCount(REQUEST_INBOX_PAGE_SIZE);
-  }, [requestOrdering]);
+  }, [requestArchived]);
 
   useEffect(() => {
     if (!heartRef.current) return;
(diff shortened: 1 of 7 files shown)


ch0rizo:
I think aesthetically it would look better if the archive wasn't so visible and prominent. Instead of a button, it would look better as a subtle icon that only appears on hover - maybe at the end of the object's title.

Bestem0r:
> I think aesthetically it would look better if the archive wasn't so visible and prominent. Instead of a button, it would look better as a subtle icon that only appears on hover - maybe at the end of the object's title.

@ch0rizo What do you think about this? I am slightly against hiding the action through hover as it reduces accessibility.
image

lego-webapp/pages/lending/_components/RequestInbox.tsx
```diff
@@ -10,11 +10,11 @@ import { canLoadMoreRequests } from './requestInboxPagination';
...
 import type { TransformedLendingRequest } from '~/redux/models/LendingRequest';
 
-export type LendingRequestOrdering = 'created_at' | '-created_at';
+export type LendingRequestArchivedFilter = 'true' | 'false';
 
-const orderingOptions: PillSwitchOption<LendingRequestOrdering>[] = [
-  { label: 'Nyeste', value: '-created_at' },
-  { label: 'Eldste', value: 'created_at' },
+const archivedOptions: PillSwitchOption<LendingRequestArchivedFilter>[] = [
+  { label: 'Aktiv', value: 'false' },
```
ch0rizo, line 16: Its more accurate with "Aktiv", nice change!

lego-webapp/pages/lending/_components/LendingRequestCard.tsx
```diff
@@ -1,60 +1,93 @@
...
-    <a
-      href={`/lending/${lendingRequest.lendableObject.id}/request/${lendingRequest.id} ${
-        isFromAdmin ? '?fromAdmin=true' : ''
-      }
+    <Card className={styles.lendingRequestCard}>
+      <Flex column width="100%" gap="var(--spacing-sm)">
+        <a
+          href={`/lending/${lendingRequest.lendableObject.id}/request/${lendingRequest.id} ${
+            isFromAdmin ? '?fromAdmin=true' : ''
+          }
```
ch0rizo, line 37: Pre-existing, but there stray space here and a trailing newline. Breaks cards in /admin

lego-webapp/pages/lending/_components/LendingRequestCard.tsx
```diff
@@ -1,60 +1,93 @@
...
-          alt={lendingRequest.lendableObject.title}
-        />
-      </Card>
-    </a>
+        </a>
+        {!isFromAdmin && (
+          <div
+            className={styles.archiveButtonWrapper}
+            onClick={preventCardNavigation}
+          >
```
ch0rizo, line 75: The button is gated on `!isFromAdmin`, but `onArchive` is optional and called as `onArchive?.()`. Any non-admin caller renders a visible, permanently dead "Arkiver" button. Gate on `onArchive && `instead

lego-webapp/pages/lending/_components/LendingRequestCard.tsx
```diff
@@ -1,60 +1,93 @@
...
+  ) => void;
 }) => {
   if (!lendingRequest.lendableObject) {
     return null;
   }
 
+  const preventCardNavigation = (event: MouseEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    event.stopPropagation();
+  };
```
ch0rizo, line 29: Dead after your implementation.

lego-webapp/pages/lending/_components/LendingRequestCard.tsx
```diff
@@ -1,60 +1,93 @@
...
-                  {truncateString(lendingRequest.lendableObject.title, 30)}
-                </h4>
-                <Flex gap="var(--spacing-sm)">
-                  <Time time={lendingRequest.startDate} format="DD. MMM" />
-                  <Icon iconNode={<MoveRight />} size={19} />
-                  <Time time={lendingRequest.endDate} format="DD. MMM" />
+        >
+          <Flex width="100%">
+            <Flex width="100%">
+              <Flex>
```
ch0rizo, line 42: These seems unneccessary

lego-webapp/pages/lending/_components/LendingRequestCard.tsx
```diff
@@ -1,60 +1,93 @@
...
+                    <h4>
+                      {truncateString(lendingRequest.lendableObject.title, 30)}
+                    </h4>
+                    <Flex gap="var(--spacing-sm)">
+                      <Time time={lendingRequest.startDate} format="DD. MMM" />
+                      <Icon iconNode={<MoveRight />} size={19} />
+                      <Time time={lendingRequest.endDate} format="DD. MMM" />
+                    </Flex>
+                  </Flex>
+                  <div className={styles.tagContainer}>
```
ch0rizo, line 54: Undefined css-class.

lego-webapp/pages/lending/_components/RequestInbox.tsx
```diff
@@ -73,15 +80,20 @@ const RequestInbox = ({
...
-          {lendingRequests.map((req) => (
-            <div
-              key={req.id}
-              data-request-id={String(req.id)}
-              className={styles.requestCardWrapper}
-            >
-              <LendingRequestCard lendingRequest={req} />
-            </div>
-          ))}
+          {lendingRequests
```
ch0rizo, line 83: This filter runs after `+Page.tsx:99` has already sliced to `visibleRequestCount`, so archived-but-still-in-the-list entries eat visible slots and we render fewer than a full page.

The pagination ids are never pruned. `buildPaginationReducer` does ids = [...new Set(ids.concat(result))], so after archiving, the id stays in the `?archived=false` set until a reload.

lego-webapp/pages/lending/_components/RequestInbox.tsx
ch0rizo, line 49: `hasRequests/showEmptyState` are computed from the unfiltered list, so when the filter removes everything we render an empty `<div>` and never show the empty state. Archive your only active request and you get a blank gap instead of "Du har ingen utlånsforespørsler". Same on Arkivert after un-archiving the last one.

ch0rizo, line 54: Both counts are pre-filter, so "Se mer" can show when there's nothing left to reveal

lego-webapp/pages/lending/_components/RequestInbox.tsx
```diff
@@ -60,9 +65,11 @@ const RequestInbox = ({
...
         <PillSwitch
-          options={orderingOptions}
-          value={ordering}
-          onChange={onOrderingChange}
+          options={archivedOptions}
+          value={archived}
+          onChange={(value) =>
+            onArchivedChange(value === 'true' ? 'true' : 'false')
+          }
           ariaLabel="Sorter utlånsforespørsler"
```
ch0rizo, line 73: Its a filter now😅 (ariaLabel)

lego-webapp/pages/lending/_components/RequestInbox.tsx
ch0rizo, line 48: `requestIds` drives the gsap animation but is built pre-filter, so the id set doesn't match the rendered DOM nodes.

ch0rizo: When the refetch inserts a new id mid-list, `getAnimationStep` sees a non-append change and returns all ids causing every card replays the fromTo(y: -18, opacity: 0) entrance and the whole inbox flashes.

lego-webapp/pages/lending/+Page.tsx
```diff
@@ -105,6 +114,17 @@ const LendableObjectList = () => {
     setVisibleCount(nextVisibleCount);
   };
 
+  const handleArchiveRequest = async (
+    requestId: TransformedLendingRequest['id'],
+    archived: boolean,
+  ) => {
+    await dispatch(editLendingRequest({ id: requestId, archived }));
+    await Promise.all([
+      dispatch(fetchLendingRequests({ query: activeRequestQuery })),
```
ch0rizo, line 123: This refetches page 1 without next: true, and `buildPaginationReducer` unconditionally does `paginationNext.next = parse(payload.next…)`. The cursor rewinds to page 2 for a list the user may have paged well past.

Repro: click "Se mer" twice (12 loaded, cursor at page 4), archive one -> cursor is page 2. The next "Se mer" refetches page 2, every id is already in the Set, nothing new appears, but `visibleCount` still climbs. The button goes dead.

lego-webapp/pages/lending/+Page.tsx
```diff
@@ -105,6 +114,17 @@ const LendableObjectList = () => {
     setVisibleCount(nextVisibleCount);
   };
 
+  const handleArchiveRequest = async (
+    requestId: TransformedLendingRequest['id'],
+    archived: boolean,
+  ) => {
+    await dispatch(editLendingRequest({ id: requestId, archived }));
```
ch0rizo, line 121: `promiseMiddleware` rejects on FAILURE. The refetches below never run and, since `onPress` discards the returned promise, it surfaces as an `unhandledrejection `(Sentry noise). Worth a try/catch.

lego-webapp/pages/lending/+Page.tsx
```diff
@@ -24,19 +25,27 @@ import { selectPaginationNext } from '~/redux/slices/selectors';
...
   const requestQuery = {
-    ordering: requestOrdering,
+    archived: requestArchived,
+  };
+  const activeRequestQuery = {
+    archived: 'false' as const,
+  };
+  const archivedRequestQuery = {
+    archived: 'true' as const,
   };
```
ch0rizo, line 49: On the Aktiv tab `requestQuery` is already `{archived: 'false'}`

The other list gets refetched by `usePreparedEffect` `(deps [requestArchived])` on tab switch anyway.

These constants would be better next to `defaultLendingQuery` I think

ch0rizo (changes requested):
Love the change, archived makes much more sense than "Nyeste/Eldste"!

One structural thing before this lands: the archived filter is applied client-side in `RequestInbox` after `+Page.tsx` has sliced and counted the list, while pagination ids are never pruned. That single placement causes four separate symptoms - short pages, a missing empty state, a wrong "Se mer", and the inbox re-animating.

Filtering before the slice (or server-side, since archived is already in the query) collapses all four. PR also needs a rebase for `migrate_search` infra break

ch0rizo (approved):

PR #5945, opened 2026-04-21
Modern footer

ch0rizo:
# Description

I feel it's time to modernize certain components. This is a proposal for how the footer could look. It might seem like a bit of a drastic change or out of place, but I think you have to start somewhere - and why not with a global component. What do you all think?

I suggest trying this component and get a feel for it youself. Feedback is most welcome!

# Result

- [] Changes look good with different viewports (mobile, tablet, etc.).

> [!CAUTION]

https://github.com/user-attachments/assets/be80e06e-3488-4e50-afac-03a456b8ceef

# Testing

---

### DIFF
--- a/lego-webapp/components/ModernFooter/ModernFooter.module.css
+++ b/lego-webapp/components/ModernFooter/ModernFooter.module.css
@@ -1,0 +1,254 @@
+.footer {
+  background: var(--dark-background-color);
+  color: var(--dark-font-color);
+  padding: 60px 0 32px;
+}
+
+.inner {
+  max-width: var(--modern-lego-max-width);
+  margin: 0 auto;
+  padding: 0 var(--spacing-xl);
+}
+
+.grid {
+  display: grid;
+  grid-template-columns: 2fr 1fr 1fr 1fr;
+  gap: 48px;
+}
+
+/* Brand column */
+.logo {
+  width: 130px;
+  height: auto;
+  object-fit: contain;
+  margin-bottom: var(--spacing-md);
+}
+
+.description {
+  font-family: monospace;
+  font-size: var(--font-size-xs);
+  line-height: 1.6;
+  opacity: 0.7;
+  max-width: 360px;
+  margin: 0 0 var(--spacing-md);
+  color: var(--color-absolute-white);
+}
+
+.address {
+  font-family: monospace;
+  font-size: 11px;
+  letter-spacing: 0.1em;
+  opacity: 0.5;
+  margin: 0;
+  color: var(--color-absolute-white);
+}
+
+.statusLink {
+  display: inline-flex;
+  align-items: center;
+  gap: var(--spacing-xs);
+  margin-top: var(--spacing-md);
+  padding: var(--spacing-xs) var(--spacing-sm);
+  border-radius: 999px;
+  text-decoration: none;
+  color: var(--color-absolute-white);
+  opacity: 0.7;
+  background: transparent;
+  transition:
+    opacity var(--easing-fast),
+    background var(--easing-medium),
+    color var(--easing-medium);
+}
+
+.statusLink:hover {
+  opacity: 1;
+  background: rgba(255, 255, 255, 0.08);
+  color: var(--status-color);
+}
+
+.statusLink:hover .statusMessage {
+  color: var(--status-color);
+}
+
+.statusDot {
+  position: relative;
+  display: inline-block;
+  width: 7px;
+  height: 7px;
+  border-radius: 50%;
+  flex-shrink: 0;
+}
+
+.statusDot::before {
+  content: '';
+  position: absolute;
+  inset: 0;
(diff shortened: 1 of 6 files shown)


falbru:
I think it's a very good idea to start modernizing components. Here's my thoughts:
- For the footer of a student organization such as Abakus, I think the most important part is having contact information visible (that is: socials, kontaktskjema, adresse etc.). That's what footers are used for typically for an organization. For me, the contact information becomes too cluttured among all the other non-contact links (Historie, Fondet, Styret, Jobbannonse, itDAGENE). I think it should atleast be its own Contact-column
- Also, I think we shouldn't make the footer yet another place of navigating abakus.no. Styret, Fondet, readme, historie and such should be removed imo. I think it would be better to try to modernize the navbar/menu instead, and add new links to them if the current ones are insufficient. We should rather improve our current navigation menus than creating new ones.
- I also think it's too big, especially with the watermark.
- The watermark looks like a new logo for Abakus, and Abakus is very strict when it comes to their logos. If we start designing our own logos for Abakus, we'll be in trouble (PR will be angry). Look at https://abakus.no/pages/organisasjon/108-grafisk-profil
- We may be obligated to have Bekk in the footer. This should be in the HSP contract. Talk to the leader of Abakus about it.
- I think we should be careful with giving links and text custom styling and font families. If we are not careful, the whole site will become a Frankenstein of different styles and designs with no harmony. Consistency is better than more interesting design in my opinion. I get that we have to start somewhere though...

One last thing is that even though we want to make our site more modern, we also want to preserve its personality. One of the things I don't really like about Online's new site is that even though it looks very modern and clean, it looks a lot like any Next.js/Vercel app. I feel like the buildings in the background of the old footer was a cool aesthetic even though it wasn't that "clean".

I'd love to discuss more about this! Modernizing components is definitely something we should work more on!

ivarnakken:
love the initiative! 🙌🏼

Viljen789 (commented):
REALLY LIKE IT! Might have to check in with PR though

ch0rizo:
> I think it's a very good idea to start modernizing components. Here's my thoughts:
> 
> * For the footer of a student organization such as Abakus, I think the most important part is having contact information visible (that is: socials, kontaktskjema, adresse etc.). That's what footers are used for typically for an organization. For me, the contact information becomes too cluttured among all the other non-contact links (Historie, Fondet, Styret, Jobbannonse, itDAGENE). I think it should atleast be its own Contact-column
> * Also, I think we shouldn't make the footer yet another place of navigating abakus.no. Styret, Fondet, readme, historie and such should be removed imo. I think it would be better to try to modernize the navbar/menu instead, and add new links to them if the current ones are insufficient. We should rather improve our current navigation menus than creating new ones.
> * I also think it's too big, especially with the watermark.
> * The watermark looks like a new logo for Abakus, and Abakus is very strict when it comes to their logos. If we start designing our own logos for Abakus, we'll be in trouble (PR will be angry). Look at [abakus.no/pages/organisasjon/108-grafisk-profil](https://abakus.no/pages/organisasjon/108-grafisk-profil)
> * We may be obligated to have Bekk in the footer. This should be in the HSP contract. Talk to the leader of Abakus about it.
> * I think we should be careful with giving links and text custom styling and font families. If we are not careful, the whole site will become a Frankenstein of different styles and designs with no harmony. Consistency is better than more interesting design in my opinion. I get that we have to start somewhere though...
> 
> One last thing is that even though we want to make our site more modern, we also want to preserve its personality. One of the things I don't really like about Online's new site is that even though it looks very modern and clean, it looks a lot like any Next.js/Vercel app. I feel like the buildings in the background of the old footer was a cool aesthetic even though it wasn't that "clean".
> 
> I'd love to discuss more about this! Modernizing components is definitely something we should work more on!

Thanks for the feedback @falbru ! I fully agree on points 1, 2, and 5, and I share your concern about preserving the Abakus personality - whatever that ultimately means for us.

That said, I do think we need to be bolder with typography and designs. The site feels a bit bland right now, and I think there's a lot of untapped potential in experimenting with font weights, sizes, and typefaces. That doesn't mean going overboard - consistency is key. What we really need is a proper design system: not just implemented in CSS, but also in a visual platform like Figma, where components are easy to adjust and serve as great context for AI tools.

Building a good design system is easier than ever with AI. Having Claude design within a system that reflects our perception of Abakus' personality and charm is genuinely realistic.

On the watermark - I'll loop in PR, but I don't see a major issue with it given that it only appears in the footer alongside the official Abakus logo.

Finally, I think a dedicated Slack channel for design discussions would go a long way for keeping things organised and easy to follow.

falbru:
I agree that we should strive to be bolder! This is definitely a good start! I think trying to redesign our brand in Figma is a cool idea! Perhaps we could get some input from PR or even Bekk when it comes to design. Would love to join the Slack channel and get other people's opinion on this!

PR #5952, opened 2026-05-07
Fix some TypeScript errors

falbru:
I think we should start aiming for fixing ALL TypeScript errors.

We'll continue to introduce new TypeScript errors as long as the TypeScript linter in the CI is turned off, and we can't turn it on until all TypeScript errors are fixed. It's a negative feedback loop.

We might try to compare amount of TS errors before and after for a PR. Not sure if it's a good idea or not :thinking: 

Here are some fixes at least.


ch0rizo (approved):
Yes!

TonyWorep (approved):
I agree

PR #5971, opened 2026-05-18
Refactor and optimize droneCI setup

### LINKED ISSUE #5970: Optimize droneCI lego-webapp

ch0rizo:
# Description

I would like inputs from the more experience regarding our droneCI. I done some research on how to make droneCI more effective. It has been a pain waiting for checks, and I feel like a lot of it is redundant or at least could have done in parallell. This is what I have attempted to implement:

1. Define check step and refer to it by the other tests. More of a refactor.
2. Decouple setup from the backend services (biggest speedup)
	- `pnpm install` has no reason to wait for postgres/api to spin up
	- `setup` -> build/test/lint/typescript/install_cypress now run in parallell with the slow api step (reset_db/migrate/load_fixtures/rebuild_index), instead of after it.
3. Persist the Cypress binary across builds
	- New host volume, mounted into the three Cypress steps so the 100MB binary download happens once
4. New non-gating `pnpm audit` step for security vulnerabilities 

All in all I dont think this is going to reduce a lot of time, but at least its faster😅

# Testing

---

Resolves #5970

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -15,7 +15,7 @@ steps:
     environment:
       POSTGRES_USER: lego
       POSTGRES_HOST_AUTH_METHOD: trust
-    when:
+    when: &checks
       event: [push]
       branch:
         exclude: [build]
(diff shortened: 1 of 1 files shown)


ch0rizo:
Wow, tests passed. Seems to be 3-4min faster. Still slow😪

ch0rizo:
> Wow, tests passed. Seems to be 3-4min faster. Still slow😪

Main drawback is still cypress

christiangryt (approved):
Hella swag.

More of this. Do we need to worry about cypress versions due to having a persistent binary?

PR #6010, opened 2026-07-18
Refactor theme handling into a single resolver

### LINKED ISSUE #6009: Refactor color theme
Setting color themes has 4 different writers. There is single owner. The SSR render (defaulting to light), the pre-paint script in +Head (defaulting to auto/OS), the hydration listener in themeUtils (defaulting to light from localStorage), and a patch-effect in the Header that flips it back to the account theme for logged-in users. Each has its own resolution logic and its own idea of what "no preference" means

ch0rizo:
# Description

Fix theme flash and refactor theme handling into a single resolver

Landing on the public frontpage in dark mode flashed dark → light: four independent theme writers (SSR attribute, pre-paint script, two hydration effects) each resolved the theme differently.

- One precedence rule in resolveTheme: explicit account theme → "auto" follows OS → logged out: device choice → OS
- The pre-paint script is generated from resolveTheme itself; a spec runs the script against the function to prevent drift
- ThemeManager owns the theme after hydration (account changes, cross-tab, live OS switching); the old listener, Header effect, and CustomEvent bus are gone
- New localStorage key — the old one was stamped on every visit and would pin existing devices to a theme they never chose

# Result

# Testing

- [x] I have created tests for my changes.

---

Resolves #6009

### DIFF
--- a/lego-webapp/components/Header/ToggleTheme.tsx
+++ b/lego-webapp/components/Header/ToggleTheme.tsx
@@ -2,6 +2,7 @@ import { Icon } from '@webkom/lego-bricks';
 import cx from 'classnames';
 import { MoonStar, Sun } from 'lucide-react';
 import { useEffect, useState } from 'react';
+import { useAppDispatch } from '~/redux/hooks';
 import { applySelectedTheme, getTheme, useTheme } from '~/utils/themeUtils';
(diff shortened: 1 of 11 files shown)


TonyWorep (approved):
nice refactor.

PR #6011, opened 2026-07-18
Fix undefined and import type errors

ch0rizo:
## Description

Fixes 12 of the ~141 type errors from `pnpm types` - the mechanical ones that don't require structural changes:

- **`packages/lego-editor/src/index.tsx`**: remove explicit `.ts`/`.tsx` extensions from 5 import paths (TS5097)
- **`utils/appConfig.ts`**: export `defaultAppConfig`, which `vitest.setup.ts` imports (TS2614)
- **`redux/actions/UserActions.ts`**: update stale `ChangePassword` import path after the component moved to `pages/users/@username/settings/profile/` (TS2307)
- **`redux/legoAdapter/createLegoAdapter.ts`**: remove unused `@ts-expect-error` directive (TS2578)
- **`cypress/support/commands.ts`**: declare the custom `mount` command with its real signature (including the `reduxStore` option) instead of referencing an unimported `mount` (TS2304)
- **`redux/slices/surveys.ts`**, **`pages/surveys/.../Results.tsx`**, **`pages/surveys/index/SurveyItem.tsx`**: coalesce `survey.event ?? undefined` when passing to `selectEventById`, which takes `EntityId | undefined` (TS2345)

## Test steps

- `pnpm types` - error count drops from 141 to 129 on the branch base; each of the 12 targeted errors is gone

### DIFF
--- a/lego-webapp/cypress/support/commands.ts
+++ b/lego-webapp/cypress/support/commands.ts
@@ -1,4 +1,7 @@
 import { apiBaseUrl } from './utils';
+import type { MountOptions, MountReturn } from 'cypress/react';
+import type { ReactNode } from 'react';
+import type store from '~/cypress/fixtures/store';
 
 declare global {
   namespace Cypress {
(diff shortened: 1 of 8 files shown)


falbru (approved):
Looks good!

PR #6045, opened 2026-08-07
Fix: Admin unregister and notifications for it

### LINKED ISSUE #6044: No notification when admin unregisters an user from an event
When an admin unregisters an user, the user gets no notification that they were removed from the event.

Users get notifications when they are registered to an event by an administrator so this should happen when they are unregistered as well.

TonyWorep:
## Description

When an admin unregisters an user from an event, the user does not get any notification that they were removed.
This is because the unregister happened at the standard unregister endpoint, not the admin_unregister endpoint.
In addition, the user did not get the notification on their feed when they were unregistered.
The pull request fixes both of these issues.

## Result

> [!CAUTION]

<table>
  <thead>
    <tr>
      <th width="25%">Description</th>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Confirm Modal</td>
      <td>
bilde
</td>
      <td>
Skjermbilde_20260807_111451
</td>
    </tr>
<tr>
<td>Notification Feed</td>
<td>
</td>
<td>
bilde
</td>
</tr>
  </tbody>
</table>

## Testing

---

Resolves #6044

### DIFF
--- a/lego-webapp/components/Feed/index.tsx
+++ b/lego-webapp/components/Feed/index.tsx
@@ -7,6 +7,7 @@ import { FeedActivityVerb } from '~/redux/models/FeedActivity';
 import { selectFeedActivitiesByFeedId } from '~/redux/slices/feeds';
 import Activity from './activity';
 import AdminRegistrationRenderer from './renders/adminRegistration';
+import AdminUnregistrationRenderer from './renders/adminUnregistration';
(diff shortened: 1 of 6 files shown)


ch0rizo (approved):

ch0rizo:
This is great! And we already had admin unregister endpoint we didn't use🙈

PR #6070, opened 2026-08-28
fix money application

Krozzinator:
## Description

Moved money application and practical info from /interest-groups to /events/interest, for better consistency and accessibility. The project structure has also been changed to remove the old interest-groups page (which has been removed). Redundant code to clean up, but full functionality is preserved.

## Result

- [x] Changes look good on both light and dark theme.
- [x] Changes look good with different viewports (mobile, tablet, etc.).
- [x] Changes look good with slower Internet connections.

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>

[Screencast from 2026-08-28 14-36-44.webm](https://github.com/user-attachments/assets/f9e4f2b7-f62d-414e-be7a-b793bc621866)

</td>
      <td>

[Screencast from 2026-08-28 14-34-15.webm](https://github.com/user-attachments/assets/ada51875-144d-4039-9af2-ff47fe299210)

</td>
    </tr>
  </tbody>
</table>

## Testing

- [x] I have thoroughly tested my changes.

Please describe what and how the changes have been tested, and provide instructions to reproduce if necessary.

PR #6088, opened 2026-09-04
Fix broken links and remove dead code after interest groups move

### LINKED ISSUE #6087: Asymmetric namespace for interest-group
/events/interest/interest-groups is not the right namespace. Interest-groups should have their own namespace.

The idiomatic fix is to keep /events/interest as the hub UI, but give everything group-related its own top-level namespace back. Which is also what the rest of the app does (committees at /pages/komiteer/@id, revue at /pages/revy/@id, boards at /pages/styrer/@id via resolveGroupLink)

ch0rizo:
## Description

#6070 moved the interest group pages under `/events/interest/interest-groups/...`, which left broken
navigations (saving a group, admin permissions, feed notifications all pointed at the removed
`/interest-groups` routes), a dead back link, no entry point to the create-group page, and duplicated
dead components/CSS.

This restores `/interest-groups/...` as the canonical namespace for everything group-related
(detail, edit, new, info, money-application, create-application), keeping `/events/interest` as the
hub. All internal links are repointed, the dead code is removed, and the old `/interestgroups` alias
is dropped along with its target. A small spec pins the redirect map: live routes are never
swallowed by a redirect, and redirects never chain.

URLs that only existed during the #6070 window (`/events/interest/interest-groups/...` etc.) now 404
intentionally.

## Result

> [!CAUTION]

## Testing

---

Resolves #6087


lego-webapp/pages/+config.ts
```diff
@@ -8,7 +8,6 @@ export default {
 
   redirects: {
     '/kontakt': '/contact',
-    '/interestgroups': '/interest-groups',
```
itsisak, line 11: why not have this?

itsisak (approved):

PR #6099, opened 2026-09-12
Add transient socket event support and refactor socket group

### LINKED ISSUE #6098: Add support for using websocket messages outside of redux
We want the website to fell more "alive" this requires support for more features similar to the toasts that is not directly tied to state and goes through redux.

itsisak:
## Description

Adds a hook to subscribe/listen to one-off effects over websockets. This enables behavior similar to the toasts but anywhere in the app. Also refactored the websocket group provider to a hook to standardize websocket structure a bit more.

Currently we have different ways websocket messages is handled:

<b>live update state (eg. registrations)</b>
 	 is dumped right into redux which makes components rendering this data update in real time

<b>live feedback for global providers (toasts) </b>
     creates new toasts to a global toast provider that always exists

<b> (this pr)live feedback for any component </b>
	Any component can subscribe, not the global requirement that toasts use.

This is hence inherently different behavior than previous setup which is why this is needed.

## Testing

Wrote some tests.

Resolves #6098


lego-webapp/components/WebsocketStatus/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const iconNode = connected ? (
+    <Radio />
+  ) : pending ? (
+    <LoaderCircle className={styles.spin} />
+  ) : (
+    <Unplug />
+  );
+
+  const tooltipContent = connected
+    ? 'Kommentarfeltet er live. Du trenger ikke refreshe for å se andre oppdateringer.'
```
jonasdeluna, line 20: This comment seems to be comment specific is that intentional?

jonasdeluna (approved):
📈📈📈

lego-webapp/components/WebsocketStatus/index.tsx
itsisak: its just me being lazy in a previous contribution..

PR #6108, opened 2026-09-12
Init playwright

### LINKED ISSUE #6095: Migrate to Playwright
I would like to open the discussion to migrate to playwright from cypress as it does not fulfill our requirements. 

Cypress has been a great tool to ensure our frontend components and UI flow works. But as the tests stacks up its runtime scales poorly. Cypress opens its own browser in test env and clicks around. It does not have a free plan for parallelization, which is frustrating. This causes long CI queues, which halts our efficiency. This paired with our current infrastructure only have two workers for CI (builds) pipeline causes extremely long queues, hence the long wait time.

I propose to migrate to [Playwright](https://playwright.dev/docs/intro) because of its efficiency, parallelism, and better debugging in CI. What increases efficiency for Playwright is the execution model; drives the browser out-of-process over a protocol. Because of that, Playwright handles multiple tabs, multiple origins, popups, and iframes natively. Playwright tests Chromium, Firefox, and WebKit, so you get a real Safari-engine proxy.

ch0rizo:
## Description

Add playwright library to lego-webapp. Is to gradually replace Cypress. Referring to the issue where I briefly discuss Playwright.

I am a little worried about the multiple minor bumps!

---

Resolves #6095


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Install Playwright Browsers
+      run: pnpm exec playwright install --with-deps
+    - name: Run Playwright tests
+      run: pnpm exec playwright test
+    - uses: actions/upload-artifact@v4
+      if: ${{ !cancelled() }}
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
```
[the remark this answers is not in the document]
ch0rizo: Redundant. Its removed as it will run on drone

lego-webapp/playwright.config.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
+    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
```
jonasdeluna: Why are these commented out and not removed?

jonasdeluna (approved):
Nice!!

jonasdeluna:
:pray: :pray: :pray: :pray: :pray:

lego-webapp/playwright.config.ts
ch0rizo: I mean its a draft. And I wanted to take up the discussion if we want these test as well

package.json
```diff
@@ -39,5 +39,8 @@
       "esbuild": "^0.25.0"
     }
   },
-  "packageManager": "pnpm@10.6.5"
+  "packageManager": "pnpm@10.6.5",
+  "devDependencies": {
+    "@types/node": "^22.20.2"
+  }
 }
```
itsisak: Move this to `lego-webapp/package.json`

itsisak (approved):

PR #6109, opened 2026-09-13
Move WebsocketsGroup status to useSocketGroup

itsisak:
## Description

With the current websocket setup, its status and group status does not really need to exist within redux. So this keeps the status to the hook that needs it. 

## Testing


ch0rizo (approved):
Branch convention 😠

ch0rizo:
> Branch convention 😠

But nice!

PR #6113, opened 2026-09-14
Banner design improvements

### LINKED ISSUE #4301: Make banner look nicer
Add nicer graphic design and styling

itsisak:
I am really tired of our banner looking as bad as it does, (and yes, i made the current red preset so I blame none but me)

There are a couple fixes this introduces:

- On large devises there is so much unused space since the header, subheader and countdown are all stacked in the middle. this separates the headers and the counter for non mobile devices.
- I think it feels confusing to have a hover effect when there is no link, so now the hover effect is only shown when the banner is an actual link
- Creates a new banner preset with a more modern style. 

The new preset requires a small backend pr too :)

I am not super sure about the animated background. It does look a little out of place since the weak red coloring is not really used anywhere else, but I do like how this makes the banner more subtle yet modern and "alive". I dont really like making the backgorund color fully lego-red-color since it is so intense and just using the border also does not look to good, so this is my proposal for a sort of middle ground.

I also worry that it maybe looks too vibe-coded, but the background is inspired by a 12 year old stack overflow post. 

Gimme feedback, pretty please :)

## Result

*before*

https://github.com/user-attachments/assets/1cbfd000-56c4-414a-909e-59d2cdf0fb84

*after* with new modern preset and split header/countdown for large devices

https://github.com/user-attachments/assets/d9780151-cb01-4d27-a13f-b5629073d165

## Testing

---

Resolves webkom/lego#4301


ch0rizo:
Personally I think it kinda looks vibe-coded. And the reason is the gradient combined with the hover glance effect. 

On a side note, it also gives me the vibe of this github component:
Screenshot 2026-09-14 at 19 50 40

ch0rizo:
> Personally I think it kinda looks vibe-coded. And the reason is the gradient combined with the hover glance effect.
> 
> On a side note, it also gives me the vibe of this github component: Screenshot 2026-09-14 at 19 50 40

But I really like that it does not have any border lines. But then I guess we need to remove wiggle as that would not look nice if its not a visible box, my guess. Like the space between the text content and timer!

Too suggest something concrete I would keep the the red color gradient, but keep it static and at an angle on the left side where the text contents is. On the right side I would maybe try to implement the abakus-grid we have on the public frontpage. Kind of the vibe the github component over.

itsisak:
https://github.com/user-attachments/assets/9cb84f17-25e2-4e10-87f0-78ae5e410cc9

I propose using this simple design for now and keep in mind further improvements for later.

Also slightly decreased the wiggle effect.

ch0rizo (approved):

frikkbol (approved):

lego-webapp/pages/admin/banners/BannerEditor.tsx
```diff
@@ -35,6 +35,7 @@ import styles from './BannerOverview.module.css';
 
 const colorToRepresentation: Record<Color, string> = {
   red: 'Rød',
+  modernRed: 'Moderne Rød',
```
Bestem0r: Maybe just replace the previous red with the new one to prevent the ugly one being used at all?

Bestem0r (approved):
Agree with @ch0rizo. I think the gradient looks a bit out of place, but the latest version looks fabulous 🖼️

lego-webapp/pages/admin/banners/BannerEditor.tsx
itsisak: I agree, this made more sense for the initial animated design suggestion

Issue #6122, opened 2026-09-16
Fix flaky cypress


itsisak:

itsisak:
’tis not cypress that is broken (although that too may have its issues) but it seems that the frontend crashes sometimes. Potential fix described in closing comment of #6124

PR #6124, opened 2026-09-16
Import gsap via default export

itsisak:
## Description

I wanna see if this helps fix our pipeline. It seems that that whenever cypress tests fail we get an error message in the frontend complaining about the gsap imports. Pretty strange that this only happens sometimes tho..

Might #6122


itsisak:
This one passed on first attempt at least which is a good sign :)

ch0rizo (approved):
Weird if this fixes it

itsisak:
> Weird if this fixes it

Yea ik, but if it randomly does it would be so nice. 

Now the tests failed but not for same reason at least, I think there is too much load atm so the services just timeout

itsisak:
This does not fix all errors that sometimes show up and would also not be a proper fix. 

The problem is tied to ESM vs CJS imports in some libriaries which have been resolved in newer versions. Updating `vike` to `^0.6.0` would solve this issue I think, but it requires react 19.

Screenshot 2026-09-17 at 09 30 51

PR #6143, opened 2026-09-21
Bump storybook

### LINKED ISSUE #6142: Bump (and fix) Storybook in lego-bricks

itsisak:
## Description

Pretty sure `storybook` is currently broken. This fixes it.  

Bumps `storybook` and updates its dependencies.
`react-docgen-typescript` -> `react-docgen` its faster and does actually work. Think the typescript one does not work because ts config is off or something.
Fix some bad stories.

Resolves #6142


ch0rizo (approved):
Great!

